PR #7, opened 2021-07-06
Create UI for authentication login TMail

dab246:
https://user-images.githubusercontent.com/80730648/124576785-d7070200-de76-11eb-9283-3d1f7b82faa4.mov


chibenwa:
Hint: if someone types `https://openpaas.linagora.com/`, you get it, and it is not the JMAP session you can then add `.well-known/jmap` to it : `https://openpaas.linagora.com/.well-known/jmap` :-) 

This tricks simplifies user life :-p

PR #16, opened 2021-07-29
Get session for TMail

dab246:
- Get session when logged success

### DIFF
--- a/lib/features/session/domain/usecases/get_session_interactor.dart
+++ b/lib/features/session/domain/usecases/get_session_interactor.dart
@@ -1,0 +1,19 @@
+import 'package:core/core.dart';
+import 'package:dartz/dartz.dart';
+import 'package:tmail_ui_user/features/session/domain/repository/session_repository.dart';
+import 'package:tmail_ui_user/features/session/domain/state/get_session_state.dart';
+
+class GetSessionInteractor {
+  final SessionRepository sessionRepository;
+
+  GetSessionInteractor(this.sessionRepository);
+
+  Future<Either<Failure, Success>> execute() async {
+    try {
+      final session = await sessionRepository.getSession();
+      return Right<Failure, Success>(GetSessionSuccess(session));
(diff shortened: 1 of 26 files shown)


core/lib/data/network/config/endpoint.dart
```diff
@@ -2,4 +2,5 @@ import 'package:core/data/network/config/service_path.dart';
 
 class Endpoint {
   static final ServicePath authentication = ServicePath('/api/login');
+  static final ServicePath session = ServicePath('/jmap/session');
```
chibenwa, line 5: Pro-tip! Prefer `.well-known/jmap`

hoangdat: thanks for your notice

chibenwa: It allows auto discovery and redirects to jmap/session 

See https://jmap.io/spec-core.html#service-autodiscovery https://datatracker.ietf.org/doc/rfc8615/

lib/features/session/domain/usecases/get_session_interactor.dart
```diff
@@ -0,0 +1,19 @@
...
+import 'package:dartz/dartz.dart';
+import 'package:tmail_ui_user/features/session/domain/repository/session_repository.dart';
+import 'package:tmail_ui_user/features/session/domain/state/get_session_state.dart';
+
+class GetSessionInteractor {
+  final SessionRepository sessionRepository;
+
+  GetSessionInteractor(this.sessionRepository);
+
+  Future<Either<Failure, Success>> execute() async {
```
hoangdat, line 11: Why not return a stream.
Loading -> State ...

dab246: The processing of this task occurs on a loading screen. So I don't think the state needs to be concerned here.

hoangdat (approved):

PR #23, opened 2021-08-13
Read email content for each email in Thread

dab246:
- [x] Implement get email content when click each email
- [x] Mark `unread` email, `flagged `email in list email
- [x] Display message content  with `text/plain` format
- [x] Display address `To`, `Cc`, `Bcc` header  with `Collapse/Expand` mode
- [x] Display message content  with `text/html` format

https://user-images.githubusercontent.com/80730648/129350690-956c58b5-d627-4d1d-b788-b4f79b1bcaea.mov

**Demo Display email content with html format**

https://user-images.githubusercontent.com/80730648/129889395-48c844db-24ab-4606-9a1d-ec72cf260f94.mov

https://user-images.githubusercontent.com/80730648/130059536-d76a4f2d-39da-4bf6-8761-153531ffa535.mov


chibenwa:
Looks good so far.

Tiny remarks:
 - Inlined attachment should not be rendered as attachments: exclude them from attachment list
 - I'm slightly concered about display when we have say 20+ attachments: they would take all of the space.
     - I would recommand that when we have 4 or + attachments we only display `20 attachments (expend button)` (mobile, on tablet today behaviour likely is acceptable.)
  - Be sure you do display `Cc and Bcc` headers if present on the detailed email view.
  
Use sentDate for the date display in `email detailed view` and `email list view`.

chibenwa:
Using `Mailbox/query` you can be able to load the mailbox list in parallel to the email list: this will fasten the loading time.

(I have the impression you load the email list then the mailbox list...)

chibenwa:
One other display remark: it is hard for me to distinguish from the email list `read emails` from `unread emails`. 

Let me suggest you to render the subject in bold when the email is unread, normally when it is read, this might be enough to be doing the trick.

chibenwa:
Do we support display of `Important / flagged` messages ? IE the Star ?

You can star messages in OpenPaaS and try to support that as well in TMail.

chibenwa:
(And of course I trust @hoangdat or prioritizing my remarks :-) )

Arsnael:
Wow, impressive! Good job guys, looks damn great!

chibenwa:
I am not convinced of the folding of To, Cc Bcc... I would prefer to have it always expended to be honest.

dab246:
> I am not convinced of the folding of To, Cc Bcc... I would prefer to have it always expended to be honest.
> 
> (At least when it fits in 3-4 lines.)
> 
> It is important information and in the detailed email view we have room. We likely should use it.

What do you mean? Please be more specific. Can you draw detailed pictures so I can understand you.

chibenwa:
THis one is cool:

cool

This one can be improved (display of important information takes extra clicks while it would not harm the display...

Screenshot from 2021-08-16 09-03-59

model/lib/mailbox/presentation_mailbox.dart
```diff
@@ -42,18 +45,16 @@ class PresentationMailbox with EquatableMixin {
...
-      if (unreadEmails!.value.value <= 0) {
-        return '';
-      } else {
-        return unreadEmails!.value.value <= 999
-          ? '${unreadEmails!.value.value}'
-          : '999+';
-      }
     }
+
+    if (unreadThreads == null || unreadThreads!.value.value <= 0) {
```
hoangdat, line 53: IMO, `unreadEmails` is better.

android/app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,10 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
     package="com.linagora.android.tmail">
+
+    <uses-permission android:name="android.permission.INTERNET"/>
+
    <application
+        android:name="io.flutter.app.FlutterApplication"
```
hoangdat, line 7: Why not `TMail`?

core/lib/presentation/views/list/sliver_grid_delegate_fixed_height.dart
```diff
@@ -0,0 +1,64 @@
+import 'package:flutter/rendering.dart';
+
+class SliverGridDelegateFixedHeight
```
hoangdat, line 3: this is for what?

dab246: Custom widget builder for tile of gridview

chibenwa:
Can you update the video now that text/html is supported? <3

dab246:
> Can you update the video now that text/html is supported? <3

Please see the video I attached above. It's not complete yet, I'm testing and fixing some bugs.

chibenwa:
Huge thanks and nice work!

chibenwa:
Remember to get rid of `John Doe // user@exemple.com` on the top of the mailbox view ;-)

android/app/src/main/kotlin/com/linagora/android/tmail/MainActivity.kt
```diff
@@ -0,0 +1,11 @@
+package com.linagora.android.tmail
+
+import io.flutter.embedding.android.FlutterFragmentActivity
+import io.flutter.embedding.engine.FlutterEngine
+import io.flutter.plugins.GeneratedPluginRegistrant
+
+class MainActivity: FlutterFragmentActivity() {
```
hoangdat, line 7: why we need to create this file?

pubspec.yaml
```diff
@@ -80,6 +80,21 @@ dependencies:
...
+  http_parser: 4.0.0
+
+  # file_size
+  filesize: 2.0.1
+
+  # flutter_widget_from_html
+  flutter_widget_from_html: 0.6.2
+
+  # photo_view
+  photo_view: 0.12.0
```
hoangdat, line 93: use for what?

pubspec.yaml
```diff
@@ -80,6 +80,21 @@ dependencies:
       url: git://github.com/linagora/jmap-dart-client.git
       ref: master
 
+  # http_parser
+  http_parser: 4.0.0
```
hoangdat, line 84: for what?

pubspec.yaml
```diff
@@ -80,6 +80,21 @@ dependencies:
...
+
+  # file_size
+  filesize: 2.0.1
+
+  # flutter_widget_from_html
+  flutter_widget_from_html: 0.6.2
```
hoangdat, line 90: how about https://pub.dev/packages/flutter_html?

lib/features/email/domain/extensions/presentation_email_extension.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:intl/intl.dart';
+import 'package:model/model.dart';
+import 'package:core/core.dart';
+
+extension PresentationEmailExtension on PresentationEmail {
+
+  String getTimeSentEmail() {
+    if (sentAt != null) {
+      if (sentAt!.value.isToday()) {
+        return DateFormat('h:mm a', 'en_US').format(sentAt!.value);
```
hoangdat, line 10: Please make sure the locale of user.

lib/features/email/domain/usecases/get_email_content_interactor.dart
```diff
@@ -0,0 +1,23 @@
...
+class GetEmailContentInteractor {
+  final EmailRepository emailRepository;
+
+  GetEmailContentInteractor(this.emailRepository);
+
+  Stream<Either<Failure, Success>> execute(AccountId accountId, EmailId emailId) async* {
+    try {
+      yield Right<Failure, Success>(UIState.loading);
+      final email = await emailRepository.getEmailContent(accountId, emailId);
+      yield Right<Failure, Success>(GetEmailContentSuccess(email?.toEmailContent()));
```
hoangdat, line 18: IMO, if `email == null` we should `yield` a failure.

lib/features/email/domain/repository/email_repository.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:jmap_dart_client/jmap/account_id.dart';
+import 'package:jmap_dart_client/jmap/mail/email/email.dart';
+
+abstract class EmailRepository {
+  Future<Email?> getEmailContent(AccountId accountId, EmailId emailId);
```
hoangdat, line 5: Should be 
```
Future<Email> getEmailContent(AccountId accountId, EmailId emailId);
```
And throw the exception if we can not get email.

lib/features/email/data/datasource/email_datasource.dart
```diff
@@ -0,0 +1,6 @@
+import 'package:jmap_dart_client/jmap/account_id.dart';
+import 'package:jmap_dart_client/jmap/mail/email/email.dart';
+
+abstract class EmailDataSource {
+  Future<Email?> getEmailContent(AccountId accountId, EmailId emailId);
```
hoangdat, line 5:
```
Future<Email> getEmailContent(AccountId accountId, EmailId emailId);
```

lib/features/email/data/network/email_api.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:jmap_dart_client/jmap/mail/email/get/get_email_method.dart';
+import 'package:jmap_dart_client/jmap/mail/email/get/get_email_response.dart';
+
+class EmailAPI {
+
+  final HttpClient httpClient;
+
+  EmailAPI(this.httpClient);
+
+  Future<Email?> getEmailContent(AccountId accountId, EmailId emailId) async {
```
hoangdat, line 17: `Future<Email>`

pubspec.yaml
dab246: render error

android/app/src/main/kotlin/com/linagora/android/tmail/MainActivity.kt
dab246: run android

pubspec.yaml
dab246: click view photo in email content

pubspec.yaml
dab246: using MediaType property

lib/features/email/presentation/email_view.dart
```diff
@@ -1,16 +1,21 @@
...
+import 'package:tmail_ui_user/features/email/presentation/widgets/attachment_file_tile_builder.dart';
 import 'package:tmail_ui_user/features/email/presentation/widgets/bottom_bar_mail_widget_builder.dart';
-import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/features/email/presentation/extensions/email_content_extension.dart';
+import 'package:tmail_ui_user/features/email/presentation/widgets/message_content_tile_builder.dart';
+import 'package:tmail_ui_user/features/email/presentation/widgets/sender_and_receiver_information_tile_builder.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 
-class EmailView extends GetWidget<EmailController> {
+class EmailView extends GetWidget {
```
hoangdat, line 16: Why not binding `EmailController` to `EmailView`?

lib/features/email/presentation/email_view.dart
```diff
@@ -38,35 +43,157 @@ class EmailView extends GetWidget<EmailController> {
...
+      AppLocalizations.of(context).no_mail_selected,
+      textAlign: TextAlign.center,
+      style: TextStyle(fontSize: 25, color: AppColor.mailboxTextColor, fontWeight: FontWeight.bold));
+  }
+
+  Widget _buildEmailSubject() {
+    return Padding(
+      padding: EdgeInsets.only(left: 8, top: 4, bottom: 16),
+      child: Text(
+          '${emailController.mailboxDashBoardController.selectedEmail.value?.subject}',
```
hoangdat, line 103: Is it reaction without `Obs()`?

dab246: Obs have been a bao quanh cha của nó.

lib/features/email/presentation/email_view.dart
dab246: On mobile devices EmailView will be canceled when exiting the screen. If you attach an EmailController to it, the controller will be destroyed. On the tablet screen it won't find it.

pubspec.yaml
hoangdat: You did the work out of the scope :). Please focus to the scope of the ticket next time.

lib/features/email/presentation/email_view.dart
hoangdat: > On mobile devices EmailView will be canceled when exiting the screen. If you attach an EmailController to it, the controller will be destroyed

yes, but I think it is true, why we need to care to the controller whenever the view is disposed?

hoangdat: > On the tablet screen it won't find it.

In the tablet, the view still here, why it is not found?

pubspec.yaml
hoangdat: What kind of error?

lib/features/email/presentation/email_view.dart
dab246: Because it is destroyed when we call Get.back(); on mobile

pubspec.yaml
dab246: Unable to render when tags are nested.

lib/features/email/presentation/email_view.dart
hoangdat: on mobile, when open another mail, why the controller is not re-create?

lib/features/email/presentation/widgets/message_content_tile_builder.dart
```diff
@@ -0,0 +1,95 @@
...
+    return Theme(
+      data: ThemeData(
+        splashColor: Colors.transparent,
+        highlightColor: Colors.transparent),
+      child: MediaQuery(
+        data: MediaQueryData(padding: EdgeInsets.zero),
+        child: Center(
+          child: _messageContent.textFormat == TextFormat.PLAIN
+            ? Text('${_messageContent.content}', style: TextStyle(fontSize: 12, color: AppColor.mailboxTextColor))
+            : HtmlWidget(
```
hoangdat, line 37: Please set up `webViewJs = false`, we don't want to run any script here.

lib/features/email/presentation/widgets/message_content_tile_builder.dart
```diff
@@ -0,0 +1,95 @@
...
+                buildAsyncBuilder: (_, snapshot) => snapshot.data ?? SizedBox(
+                  height: 30,
+                  width: 30,
+                  child: CircularProgressIndicator(color: AppColor.primaryColor)),
+                onTapUrl: (url) {
+                  if (_onOpenExternalUrlActionClick != null) {
+                    _onOpenExternalUrlActionClick!(url);
+                  }
+                },
+                webView: true
```
hoangdat, line 52: IMO, we don't need this also, no `iframe` on email content

lib/features/email/presentation/widgets/message_content_tile_builder.dart
hoangdat: beside that. IMO, this library quite cool. Please try with it.

hoangdat: but the parsing time is quite low, we need to make a comparison with other library? WDYT? @chibenwa @dab246

dab246: What do you base that claim on?

lib/features/email/presentation/email_view.dart
dab246: If we do so, we will have to re-create in some cases such as switching between tablet and mobile.

android/app/src/main/kotlin/com/linagora/android/tmail/MainActivity.kt
hoangdat: But you delete another `MainActivity`. Please help me to understand more in one comment, no need to save the time with short sentences.

lib/features/email/presentation/model/message_content.dart
```diff
@@ -9,6 +11,23 @@ class MessageContent with EquatableMixin {
...
+  String getContentHasInlineAttachment(Uri baseDownloadUrl, AccountId accountId, List<AttachmentFile> attachmentFiles) {
+    var contentValid = content;
+    attachmentFiles.forEach((attachment) {
+      if(attachment.cid != null) {
+        final urlDownloadImage = baseDownloadUrl.origin
+            .replaceAll('{accountId}', '${accountId.id.value}')
+            .replaceAll('{blobId}', '${attachment.blobId?.value}')
+            .replaceAll('{type}', '${attachment.type?.mimeType}')
+            .replaceAll('{name}', '${attachment.name}');
+        contentValid = content.replaceAll('cid:${attachment.cid}', '$urlDownloadImage');
```
hoangdat, line 25: please using `UriTemplate`: https://pub.dev/documentation/uri/latest/uri/UriTemplate-class.html. And please create method to create download URL(Maybe in Session?).
```dart 
Url getDownloadUrl(String accountId, String blobId, String name, String type)
```

dab246:
**NOTE**
- Some Emails with content in the form of a full web page have not been displayed, including full tags such as:
````
<!DOCTYPE html>
<html>
<head>
<title>Page Title</title>
</head>
<body>
</body>
</html>
````
- Specific email address:
_noreply@qa.open-paas.org_
_robin@thebusinessanalystjobdescription.com_

**REASON**
- Many tags are formatted with css properties that cannot be parsed by the library resulting in nullable objects.

**Comment by chibenwa** [comment]
>     Many tags are formatted with css properties that cannot be parsed by the library resulting in nullable objects.

As a safe guard can we sanitize such CSS properties before the display?

Did you reported the bug upstream?

Do we have other HTML libraries to try out?

**Comment by dab246** [comment]
> > ```
> > Many tags are formatted with css properties that cannot be parsed by the library resulting in nullable objects.
> > ```
> 
> As a safe guard can we sanitize such CSS properties before the display?
> 
> Did you reported the bug upstream?
> 
> Do we have other HTML libraries to try out?

- If you delete some important css, it will affect the display interface.

- I tried the two most popular libraries on pub dev, but the results are still not satisfactory. I gave them feedback to fix the problem. Meanwhile I'm trying to get their project on git to debug and fix the problem.

- If possible can you give me the code for handling this display on the font-end side of the web team

**Comment by chibenwa** [comment]
> If possible can you give me the code for handling this display on the font-end side of the web team

cc @fabienmoyon  ?

**Comment by fabienmoyon** [comment]
@dab246 and @chibenwa for this we used https://github.com/cure53/DOMPurify like others projects like Protonmail ;)

**Comment by hoangdat** [comment]
> @dab246 and @chibenwa for this we used https://github.com/cure53/DOMPurify like others projects like Protonmail ;)

It quite complicated, and IMO, we should create an issue and move it to the relevant sprint to solve that. We still have an useable reading mail widget. WDYT? @dab246 @chibenwa

**Comment by hoangdat** [comment]
btw, how about it: https://pub.dev/packages/sanitize_html

**Comment by dab246** [comment]
> @dab246 and @chibenwa for this we used https://github.com/cure53/DOMPurify like others projects like Protonmail ;)

I sanitizing the HTML content and it showed up. But it has removed some of the styles of the tags so the interface is not displayed correctly. How did you handle it after sanitizing the HTML @fabienmoyon

**Comment by dab246** [comment]
**Issues:** https://github.com/linagora/tmail-flutter/issues/51

File: android/app/src/main/kotlin/com/linagora/android/tmail/MainActivity.kt (hunk 4 above)
Comment (dab246) [reply android/app/src/main/kotlin/com/linagora/android/tmail/MainActivity.kt:7]: Because I renamed packagegage, git saved it like that.

**Comment by fabienmoyon** [comment]
> I sanitizing the HTML content and it showed up. But it has removed some of the styles of the tags so the interface is not displayed correctly. How did you handle it after sanitizing the HTML @fabienmoyon

@dab246 Yes you need to authorize some CSS and HTML tag like https://github.com/linagora/esn-frontend-common-libs/blob/ffdd13b42ca611d480f90761a879b0166348ba06/src/frontend/js/modules/core.js#L235

File: core/lib/presentation/validator/sane_html_validator.dart [hunk 18]
```diff
@@ -0,0 +1,274 @@
...
+  final Iterable<String>? Function(String)? addLinkRel;
+
+  SaneHtmlValidator({
+    required this.allowElementId,
+    required this.allowClassName,
+    required this.allowAttributes,
+    required this.addLinkRel,
+  });
+
+  String sanitize(String htmlString) {
```
hoangdat, line 211: please write the test for this method.

lib/features/email/presentation/widgets/message_content_tile_builder.dart
```diff
@@ -57,47 +51,21 @@ class MessageContentTileBuilder {
...
+  String purifyHtmlMessage(String html) {
+    return sanitizeHtml(
+        html,
+        allowAttributes: (attribute) => attribute == 'style' || attribute == 'input' || attribute == 'form')
+      .changeStyleBackground()
+      .removeFontSizeZeroPixel()
+      .removeMaxHeightZeroPixel()
+      .removeMaxWidthZeroPixel()
+      .removeHeightZeroPixel()
+      .removeWidthZeroPixel();
```
hoangdat, line 69: please split to another class for `Purifying`, some other place can re-use it. `HtmlMessagePurifier`

core/pubspec.yaml
```diff
@@ -42,6 +42,8 @@ dependencies:
 
   built_collection: 5.1.0
 
+  html: 0.15.0
```
hoangdat, line 45: Why you not use `sanitize_html` instead of parsing and sanitize yourself?

core/lib/presentation/validator/sanitize_html.dart
```diff
@@ -0,0 +1,16 @@
...
+      bool Function(String)? allowAttributes,
+      Iterable<String>? Function(String)? addLinkRel,
+    }) {
+  return SaneHtmlValidator(
+    allowElementId: allowElementId,
+    allowClassName: allowClassName,
+    allowAttributes: allowAttributes,
+    addLinkRel: addLinkRel,
+  ).sanitize(htmlString);
+}
```
hoangdat, line 16: Please create a class and put it in, we can easier to test the invoker.

core/pubspec.yaml
dab246: Since sanitize_html removes all the attributes in the tags that I need to keep. (Ex: style)

hoangdat: No way to modified it?. How about fork it and make a change. We should to avoid to re-invent the cycle as much as possible

hoangdat (approved):

PR #55, opened 2021-08-26
TF-46 Create composer widget

dab246:
https://user-images.githubusercontent.com/80730648/130947165-4447ef6c-77e6-489d-b241-487c0245e13e.mov

https://user-images.githubusercontent.com/80730648/131120420-ce262e8a-1e0c-4153-b9ff-39fe79a3d8f0.mov


model/lib/email/email_action_type.dart
```diff
@@ -2,5 +2,6 @@
 enum EmailActionType {
   reply,
   forward,
-  replyAll
+  replyAll,
+  createNew,
```
hoangdat, line 6: `compose`

lib/features/composer/presentation/composer_view.dart
```diff
@@ -109,68 +115,43 @@ class ComposerView extends GetWidget<ComposerController> {
...
+        ..textInputAction(TextInputAction.newline)
+        ..textStyle(TextStyle(color: AppColor.nameUserColor, fontSize: 14))
+        ..addOnTapActionClick(() => keyboardUtils.hideKeyboard(context))
+        ..textDecoration(InputDecoration(
+            hintText: AppLocalizations.of(context).hint_content_email_composer,
+            hintStyle: TextStyle(color: AppColor.baseTextColor, fontSize: 14),
+            contentPadding: EdgeInsets.only(bottom: 24, left: 28, top: 24, right: 28),
+            filled: true,
+            border: InputBorder.none,
+            fillColor: AppColor.bgComposer)))
```
hoangdat, line 140: why not use `HtmlEditor` also, because in the future, we will support rich text for composer also?

android/app/build.gradle
```diff
@@ -39,6 +39,8 @@ android {
         targetSdkVersion 30
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
+        // The number of method references in a .dex file cannot exceed 64K
+        multiDexEnabled true
```
hoangdat, line 43: Why we need to enable this?

hoangdat (approved):

android/app/build.gradle
dab246: Because the number of method references in a .dex file exceed 64K

hoangdat: Ok, noted to improve

Issue #74, opened 2021-09-09
[Bug] Should keep only one input focus on composer at one time


hoangdat:
### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

Devices, OS

### Additional information

frequency, log files,...

Issue #73, opened 2021-09-09
[Bug] Can not scroll to write email in composer


hoangdat:
### Description
Expand To/Cc/Bcc -> then compose email -> Keyboard show and cover all the display of composer

### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

Devices, OS

### Additional information

frequency, log files,...

PR #77, opened 2021-09-10
Fix some bug compose email

dab246:
- [x] #73 Can not scroll to write email in composer
- [x] #74 Should keep only one input focus on composer at one time

https://user-images.githubusercontent.com/80730648/132826080-7c94c71a-ee50-4c47-b79f-4979568c6204.mp4


hoangdat:
Uploading signal-2021-09-13-141716_001.mp4…

I have the problem, blink all the screen when select all the text.

hoangdat:
> I have the problem, blink all the screen when select all the text. Can not upload, so i send it directly to @dab246

Create another issue: https://github.com/linagora/tmail-flutter/issues/85 because it is not have in other device

hoangdat (approved):

PR #96, opened 2021-09-16
TF-32 Download an attachments from email detailed view

dab246:
- [x] Display attachments list

https://user-images.githubusercontent.com/80730648/133751066-f1a25072-e04b-4435-98c8-389f8083dd16.mov

- [x] Download attachment for android device

https://user-images.githubusercontent.com/80730648/133544973-c8b87341-7f20-4079-a918-676ee89532cc.mov

- [x] Download attachment for iOS device

https://user-images.githubusercontent.com/80730648/133545008-922fb81a-bf18-44f8-ab4b-6fd840387a9f.mov


chibenwa:
As a side note when there is too many attachments we should be able to hide them:

```
> 16 attachments. See...
```

(This makes the content harder to read.)

chibenwa:
Also attachements should be displayed between the headers and the body (not below the body)

hoangdat:
> Also attachements should be displayed between the headers and the body (not below the body)

It quite not good UX, email is not a file sharing tool, it is focus to the content with other information. So that, IMO, attachment will be below body. (user need to understand the context before access the attachment)

android/app/src/main/AndroidManifest.xml
```diff
@@ -3,6 +3,7 @@
     package="com.linagora.android.tmail">
 
     <uses-permission android:name="android.permission.INTERNET"/>
+    <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="28"/>
```
hoangdat, line 6: @huynguyennovem when we use: `flutter_downloader: 1.7.0` do we need to set `android:maxSdkVersion="28"`?

android/app/src/main/AndroidManifest.xml
```diff
@@ -17,7 +18,8 @@
     </queries>
 
    <application
-        android:label="TMail"
+        android:label="@string/app_name"
+        android:requestLegacyExternalStorage="true"
```
hoangdat, line 22: we also no need this flag @huynguyennovem ?

android/app/src/main/AndroidManifest.xml
huynguyennovem: we should add this to ignore the warning by the WRITE permission will no longer provided when targeting Android 10 and above

android/app/src/main/AndroidManifest.xml
huynguyennovem: No, you can remove it

android/app/src/main/AndroidManifest.xml
hoangdat: ah, you mean, because we target to Android 10 and above, so we need to set this flag to avoid warning (this still need for Android before 10).

android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,9 @@
+<resources>
+    <string name="app_name">TMail</string>
```
hoangdat, line 2: please update with Team Mail.

android/app/src/main/AndroidManifest.xml
huynguyennovem: Right. It will need for Android 9 and below. From Android 10 and above, it does not need

huynguyennovem: Right. The flag means that the permission will need for Android 9 and below. From Android 10 and above, it does not need

pubspec.yaml
```diff
@@ -95,6 +95,24 @@ dependencies:
...
   uuid: 3.0.4
 
+  # flutter_downloader
+  flutter_downloader: 1.7.0
+
+  # external_path
+  external_path: 1.0.1
+
+  # path_provider
+  path_provider: 2.0.3
```
hoangdat, line 105: this is for what?

core/lib/data/utils/device_manager.dart
```diff
@@ -0,0 +1,39 @@
...
+    }
+    return deviceUUID;
+  }
+
+  String getPlatformString() {
+    if (Platform.isIOS) {
+      return 'IOS-App';
+    }
+    return 'Android-App';
+  }
```
hoangdat, line 33: At the moment, we don't need these method

chibenwa:
> It quite not good UX, email is not a file sharing tool, it is focus to the content with other information. So that, IMO, attachment will be below body. (user need to understand the context before access the attachment)

@hoangdat you propose the exact same design than what was already done on OpenPaaS. Long story short: this behaviour was changed: the attachments are now between the headers and the body, folded if needed and no longer at the bottom.

Why? I magine a looong thread (hundred successive replies), it takes 20+ scrolls to go and see the attachment at the bottom and the UX becomes complete crap.

Do as I say, I have my reasons, I have the experience, and I can avoid us the common traps that I did end up falling into in the past.

**Put attachments between headers and body, folded if need be.**

pubspec.yaml
dab246: To get directory path on ios devvice

hoangdat: ah, okie, I remember

hoangdat:
> > It quite not good UX, email is not a file sharing tool, it is focus to the content with other information. So that, IMO, attachment will be below body. (user need to understand the context before access the attachment)
> 
> @hoangdat you propose the exact same design than what was already done on OpenPaaS. Long story short: this behaviour was changed: the attachments are now between the headers and the body, folded if needed and no longer at the bottom.
> 
> Why? I magine a looong thread (hundred successive replies), it takes 20+ scrolls to go and see the attachment at the bottom and the UX becomes complete crap.
> 
> Do as I say, I have my reasons, I have the experience, and I can avoid us the common traps that I did end up falling into in the past.
> 
> **Put attachments between headers and body, folded if need be.**

okie, I have checked the email content again, I agree with your point. We will:
```
Put attachments between headers and body and default to collapse, expand only when user want to do it.
```
@dieptran88 please help me make the mockup, @dab246 please fix it.

chibenwa:
default to collapse **When there is too much attachments.

Please do not collapse 1-2 attachments

dieptran88:
https://app.zenhub.com/workspaces/tmail-flutter-60d039db6ef62a0013432c14/issues/linagora/tmail-flutter/81
Story updated with mockup

hoangdat:
@dab246 please share the video clip to open discussion.

hoangdat:
Cool for the attachment list, @chibenwa and @dieptran88 please check it in the video demo

chibenwa:
> Cool for the attachment list, @chibenwa and @dieptran88 please check it in the video demo

Perfect!

It looks awesome, very nice initiatives @dab246 for the design, I love it!

Maybe you could share some screenshots of tablet mode too?

dab246:
> > Cool for the attachment list, @chibenwa and @dieptran88 please check it in the video demo
> 
> Perfect!
> 
> It looks awesome, very nice initiatives @dab246 for the design, I love it!
> 
> Maybe you could share some screenshots of tablet mode too?

Gifts for you: 

https://user-images.githubusercontent.com/80730648/133870177-8c7fdfa7-60bf-4e3e-9ccf-3383d6015107.mov

chibenwa:
Awesome!

hoangdat (approved):

Issue #123, opened 2021-09-24
Caching Mailbox


hoangdat:
### Desc:
- State should be store in persistent storage `SharedPreference` or DB
```
StateDao {
   final StateType type;
   final State;
}

enum StateType {
   Mailbox,
   Email
}
```

- Mailbox should be store in a `table(box)`: `mailboxes`

- Each mailbox have to store all the properties with `lastOpen` property: 
```
  final MailboxId id;
  final MailboxName? name;
  final MailboxId? parentId;
  final Role? role;
  final SortOrder? sortOrder;
  final TotalEmails? totalEmails;
  final UnreadEmails? unreadEmails;
  final TotalThreads? totalThreads;
  final UnreadThreads? unreadThreads;
  final MailboxRights? myRights;
  final IsSubscribed? isSubscribed;
  final Date? lastOpened
```
- With `MailboxRepository` we need to implement 2 method:
```
abstract class MailboxRepository {
  Stream<List<Mailbox>> getAllMailbox(AccountId accountId, {Properties? properties}) {
      final localList = local.getAll();  
      yeild local;  
      if (localList > 0) {
          final changes = network.changes();
          await asyncUpdateDb(changes);
          yield local.getAll();
      } else {
          final news = network.getAll();
          await asyncUpdateDb(news);
          yeild local.getAll();
      }
  }
  
  Future<List<Maibox>> refresh(AccountId accountId, State currentState) {
      final changes = network.changes();
      await asyncUpdateDb(changes);
      yield local.getAll();
  }
}
```

- `MailboxAPI` need to be modified:
```
class MailboxAPI {
     Future<GetMailboxResponse> getAllMailbox(AccountId accountId, {Properties? properties});

     Future<MailboxChangeResponse> getChanges(AccountId account, State lastState);
}

class MailboxChangeResponse {
  List<Mailbox> updated;
  List<Mailbox> created;
  List<Mailbox> destroyed;
  State newState;
}
```

- Update db: 
```
class HiveMailboxDataSource extends MailboxDataSource {
    //crud
    insert;
    get;
    update;
    delete;
    Future<void> asyncUpdateDb();
}
```

hoangdat:
PR: #127

PR #137, opened 2021-09-30
Writing test for mailbox cache and get all mailbox interactor

hoangdat:
@dab246 please review it


test/features/caching/mailbox_cache_client_test.dart
```diff
@@ -0,0 +1,181 @@
...
+    HiveCacheConfig().setUp(cachePath: Directory.current.path);
+  });
+
+  setUp(() {
+    _mailboxCacheClient = MailboxCacheClient();
+  });
+
+  group('[delete]', () {
+    test('cache should delete item successfully when cache empty', () async {
+      await _mailboxCacheClient.deleteItem(MailboxFixtures.inboxMailbox.id.toString());
```
dab246, line 24: This is a bit absurd. There is no data in the database so you can delete it and then compare again. I think in this case you are always right to leave this out,

test/features/caching/mailbox_cache_client_test.dart
```diff
@@ -0,0 +1,181 @@
...
+      await _mailboxCacheClient.insertItem(
+        MailboxFixtures.sentMailbox.id.toString(),
+        MailboxFixtures.sentMailbox.toMailboxCache());
+
+      await _mailboxCacheClient.deleteItem(MailboxFixtures.inboxMailbox.id.toString());
+
+      final remainingItems = await _mailboxCacheClient.getAll();
+
+      expect(remainingItems.length, equals(1));
+      expect(remainingItems[0], equals(MailboxFixtures.sentMailbox.toMailboxCache()));
```
dab246, line 67: IMO It is advisable to check that `remainingItems` still contains an `inboxMailbox` that is not correct in the meaning of this case.

test/features/caching/mailbox_cache_client_test.dart
```diff
@@ -0,0 +1,181 @@
...
+          MailboxFixtures.sentMailbox.id.toString(),
+          MailboxFixtures.sentMailbox.toMailboxCache());
+
+      await _mailboxCacheClient.deleteItem(MailboxFixtures.inboxMailbox.id.toString());
+      await _mailboxCacheClient.deleteItem(MailboxFixtures.inboxMailbox.id.toString());
+
+      final remainingItems = await _mailboxCacheClient.getAll();
+
+      expect(remainingItems.length, equals(1));
+      expect(remainingItems[0], equals(MailboxFixtures.sentMailbox.toMailboxCache()));
```
dab246, line 85: I find this case you test a bit ridiculous. You have to compare it with an exception.

test/features/caching/mailbox_cache_client_test.dart
```diff
@@ -0,0 +1,181 @@
...
+          MailboxFixtures.sentMailbox.toMailboxCache());
+
+      await _mailboxCacheClient.updateItem(
+          MailboxFixtures.inboxMailbox.id.toString(),
+          MailboxFixtures.folder1.toMailboxCache());
+
+      final remainingItems = await _mailboxCacheClient.getAll();
+
+      expect(remainingItems.length, equals(2));
+      expect(
```
dab246, line 169: You should compare the updated value, not like that.

hoangdat:
cherry to #123

PR #149, opened 2021-10-05
[ADR] Caching mailbox and email

hoangdat:
@chibenwa @dab246 please help me review this ADR

### DIFF
--- a/docs/adr/0002-caching.md
+++ b/docs/adr/0002-caching.md
@@ -1,0 +1,28 @@
+# 2. Caching
+
+Date: 2021-10-05
+
+## Status
+
+Accepted
+
+## Context
+
+- Client always make requests to back-end to refresh the data, what put a lot of pressure on it.
+- User need to wait a long time to see the mailbox and email.
+
+## Decision
+
+- Caching the data with the [Hive](https://github.com/hivedb/hive) dependencies 
+Hive is a lightweight and blazing fast key-value database written in pure Dart. It support cross-platform: mobile, desktop and browser
+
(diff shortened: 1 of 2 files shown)


docs/adr/0002-caching.md
```diff
@@ -0,0 +1,24 @@
...
+Date: 2021-10-05
+
+## Status
+
+Accepted
+
+## Context
+
+- Client always make requests to back-end to refresh the data
+- User need to wait a long time to see the mailbox and email
```
chibenwa, line 12: Always requesting the same entities again and over again put a lot of pressure on the backend.

docs/adr/0002-caching.md
```diff
@@ -0,0 +1,24 @@
...
+
+- Caching the data with the [Hive](https://github.com/hivedb/hive) dependencies 
+Hive is a lightweight and blazing fast key-value database written in pure Dart. It support cross-platform: mobile, desktop and browser
+
+- `changes` method allows a client to efficiently update the state of data(Mailbox and Email)
+
+## Consequences
+
+- Minimizing the requests to the back-end 
+- Reducing the load latency in the client
```
chibenwa, line 24:
```suggestion
- Reducing the latency in the client
```

chibenwa: Also what would we need to handle offline gracefully? Pre-loading the bodies of the 10 first messages?

chibenwa: We can have a quick note here. IE in a `Follow up work`.

dab246 (approved):

chibenwa (approved):

PR #150, opened 2021-10-06
TF-105 Fix html not reading well

dab246:
- [x] Show better message content on all device.
- [ ] Display inline image has `cid` 

https://user-images.githubusercontent.com/80730648/136134482-a52cfb05-c393-43b7-95a2-ef00888e6064.mov

https://user-images.githubusercontent.com/80730648/136134730-f86d6f82-faef-4ebf-bcef-c02bf6dfe368.mov


dab246:
> Please care about:
> 
> ```
> Because every version of html_editor_enhanced from git depends on flutter_inappwebview from hosted and every version of core from path depends on flutter_inappwebview from git, html_editor_enhanced from git is incompatible with core from path.
> ```

I already know this. We can replace it to sync with htmlcontentviewer

dab246:
@hoangdat  Build successful ! Please review it

hoangdat:
Can not scroll in composer

hoangdat:
Some image is scaled not in ratio

dab246:
> Link should be expanded within the width of screen

Please provide specific pictures and videos of the error found.

PR #206, opened 2021-10-21
Add icon and create splash screen for iOS

dab246:
https://user-images.githubusercontent.com/80730648/138203866-644a4618-e17e-4a47-8afd-5a4032ae6af0.mov

@hoangdat  please review


hoangdat:
Did you try on tablet? We have any aliased in the bigger screen for splash screen? No way to use svg for splash screen in iOS?

PR #219, opened 2021-10-22
[BUG] Fix upload attachment

dab246:
https://user-images.githubusercontent.com/80730648/151537034-a2df591f-8523-4c27-abdb-dae265465bab.mov

### DIFF
--- a/lib/features/composer/data/network/composer_api.dart
+++ b/lib/features/composer/data/network/composer_api.dart
@@ -1,7 +1,9 @@
 
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:core/core.dart';
+import 'package:dio/dio.dart';
 import 'package:model/model.dart';
 
 class ComposerAPI {
@@ -11,9 +13,15 @@ class ComposerAPI {
   ComposerAPI(this._dioClient);
 
   Future<UploadResponse> uploadAttachment(UploadRequest uploadRequest) async {
+    final file = File(uploadRequest.fileInfo.filePath);
+    final headerParam = _dioClient.getHeaders();
+    headerParam[HttpHeaders.contentTypeHeader] = uploadRequest.fileInfo.mimeType;
+    headerParam[HttpHeaders.contentLengthHeader] = file.readAsBytesSync().length;
+
     final resultJson = await _dioClient.post(
         Uri.decodeFull(uploadRequest.uploadUrl.toString()),
-        data: File(uploadRequest.fileInfo.filePath).readAsBytesSync());
-    return UploadResponse.fromJson(resultJson);
+        options: Options(headers: headerParam),
+        data: file.openRead());
+    return UploadResponse.fromJson(jsonDecode(resultJson));
   }
 }
--- a/model/lib/upload/file_info.dart
+++ b/model/lib/upload/file_info.dart
@@ -1,5 +1,6 @@
 
 import 'package:equatable/equatable.dart';
+import 'package:mime/mime.dart';
 
 class FileInfo with EquatableMixin {
   final String fileName;
@@ -11,6 +12,8 @@ class FileInfo with EquatableMixin {
   factory FileInfo.empty() {
     return FileInfo('', '', 0);
   }
+
+  String get mimeType => lookupMimeType(filePath) ?? 'application/json; charset=UTF-8';
 
   @override
   List<Object> get props => [fileName, filePath, fileSize];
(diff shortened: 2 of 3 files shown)


dab246:
@chibenwa @hoangdat  Please check the upload api again for me. The uploaded file cannot be opened. See details in the video below.

https://user-images.githubusercontent.com/80730648/150954727-d2340a92-412c-491b-a670-6091afed181f.mov

chibenwa:
mmm Why are we having a Content/type jpeg on the response in the first place?

Uploads are not supposed to return the uploaded image but a short JSON with the blob id, and other information related to the uploaded blob.

chibenwa:
cf https://github.com/apache/james-project/pull/862

dab246:
> cf [apache/james-project#862](https://github.com/apache/james-project/pull/862)

@chibenwa Has this bug been fixed yet? I still can't do it successfully.

chibenwa:
1. No it's in progress

2. If you just ignore the (wrong) content type and pretend it is JSON (which it is!) then there is no blocker at all!

chibenwa:
Try to cat the response instead, to see the underlying JSON.

dab246:
> Try to cat the response instead, to see the underlying JSON.

I tried, but the file size returned in json after successful upload and the actual uploaded file size is different.

chibenwa:
Can you post the JSON result?

dab246:
> Can you post the JSON result?

@chibenwa  Can you explain this to me?

````
{
  "accountId": "3ce33c876a726662c627746eb9537a1d13c2338193ef27bd051a3ce5c0fe5b12",
  "blobId": "uploads-4baffc90-7e79-11ec-ac27-732dd96cc126",
  "type": "image/jpeg",
  "size": 586388
}
````

https://user-images.githubusercontent.com/80730648/151120942-dba610a6-43e7-411f-8150-cdb76c77c2e1.mov

chibenwa:
I'm having a look.

BTW @dab246 don't be impatient and take the time to clearly states your observations. You can structure your report in the following way: 1. what I observe 2. what I think is expected 3. useful complementary information. Just stating "your crap don't work, fix it" is not enough and thankfully people in the Quality Assurance area don't do that. It avoids other people not to guess, but jump straight to the diagnostic.

Can we agree on that?

Here is my take on your issue:

1. The size in the upload response do not match the size of the uploaded file
2. It should be the same
3. Pateform: OpenPaaS DEV

For yesterday 

1. The uploaded file is not returned in the upload response
2. It should 

We can agree that 2. is not spec compliant (and illogical) and that this confusion was caused by the content type of the response being mistakenly positionned to the content type of the uploaded object instead of JSON

Wich gives:

1. The upload response Content Type is of the type of the uploaded object
2. This should be application/json

And we have already a PR to fix that.

dab246:
> I'm having a look.
> 
> BTW @dab246 don't be impatient and take the time to clearly states your observations. You can structure your report in the following way: 1. what I observe 2. what I think is expected 3. useful complementary information. Just stating "your crap don't work, fix it" is not enough and thankfully people in the Quality Assurance area don't do that. It avoids other people not to guess, but jump straight to the diagnostic.
> 
> Can we agree on that?
> 
> Here is my take on your issue:
> 
> 1. The size in the upload response do not match the size of the uploaded file
> 2. It should be the same
> 3. Pateform: OpenPaaS DEV
> 
> For yesterday
> 
> 1. The uploaded file is not returned in the upload response
> 2. It should
> 
> We can agree that 2. is not spec compliant (and illogical) and that this confusion was caused by the content type of the response being mistakenly positionned to the content type of the uploaded object instead of JSON
> 
> Wich gives:
> 
> 1. The upload response Content Type is of the type of the uploaded object
> 2. This should be application/json
> 
> And we have already a PR to fix that.

**1. Please send the PR link that you are fixing that error. I need to see.
2. The problem here I want to know is that the error is caused by the back-end side? I just need to know that.**

chibenwa:
> 1. Please send the PR link that you are fixing that error. I need to see.

cf apache/james-project#862

(earlier in this discussion)

> 2. The problem here I want to know is that the error is caused by the back-end side? I just need to know that.

I'm investigating that since yesterday.

I don't succeed reproducing running James alone. If that's a James code issue I dont have enough details to reproduce it.

It *may* be a deployment issue. Like a NGinx cutting things in the middle. We shall continue to investigate.

hoangdat:
@dab246 can you help Benoit to reproduce this issue.

(maybe the headers, content, ....)

hoangdat (approved):

PR #224, opened 2021-11-09
Upgrade Flutter sdk version to 2.5.3

Arsnael:
Took a bit of time today to try myself on Flutter :)

I guess I expect this to fail on the CI, as it should use the 2.2.3 version of the sdk :)

I did the work to upgrade to flutter sdk 2.5.3 (using dart sdk 2.14) and I upgraded some dart deps as well. 

Some dart deps bumps are depending on dart 2.14 which made me bump the android compileSdkVersion to 31 as well.

There is a few other deps I tried to upgrade, but I got conflicts with jmap_dart_client. How do you deal with deps update synced between your main project and submodules?

@hoangdat Well let me know if it interests you or not, or if there is more fixes needed to it. I could compile and launch the app in an emulated android device but I didn't test much, I'm not sure if it generated some issues or not.

### DIFF
--- a/model/pubspec.yaml
+++ b/model/pubspec.yaml
@@ -20,7 +20,7 @@ publish_to: none
 version: 1.0.0+1
 
 environment:
-  sdk: ">=2.12.0 <3.0.0"
+  sdk: ">=2.14.0 <3.0.0"
 
 dependencies:
   flutter:
--- a/model/pubspec.yaml
+++ b/model/pubspec.yaml
@@ -63,7 +63,7 @@ dev_dependencies:
 
   build_runner: 2.1.5
 
-  json_serializable: 4.1.3
+  json_serializable: 6.0.1
 # For information on the generic Dart part of this file, see the
 # following page: https://dart.dev/tools/pub/pubspec
 
--- a/model/pubspec.yaml
+++ b/model/pubspec.yaml
@@ -45,8 +45,8 @@ dependencies:
   # jmap_dart_client
   jmap_dart_client:
     git:
-      url: git://github.com/linagora/jmap-dart-client.git
-      ref: master
+      url: git://github.com/Arsnael/jmap-dart-client.git
+      ref: upgrade-flutter-version
 
   # uri
   uri: 1.0.0
--- a/model/pubspec.yaml
+++ b/model/pubspec.yaml
@@ -37,7 +37,7 @@ dependencies:
   equatable: 2.0.3
 
   # quiver
-  quiver: 3.0.1
+  quiver: 3.0.1+1
 
   # intl
   intl: 0.17.0
--- a/model/pubspec.yaml
+++ b/model/pubspec.yaml
@@ -31,7 +31,7 @@ dependencies:
 
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
-  cupertino_icons: ^1.0.2
+  cupertino_icons: ^1.0.4
 
   # equatable
   equatable: 2.0.3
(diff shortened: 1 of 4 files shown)


Arsnael:
I'm not gonna push on this for now... I realized that compared to the actual master where the debug didn't seem to be alerting me on issues, this one I can often logs like:

```
D/skia    ( 4809): Shader compilation error
D/skia    ( 4809): ------------------------
D/skia    ( 4809): Errors:
```

when launching the app and doing stuff around on the emulator. I'm not sure what it means but maybe it's more complex to upgrade than I thought, and I don't have the skills to get what's this about. I will let this PR around as it can still be useful but will not push further for now

chibenwa:
Shaders are scripts ran on a graphic card, for light tracing, color rendering, texture and the like.

Arsnael:
> Shaders are scripts ran on a graphic card, for light tracing, color rendering, texture and the like.

Yeah but I'm not too sure what it implies in that context...if it's the emulated device maybe with some issues or the compiled code.

Arsnael:
Might be just some emulator schenanigan after all... I get those errors on an other emulated device in the master branch so... maybe it's nothing

Arsnael:
After upgrading the jmap-dart-client I could upgrade all the other libs. Tested a bit around it seems working. I don't know about bumping up the version of the app though, but as it has some major changes (bump up sdks, android compile version, jmap-dart-client version) it feels it's necessary if this goes on...

Don't hesitate to feedback, I'm not sure at all about what I'm doing :D

pubspec.yaml
```diff
@@ -15,7 +15,7 @@ publish_to: 'none' # Remove this line if you wish to publish to pub.dev
 # In iOS, build-name is used as CFBundleShortVersionString while build-number used as CFBundleVersion.
 # Read more about iOS versioning at
 # https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html
-version: 1.0.0+1
+version: 1.1.0
```
hoangdat, line 18: no, please check the tags, that the file we change the version of app

pubspec.yaml
```diff
@@ -111,7 +111,7 @@ dependencies:
       ref: master
 
   # file_picker
-  file_picker: 4.2.1
+  file_picker: 4.2.2+1
```
hoangdat, line 114: I am not sure with it, so please don't change this until we can test it

pubspec.yaml
```diff
@@ -99,7 +99,7 @@ dependencies:
   device_info: 2.0.3
 
   # permission_handler
-  permission_handler: 8.1.6
+  permission_handler: 8.2.6
```
hoangdat, line 102: same here, we need to understand what is update with new version and carefully test in many OS versions

Arsnael:
Probably better to just not update this yet then? It seems to create too many concerns^^'

model/pubspec.yaml
```diff
@@ -37,22 +37,22 @@ dependencies:
...
   # intl
   intl: 0.17.0
 
   # jmap_dart_client
   jmap_dart_client:
     git:
-      url: git://github.com/linagora/jmap-dart-client.git
-      ref: master
+      url: git://github.com/Arsnael/jmap-dart-client.git
+      ref: upgrade-flutter-version
```
chibenwa, line 49: Should not merge as is ;-)

Arsnael: yes of course ^^

hoangdat:
because we upgrade to flutter 2.8.1

PR #225, opened 2021-11-11
Change label "username" to "Email" in login form

Arsnael:
That has been in some feedbacks that it was confusing.

I also saw that the text input type for the url in the login form was emailAddress, which is wrong (I think), so I changed it to url type.

email-label

### DIFF
--- a/lib/l10n/intl_en.arb
+++ b/lib/l10n/intl_en.arb
@@ -18,8 +18,8 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "username": "username",
-  "@username": {
+  "email": "email",
+  "@email": {
     "type": "text",
     "placeholders_order": [],
     "placeholders": {}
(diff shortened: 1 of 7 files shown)


lib/l10n/intl_en.arb
```diff
@@ -18,7 +18,7 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "username": "username",
+  "username": "Email",
```
hoangdat, line 21: why you not change the variable name `username` to `email` too?

lib/main/localizations/app_localizations.dart
```diff
@@ -35,7 +35,7 @@ class AppLocalizations {
   }
 
   String get username {
```
hoangdat, line 37: change this field to `email`

lib/l10n/intl_en.arb
Arsnael: Thought about it, but wasn't sure about all the links in the code... But it would make sense I agree, will give it a shot

hoangdat (approved):

Arsnael:
Squashed the fixup

PR #226, opened 2021-11-12
TF-220 Sort list of mailboxes by SortOrder before display

Arsnael:
Doing a sort using SortOrder on the list of mailboxes before display.

INBOX has always the lowest SortOrder, followed gradually by other system mailboxes and then finally the other folders. We should use the SortOrder in the jmap mailbox object, it's for this kind of purpose I think. 

I could reproduce the issue on preprod with user `firstname1.surname1`.

Master branch:
mailboxLIstMaster

This fix:
listMailboxesFix

### DIFF
--- a/lib/features/mailbox/data/datasource_impl/mailbox_cache_datasource_impl.dart
+++ b/lib/features/mailbox/data/datasource_impl/mailbox_cache_datasource_impl.dart
@@ -5,6 +5,7 @@ import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
 import 'package:tmail_ui_user/features/mailbox/data/datasource/mailbox_datasource.dart';
 import 'package:tmail_ui_user/features/mailbox/data/model/mailbox_change_response.dart';
 import 'package:tmail_ui_user/features/mailbox/data/local/mailbox_cache_manager.dart';
+import 'package:tmail_ui_user/features/mailbox/data/extensions/mailbox_extension.dart';
 import 'package:tmail_ui_user/features/mailbox/domain/model/mailbox_response.dart';
 
 class MailboxCacheDataSourceImpl extends MailboxDataSource {
@@ -35,7 +36,9 @@ class MailboxCacheDataSourceImpl extends MailboxDataSource {
   @override
   Future<List<Mailbox>> getAllMailboxCache() {
     return Future.sync(() async {
-      return await _mailboxCacheManager.getAllMailbox();
+      final listMailboxes = await _mailboxCacheManager.getAllMailbox();
+      listMailboxes.sort((mailboxA, mailboxB) => mailboxA.compareTo(mailboxB));
+      return listMailboxes;
     }).catchError((error) {
       throw error;
     });
--- a/lib/features/mailbox/data/extensions/mailbox_extension.dart
+++ b/lib/features/mailbox/data/extensions/mailbox_extension.dart
@@ -19,4 +19,8 @@ extension MailboxExtension on Mailbox {
       isSubscribed: isSubscribed?.value
     );
   }
+
+  int compareTo(Mailbox other) {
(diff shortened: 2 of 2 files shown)


lib/features/mailbox/data/datasource_impl/mailbox_cache_datasource_impl.dart
```diff
@@ -35,7 +35,9 @@ class MailboxCacheDataSourceImpl extends MailboxDataSource {
   @override
   Future<List<Mailbox>> getAllMailboxCache() {
     return Future.sync(() async {
-      return await _mailboxCacheManager.getAllMailbox();
+      final listMailboxes = await _mailboxCacheManager.getAllMailbox();
+      listMailboxes.sort((mailboxA, mailboxB) => mailboxA.sortOrder!.value.value.compareTo(mailboxB.sortOrder!.value.value));
```
Arsnael, line 39: There is maybe a better or cleaner way to do this? Like implementing Mailbox or SortOrder with the Comparable interface? (but that would require to modify the jmap-dart-client repo)

Or maybe better ideas? Not sure if this is the best place to do that as well as it means we sort everytime we take the list of mailboxes from the cache, but looked like the easiest.

Well let me know

hoangdat: please try with `extensions`, may be you should not modify on dependencies :)

chibenwa (approved):

lib/features/mailbox/data/datasource_impl/mailbox_cache_datasource_impl.dart
chibenwa: Shouldn't the cache store the mailbox sorted on the first place?

Arsnael: > please try with extensions, may be you should not modify on dependencies :)

I'm not modifying on dependencies here, that's tmail-flutter code. What do you mean by trying with `extensions`?

> Shouldn't the cache store the mailbox sorted on the first place?

Yeah sounds more logical, but seemed a bit more complicated than just sorting the list when getting it (you have different cases, the getAllMailboxes and /changes). But I could spend more time and try

hoangdat: with current db `Hive` maybe we can not store `sorted`. We stored by key and value only

> What do you mean by trying with extensions?

With `extension`, maybe we can implement Mailbox or Sorter with Compare interface without change the code of dependencies.

Arsnael: oooooh you mean this? https://dart.dev/guides/language/extension-methods

Nice suggestion, didn't know it existed (I still don't know much about Dart), that sounds like a very cool thing :D I will try^^

Arsnael: so the "cache" is an embedded db Hive then?

hoangdat: not all, but this cache, we depend on Hive

Arsnael: I tried with `extension`, let me know :)

Arsnael:
@hoangdat a final review on this please? :)

hoangdat (approved):

hoangdat:
> @hoangdat a final review on this please? :)

oh srr, I forgot to submit the approval from that day :(

Arsnael:
> oh srr, I forgot to submit the approval from that day :(

No worries ;)

hoangdat:
@Arsnael plese rebase it, I will merge

Arsnael:
Squashed :)

PR #262, opened 2022-02-15
TF-255 Read Email in Browser

dab246:
https://user-images.githubusercontent.com/80730648/154023490-8d76dedd-f4ac-4809-8e72-d0d99c0a1636.mov


hoangdat:
@dab246 How about other special email body? `easywebview` support it?

dab246:
> 

It supports all html formats. I have checked most of the special emails in TMail it works fine.

hoangdat (approved):

PR #277, opened 2022-02-21
Start-up web-app

hoangdat:
@Arsnael @chibenwa @dab246 @tnstung  Please help me review and support us on it. We are stuck in
```
 => [ 2/13] RUN apt-get update                                                                                                                                                                                     6.9s
 => [ 3/13] RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libglu1-mesa fonts-droid-fallback python3                                                                                                 67.3s 
 => [ 4/13] RUN apt-get clean                                                                                                                                                                                      0.2s 
 => ERROR [ 5/13] RUN curl -L https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_2.10.0-stable.tar.xz | tar -C /opt -xJ       
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,15 +6,16 @@ RUN apt-get update
 RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libglu1-mesa fonts-droid-fallback python3
 RUN apt-get clean
 
-# Clone the flutter repo
-RUN git clone https://github.com/flutter/flutter.git /usr/local/flutter
+# Install flutter
+RUN curl -L https://storage.googleapis.com/flutter_infra_release/releases/stable/linux/flutter_linux_2.10.0-stable.tar.xz | tar -C /opt -xJ
 
 # Set flutter path
 # RUN /usr/local/flutter/bin/flutter doctor -v
-ENV PATH="/usr/local/flutter/bin:/usr/local/flutter/bin/cache/dart-sdk/bin:${PATH}"
+ENV PATH="/opt/flutter/bin:/opt/flutter/bin/cache/dart-sdk/bin:${PATH}"
 
 # Run flutter doctor
 RUN flutter doctor -v
+
 # Enable flutter web
 RUN flutter channel master
 RUN flutter upgrade
@@ -24,11 +25,9 @@ RUN flutter config --enable-web
 RUN mkdir /app/
 COPY . /app/
 WORKDIR /app/
-RUN flutter build web
 
 # Record the exposed port
 EXPOSE 3000
 
-# Stage 2 - Create the run-time image
-FROM nginx:1.21.1-alpine
-COPY --from=build-env /app/build/web /usr/share/nginx/html
+RUN ["chmod", "+x", "/app/server/startup.sh"]
+ENTRYPOINT [ "/app/server/startup.sh"]
--- a/server/startup.sh
+++ b/server/startup.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+PORT=3000
+
+echo pwd
+cd core
+flutter pub get
+
+cd ../model
+flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs
+
+cd ..
+flutter pub get && flutter pub run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart
+
+flutter pub get && flutter pub run intl_generator:generate_from_arb --output-dir=lib/l10n --no-use-deferred-loading lib/main/localizations/app_localizations.dart lib/l10n/intl*.arb
+
+flutter build web --dart-define=SERVER_URL='https://dev.open-paas.org'
+
+echo 'preparing port' $PORT '...'
+fuser -k 3000/tcp
+
+cd build/web/
+
+# Start the server
+echo 'Server starting on port' $PORT '...'
+python3 -m http.server $PORT


server/startup.sh
```diff
@@ -0,0 +1,26 @@
...
+
+cd ../model
+flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs
+
+cd ..
+flutter pub get && flutter pub run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart
+
+flutter pub get && flutter pub run intl_generator:generate_from_arb --output-dir=lib/l10n --no-use-deferred-loading lib/main/localizations/app_localizations.dart lib/l10n/intl*.arb
+
+flutter build web --dart-define=SERVER_URL='https://dev.open-paas.org'
```
Arsnael, line 17: That's an other matter but we shouldn't hardcode the server url :)

server/startup.sh
```diff
@@ -0,0 +1,26 @@
...
+
+echo 'preparing port' $PORT '...'
+fuser -k 3000/tcp
+
+cd build/web/
+
+# Start the server
+echo 'Server starting on port' $PORT '...'
+python3 -m http.server $PORT
```
Arsnael, line 26: urg python, really? T_T

Arsnael:
Docker does not like your curl command for sure ^^' not sure why though, I will try to take the time tomorrow to look at it a bit more into details ;)

server/startup.sh
hoangdat: please help me fix up :(

server/startup.sh
hoangdat: yes, I am finding the way to use env file instead

server/startup.sh
chibenwa: If we are just serving static content we may consider something like nginx

server/startup.sh
chibenwa: Building the flutter web app every time you start the SDK is not a best practice.

We should build the web app outside of docker, then only import the resulting files within docker.

hoangdat: > We should build the web app outside of docker

How to do it?

Arsnael: I'm not sure what the target looks like after being built honestly... might not be that easy. We need to check a bit deeper I think before taking this kind of decision

hoangdat: How about, we build it on CI/CD, then import it and build docker image in CI/CD, then publish it to linagora docker hub?

hoangdat: @tnstung Can you give me some idea for this?

server/startup.sh
Arsnael: Me it's more like I hate Python :D After it depends how flutter build this for webapp content, I have no idea

hoangdat: For the web server, no constrain with `flutter build`. I chose `python3` because it is in tutorial, that's all. Please guide me the better option for our application.

Arsnael: https://docs.flutter.dev/deployment/web#building-the-app-for-release that's the tuto right? Yeah i'm rusty in frontend webapp but let's not use python :) let's check something else

Arsnael: https://rebar-ahmad.medium.com/env-aware-flutter-apps-in-docker-b899d0deebfd

The same but with nginx :)

server/startup.sh
Arsnael: You can inject env variables when running a docker container btw (the link I posted in the other comment, you can see it too: https://rebar-ahmad.medium.com/env-aware-flutter-apps-in-docker-b899d0deebfd)

I stumbled upon this flutter package also if you prefer using env file: https://pub.dev/packages/flutter_dotenv

From all the articles and doc I could read though, it's obvious that the env variable injection is only possible during the build phase. It seems the best way if you want your docker image to be env custom is to build it inside the image. 

@chibenwa but is that really an issue? Maybe not best practice, but with this limitation on flutter webapp, this might still be the best choice.

It is easier to maintain like this than changing constantly your CI conf depending on how many environments you want to serve. If you get multiples customers with different domains in the future, it could become very chaotic and error prone to maintain for example

Dockerfile
```diff
@@ -6,15 +6,16 @@ RUN apt-get update
...
 # RUN /usr/local/flutter/bin/flutter doctor -v
-ENV PATH="/usr/local/flutter/bin:/usr/local/flutter/bin/cache/dart-sdk/bin:${PATH}"
+ENV PATH="/opt/flutter/bin:/opt/flutter/bin/cache/dart-sdk/bin:${PATH}"
 
 # Run flutter doctor
 RUN flutter doctor -v
+
 # Enable flutter web
 RUN flutter channel master
 RUN flutter upgrade
```
Arsnael, line 21: just noticed this... that's gonna update your sdk to the latest :) (ok it's 2.10 now but in a near future..^^)

Arsnael:
@hoangdat played a bit around this afternoon with it :) I pushed a branch on my repo, to not disturb your work : https://github.com/Arsnael/tmail-flutter/commit/22989dc6eb3c785a90855c712dab7d55ea497aa1

Build: 
```
docker build -t tmail-web:latest .
```

Run: 
```
docker run -ti -d -p 8080:80 -e SERVER_URL='[JMAP_URL]' --name web tmail-web:latest
```
replace [JMAP_URL] with well dev, or preprod... Go to http://localhost:8080 I tested with preprod and could connect with firstname1.surname1 account. Could view mailboxes, emails (I didn't test further) :) 

It's not optimized, I think we should think how to do better but... it works, it's a good first step and should help greatly :)

So: 
* I use nginx 
* it's better  to build the docker image with the flutter deps being cleaned up before ('flutter clean`). As we copy the all folder to the image, it's lighter that way and we recompile everything anyway (for now). Even with that the image is still 2.5GB... But before I did that it was 3.4GB sth like that
* I also moved out of the script the prebuild steps. We only want the `flutter build web` in the script as it's where we inject the env variable. Anything before can be done when building the image I think

=> takes around 1 minute on my local before nginx serves the app

What I think could be interesting to investigate:
-> https://pub.dev/packages/flutter_dotenv
Allows to load env file on runtime. I'm not sure how it works with the web, it's not clear. I think it can be interesting to test and see if we can make it work with this module and that there is a fix path on this env filewhen running the app with nginx. So we can load the env file through a volume on docker. I'm not sure how it works, but worth to check IMO

server/startup.sh
chibenwa: Fair.

But then should't we consider multi stage builds? https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

Rationals: docker images should not be big for convenience and security resons.

We could have a docker file that first build the static assets then just copy the build result to a nice minimal nginx image.

chibenwa:
How long the build takes? Thus how long does it take to start this container? How big in MB the image is?

Did someone performed a security review of the resulting docker container, for instance using something like Snyk.

I really think we would better be served by a multi-staged image build and no build at runtime.

chibenwa:
Sure but you understand we can't accept a 2.5GB frontend that needs 1 minute to boot and nails the k8s cluster down doing so.

Arsnael:
> Sure but you understand we can't accept a 2.5GB frontend that needs 1 minute to boot and nails the k8s cluster down doing so.

Thus why I suggested to investigate https://pub.dev/packages/flutter_dotenv as it might maybe solve this :) (note I say maybe^^)

server/startup.sh
Arsnael: Forgot the multi stage build but yeah very good suggestion!

Arsnael:
@chibenwa @hoangdat I moved the discussion to my parallel PR on this work: https://github.com/linagora/tmail-flutter/pull/279

hoangdat:
> @chibenwa @hoangdat I moved the discussion to my parallel PR on this work: #279

IMO, we can close this PR. I saw a big improvement from your #279

PR #278, opened 2022-02-21
TF-82 Save an email as drafts

dab246:
https://user-images.githubusercontent.com/80730648/154935640-9da4cd33-6a4d-43e6-9990-bb7a8f64f2d0.mp4

### DIFF
--- a/core/lib/core.dart
+++ b/core/lib/core.dart
@@ -7,6 +7,8 @@ export 'presentation/extensions/capitalize_extension.dart';
 export 'presentation/extensions/list_extensions.dart';
 export 'domain/extensions/datetime_extension.dart';
 export 'presentation/extensions/html_extension.dart';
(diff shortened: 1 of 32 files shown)


chibenwa:
Looks good to me!

One trap with drafts I saw before was open the composer and immediatly close it. In such case, in the absence of user input, not draft shall be created.

Also, do we support drafts with attachments?

dab246:
> Looks good to me!
> 
> One trap with drafts I saw before was open the composer and immediatly close it. In such case, in the absence of user input, not draft shall be created.
> 
> Also, do we support drafts with attachments?

1. Every time a user opens an editor, the user can enter or not enter information and closing it will automatically create a draft.
2. We have draft support with attachments

chibenwa:
If the user did not input anything please don't save an empty draft.

Please add the logic not to do that.

hoangdat (approved):

PR #279, opened 2022-02-22
Startup web mail

Arsnael:
Parallel work of #277

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,44 @@
 #Stage 1 - Install dependencies and build the app
 FROM debian:latest AS build-env
 
-# Install flutter dependencies
-RUN apt-get update
-RUN apt-get install -y curl git wget unzip libgconf-2-4 gdb libglu1-mesa fonts-droid-fallback python3
-RUN apt-get clean
+ENV FLUTTER_CHANNEL="stable"
+ENV FLUTTER_VERSION="2.10.0"
+ENV FLUTTER_URL="https://storage.googleapis.com/flutter_infra_release/releases/$FLUTTER_CHANNEL/linux/flutter_linux_$FLUTTER_VERSION-$FLUTTER_CHANNEL.tar.xz"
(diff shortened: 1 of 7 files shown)


Arsnael:
So... Now using the package dot_env (so using a `env.file` at the root here), multi-staged build, first stage we build the webapps, then we copy the files into a nginx container. 

At the end, final image is 170mb and of course runs right away (just about starting nginx).

Build:
```
docker build -t tmail-web:latest .
```

Before building or you can leave the env.file to  the default value (http://localhost) or you can put directly the value you want (like jmap preprod url for example). Will just change the way you run the container.

Run (if put the wanted value in env.file when building):
```
docker run -d -ti -p 8080:80 --name web tmail-web:latest 
```

or create a file env.preprod.file with the correct value for SERVER_URL and mount the file to `/usr/share/nginx/html/assets/env.file` when running the container, like this:
```
docker run -d -ti -p 8080:80 --mount type=bind,source="$(pwd)"/env.preprod.file,target=/usr/share/nginx/html/assets/env.file --name web tmail-web:latest 
```

It works. 

A little warning though, I noticed by checking the network requests done through the dev console... the env file is being asked, so the client can see it clearly. So for now we have the jmap server url, which is not a secret as you need to fill it on the mobile app anyway, so maybe okish? But shouldn't put sensitive creds in that file for sure!!!

@chibenwa I know you gonna tell me you probably prefer getting rid of the multi stage build, like you build in local and then the dockerfile just copies what you built into the nginx image... Well I don't know, I find the build a bit complex (like numerous commands before getting your web stuff). It's easier for someone not very knowledgeable to run a docker build without a care in the world... I don't know how is the CI building stuff exactly too. Well I guess this last point is up to discussion :)

Anyway, happy with the result though :)

chibenwa (approved):

hoangdat (approved):

hoangdat (approved):

chibenwa (approved):

PR #291, opened 2022-02-25
Add a Jenkinsfile for web app docker build and push

Arsnael:
I think it should work but we still need to check after merge.

Indeed, Jenkins is expecting a Jenkinsfile being present in the branch it's trying to build. So we need to merge this on master and also on release if you want both docker builds and pushes for tmail flutter webapp :)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,41 @@
+pipeline {
+  agent any
+
+  options {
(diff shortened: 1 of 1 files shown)


hoangdat:
IMO, it should be better if we have the version in image name :)

chibenwa (approved):

PR #299, opened 2022-03-01
Add hint text to avoid show null in editor

hoangdat:
it is not a crash on composer. if  `initialText` is empty, editor will put `hint` on.

https://github.com/tneotia/html-editor-enhanced/blob/31406a4c0d6df0e9e8833057828dd8bf116df355/lib/src/widgets/html_editor_widget_web.dart#L212
```
            placeholder: "${widget.htmlEditorOptions.hint}",
```

image

@dab246 please review


dab246 (approved):
/lgtm

PR #301, opened 2022-03-02
Config firebase

hoangdat:
- [x] config Android to work with `firebase`
- [x] config iOS to work with `firebase`
  - https://firebase.flutter.dev/docs/messaging/apple-integration/

### DIFF
--- a/ios/Runner/Info.plist
+++ b/ios/Runner/Info.plist
@@ -22,16 +22,21 @@
 	<string>$(FLUTTER_BUILD_NUMBER)</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSContactsUsageDescription</key>
+	<string>To send email with these contacts</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>To save photo to library, we need permission to access your photo library.</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>To select and upload photo, we need permission to access your photo library.</string>
+	<key>UIBackgroundModes</key>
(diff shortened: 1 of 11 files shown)


pubspec.yaml
```diff
@@ -130,6 +130,7 @@ dependencies:
   # fk_user_agent
   fk_user_agent: 2.1.0
 
+  firebase_analytics: 9.1.2
```
hoangdat, line 133: why we should add analytic also?

dab246: I think there should be so we can see the log of events and actions from firebase. This firebase library is also recommended.

dab246:
@hoangdat  Please test it on real iOS device

ios/Runner/Info.plist
```diff
@@ -22,16 +22,21 @@
 	<string>$(FLUTTER_BUILD_NUMBER)</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
+	<key>NSContactsUsageDescription</key>
+	<string>To send email with these contacts</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>To save photo to library, we need permission to access your photo library.</string>
+	<key>NSPhotoLibraryUsageDescription</key>
+	<string>To select and upload photo, we need permission to access your photo library.</string>
+	<key>UIBackgroundModes</key>
```
hoangdat, line 31: Did you forget to enable `Push notification`?

dab246: @@ it's enabled but it doesn't need to be declared here

PR #303, opened 2022-03-03
TF-296 Support upload download file on browser

dab246:
https://user-images.githubusercontent.com/80730648/156513006-05a1ccf5-2eb3-41c9-82ac-0c26112a49b6.mov

### DIFF
--- a/env.file
+++ b/env.file
@@ -1,1 +1,1 @@
-SERVER_URL=https://dev.open-paas.org
+SERVER_URL=http://localhost
--- a/env.file
+++ b/env.file
@@ -1,1 +1,1 @@
-SERVER_URL=http://localhost
+SERVER_URL=https://dev.open-paas.org
--- a/lib/features/email/presentation/email_controller.dart
+++ b/lib/features/email/presentation/email_controller.dart
@@ -7,14 +7,17 @@ import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/mail/email/email.dart';
 import 'package:model/model.dart';
 import 'package:permission_handler/permission_handler.dart';
+import 'package:pointer_interceptor/pointer_interceptor.dart';
 import 'package:tmail_ui_user/features/base/base_controller.dart';
 import 'package:tmail_ui_user/features/destination_picker/presentation/model/destination_picker_arguments.dart';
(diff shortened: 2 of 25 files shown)


lib/features/email/presentation/email_controller.dart
```diff
@@ -269,6 +290,29 @@ class EmailController extends BaseController {
     }
   }
 
+  void downloadAttachmentForWeb(BuildContext context, Attachment attachment) {
+    _showDownloadingFileDialog(context, attachment);
```
hoangdat, line 294: IMO, we don't need to show loading dialog in browser

hoangdat:
some tests are failed. Please check it @dab246

bitrise.yml
```diff
@@ -85,7 +85,7 @@ workflows:
         - project_location: "$BITRISE_FLUTTER_PROJECT_LOCATION/core"
     - flutter-test@1:
         inputs:
-        - additional_params: "--verbose"
+        - additional_params: "--platform chrome --verbose"
```
hoangdat, line 88: 👍

env.file
```diff
@@ -1 +1 @@
-SERVER_URL=https://dev.open-paas.org
\ No newline at end of file
+SERVER_URL=http://localhost
```
hoangdat, line 1: you should add this file to your global gitignore in local :)

test/features/caching/email_cache_client_test.dart
```diff
@@ -1,3 +1,5 @@
+@TestOn('vm')
```
hoangdat, line 1: what is it?

dab246: Mark this file for testing only on Dart VM. More details here. [https://github.com/dart-lang/test/tree/master/pkgs/test#platform-selectors](https://github.com/dart-lang/test/tree/master/pkgs/test#platform-selectors)

bitrise.yml
hoangdat: why need to add this flag?

dab246: we use library `dart.html` to download file on browser. But it not found in dart test. So we need add flag to target to web

hoangdat:
`dart:html` only use in web environment, we should use `universal:html` instead

PR #308, opened 2022-03-07
TF-287 Fix bug replies do not preserve line breaks

dab246:
https://user-images.githubusercontent.com/80730648/156957444-a3d0e8cd-f878-4a7a-8fcc-ec71f97aa6a7.mov

### DIFF
--- a/model/lib/email/email_content.dart
+++ b/model/lib/email/email_content.dart
@@ -13,3 +13,18 @@ class EmailContent with EquatableMixin {
   @override
   List<Object?> get props => [type, content];
 }
+
+extension EmailContentExtension on EmailContent {
+  String get asHtml {
+    if (type == EmailContentType.textPlain) {
+      return content
+          .replaceAll("'", r"\'")
+          .replaceAll('"', r'\"')
+          .replaceAll('\r', '')
+          .replaceAll('\r\n', '')
+          .replaceAll('\n', '<br/>')
+          .replaceAll('\n\n', '<br/>');
+    }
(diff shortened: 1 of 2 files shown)


hoangdat (approved):

model/lib/email/email_content.dart
```diff
@@ -12,4 +12,19 @@ class EmailContent with EquatableMixin {
...
+      return content
+          .replaceAll("'", r"\'")
+          .replaceAll('"', r'\"')
+          .replaceAll('\r', '')
+          .replaceAll('\r\n', '')
+          .replaceAll('\n', '<br/>')
+          .replaceAll('\n\n', '<br/>');
+    }
+    return content;
+  }
```
chibenwa, line 29: I'd prefer to rely on a library to do text => HTML because now we are vulnerable to HTML injections

chibenwa:
This is a real blocker to me BTW

Cc @hoangdat

model/lib/email/email_content.dart
hoangdat: #309

hoangdat: Hi @chibenwa, back to that problem, we found that no library support us on that (build up the block quote html from plaintext). 
So how about we still keep our method like that, then sanitize the html before rendering?

hoangdat: please continue discuss on https://github.com/linagora/tmail-flutter/issues/309#issuecomment-1062698198

PR #311, opened 2022-03-08
TF-267 Fix bug display a full page for sorting results

dab246:
https://user-images.githubusercontent.com/80730648/157153135-34be0398-1e09-4d60-9148-ce2d998b553d.mov


hoangdat (approved):

chibenwa:
Does this change leverage backend filtering or does it just  fix the current frontend filtering implementation?

If it only fix the current frontend filtering implementation we need a ticket not to forget to come back and fix it once we want a performant, scalable client!

hoangdat:
1. first, we will only filter all we have in the cache, 
2. then `loadMore` with filtering to the backend, for sure

image

chibenwa:
Ok so loadMore pushes the filter to the backend.

Great.

Awesome.

Wonderful implem :+1:

PR #346, opened 2022-03-14
Delete mailbox with children

hoangdat:
https://user-images.githubusercontent.com/6462404/158109619-d8ec3ea0-8055-45d3-aa73-b092e43d5204.mp4

@dab246 please reivew


lib/features/mailbox/data/network/mailbox_api.dart
```diff
@@ -131,29 +136,46 @@ class MailboxAPI {
...
-  Future<bool> deleteMultipleMailbox(AccountId accountId, List<MailboxId> mailboxIds) async {
-    final setMailboxMethod = SetMailboxMethod(accountId)
-      ..addDestroy(mailboxIds.map((mailboxId) => mailboxId.id).toSet())
-      ..addOnDestroyRemoveEmails(true);
+  Future<bool> deleteMultipleMailbox(Session session, AccountId accountId, List<MailboxId> mailboxIds) async {
+    requireCapability(session, [CapabilityIdentifier.jmapCore, CapabilityIdentifier.jmapMail]);
 
-    final requestBuilder = JmapRequestBuilder(httpClient, ProcessingInvocation());
+    final coreCapability = (session.capabilities[CapabilityIdentifier.jmapCore] as CoreCapability);
+    final maxMethodCount = coreCapability.maxCallsInRequest.value.toInt();
```
dab246, line 143: What is this? Why need it here? Why not use other api?

hoangdat: as the requirement, we need to separate each mailbox in each jmap method `mailbox/set`. But for each Jmap server, we are limited the method in each request by `maxCallsInRequest`, so that we only add the `method number <= maxCallsInRequest` for each request send to server. 
Ex: if mailbox 1 has 18 descendants, we will combine 16 (dev environment `maxCallsInRequest = 16`) descendants in first request, then 2 remaining mailbox in the second request.

dab246 (approved):

PR #366, opened 2022-03-17
Fix can't scroll email content in browser

dab246:
https://user-images.githubusercontent.com/80730648/158779033-0f951365-ec52-43b1-8d3c-5c8d1ce940a7.mov

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -7,8 +7,6 @@
 
 # The following line activates a set of recommended lints for Flutter apps,
 # packages, and plugins designed to encourage good coding practices.
-include: package:flutter_lints/flutter.yaml
-
 linter:
   # The lint rules applied to this project can be customized in the
   # section below to disable rules from the `package:flutter_lints/flutter.yaml`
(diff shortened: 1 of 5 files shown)


analysis_options.yaml
```diff
@@ -7,8 +7,6 @@
 
 # The following line activates a set of recommended lints for Flutter apps,
 # packages, and plugins designed to encourage good coding practices.
-include: package:flutter_lints/flutter.yaml
```
hoangdat, line 10: why you remove that?

dab246: It's not necessary here

hoangdat: We need to have lint check, IMO we need it :)

dab246: The problem here is that we have not added library lint so it is always not found when doing `flutter analyze`

hoangdat: okie, move forward, we will back on that problem later : #367

hoangdat (approved):

Issue #362, opened 2022-03-17
[Web] After you refresh the page (F5), the application will stay in the loading state forever


huy-ta:
### Reproduction steps

1. Login to the application
2. Reload/Refresh the page (F5)

### Current behavior

- The application stays in the loading state forever.

### Expected result

- The application should display contents as soon as it finishes loading.

### Devices, OS

- Chrome / Ubuntu Desktop / Web

### Reproduction video

https://user-images.githubusercontent.com/24670327/158735315-495a440d-9dd0-4905-9942-f1f71a50fd10.mp4

PR #373, opened 2022-03-18
Reload web #362

hoangdat:
https://user-images.githubusercontent.com/6462404/158948193-0ff2f61a-6ce9-4787-ad57-1bf1708e7a23.mov

When user click on `reload` button in browser, a widget will be reloaded, and we catch the event on `onReady` of controller to set up all the things:
- url interceptor
- authentication interceptor
- get new session

### DIFF
--- a/lib/features/base/reloadable/reloadable_controller.dart
+++ b/lib/features/base/reloadable/reloadable_controller.dart
@@ -1,0 +1,93 @@
+import 'package:core/data/network/config/authorization_interceptors.dart';
+import 'package:core/data/network/config/dynamic_url_interceptors.dart';
+import 'package:core/presentation/state/failure.dart';
+import 'package:core/presentation/state/success.dart';
+import 'package:dartz/dartz.dart';
+import 'package:get/get.dart';
+import 'package:jmap_dart_client/jmap/core/session/session.dart';
+import 'package:tmail_ui_user/features/base/base_controller.dart';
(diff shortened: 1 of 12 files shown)


dab246:
Looks fine, but there's a problem here:
- When you select an email to view the content, then you reload the page, in my opinion the email will still be opened after reloading.
- I think the problem is related to the web interface is not reasonable

hoangdat:
> When you select an email to view the content, then you reload the page, in my opinion the email will still be opened after reloading.

yes, good idea, but with the new UI, email view will not be in the same screen with Mailbox and ThreadView, so I want to keep that PR simple as simple to resolve the critical bug, can not work with mail after refresh

hoangdat:
> I think the problem is related to the web interface is not reasonable

not get your idea

dab246:
> > When you select an email to view the content, then you reload the page, in my opinion the email will still be opened after reloading.
> 
> yes, good idea, but with the new UI, email view will not be in the same screen with Mailbox and ThreadView, so I want to keep that PR simple as simple to resolve the critical bug, can not work with mail after refresh

Oh, If I reload, the current mailbox should be selected again, right? I see the default is still `Inbox`

chibenwa:
> When you select an email to view the content, then you reload the page, in my opinion the email will still be opened after reloading.

True fro `f5`, false for `ctrl + f5`

IMO the behaviour proposed by Dat is good enough for now.

dab246 (approved):

PR #431, opened 2022-03-30
TF-357 Fix avoid loading so many fonts

dab246:
- App only use `Inter` fonts


chibenwa (approved):
Thanks for having a look ;-)

hoangdat (approved):

PR #433, opened 2022-03-30
TF-386 Rework the select items in a message list in edit mode

dab246:
https://user-images.githubusercontent.com/80730648/160787686-4f6dde44-696b-4eba-bdc1-0f9ed688971d.mp4


hoangdat:
@dab246 IMO, it is better for user to touch everywhere in the item (subject, preview also) to select when `Selection mode` is turned on

dab246:
> IMO, it is better for user to touch everywhere in the item (subject, preview also) to select when `Selection mode` is turned on

You should refer to the Gmail app. I think it's reasonable for now.

hoangdat (approved):

PR #446, opened 2022-04-04
TF-358 Reduce `main.dart.js` file size at web browser startup

dab246:
- Issues:  
    - The file size `main.dart.js` generated when building the web is too large.
    - This issue is discussed [here](https://github.com/flutter/flutter/issues/46589)

**Current `main.dart.js` file size : 3.8MB** 

Screen Shot 2022-04-04 at 12 23 55

- Cause: Simultaneous download of all pages 
- Solve: Use [Lazily loading](https://dart.dev/guides/language/language-tour#lazily-loading-a-library) which could split your code in multiple JavaScript files
- Result: 

**After implement Lazily loading. File `main.dart.js` has size : 3.4MB** 

Screen Shot 2022-04-04 at 12 26 05

### DIFF
--- a/lib/main/pages/deferred_widget.dart
+++ b/lib/main/pages/deferred_widget.dart
@@ -1,0 +1,73 @@
+import 'dart:async';
+import 'package:flutter/material.dart';
+
+typedef LibraryLoader = Future<void> Function();
+typedef DeferredWidgetBuilder = Widget Function();
+
+/// Wraps the child inside a deferred module loader.
+///
+/// The child is created and a single instance of the Widget is maintained in
+/// state as long as closure to create widget stays the same.
+///
+class DeferredWidget extends StatefulWidget {
+  DeferredWidget(
+      this.libraryLoader,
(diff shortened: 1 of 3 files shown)


lib/main/pages/deferred_widget.dart
```diff
@@ -0,0 +1,73 @@
...
+
+typedef LibraryLoader = Future<void> Function();
+typedef DeferredWidgetBuilder = Widget Function();
+
+/// Wraps the child inside a deferred module loader.
+///
+/// The child is created and a single instance of the Widget is maintained in
+/// state as long as closure to create widget stays the same.
+///
+class DeferredWidget extends StatefulWidget {
```
hoangdat, line 12: it is very cool, can you write an `ADR` to describe the solution, it will be a valuable documentation for other people can understand it in the future.

dab246: Added.

hoangdat (approved):

PR #476, opened 2022-04-08
Enable GZIP for Nginx

quantranhong1999:
Tested and gzip works locally.
The new `nginx.conf` file = default `nginx.conf` + server block + gzip on.
The GZIP setting includes compressed mime types review is welcome.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ENV FLUTTER_HOME="/opt/flutter"
 ENV PATH "$PATH:$FLUTTER_HOME/bin"
 
 # Prerequisites
-RUN apt update && apt install -y curl git unzip xz-utils zip libglu1-mesa \
+RUN apt update && apt install -y curl git unzip xz-utils zip gzip libglu1-mesa \
  && mkdir -p $FLUTTER_HOME \
  && curl -o flutter.tar.xz $FLUTTER_URL \
  && tar xf flutter.tar.xz -C /opt \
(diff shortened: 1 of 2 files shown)


Arsnael (approved):

chibenwa (approved):
Could you please use browser dev tools to show us the gains?

hoangdat (approved):

hoangdat:
please let me know when it is available to merge (don't worry, our CI are off for some days)

tanandy:
so you are only interested by gzip and not gzip_static ?

You dont have files already compressed ? and force dynamic compression in any case ? (instead of static)

quantranhong1999:
> Could you please use browser dev tools to show us the gains?

image

- PNG zip is not useful (even add more zipped size). Maybe cause PNG is already compressed enough.
- Many font files (.ttf, .otf) do not zip (although already have that type in config, seems weird). There is a .ttf file that has been zipped (return type ttf) and other .ttf not zipped with default return type octet-stream => It looks like Nginx could have some issue recognizing .ttf files.

quantranhong1999:
Added `application/octet-stream` and zipping works for font files now (but actually it should have worked with .ttf etc).
image

This GZIP only works for Frontend's static files. Maybe we should config GZIP for other NGINX which stays in front of James Jmap also.

chibenwa:
> note: considers reading BREACH/CRIME attack if not done already

Clearly not an issue when serving SPA assets. No info to leak...

> so you are only interested by gzip and not gzip_static ?

That would be freaking cool if we could gzip statically, yes! A very nice to have. Quan can you have a look?

> PNG zip is not useful

Yes like on all heavily compressed files...

quantranhong1999:
Static GZIP (all type of files) result:
image

image

Arsnael (approved):

quantranhong1999:
Can we merge this?

hoangdat:
please rebase. i will merge it

PR #502, opened 2022-04-19
TF-495 Apply Isolate for LoadMore use-case

dab246:
### The problem is solved in this PR
- [x] Apply Isolate for LoadMore use-case
   - Increase email loading speed avoid blocking user interaction and other tasks.
- [x] Integration `linter` plugin
   - To encourage good coding practices.
   - Performance improvement
 
278493242_5049713618455851_132285591329688861_n

### Result

- The app stays at ` 60 fps` most of the time, or in other words, changing frames every `16 milliseconds`.
- Device Test: Oppo - Android 10

278781581_4827297477367408_1328088794709915929_n

### DIFF
--- a/core/lib/utils/benchmark.dart
+++ b/core/lib/utils/benchmark.dart
@@ -1,0 +1,29 @@
+
+import 'package:core/utils/app_logger.dart';
+
+class _Benchmark {
+  final Map<String, int> _starts = <String, int>{};
+
+  void start(dynamic id) {
+    final String benchId = id.toString();
+    if (_starts.containsKey(benchId)) {
+      log('_Benchmark::start(): Benchmark already have comparing with id=$benchId in time');
+    } else {
+      _starts[benchId] = DateTime.now().microsecondsSinceEpoch;
+    }
+  }
+
+  double end(dynamic id) {
+    final String benchId = id.toString();
+    if (!_starts.containsKey(benchId)) {
+      throw Exception('In Benchmark not placed comparing with id=$benchId');
+    }
+    final double diff = (DateTime.now().microsecondsSinceEpoch - _starts[benchId]!) / 1000;
+    final String info = '$benchId need ${diff}ms';
+    log('_Benchmark::end(): $info');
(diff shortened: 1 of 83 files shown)


hoangdat:
@dab246 please add the image for the `perfomance_overlay` before apply your solution.

And also please add more about your device: RAM, ROM, refresh rate :)

dab246:
> @dab246 Great! please add the image for the `perfomance_overlay` before apply your solution.
> 
> And also please add more about your device: RAM, ROM, refresh rate :)
 
I updated

hoangdat (approved):

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -382,13 +387,15 @@ class ThreadController extends BaseController {
   void loadMoreEmails() {
     if (canLoadMore && _accountId != null) {
       log('ThreadController::loadMoreEmails(): latest: ${emailList.last.receivedAt}');
+      startFpsMeter();
```
hoangdat, line 390: how to only apply it on debug or profile mode?

core/lib/utils/benchmark.dart
```diff
@@ -0,0 +1,29 @@
+
+import 'package:core/utils/app_logger.dart';
+
+class _Benchmark {
```
hoangdat, line 4: should run it only in debug or profile mode

dab246: oke i will update it

lib/features/thread/presentation/thread_controller.dart
dab246: Only check `Foundation.kDebugMode` or `Foundation.kProfileMode`

hoangdat: yes, please apply it from the Fps class, or here, or inside the startFpsMeter()

hoangdat (approved):

hoangdat:
image

Issue #339, opened 2022-03-14
Empty  entries in email list view


chibenwa:
### Description

https://user-images.githubusercontent.com/6928740/158096991-f1555845-82f3-45f2-aee8-d3650b2f2702.mp4

Happens regulary to me. Switching mailboxes, setting filters, scrolling and I sometime gets this.

Arsnael:
Noticed as well on ios version:
![signal-2022-03-15-102920](https://user-images.githubusercontent.com/9005025/158300669-0698a60f-e9aa-4ea3-bd10-881f40ac68d3.jpeg)

hoangdat:
Hi @chibenwa, i found you not use the latest version for this issue, can you show me your currently `commitId`

chibenwa:
No idea how to get that from the docker container. Deployed it on friday 11/03. 

Now `tmail.linagora.*` is updated nightly.

I will see if this still happens.

chibenwa:
![Screenshot from 2022-03-16 14-34-56](https://user-images.githubusercontent.com/6928740/158539393-2e3c282c-ca57-43d2-827b-c0feb8e80cbe.png)

hoangdat:
@chibenwa how to reproduce it

chibenwa:
I don't know I'm not doing anything special

Arsnael:
@hoangdat Happens randomly to me as well... I was just checking on different mailboxes a few times before seeing this sometimes

hoangdat:
> Note: ctrl + f5 don't drop the cache, I only see Mailbox/changes , Email/changes being issued so I can't go back in history and provide you JMAP requests.

for the debug, all the log are dumped to console, if you have more information, please share to me

Kevinle2312:
Failed on CI platform. I found two emails having this issue:
![image](https://user-images.githubusercontent.com/32490981/162145631-e5a3b0fe-8a18-4ac7-9d39-cb67b5576040.png)

Email content: 
![image](https://user-images.githubusercontent.com/32490981/162145692-ab4bfd51-5605-4eac-8a19-de0103c33b22.png)
![image](https://user-images.githubusercontent.com/32490981/162145784-4092799f-d4f2-4a3e-8fa6-a46a89fb13a2.png)

hoangdat:
@Kevinle2312 please share us the way and the platform you reproduce that?

Kevinle2312:
> @Kevinle2312 please share us the way and the platform you reproduce that?
[...]

PR #512, opened 2022-04-26
#339 Fix Empty item on ThreadView

hoangdat:


chibenwa:
> Clean up the cache whenever start the app

This I imagine can be debatable. I am not that sure about it, but I let you choose.

hoangdat:
> > Clean up the cache whenever start the app
> 
> This I imagine can be debatable. I am not that sure about it, but I let you choose.

FYI, 
- before: we delete all the item in the cache have `received at` more than 30 days from before today, it means that, the first time user START the app today, we will execute clean up.

- now: we delete all the items in the cache have `received at` more than 10 days from now, it means that, every time which the user START the app, we will execute clean up

START = app is terminated before, not in the stack of multitask on mobile, and it is different with app run in background.

dab246 (approved):
LGTM.

johnnguyenit39 (commented):
read it

Issue #457, opened 2022-04-05
[Story] As a user, I want to see my signature in the composer when I composing email


hoangdat:
# Summary

* [Related EPIC](#related-epic)
* [Definition](#definition)
* [Screenshots](#screenshots)
* [Misc](#misc)

## Related EPIC

* TODO: {link}

## Definition

- Given that I am a Tmail user 
- After log-in successfully to the Tmail web, tablet (only see signature on web and tablet, mobile don't display signature, but signature will be appended when sending email)
- I compose/reply/foward email
- I can choose between my identifies via `From` field as a drop-down list
- After I choose an identifier, all the related fields are update base on the identifier
- Signature also be updated
- Bcc also add item `bcc` of selected Identifier

https://user-images.githubusercontent.com/6462404/161738795-8984ee97-d0da-4727-a990-b02f6e11a802.mov

[Back to Summary](#summary)

## Design ticket

https://ci.linagora.com/linagora/lgs/design/-/issues/174

[Back to Summary](#summary)

## Misc

None

[Back to Summary](#summary)

Kevinle2312:
Passed on the Production platform.

PR #519, opened 2022-04-27
#517 Apply identity in composer

dab246:
https://user-images.githubusercontent.com/80730648/165474952-4986abec-1ef2-40e9-bbdc-a452f40f62b8.mov


lib/features/composer/presentation/composer_controller.dart
```diff
@@ -173,6 +180,11 @@ class ComposerController extends BaseController {
           _uploadAttachmentsSuccess(success);
         } else if (success is GetEmailContentSuccess) {
           _getEmailContentSuccess(success);
+        } if (success is GetAllIdentitiesSuccess) {
```
hoangdat, line 183: How about if we can not `get all identities success`? I propose we can use user as we do before.

dab246: It still works the way you want it to.

hoangdat:
IMO, you forgot to add `Bcc` of Identity when you generate email in composer

image

dab246:
> I did not see the signature is add to the composer when changing the identity. please share the demo for this also

It is not handled in this ticket. Please note that the ticket is referred to

dab246:
> IMO, you forgot to add `Bcc` of Identity when you generate email in composer
> 
> image
> 
> Refs: https://jmap.io/spec-mail.html#identities

Please see it in more detail inside jmap_dart_client

dab246:
> > > I did not see the signature is add to the composer when changing the identity. please share the demo for this also
> > 
> > 
> > It is not handled in this ticket. Please note that the ticket is referred to
> 
> srr, please do it later in: #457

hoangdat (approved):

PR #576, opened 2022-05-18
Fix padding login

hoangdat:
Screen Shot 2022-05-18 at 21 10 47

Screen Shot 2022-05-18 at 21 10 38

### DIFF
--- a/lib/features/login/presentation/login_view_web.dart
+++ b/lib/features/login/presentation/login_view_web.dart
@@ -35,13 +35,7 @@ class LoginView extends BaseLoginView {
             children: [
               Padding(
                 padding: const EdgeInsets.only(top: 67),
-                child: (SloganBuilder(arrangedByHorizontal: true)
-                    ..setSloganText(AppLocalizations.of(context).app_name)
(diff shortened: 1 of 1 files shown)


lib/features/login/presentation/login_view_web.dart
```diff
@@ -32,16 +32,11 @@ class LoginView extends BaseLoginView {
           child: Column(
             mainAxisAlignment: MainAxisAlignment.start,
             mainAxisSize: MainAxisSize.max,
+            crossAxisAlignment: CrossAxisAlignment.center,
```
dab246, line 35: You don't need to set it because the default is already there.

lib/features/login/presentation/login_view_web.dart
```diff
@@ -72,73 +67,75 @@ class LoginView extends BaseLoginView {
   Widget _buildWebForm(BuildContext context) {
     return Stack(
```
dab246, line 68: If there is 1 element, you don't need to use `Stack` here.

dab246 (approved):
/lgtm

Issue #558, opened 2022-05-12
Links broken


chibenwa:
https://user-images.githubusercontent.com/6928740/168000393-71e5569c-700b-44c4-8e40-26776e9840f2.mp4

tmail.linagora.com

hoangdat:
Hi @chibenwa can you share us the eml, or send me this emal

chibenwa:
```
Return-Path: <SRS0=F3Np=V2=james.apache.org=server-dev-return-83397-btellier=linagora.com@ik2.com>
Received: from 172.17.0.1 (EHLO incoming.linagora.com) ([172.17.0.1])
          by incoming.linagora.com (JAMES SMTP Server ) with ESMTP ID 2d3706e1
          for <btellier@linagora.com>;
          Wed, 18 May 2022 05:47:07 +0000 (UTC)
Received: from obm3-ui.linagora.dc1 (obm3-ui.linagora.dc2 [172.24.128.227])
	by incoming.linagora.com (Postfix) with ESMTP id 1C54A9C623
	for <btellier@linagora.com>; Wed, 18 May 2022 05:47:07 +0000 (UTC)
Received: from s480e.ik2.com (s480e.ik2.com [64.38.239.84])
	by obm3-ui.linagora.dc1 (Postfix) with ESMTP id B2BDF1D5A3
	for <btellier@linagora.com>; Wed, 18 May 2022 07:47:06 +0200 (CEST)
Received: from s480d.ik2.com
	by s480e.ik2.com with esmtps (TLS1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(envelope-from <server-dev-return-83397-btellier=linagora.com@james.apache.org>)
	id 1nrCWV-0000qS-O8
	for btellier@linagora.com; Wed, 18 May 2022 05:47:06 +0000
Received: from 3.227.148.255 by s480d.ik2.com (IK2 SMTP Server); Wed, 18 May 2022 05:47:03 +0000
Received: from mail.apache.org (mailroute1-lw-us.apache.org [207.244.88.153])
	by mxout1-ec2-va.apache.org (ASF Mail Server at mxout1-ec2-va.apache.org) with SMTP id 3C1C13F8A4
	for <btellier@linagora.com>; Wed, 18 May 2022 05:47:02 +0000 (UTC)
Received: (qmail 53660 invoked by uid 500); 18 May 2022 05:47:01 -0000
[...]

PR #596, opened 2022-05-26
#558 Fix links broken in email view

dab246:
https://user-images.githubusercontent.com/80730648/170466961-a0ca73ea-4841-48a3-9cc0-619353605daf.mov

### DIFF
--- a/model/lib/email/email_content.dart
+++ b/model/lib/email/email_content.dart
@@ -18,8 +18,6 @@ extension EmailContentExtension on EmailContent {
   String get asHtml {
     if (type == EmailContentType.textPlain) {
       return content
-          .replaceAll("'", r"\'")
-          .replaceAll('"', r'\"')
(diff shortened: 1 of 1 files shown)


model/lib/email/email_content.dart
```diff
@@ -18,8 +18,6 @@ extension EmailContentExtension on EmailContent {
   String get asHtml {
     if (type == EmailContentType.textPlain) {
       return content
-          .replaceAll("'", r"\'")
-          .replaceAll('"', r'\"')
```
hoangdat, line 22: I am worried this may have side effect :(

hoangdat (approved):

model/lib/email/email_content.dart
dab246: IMO, It won't be affected whatsoever. WebView has a good handling mechanism to recognize these characters.

PR #617, opened 2022-06-02
#606 Use refreshToken to get new token when accessToken expired

dab246:


lib/features/login/data/model/oidc_configuration_cache.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:equatable/equatable.dart';
+import 'package:hive/hive.dart';
+import 'package:tmail_ui_user/features/caching/utils/caching_constants.dart';
+
+part 'oidc_configuration_cache.g.dart';
+
+@HiveType(typeId: CachingConstants.OIDC_CONFIGURATION_HIVE_CACHE_IDENTIFY)
+class OidcConfigurationCache extends HiveObject with EquatableMixin {
```
hoangdat, line 8: why we need cache that?

hoangdat: got it, but please change the name to `AuthorityOIDCCache`

lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -1,32 +1,123 @@
...
+      case AuthenticationType.none:
+        break;
     }
     super.onRequest(options, handler);
   }
 
+  @override
+  void onError(DioError err, ErrorInterceptorHandler handler) async {
+    final requestOptions = err.requestOptions;
+    if (_isAuthenticationError(err) && _authenticationType == AuthenticationType.oidc) {
```
hoangdat, line 70: We will have varied kind of authentication Error, please more specific in the method name. `_isTokenExpired`

lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -1,32 +1,123 @@
...
+        final newToken = await _oidcHttpClient.refreshingTokensOIDC(
+            _configOIDC!.clientId,
+            _configOIDC!.redirectUrl,
+            _configOIDC!.discoveryUrl,
+            _configOIDC!.scopes,
+            _token!.refreshToken);
+
+        await Future.wait([
+          _tokenOidcCacheManager.persistOneTokenOidc(newToken),
+          _accountCacheManager.setSelectedAccount(Account(
```
hoangdat, line 81: The `tokenId` is changed when you refresh?. I don't think so

lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -1,32 +1,123 @@
...
+
+      final response = await _dio.fetch(requestOptions);
+      return handler.resolve(response);
+    } else {
+      super.onError(err, handler);
+    }
+  }
+
+  bool _isAuthenticationError(DioError dioError) {
+    if (dioError.type == DioErrorType.response &&
```
hoangdat, line 104: `response` don't have a dedicated error for `expired`?

lib/features/login/domain/repository/authentication_oidc_repository.dart
```diff
@@ -11,4 +11,8 @@ abstract class AuthenticationOIDCRepository {
   Future<void> persistTokenOIDC(TokenOIDC tokenOidc);
 
   Future<TokenOIDC> getStoredTokenOIDC(String tokenIdHash);
+
+  Future<void> persistOidcConfiguration(OIDCConfiguration config);
```
hoangdat, line 15: why not create other model to store `TokenOIDC` with its own `authority`? I propose it, because we don't need to `openBox` too much time.

lib/features/login/presentation/login_controller.dart
```diff
@@ -186,11 +186,13 @@ class LoginController extends GetxController {
...
   }
 
-  void _getTokenOIDCSuccess(GetTokenOIDCSuccess success) {
+  void _getTokenOIDCSuccess(GetTokenOIDCSuccess success, OIDCConfiguration config) {
     log('LoginController::_getTokenOIDCSuccess(): ${success.tokenOIDC.toString()}');
     loginState.value = LoginState(Right(success));
     _dynamicUrlInterceptors.changeBaseUrl(kIsWeb ? AppConfig.baseUrl : _urlText);
     _authorizationInterceptors.setToken(success.tokenOIDC.toToken());
+    _authorizationInterceptors.setAuthenticationType(AuthenticationType.oidc);
+    _authorizationInterceptors.setOIDCConfiguration(config);
```
hoangdat, line 195: why not set it at the same time, in one method?

lib/features/login/domain/state/get_stored_token_oidc_state.dart
```diff
@@ -1,19 +1,20 @@
 import 'package:core/presentation/state/failure.dart';
 import 'package:core/presentation/state/success.dart';
-import 'package:model/oidc/token_oidc.dart';
+import 'package:model/model.dart';
```
hoangdat, line 3: I think, we should import exactly what we use, it is optimized for our code.

hoangdat:
we have the problem with CICD at the moment, I am trying to fix it. Update soon.

lib/features/login/domain/state/get_stored_token_oidc_state.dart
dab246: We are using multiple classes in this module. I don't want to write too much code for it. This is the best way for us in the future.

lib/features/login/presentation/login_controller.dart
dab246: I wish it was clearer what we update.

lib/features/login/domain/repository/authentication_oidc_repository.dart
dab246: If you think `authority` should not be secured, we can use another storage method.

lib/features/login/data/network/config/authorization_interceptors.dart
dab246: It is a necessary condition, when it cannot return data from a `token expire`. I have checked `token is expire` in this method.

lib/features/login/data/network/config/authorization_interceptors.dart
dab246: Are you sure of that? Just to be on the safe side I want to update it so that an unexpected error never occurs.

lib/features/login/domain/repository/authentication_oidc_repository.dart
hoangdat: no problem with the security, URL for `authServer` is public :).
I only keep the `Token` and `its authority` in the same place. Easy to manage, less operator to the db. That is my idea

lib/features/login/data/network/config/authorization_interceptors.dart
hoangdat: I am not sure, if `tokenId` not change, we don't need to store account again. We should not keep unneeded account in `Account Cache table`. If `tokenId` change, we should delete the old account.

lib/features/login/data/network/config/authorization_interceptors.dart
hoangdat: IMO, it should be more safety for us when the server says that we `expired`.

lib/features/login/data/network/config/authorization_interceptors.dart
hoangdat: not get your point.

model/lib/oidc/token_id.dart
```diff
@@ -8,4 +8,5 @@ class TokenId with EquatableMixin {
 
   @override
   List<Object> get props => [uuid];
+
```
dab246, line 11: Why did you fix it like that?

hoangdat: no, temporary change to trigger CICD :), please help me remove it

model/lib/oidc/token_oidc.dart
```diff
@@ -32,4 +32,6 @@ extension TokenOIDCExtension on TokenOIDC {
   Token toToken() {
     return Token(token, tokenId, refreshToken, expiredTime: expiredTime);
   }
+
+  String get tokenIdHash => tokenId.uuid.hashCode.toString();
```
hoangdat, line 36: 👍

lib/features/login/data/model/oidc_configuration_cache.dart
dab246: I switched to using `SharedPreference` to store `authority`

lib/features/login/data/network/config/authorization_interceptors.dart
dab246:
````dart
if ((_isTokenExpired() || err.response?.statusCode == 401) 
   && _authenticationType == AuthenticationType.oidc) {
 
}
````

File: lib/features/login/domain/repository/authentication_oidc_repository.dart (hunk 5 above)
Comment (dab246) [reply lib/features/login/domain/repository/authentication_oidc_repository.dart]: We will only save `authority` in `SharedPreference` 
```dart
  Future<void> persistAuthorityOidc(String authority);
```

lib/features/login/presentation/login_controller.dart
dab246:
```dart
void setTokenAndAuthorityOidc({Token? newToken, OIDCConfiguration? newConfig}) {
    _token = newToken;
    _configOIDC = newConfig;
    _authenticationType = AuthenticationType.oidc;
  }

```

lib/features/login/domain/state/get_stored_token_oidc_state.dart
dab246: Ok, we will only `import` the `classes` that we use.

lib/features/login/data/network/config/authorization_interceptors.dart
dab246: if ((_isTokenExpired() || err.response?.statusCode == 401) 
   && _authenticationType == AuthenticationType.oidc) {
 
}

lib/features/login/data/network/config/authorization_interceptors.dart
dab246: After double checking, I noticed that the `tokenId` value is a string of characters, the same as the `accessToken` value. So after `refreshToken` we always have `tokenId ` changed. So we will do:
- Delete `account` with old `token`.
- Add `account` with new `token`

hoangdat: okie, go ahead

hoangdat (approved):

PR #631, opened 2022-06-07
Authentication with OIDC on browser

dab246:
- [x] Login on browser

 - Use OIDC

https://user-images.githubusercontent.com/80730648/172589125-055b9aa6-a41f-4187-963c-d9b649945bdf.mov

 - Use Credential

https://user-images.githubusercontent.com/80730648/172589174-3d7b1ef4-d38d-4caf-b28f-8e85b356a2f1.mov

- [x] Logout OIDC on browser

https://user-images.githubusercontent.com/80730648/172604718-5b7de7a3-4a12-493a-8d50-b7fc21d4988e.mov


chibenwa:
Imo when we have sso we should not have username pwd fields.

model/lib/oidc/oidc_configuration.dart
```diff
@@ -1,13 +1,14 @@
 
+import 'package:core/utils/build_utils.dart';
 import 'package:equatable/equatable.dart';
 
 class OIDCConfiguration with EquatableMixin {
-  static const redirectOidc = 'teammail.mobile://oauthredirect';
+  static const redirectOidcMobile = 'teammail.mobile://oauthredirect';
+  static const redirectOidcWeb = 'http://localhost:3000/login-callback.html';
```
hoangdat, line 7: How about set it up when compile time via `env`?

lib/features/login/data/extensions/service_path_extension.dart
```diff
@@ -1,14 +1,14 @@
...
+import 'package:core/data/model/query/query_parameter.dart';
+import 'package:core/data/network/config/service_path.dart';
 
 extension ServicePathExtension on ServicePath {
   String generateEndpointPath() {
     return path;
   }
 
   ServicePath generateOIDCPath(Uri baseUrl) {
-    return ServicePath(baseUrl.origin + Endpoint.oidc + path);
```
hoangdat, line 11: srr, my fault, I added to test, and forget to remove :(

lib/features/login/data/network/config/oidc_constant.dart
```diff
@@ -1,5 +1,10 @@
+import 'package:core/core.dart';
+
 class OIDCConstant {
   static String get mobileOidcClientId => 'teammail-mobile';
+  static String get webOidcClientId => 'teammail-web';
```
hoangdat, line 5: should set-it-up in compile time

pubspec.yaml
```diff
@@ -156,6 +156,12 @@ dependencies:
   # OIDC Mobile
   flutter_appauth: 4.0.0
 
+  # OIDC Web
+  flutter_appauth_web:
+    git:
+      url: https://github.com/dab246/flutter_appauth_web.git
```
hoangdat, line 162: Why you choose this? Why you need folk and modify it?

lib/features/login/data/network/authentication_client/authentication_client_web.dart
```diff
@@ -0,0 +1,47 @@
+
+import 'package:get/get.dart';
+import 'package:tmail_ui_user/features/login/data/network/authentication_client/authentication_client_base.dart';
+import 'package:flutter_appauth_platform_interface/flutter_appauth_platform_interface.dart';
+import 'package:model/oidc/oidc_configuration.dart';
+import 'package:model/oidc/token_id.dart';
+import 'package:model/oidc/token_oidc.dart';
+
+import '../../utils/library_platform/app_auth_plugin/app_auth_plugin.dart';
```
hoangdat, line 15: please use absolute path instead of it

model/lib/oidc/oidc_configuration.dart
```diff
@@ -5,6 +5,7 @@ import 'package:equatable/equatable.dart';
 class OIDCConfiguration with EquatableMixin {
   static const redirectOidcMobile = 'teammail.mobile://oauthredirect';
   static const redirectOidcWeb = 'http://localhost:3000/login-callback.html';
+  static const logoutRedirectOidcWeb = 'http://localhost:3000/logout-callback.html';
```
hoangdat, line 8: how about to set-it-up with relative path?

hoangdat:
A small discussion:
1. in mobile, we store `refreshToken` on permanent cache, use it when `accessToken` expired. How about we do the same on browser? @chibenwa

chibenwa:
> in mobile, we store refreshToken on permanent cache, use it when accessToken expired, so user no need to input in login form too much. How about we do the same on browser? @chibenwa

Sounds fair to me.

But what we should truely do is ask browser experts ;-)

tuanlc:
Hi all,

In the web version, we have relied on an [OIDC client library](https://www.npmjs.com/package/oidc-client) for this. We just need to provide some metadata like OIDC server URL, then it does the remaining logic.

I looked how the library works and the idea is the same as you said:
- Store access token & refresh token in local storage
- We can set the time  to trigger the lib to renew the access token

> accessTokenExpiringNotificationTime (number, default: 60): The number of seconds before an access token is to expire to raise the accessTokenExpiring event.

chibenwa:
Thanks @tuanlc 

> We can set the time to trigger the lib to renew the access token

Do you do it?

hoangdat:
@tuanlc what are conditions for the case user need to fill in the login form?

hoangdat:
@tuanlc How the front-end manage the environment configuration? How to prevent it expose to user?

we are now :(.

image

lib/features/login/data/network/config/oidc_constant.dart
dab246: Why is it necessary, IMO it is not necessary to set this value at compile time

hoangdat: Because, every one who want to deploy their own front-end can change it. It not like mobile, many users use the app via store, so we need to hard-code `clientId` for mobile.

lib/features/login/domain/usecases/get_token_oidc_interactor.dart
```diff
@@ -7,6 +7,7 @@ import 'package:model/account/account.dart';
 import 'package:model/account/authentication_type.dart';
 import 'package:model/oidc/oidc_configuration.dart';
 import 'package:model/oidc/token_oidc.dart';
+import 'package:tmail_ui_user/features/login/data/extensions/oidc_configuration_extensions.dart';
```
hoangdat, line 10: I saw some file have the same import. is it necessary?

hoangdat:
Check upgrade with android

tuanlc:
@chibenwa 

> > We can set the time to trigger the lib to renew the access token
> 
> Do you do it?

Yes, we did it. We set it to 10 seconds. The default value is 60 seconds.

tuanlc:
@hoangdat 

> @tuanlc what are conditions for the case user need to fill in the login form?

I don't get your point?

tuanlc:
> @tuanlc How the front-end manage the environment configuration? How to prevent it expose to user? Especially, how did you handle the configuration for OIDC (clientId, redirectUrl, ....)?

@hoangdat 
> How the front-end manage the environment configuration?

There is a configuration file and like you have, clients will fetch the file when loading web apps

> How to prevent it expose to user?

Why do you need to prevent this? These configuration are public and there is no security issue for them.

> Especially, how did you handle the configuration for OIDC (clientId, redirectUrl, ....)?

We use environment configuration to setup metadata for the OIDC client lib

dab246:
> Check upgrade with android

I have built and run on android 8.1 (SamSung J7) and Simulator (Android 11.0)  and it still works fine.

hoangdat (approved):

PR #648, opened 2022-06-13
TF-644 Download attachment with OIDC in web

dab246:
- Currently `downloadUrl=https://gateway.upn.integration-open-paas.org/oidc/jmap/download/{accountId}/{blobId}?type={type}&name={name}` is not working access. We need to wait for the back-end to fix it.

- Temporarily for the demo video, I modified `downloadUrl` by removing `/jmap` in it. It will become `downloadUrl=https://gateway.upn.integration-open-paas.org/oidc/download/{accountId}/{blobId}?type={type}&name={name}`

- Result.

 - Web

https://user-images.githubusercontent.com/80730648/173502101-02f56bc8-ef73-4bf4-a6dd-2e06143c3d53.mov

 - Mobile (Android)

https://user-images.githubusercontent.com/80730648/173501827-e18099db-d9ed-47e0-93f8-5a7336bce5b8.mov

 - Mobile (iOS)

https://user-images.githubusercontent.com/80730648/173501734-f7aa23d4-5039-437d-b6a6-942bc3d8a9a3.mov


core/lib/data/network/download/download_manager.dart
```diff
@@ -74,30 +74,42 @@ class DownloadManager {
...
+  Future<bool> downloadFileForWeb(String downloadUrl, String filename, String authValue) async {
+    try {
+      final headerParam = Map<String, String>();
+      headerParam[HttpHeaders.authorizationHeader] = authValue;
+      headerParam[HttpHeaders.acceptHeader] = DioClient.jmapHeader;
 
-    http.Response res = await http.get(Uri.parse(downloadUrl), headers: headerParam);
+      log('DownloadManager::downloadFileForWeb(): downloadUrl: $downloadUrl');
+      log('DownloadManager::downloadFileForWeb(): filename: $filename');
+      log('DownloadManager::downloadFileForWeb(): authValue: $authValue');
```
hoangdat, line 85: please remove it in the log for web for safe :)

hoangdat: `authValue` is very sensitive

hoangdat:
waiting for the fix from Back-end side

core/lib/data/network/download/download_manager.dart
```diff
@@ -79,31 +82,40 @@ class DownloadManager {
     return streamController.stream.first;
   }
 
-  Future<bool> downloadFileForWeb(String downloadUrl, String filename, String basicAuth) async {
-    final headerParam = Map<String, String>();
-    headerParam[HttpHeaders.authorizationHeader] = basicAuth;
-    headerParam[HttpHeaders.acceptHeader] = DioClient.jmapHeader;
+  Future<bool> downloadFileForWeb(
```
hoangdat, line 85: how about make the download progress display in browser?

lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -77,8 +77,7 @@ class AuthorizationInterceptors extends InterceptorsWrapper {
   void onError(DioError err, ErrorInterceptorHandler handler) async {
     final requestOptions = err.requestOptions;
     log('AuthorizationInterceptors::onError(): $err');
-    if ((_isTokenExpired() || err.response?.statusCode == 401) &&
-        _isAuthenticationOidcValid()) {
+    if (_isTokenExpired() && _isAuthenticationOidcValid()) {
```
hoangdat, line 80: Why we remove 401?

dab246: There are many possible errors that return `statusCode=401`. We are currently dealing with `token` retrieval when `token expired`. So here we only need to care about the `token expired` condition to perform `refreshToken`. Avoid repeatedly calling refreshToken when some unexpected `401` error occurs.

core/lib/data/network/download/download_manager.dart
dab246: 1. To download progress display in browser (Specifically Chrome) we need `downloadUrl public` (download url does not use `authentication`)
2. Currently download attachment requires `accessToken` or `basicAuth`.
=> I don't know if the server side currently supports getting `downloadUrl public` so we can do that?

ManhNTX (approved):
LGTM

core/lib/data/network/download/download_manager.dart
hoangdat: create another ticket #654 for it

chibenwa: The dowload response should have a content length header allowing you to know the size of what you need to download.

Then you just need to count the downloaded bytes.

> => I don't know if the server side currently supports getting downloadUrl public so we can do that?

Thats off spec...

PR #659, opened 2022-06-15
TF-658 Fix composer disappears when performing actions on browser

dab246:
https://user-images.githubusercontent.com/80730648/173738324-e3ed1ab4-6a6c-4be3-9aa7-bb576774a594.mov


hoangdat:
How about if user typed something in composer?

dab246:
> 

- It has no effect on user input in composer. This is just a matter of displaying `Composer overlay` on `MailboxDashBoard`.
- Previously, the display of `Composer overlay` was based on the `DashAction` variable of the mailboxdashboard. So it will affect when the action suddenly changes.
- Currently, we only use one variable `ComposerOverlayState` to listen for its display.

ManhNTX (approved):

PR #665, opened 2022-06-16
TF-602 Fix size changes when selection email in browser

dab246:
https://user-images.githubusercontent.com/80730648/174059415-9fe45089-4472-4a8d-b173-cd018ca04706.mov

### DIFF
--- a/lib/features/thread/presentation/widgets/email_tile_web_builder.dart
+++ b/lib/features/thread/presentation/widgets/email_tile_web_builder.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:core/core.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_svg/flutter_svg.dart';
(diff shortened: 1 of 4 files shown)


lib/features/thread/presentation/widgets/email_tile_web_builder.dart
```diff
@@ -46,6 +50,24 @@ class EmailTileBuilder {
     _onMoreActionClick = onMoreActionClick;
   }
 
+  void _onHoverIconChanged(bool isHover, StateSetter setState) {
+    if (_debounceTimeIcon?.isActive ?? false) _debounceTimeIcon?.cancel();
+    _debounceTimeIcon = Timer(const Duration(milliseconds: 300), () {
+      setState(() {
```
hoangdat, line 56: we don't have other way to do that? with `setState` here, IMO we maybe need to re-render all MailboxDashboard 😨

dab246: Do you understand what setState is here?. Please read more code signing.

hoangdat (approved):

PR #690, opened 2022-06-29
Add a horizontal and vertical scrollbar to an HTML

ManhNTX:
Add a horizontal scrollbar to an HTML

https://user-images.githubusercontent.com/107173849/178128168-dff8e91a-7ff5-442a-bf31-d88e105daa45.mov


core/lib/presentation/views/html_viewer/html_content_viewer_widget.dart
```diff
@@ -95,6 +95,7 @@ class _HtmlContentViewState extends State<HtmlContentViewer> {
       backgroundColor: Colors.white,
       onWebViewCreated: (controller) async {
         _webViewController = controller;
+       _htmlData = _htmlData?.replaceFirst('overflow: hidden', 'overflow: auto');
```
dab246, line 98: - I don't think it solves this problem.
- You should note that if you fix this, some emails may be displayed with 2 scrollbars while reading emails on mobile. You should retry the view cases.
- Please show us the demo video to see the results more clearly.

ManhNTX: I added at top comment. Please check it

core/lib/presentation/views/html_viewer/html_content_viewer_widget.dart
```diff
@@ -55,7 +55,7 @@ class _HtmlContentViewState extends State<HtmlContentViewer> {
   }
 
   String _generateHtmlDocument(String content) {
-    final htmlTemplate = generateHtml(content, hideScrollBar: true);
+    final htmlTemplate = generateHtml(content, hideScrollBar: false);
```
dab246, line 58: 1. Could you please explain more about this change?
2. Like the demo video, I feel that when scrolling horizontally it is not very smooth. Is there any way to improve it?

ManhNTX: 1. Yeah! 
hidden - The overflow is clipped, and the rest of the content will be invisible (it mean cant scroll to visible all content)
auto - The overflow is clipped, and a scrollbar is added to see the rest of the content, but it adds scrollbars only when necessary
You can check link in here: 
 https://www.w3schools.com/css/css_overflow.asp
 2. Yeah I add some properties for smooth scroll animation and re-upload video. Please check this

dab246 (approved):

PR #698, opened 2022-07-05
TF-675 Improve upload multiple attachments

dab246:
@hoangdat @ManhNTX  Please review it

- Mobile

https://user-images.githubusercontent.com/80730648/177329499-8d140285-4a0a-4290-86ba-2cbd50d70401.mov

- Web

https://user-images.githubusercontent.com/80730648/177380329-1b7589a6-1b3c-47b4-bea6-1451b5656118.mov


lib/features/composer/presentation/composer_controller.dart
```diff
@@ -438,8 +433,8 @@ class ComposerController extends BaseController {
     if (identitySelected.value?.replyTo?.isNotEmpty == true) {
       listReplyToEmailAddress = identitySelected.value!.replyTo!;
     }
-    final generatePartId = PartId(_uuid.v1());
-    final generateBlobId = Id(_uuid.v1());
+    final generatePartId = PartId(const Uuid().v1());
```
hoangdat, line 436: Why you need to change that? IMO, you will have trouble with it in test, can not mock, test it

dab246: We don't need to worry about it. Since the controller it manages quite a lot, writing unit tests is not necessary here. On the other hand `Uuid ` is not too complicated so we don't need to create `singleton` for it

dab246: Ok, I made a mistake here, I'll fix it

hoangdat (approved):

ManhNTX (approved):

hoangdat (approved):

PR #712, opened 2022-07-12
TF-61 Compose email with text style for web

dab246:
- Supported style `bold, italic, underline, and default (no select any option)`


lib/features/composer/presentation/model/rich_text_option.dart
```diff
@@ -0,0 +1,30 @@
+
+import 'package:html_editor_enhanced/html_editor.dart';
+import 'package:model/mailbox/select_mode.dart';
+
+class RichTextOption {
+  final ButtonType buttonType;
+  final SelectMode selectMode;
```
hoangdat, line 7: what is `selectMode` in this case? you mean `selected` item in list?

lib/features/composer/presentation/controller/rich_text_controller.dart
```diff
@@ -0,0 +1,42 @@
...
+  void toggleRichTextButton() {
+    final currentState = formattingOptionsState.value;
+    final newState = currentState == SelectMode.ACTIVE
+        ? SelectMode.INACTIVE
+        : SelectMode.ACTIVE;
+    formattingOptionsState.value = newState;
+  }
+
+  bool get isFormattingOptionsEnabled =>
+      formattingOptionsState.value == SelectMode.ACTIVE;
```
hoangdat, line 27: IMO, you should add another `selectMode` for composer in separate with other selectMode

hoangdat:
What happens with the style button when you attach files?

lib/features/composer/presentation/model/rich_text_option.dart
dab246: yes

lib/features/composer/presentation/controller/rich_text_controller.dart
dab246: I think it is not so important here. SelectMode represents a state object. It's too obvious when setting an object's state.

dab246:
> What happens with the style button when you attach files?

It will be under `attachments` when you display it.

dab246:
> how about apply style when select text?

You can `select` text and apply styles to it. Take a look at the `demo video'

lib/features/composer/presentation/controller/rich_text_controller.dart
hoangdat: because when we have other change for the list or the composer, I don't want it impact to other

dab246: What are other changes? What options do you have for it? Please clarify what you mean here?

dab246: Ok. I created a separate state object for it
```dart
enum FormattingOptionsState {
   enable,
   disable
}
```

dab246: According to the design we don't need this object

lib/features/composer/presentation/model/rich_text_option.dart
```diff
@@ -0,0 +1,41 @@
+
+import 'dart:ui';
+
+import 'package:core/presentation/extensions/color_extension.dart';
+import 'package:html_editor_enhanced/html_editor.dart';
+import 'package:model/mailbox/select_mode.dart';
+
+class RichTextOption {
+  final ButtonType buttonType;
+  final SelectMode state;
```
hoangdat, line 10: it is still here :(

dab246: I updated it. Please review again.

hoangdat (approved):

ManhNTX (approved):

PR #718, opened 2022-07-13
Reset _isLoadingMore when change mailbox to avoid blocking load more

hoangdat:


hoangdat:
@ManhNTX please review it

ManhNTX (approved):

PR #731, opened 2022-07-18
TF-725 Add inline image to composer web

dab246:
### Resolve
- Use `cid` with `blod_id`
- Only error showing on `Gmail`

Screen Shot 2022-07-20 at 11 16 27

### Result Demo

https://user-images.githubusercontent.com/80730648/179956799-8d214592-8a43-45f2-939a-273a46ec8c93.mp4


hoangdat:
@dab246 what any issue in the eml? Why `gmail` and `apple mail` can not show our inline image :)

hoangdat:
after discussion with designer, please remove the image picker at the moment, we will support it later @dab246

lib/features/composer/domain/usecases/download_image_as_base64_interactor.dart
```diff
@@ -0,0 +1,23 @@
...
+import 'package:dartz/dartz.dart';
+import 'package:tmail_ui_user/features/composer/domain/repository/composer_repository.dart';
+import 'package:tmail_ui_user/features/composer/domain/state/download_image_as_base64_state.dart';
+
+class DownloadImageAsBase64Interactor {
+  final ComposerRepository _composerRepository;
+
+  DownloadImageAsBase64Interactor(this._composerRepository);
+
+  Future<Either<Failure, Success>> execute(String url) async {
```
hoangdat, line 11: I think you should you Stream, first item is Loading state. Maybe we should showing loading in the position of inline image when downloading? WDYT?

dab246: Ok. I think that's quite reasonable.

hoangdat (approved):

PR #741, opened 2022-07-20
hot fix crash app on android when loading webview

ManhNTX:
https://user-images.githubusercontent.com/107173849/179885716-ca2dd90f-fc31-46ba-bcee-78735ff9c824.mov

### DIFF
--- a/docs/adr/0009-on-before-unload-not-support-indexed-db.md
+++ b/docs/adr/0009-on-before-unload-not-support-indexed-db.md
@@ -1,4 +1,4 @@
-# 8. Unable to save data to indexedDB in onBeforeUnload state of web
+# 9. Unable to save data to indexedDB in onBeforeUnload state of web
 
 Date: 2022-07-11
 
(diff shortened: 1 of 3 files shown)


docs/adr/0009-on-before-unload-not-support-indexed-db.md
```diff
@@ -1,4 +1,4 @@
-# 8. Unable to save data to indexedDB in onBeforeUnload state of web
+# 9. Unable to save data to indexedDB in onBeforeUnload state of web
```
hoangdat, line 1: it seems you create this file manually

docs/adr/0010-crash-app-when-open-webview-on-android-device.md
```diff
@@ -0,0 +1,29 @@
+# 10. Crash app when open web_view on android device
+
+Date: 2022-07-20
+
+## Status
+
+Open
+
+## Context
```
hoangdat, line 9: please note the flutter version which we face the problem

hoangdat (approved):

PR #757, opened 2022-07-26
add insert inline image function for tablet and mobile

ManhNTX:
https://user-images.githubusercontent.com/107173849/180972713-90f7bf07-5956-4905-b1c9-fbddf5adedf0.MP4
Screen Shot 2022-07-26 at 16 22 32
Screen Shot 2022-07-26 at 16 21 46

https://user-images.githubusercontent.com/107173849/180984656-91cb08c4-9176-4543-b4da-e8a168bbf147.mov

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -139,7 +139,7 @@ class ComposerController extends BaseController {
       log('ComposerController::_getEmailBodyText():WEB: newContentHtml: $newContentHtml');
       return newContentHtml;
     } else {
-      final contentHtml = await htmlEditorApi?.getText() ?? '';
+      final contentHtml = (await htmlEditorApi?.getText() ?? '').replaceAll('&quot;', '"');
       log('ComposerController::_getEmailBodyText():MOBILE: $contentHtml');
       final newContentHtml = contentHtml.removeEditorStartTag();
       if (changedEmail) {
@@ -247,13 +247,23 @@ class ComposerController extends BaseController {
             selectIdentity(listIdentities.first);
           }
         } else if (success is DownloadImageAsBase64Success) {
-          richTextWebController.insertImage(
-              InlineImage(
-                  ImageSource.local,
-                  fileInfo: success.fileInfo,
-                  cid: success.cid,
-                  base64: success.imageAsBase64),
-              maxWithEditor: maxWithEditor);
+          if(kIsWeb) {
+            richTextWebController.insertImage(
+                InlineImage(
+                    ImageSource.local,
+                    fileInfo: success.fileInfo,
+                    cid: success.cid,
+                    base64: success.imageAsBase64),
+                maxWithEditor: maxWithEditor);
+          } else {
+            richTextMobileTabletController.insertImage(
+                InlineImage(
+                    ImageSource.local,
+                    fileInfo: success.fileInfo,
+                    cid: success.cid,
+                    base64: success.imageAsBase64),
+                maxWithEditor: maxWithEditor);
+          }
           maxWithEditor = null;
         }
       });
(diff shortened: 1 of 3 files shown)


core/lib/presentation/utils/html_transformer/html_template.dart
```diff
@@ -30,6 +30,7 @@ String generateHtml(String content, {
   String? resize,
   String? cursor,
 }) {
+  content = content.replaceAll('&quot;', '');
```
hoangdat, line 33: WDYT @dab246 ? any side effect?

dab246: I don't know what your intention is when replacing that character. IMO it will remove the apostrophe " characters in the email body. You should double check it.

hoangdat: please more detail on the case we need to check.

hoangdat: As Manh explain that will remove the redundant `&quot;` for the base 64, ....bla bla bla

dab246: Please provide details about this check. I don't think it's like that?

ManhNTX: @dab246  When using function getText of enough_html_editor, `\"` will replace to `&quot;`. So when you send body content to the server, it will be wrong. So I remove it for that. I change this code to the below. Please check this.
https://github.com/linagora/tmail-flutter/pull/757/files#diff-dfc6275288275045d8a26d4b6c04d7b180f0ccd025f6f1564400a14b744c5d18R142

ManhNTX: here is image code of package
Screen Shot 2022-07-26 at 17 12 10

dab246: So why do you replace it with a blank character? `content = content.replaceAll('&quot;', '');`. If you do so, double-quoted email messages will not be displayed properly.

dab246 (commented):
In the demo video, I saw you upload 3 photos, 1 image after the word `text`. But don't see it? Is that so? Is that some kind of error?

core/lib/presentation/utils/html_transformer/html_template.dart
dab246: It can affect:
- `saveDraft` when the two content before and after composing email are different.
- The content when `reply`, `forward` email is incorrect

ManhNTX: @dab246 You are right. I changed it please review it

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1111,16 +1121,41 @@ class ComposerController extends BaseController {
...
   void insertImage(BuildContext context, {double? maxWithEditor}) async {
-    await InsertImageDialogBuilder(
-        context,
-        insertActionCallback: (inlineImage) {
-          log('ComposerController::insertImage(): ${inlineImage.source}|$maxWithEditor');
-          if (BuildUtils.isWeb) {
-            this.maxWithEditor = maxWithEditor;
-            _insertImageOnWeb(inlineImage);
+    if(kIsWeb) {
+      await InsertImageDialogBuilder(
```
hoangdat, line 1125: Please remove it also :)

core/lib/presentation/utils/html_transformer/html_template.dart
ManhNTX: > It can affect:
> 
> * `saveDraft` when the two content before and after composing email are different.
> * The content when `reply`, `forward` email is incorrect

Yeb I fixed it

ManhNTX:
> In the demo video, I saw you upload 3 photos, 1 image after the word `text`. But don't see it? Is that so? Is that some kind of error?

Yeb I fixed it

dab246 (approved):

hoangdat (approved):

PR #772, opened 2022-07-28
TF-761 Down size of base 64 in composing email and make it smooth

dab246:
https://user-images.githubusercontent.com/80730648/181467140-db0dd724-93e9-49e3-adb8-d89e6cb69609.mp4

https://user-images.githubusercontent.com/80730648/181467577-04a26897-b570-4a7c-a7b8-aab712d6e345.mp4


docs/adr/0011-down-size-of-base64-image.md
```diff
@@ -0,0 +1,25 @@
...
+For large images (several MB or more) converting to base64 will take a long time and exceed the storage limit of the string. 
+Leads to long waits on the main thread, causing a feeling of not being smooth on mobile devices with weak configuration.
+
+## Decision
+
+We can reduce the image `size` and `quality` when displaying.
+- For mobile: 
+  + Use [flutter_image_compress](https://pub.dev/packages/flutter_image_compress) to compress image use native (Objective-C/Kotlin) with faster speed.
+- For web:
+  + Use [image_picker](https://pub.dev/packages/image_picker) to resize image when pick image
```
hoangdat, line 20: You mean, we will reduce the size before uploading? If yes, IMO it is not a good idea. Please keep the message as plain as the people wrote. We should only reduce the size when rendering only.

dab246: - IMO It's pretty cool solution for web platform. Because if we proceed to resize it, it will take a long time, moreover, our web has not yet implemented multithreading for it, so it is easy to block the main thread.
- This is an `inline` image in the text, so in my opinion users will not pay too much attention to the quality. Because they can send high quality photos through `attachments`.

ManhNTX (approved):

core/lib/data/utils/compress_file_utils.dart
```diff
@@ -0,0 +1,40 @@
...
+import 'dart:typed_data';
+
+import 'package:core/utils/app_logger.dart';
+import 'package:core/utils/build_utils.dart';
+import 'package:flutter_image_compress/flutter_image_compress.dart';
+
+class CompressFileUtils {
+
+  /// Maximum image size displayed (Unit: KB)
+  static const int MAXIMUM_IMAGE_SIZE = 300;
```
hoangdat, line 11: `MAXIMUM_IMAGE_SIZE_KB`

hoangdat: why you choose `300 Kb` is maximum?

core/lib/data/utils/compress_file_utils.dart
```diff
@@ -0,0 +1,40 @@
...
+  /// Quality image 0 - 100
+  static const int QUALITY_DEFAULT = 50;
+
+  bool _exceedMaximumImageSize(Uint8List bytesData) {
+    final maximumSizeBytes = MAXIMUM_IMAGE_SIZE * 1024;
+    final sizeBytesData = bytesData.length;
+    log('DownloadClient::exceedMaximumImageSize(): maximumSizeBytes: $maximumSizeBytes');
```
hoangdat, line 18: please carefully about the log, should be exactly class name for easier in debug. Please use `live templeate`.

core/lib/data/utils/compress_file_utils.dart
```diff
@@ -0,0 +1,40 @@
...
+    log('DownloadClient::exceedMaximumImageSize(): sizeBytesData: $sizeBytesData');
+    return sizeBytesData > maximumSizeBytes;
+  }
+
+  Future<Uint8List> compressBytesDataImage(Uint8List bytesData, {int? maxWidth}) async {
+    if (_exceedMaximumImageSize(bytesData)) {
+      log('DownloadClient::_compressImageData(): BEFORE_COMPRESS: bytesData: ${bytesData.lengthInBytes}');
```
hoangdat, line 25: idem

core/lib/data/utils/compress_file_utils.dart
```diff
@@ -0,0 +1,40 @@
...
+      if (BuildUtils.isWeb) {
+        return bytesData;
```
hoangdat, line 27: because we compressed when picked image, so no need to compress at this point? Am I right?

hoangdat: if yes, how the picker compress image for web? please address it in `ADR` also.

core/lib/data/network/download/download_client.dart
```diff
@@ -28,14 +35,72 @@ class DownloadClient {
     return (responseBody as ResponseBody);
   }
 
-  Future<String?> downloadImageAsBase64(String url) async {
+  Future<String?> downloadImageAsBase64(
```
hoangdat, line 38: IMO, you should define a flag to config compress or not. WDYT?

core/lib/data/utils/compress_file_utils.dart
dab246: As your previous comment. I canceled all web operations. Web does not compress anything

core/lib/data/network/download/download_client.dart
dab246: That is not necessary. We already have a check size flag.

core/lib/data/utils/compress_file_utils.dart
dab246: It's just a copy error.

core/lib/data/utils/compress_file_utils.dart
dab246: IMO 300Kb is what I test to set a stable image quality. We can change it. And of course if you have any suggestions please comment here.

core/lib/data/utils/compress_file_utils.dart
hoangdat: I agreed with the compress in picking step. dont' remove it

core/lib/data/utils/compress_file_utils.dart
hoangdat: please use `live template` to avoid that. please! please!

core/lib/data/utils/compress_file_utils.dart
hoangdat: ok remove compress for web

core/lib/data/network/download/download_client.dart
dab246: oke

hoangdat (approved):

Issue #654, opened 2022-06-15
Show the progress of download attachment in app


hoangdat:
### Desc:

At the moment, when we download an attachment in browser, it takes some delay (downloading bytes of attachment) before show in bottom of browser.

### DoD:
- [ ] Designer: design the view to indicate user downloading attachment progress
- [ ] Dev: implement progress in web-app

hoangdat:
@chibenwa please let us more related information

chibenwa:
The dowload response should have a content length header allowing you to know the size of what you need to download.

Then you just need to count the downloaded bytes.

> => I don't know if the server side currently supports getting downloadUrl public so we can do that?

Thats off spec...

hoangdat:
so you mean, we can not transfer the download task to browser like `openpaas`. We need to show the progress our-self. Am I right?

chibenwa:
Right.

hoangdat:
- Progress in toast

dab246:
1 HD

PR #809, opened 2022-08-06
TF-654 Show the progress of download attachment in app

dab246:
#654 

### Dependency 

- Upgrade version `dio` for `jmap-dart-client` at [PR#33](https://github.com/linagora/jmap-dart-client/pull/33)

### Resolved

https://user-images.githubusercontent.com/80730648/183249742-17d4e1fc-ed2c-4744-9316-428e7af32440.mov

### DIFF
--- a/core/lib/data/network/download/download_manager.dart
+++ b/core/lib/data/network/download/download_manager.dart
@@ -1,13 +1,12 @@
 import 'dart:async';
 import 'dart:io';
+import 'dart:typed_data';
 
-import 'package:core/data/network/dio_client.dart';
 import 'package:core/data/network/download/download_client.dart';
 import 'package:core/data/network/download/downloaded_response.dart';
 import 'package:core/domain/exceptions/download_file_exception.dart';
 import 'package:core/utils/app_logger.dart';
 import 'package:dio/dio.dart';
-import 'package:http/http.dart' as http;
 import 'package:http_parser/http_parser.dart';
 import 'package:universal_html/html.dart' as html;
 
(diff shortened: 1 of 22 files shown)


core/lib/data/network/download/download_manager.dart
```diff
@@ -82,39 +81,28 @@ class DownloadManager {
...
-      }
+      html.document.body?.children.remove(anchor);
+      html.Url.revokeObjectUrl(url);
 
+      return true;
+    } catch (error) {
+      log('DownloadManager::createAnchorElementDownloadFileWeb(): ERROR: $error');
       return false;
-    } catch (exception) {
-      throw exception;
```
hoangdat, line 117: why you not throw exception?

ManhNTX (approved):

core/lib/data/network/download/download_manager.dart
dab246: I'm not handling any error events here at the moment. But maybe I'll add it so I can use it later.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -578,6 +598,7 @@ class MailboxDashBoardController extends ReloadableController {
     _connectivityStreamSubscription.cancel();
     _progressStateController.close();
     _isolateExecutor.dispose();
+    Get.delete<DownloadController>();
```
hoangdat, line 601: why we need to do that? `Get` does not support us?

dab246: This controller is not attached to the view so `Getx` will not automatically clean it.

hoangdat (approved):

PR #896, opened 2022-08-29
[Temporary] Fix error with inconsistency in cache and add ReSync function

hoangdat:
### Problem:
- When set the `filterAction` (Fx: Unread), then open a mailbox A in the first time, all and only `UnRead emails` are cache.
- Then, Go to other mailbox, un-select `filterAction`, then back to A, cache will show unread, other read mail is not loaded.

### Fix:
- Reload from the network 
And add `ReSync` to fix the cache inconsistency

### DIFF
--- a/lib/features/thread/data/datasource_impl/local_thread_datasource_impl.dart
+++ b/lib/features/thread/data/datasource_impl/local_thread_datasource_impl.dart
@@ -55,6 +55,19 @@ class LocalThreadDataSourceImpl extends ThreadDataSource {
   }
 
   @override
+  Future<void> deleteEmails({MailboxId? inMailboxId}) {
+    if (inMailboxId == null) {
+      return _emailCacheManager.clearAll();
+    }
+    return _emailCacheManager
+      .deleteWhere((emailCache) {
+        return emailCache.mailboxIds != null
+          && emailCache.mailboxIds!.containsKey(inMailboxId.id.value)
+          && emailCache.mailboxIds![inMailboxId.id.value] == true;
+      });
+  }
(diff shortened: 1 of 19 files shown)


lib/features/thread/data/datasource_impl/local_thread_datasource_impl.dart
```diff
@@ -54,6 +54,19 @@ class LocalThreadDataSourceImpl extends ThreadDataSource {
     });
   }
 
+  @override
+  Future<void> deleteEmails({MailboxId? inMailboxId}) {
+    if (inMailboxId == null) {
+      return _emailCacheManager.clearAll();
```
dab246, line 60: why must `clearAll`?

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -374,34 +399,50 @@ class ThreadController extends BaseController {
...
+          before: emailList.last.receivedAt
         );
       case FilterMessageOption.unread:
         return EmailFilterCondition(
-            inMailbox: mailboxDashBoardController.selectedMailbox.value?.id,
-            notKeyword: KeyWordIdentifier.emailSeen.value,
-            before: isLoadMore ? emailList.last.receivedAt : null
+          inMailbox: mailboxDashBoardController.selectedMailbox.value?.id,
+          notKeyword: KeyWordIdentifier.emailSeen.value,
+          before: emailList.last.receivedAt
```
dab246, line 419: same

dab246 (commented):
Quick review

PR #968, opened 2022-09-21
TF-880, 912: Apply rich text toolbar at keyboard

ManhNTX:
- Apply toolbar composer and signature `htmlBody` of identity:

https://user-images.githubusercontent.com/107173849/193200290-b8aa6335-7193-4b0e-9532-eeea32ebba55.mov

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -42,6 +42,10 @@ dependencies:
   forward:
     path: forward
 
+  rich_text_composer:
+    git:
+      url: https://github.com/linagora/rich-text-composer.git
+      ref: apply_rich_text_bottom_sheet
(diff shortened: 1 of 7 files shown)


hoangdat:
image

Please apply the background color with the same color of keyboard

pubspec.yaml
```diff
@@ -42,6 +42,10 @@ dependencies:
   forward:
     path: forward
 
+  rich_text_composer:
+    git:
+      url: https://github.com/linagora/rich-text-composer.git
+      ref: apply_rich_text_bottom_sheet
```
hoangdat, line 48: should be master when `apply_rich_text_bottom_sheet` merge?

Issue #971, opened 2022-09-22
Crash vacation view when reload web in setting


dab246:
### DoD

https://images.zenhubusercontent.com/378835363/f28a34f0-7495-4fc9-913e-b71a0e710f6a/screen_recording_2022_09_22_at_19_14_34.mov

PR #972, opened 2022-09-22
TF-971 Fix crash when reload setting view on web

dab246:
#971 

### Causes

- Due to an error thrown when call method `injectAutocompleteBindings`

### Resolved

https://user-images.githubusercontent.com/80730648/191743284-e1e4f290-dc99-4c4e-bf8f-bf351c4b9996.mov


lib/features/base/base_controller.dart
```diff
@@ -66,4 +72,17 @@ abstract class BaseController extends GetxController with MessageDialogActionMix
       FpsManager().removeFpsCallback(fpsCallback!);
     }
   }
+
+  void injectAutoCompleteBindings(Session? session, AccountId? accountId) {
+    if (session == null || accountId == null) {
+      return;
+    }
```
hoangdat, line 79: IMO. it is useless, because the `requireCapability` handle this

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -721,30 +727,58 @@ class ComposerController extends BaseController {
...
-    log('ComposerController::getAutoCompleteSuggestion(): $word | $_contactSuggestionSource');
+  @override
+  void injectAutoCompleteBindings(Session? session, AccountId? accountId) {
+    if (session == null || accountId == null) {
+      return;
+    }
+    try {
+      super.injectAutoCompleteBindings(session, accountId);
+      _getAutoCompleteWithDeviceContactInteractor = Get.find<GetAutoCompleteWithDeviceContactInteractor>();
+      _getAutoCompleteInteractor = Get.find<GetAutoCompleteInteractor>();
```
hoangdat, line 738: How about the app run on the web? (no GetAutoCompleteWithDeviceContactInteractor but GetAutoCompleteInteractor), exception will throw after line 737?

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -721,30 +727,58 @@ class ComposerController extends BaseController {
...
+  }
 
-    if (!Get.isRegistered<GetAutoCompleteWithDeviceContactInteractor>() ||
-        !Get.isRegistered<GetAutoCompleteInteractor>()) {
-      mailboxDashBoardController.injectAutoCompleteBindings();
+  Future<List<EmailAddress>> getAutoCompleteSuggestion(String word) async {
+    log('ComposerController::getAutoCompleteSuggestion(): $word | $_contactSuggestionSource');
+    try {
+      _getAutoCompleteWithDeviceContactInteractor = Get.find<GetAutoCompleteWithDeviceContactInteractor>();
+      _getAutoCompleteInteractor = Get.find<GetAutoCompleteInteractor>();
```
hoangdat, line 748: idem

ManhNTX (approved):

lib/features/base/base_controller.dart
dab246: I think it's necessary. Because in the case of `reload web`. At `reload` time `session` and `accountId` will be null, so it will continue with the `requireCapability ` statement and will be fired `exception` in `try/catch`. I don't want to waste time for the system to process when it is certain that `session` and `accountId` are null. WDYT? @hoangdat h

lib/features/composer/presentation/composer_controller.dart
dab246: Don't worry, `_getAutoCompleteWithDeviceContactInteractor` always injected.

lib/features/composer/presentation/composer_controller.dart
dab246: same

lib/features/base/base_controller.dart
hoangdat: okie, but for the case `session` and `accountId` not null, system need to spend time to check these conditions 2 times :). It up to you. If you keep that, please `log` before return. Thank

PR #1016, opened 2022-09-28
Added workflow for CI/CD

tk-nguyen:
Initial work to move from Bitrise to Github Actions


hoangdat (approved):

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,94 @@
+on:
+  push:
+    branches:
+      - "release"
```
hoangdat: i want to deploy in master, after tagged

hoangdat (approved):

dab246 (commented):
Please add all file `.lock` to `.gitignore` file

hoangdat:
@tk-nguyen please explain again about `.lock` file. Why we need it?

tk-nguyen:
Lock files are used to pin dependencies so when there are upgrades to those dependencies, the builds will not break. They're also used for the cache in CI.

hoangdat:
> Lock files are used to pin dependencies so when there are upgrades to those dependencies, the builds will not break. They're also used for the cache in CI.

You mean we will check the lock file to fetch dependencies or not?

hoangdat:
> Yes. When you changed `pubspec.yaml` the lock file will change when you run `flutter pub get`. Same with `Podfile` and `Podfile.lock`.

So, the problem here, some dependencies we use github path to import in this project, how to get the update from these libraries?

image

hoangdat:
One more question: job with cache is faster than job without cache? How long? 
What happen if we don't use `lock` files?

tk-nguyen:
> One more question: job with cache is faster than job without cache? How long? What happen if we don't use `lock` files?

About 2-3 minutes. If there's no lock file, then `flutter` will have to download all dependencies with the version specified in `pubspec.yaml`. If there are dependencies with ^<version> annotation, `flutter` will download the latest version, which might break our builds if there's breaking changes.

PR #1026, opened 2022-09-29
TF-1009 cache not synchronize enough:

hoangdat:
- Cause: combine with cache for only one mailbox while `changes` response in all
- Resolve: get all email cache for synchronizing

https://user-images.githubusercontent.com/6462404/193189780-9e815610-fc52-4906-93cd-c94c668fe4a0.mov

### DIFF
--- a/lib/features/thread/data/repository/thread_repository_impl.dart
+++ b/lib/features/thread/data/repository/thread_repository_impl.dart
@@ -88,7 +88,7 @@ class ThreadRepositoryImpl extends ThreadRepository {
       );
     } else {
       if (networkEmailResponse != null) {
-        log('ThreadRepositoryImpl::getAllEmail(): filter = ${emailFilter?.mailboxId} no local state -> update from network: ${networkEmailResponse?.state}');
+        log('ThreadRepositoryImpl::getAllEmail(): filter = ${emailFilter?.mailboxId} no local state -> update from network: ${networkEmailResponse.state}');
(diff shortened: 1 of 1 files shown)


ManhNTX (approved):

lib/features/thread/data/repository/thread_repository_impl.dart
```diff
@@ -348,4 +280,59 @@ class ThreadRepositoryImpl extends ThreadRepository {
...
+  Future<void> _synchronizeCacheWithChanges(
+    AccountId accountId,
+    State currentState,
+    {
+      Properties? propertiesCreated,
+      Properties? propertiesUpdated,
+    }
+  ) async {
+    final localEmailList = await mapDataSource[DataSourceType.local]!
+        .getAllEmailCache();
```
dab246, line 293: IMO, if you do that you will have to get all the data in the email table out. Is that necessary? Does it take too long to query? Why don't you add `filter` conditions.

hoangdat: yes, it will take time to query. But if not, it will be wrong in the cache, some messages never be updated after state update (some issues you can see before).

hoangdat: this list will be used to `combine` between cache and `network response` (cache hit or not) to update cache. So, add `filter` here will make the cache update wrong

dab246 (approved):

PR #1044, opened 2022-10-04
Force synchronize cache when state available 

hoangdat:
https://user-images.githubusercontent.com/6462404/193951012-dfe4aa33-80a3-40d2-9e0c-8bc1e2ea75b0.mov

### DIFF
--- a/lib/features/thread/data/repository/thread_repository_impl.dart
+++ b/lib/features/thread/data/repository/thread_repository_impl.dart
@@ -78,11 +78,12 @@ class ThreadRepositoryImpl extends ThreadRepository {
       yield localEmailResponse;
     }
 
+    if (networkEmailResponse != null) {
+      await _updateEmailCache(newCreated: networkEmailResponse.emailList);
+    }
+
     if (localEmailResponse.hasState()) {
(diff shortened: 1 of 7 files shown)


ManhNTX (approved):

lib/features/thread/data/repository/thread_repository_impl.dart
```diff
@@ -78,8 +78,11 @@ class ThreadRepositoryImpl extends ThreadRepository {
       yield localEmailResponse;
     }
 
-    if (localEmailResponse.hasState() && networkEmailResponse == null) {
+    if (localEmailResponse.hasState()) {
       log('ThreadRepositoryImpl::getAllEmail(): filter = ${emailFilter?.mailboxId} local has state: ${localEmailResponse.state}');
+      if (networkEmailResponse != null) {
+        await _updateEmailCache(newCreated: networkEmailResponse.emailList);
```
dab246, line 84: - As I understand it, we will always update the email cache again when `networkReponse` is different from `null`. And should we update `state` again?.
- Also you should execute that statement before checking the condition `if/else`


Screen Shot 2022-10-05 at 09 19 49

hoangdat: fixed

dab246 (approved):
/lgtm

Issue #1045, opened 2022-10-04
[Load mail] Can't  load more than 20 mails


haftus98:
[Untitled_ Oct 04,2022 6_11 PM.webm](https://user-images.githubusercontent.com/97084773/193804836-0b7bc18a-a13e-41af-aa11-70956bfb7b38.webm)
 I couldn't load more mail until I removed the last ones

dab246:
2 HD

PR #1048, opened 2022-10-05
#1045 Remove wrong conditions to append more items after loading more

hoangdat:
### Root cause:
- wrong condition: if `moreItems` have any item in `emailList` will be ignored to append.

### Problem:

https://user-images.githubusercontent.com/6462404/194032802-7ba98576-853e-45bd-aa9a-e233f4725838.mov

### Solved:

https://user-images.githubusercontent.com/6462404/194032821-b573138a-d002-456e-9a54-9ccb47238f16.mov


dab246 (commented):
As I understand you are processing filtering at `Presentation` layer and not processing at `Data` layer anymore. Is not it?

dab246 (approved):

PR #1103, opened 2022-10-21
[DO NOT MERGE] Add toast to verify fetching email error

hoangdat:


hoangdat:
can not catch error from stream in interactor, no need this PR.

Issue #1115, opened 2022-10-26
Support CannotCalculateChanges


chibenwa:
### Description

Some users reported the unability to synchronise changes any longer (ie press reload button and nothing happens)

They needed to logout/login to get their newer mails.

### Expected result

When I press the reload button I should always get my newer mails.

### Current behavior

https://user-images.githubusercontent.com/6928740/197935037-70e7ec4f-8a48-4470-abf2-0ed35fb53f8a.mp4

If, for any reason, the server returns a "cannotCalculateChanges" TMail app locks itself and never ever re-updates the view.

#### Reproduction Steps

 1. Go in the trash, in another mail app, select 300+ emails and delete them. This records a number of changes in a single line greater than the resync capabilities of TMail, as this is a signle line, James can't slice it, generating intermediate states, Thus it would return an error on the next `Email/changes`.
 2. Send a new email. It is well received and synced in TB
 3. Now reload the Tmail Web App
 
Nothing happens.

Query responses shows:

```
{
	"sessionState": "2c9f1b12-b35a-43e6-9af2-0106fb53a943",
	"methodResponses": [
		["error", {
			"type": "cannotCalculateChanges",
			"description": "Current change collector limit 256 is exceeded by a single change, hence we cannot calculate changes."
		}, "c0"],
		["error", {
			"type": "invalidResultReference",
			"description": "Failed resolving back-reference: List((,List(JsonValidationError(List(MethodCallId(c0) references a MethodName(error) method),List()))))"
		}, "c1"],
		["error", {
			"type": "invalidResultReference",
			"description": "Failed resolving back-reference: List((,List(JsonValidationError(List(MethodCallId(c0) references a MethodName(error) method),List()))))"
		}, "c2"]
	]
}
```

### Acceptance criteria

When TMail APP encounters a `cannotCalculateChanges` error, TMail should throw its cache away, and resynchronise the current view.

dab246:
### DoD
- `Clean cache` and call `GetAll` again

dab246:
2 HD

PR #1124, opened 2022-10-27
TF-1115 Support cannot calculate changes

dab246:
#1115 

### Resolved

https://user-images.githubusercontent.com/80730648/198519255-736b2976-790a-4eca-b42b-3bd466ef3796.mp4


lib/features/thread/presentation/thread_controller.dart
```diff
@@ -345,6 +347,51 @@ class ThreadController extends BaseController {
...
+          textActionColor: Colors.white,
+          maxWidth: _responsiveUtils.getMaxWidthToast(currentContext!));
+      }
+    }
+  }
+
+  void _cleanEmailCacheAndGetAllEmailAgainAction() async {
+    await _cachingManager.cleanEmailCache();
+    _getAllEmail();
+  }
```
hoangdat, line 393: no need to combine it in method

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -345,6 +347,51 @@ class ThreadController extends BaseController {
     viewStateWorker.call();
   }
 
+  void _handleErrorGetOrRefreshChangesEmail(Exception? exception) {
+    logError('ThreadController::_handleErrorGetOrRefreshChangesEmail(): $exception');
+
+    if (exception is CannotCalculateChangesMethodResponseException) {
+      _cleanEmailCacheAndGetAllEmailAgainAction();
```
hoangdat, line 354:
```suggestion
      await _cachingManager.cleanEmailCache();
      _getAllEmail();
```

lib/features/base/base_controller.dart
```diff
@@ -21,11 +21,13 @@ abstract class BaseController extends GetxController
   FpsCallback? fpsCallback;
 
   void consumeState(Stream<Either<Failure, Success>> newStateStream) async {
-    newStateStream.listen(
-      (state) => onData(state),
-      onError: (error) => onError(error),
-      onDone: () => onDone()
-    );
+    newStateStream
+      .handleError((error) => onError(error))
```
hoangdat, line 25:
```suggestion
      .handleError((error, stackTrace) {
          //log error and stackTrace here
          onError(error));
      }
```

lib/features/base/base_controller.dart
```diff
@@ -21,11 +21,13 @@ abstract class BaseController extends GetxController
...
+      .listen(
+        (state) => onData(state),
+        onError: (error) => onError(error),
```
hoangdat, line 28: why we need 
```
   .handleError((error) => onError(error))
```
when we have `onError` here?

lib/features/thread/presentation/thread_controller.dart
dab246: We should separate the new function for better readability and handling later.

lib/features/base/base_controller.dart
dab246: Oh, that's right, it's not necessary

hoangdat: please add the log before the `onError`

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -281,18 +281,6 @@ class ComposerController extends BaseController {
...
-  void onError(error) {
-    if (currentOverlayContext != null && currentContext != null) {
-      _appToast.showToastWithIcon(
-          currentOverlayContext!,
-          textColor: AppColor.toastErrorBackgroundColor,
-          message: AppLocalizations.of(currentContext!).message_has_been_sent_failure,
-          icon: _imagePaths.icSendToast);
-    }
-    popBack();
-  }
```
hoangdat, line 294: Why you delete it? You mean Toast will show unknown error?

dab246: 1. It won't call here. Since we captured `try/catch` it in an `await` from the repository.
2. We also checked the error result from this interactor in the dashboard controller when it had `SendEmailFailure` error

hoangdat: got it

hoangdat (approved):

Issue #900, opened 2022-08-30
[Web] Not correct search result when close search box


hoangdat:
https://user-images.githubusercontent.com/6462404/188352117-4f506de9-18c7-4e63-a326-218ac191008f.mov

hoangdat:
- [ ] change `hasKeyword` to use `text` in Advanced search

dab246:
1 HD

dab246:
2 HD

haftus98:
[Uploading... Untitled_ Oct 11,2022 4_29 PM.webm]

Step 1: search sth
Step 2: remove the keyword then click enter => For now it's supposed to return the list of all mail but the result is the same as in filter applied result
Step3: Open the advanced search then click Search, now it returns the correct number

haftus98:
still bug on web browser google chrome Version 106.0.5249.119 V0.3.17
@hoangdat

dab246:
2 HD

PR #1131, opened 2022-10-30
tf-900 fix not correct search result when close search box

sherlockvn:
#900 
###Resolved

https://user-images.githubusercontent.com/43041967/198890038-a1d32ff4-03bb-4591-90b1-592cacc60a9f.mov


lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -92,6 +93,7 @@ class MailboxDashBoardController extends ReloadableController {
   final DownloadController downloadController = Get.find<DownloadController>();
   final NetworkConnectionController networkConnectionController = Get.find<NetworkConnectionController>();
   final AppGridDashboardController appGridDashboardController = Get.find<AppGridDashboardController>();
+  final MailboxController mailBoxController = Get.find<MailboxController>();
```
dab246, line 96: IMO, it's not advisable to do so. `DashboardController` is an intermediate object that listens for events from other controllers.

sherlockvn:
Currently it has a problem: when textfield for searching is empty, user click to searchBar icon, the mailbox isn't refreshed because of IMO, the search field is unfocused. I don't know how to resolve this ? do you have any suggestion ?

https://user-images.githubusercontent.com/43041967/198985914-4b7e4bdd-e338-44aa-b182-062cfd47dd51.mov

dab246:
> Currently it has a problem: when textfield for searching is empty, user click to searchBar icon, the mailbox isn't refreshed because of IMO, the search field is unfocused. I don't know how to resolve this ? do you have any suggestion ?
> 
>  Screen.Recording.2022-10-31.at.17.19.24.mov

When you click icon search, let check text search isEmpty or not? Is IsEmpty then call `dispatchAction(CleanSearchEmail());`

lib/features/mailbox_dashboard/presentation/controller/search_controller.dart
```diff
@@ -157,7 +157,6 @@ class SearchController extends BaseController {
         searchInputController.clear();
         cleanSearchFilter();
         searchFocus.unfocus();
-        searchState.value = searchState.value.disableSearchState();
```
sherlockvn, line 160: in order to receive onTap event from search icon in search bar, searchState have not to be disabled here but later in clearSearchFilter, is it ok that i remove this line?

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -422,6 +422,11 @@ class MailboxDashBoardController extends ReloadableController {
         && dashboardRoute.value == DashboardRoutes.emailDetailed;
   }
 
+  void clearSearchEmail() {
+    dispatchAction(ClearSearchEmailAction());
```
sherlockvn, line 426: by the way, what is different between action and state ?

lib/features/mailbox_dashboard/presentation/controller/search_controller.dart
dab246: If you need to remove it, you should double check the cases when searching.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
hoangdat: in common: 
- `*Action` is the state, but it come from the user interactions (tap on button, click, drag, ....sth like that)

dab246 (approved):
/lgtm

Issue #1008, opened 2022-09-28
Error when deleting mailbox


chibenwa:
### Description

```
Given mailbox `a`
And mailbox `a/b`

When I delete mailbox `b` in OpenPaaS
And then I delete mailbox `a` in TMail

Then Deletion fails in TMail
```

### Expected result

TMail should handle `Mailbox/set` destroy **notFound** response code. If it's not found, it no longer exist and we thus don't need to delete it. It's not a failure and we can continue the deletion chain by deleting `a`...

### Current behavior

https://user-images.githubusercontent.com/6928740/192670271-a3f1531f-121b-44bc-bcbf-0ced229b0ca4.mp4

dab246:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Tmail-Flutter-60d039db6ef62a0013432c14/issues/linagora/tmail-flutter/1008?planning-poker) with Zenhub @hoangdat @ManhNTX @sherlockvn

dab246:
1 HD

hoangdat:
1 HD

dab246:
1 HD

PR #1174, opened 2022-11-08
#1008 Add handler for SetError

hoangdat:
https://user-images.githubusercontent.com/107173849/201033786-033d5fb8-dcc9-46ee-b2c1-858ea6cfd21b.mov

### DIFF
--- a/lib/features/base/mixin/handle_error_mixin.dart
+++ b/lib/features/base/mixin/handle_error_mixin.dart
@@ -17,19 +17,22 @@ mixin HandleSetErrorMixin {
     if (notDestroyedError != null && notDestroyedError.isNotEmpty) {
       notDestroyedError.entries
         .map((e) => _handleInChain(e, notDestroyedHandlers))
-        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError));
+        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError))
+        .toSet();
     }
 
     if (notCreatedError != null && notCreatedError.isNotEmpty) {
       notCreatedError.entries
         .map((e) => _handleInChain(e, notCreatedHandlers))
-        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError));
+        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError))
+        .toSet();
     }
 
     if (notUpdatedError != null && notUpdatedError.isNotEmpty) {
       notUpdatedError.entries
         .map((e) => _handleInChain(e, notUpdatedHandlers))
-        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError));
+        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError))
+        .toSet();
     }
   }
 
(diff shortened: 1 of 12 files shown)


lib/features/mailbox/data/network/mailbox_api.dart
```diff
@@ -191,13 +194,38 @@ class MailboxAPI {
...
+
+  List<MapEntry<Id, SetError>> _handleDeleteMailboxResponse(SetMailboxResponse? response) {
+    final List<MapEntry<Id,SetError>> remainedErrors = [];
+    if (response != null) {
+      handleSetErrors(
+          notDestroyedError: response.notDestroyed,
+          notUpdatedError: response.notUpdated,
+          notCreatedError: response.notCreated,
+          notDestroyedHandlers: <SetMethodErrorHandler>{
+                (entry) {
```
dab246, line 215: remove indent

lib/features/mailbox/data/network/mailbox_api.dart
```diff
@@ -191,13 +194,38 @@ class MailboxAPI {
...
+              if (entry.value.type == SetError.notFound) {
```
dab246, line 216: IMO, you can write `return entry.value.type == SetError.notFound;`

lib/features/base/mixin/handle_error_mixin.dart
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    if (notUpdatedError != null && notUpdatedError.isNotEmpty) {
+      notUpdatedError.entries
+        .map((e) => _handleInChain(e, notUpdatedHandlers))
+        .map((optionError) => _handleRemainedError(unCatchErrorHandler, optionError));
+    }
+  }
+
+  Option<MapEntry<Id, SetError>> _handleInChain(MapEntry<Id, SetError> setError, Set<SetMethodErrorHandler>? handlers) {
```
dab246, line 36: Good, May I ask what type of variable is `Option`?

ManhNTX (approved):
LGTM

ManhNTX:
@hoangdat  Should we add unitTests for this feature?

lib/features/base/mixin/handle_error_mixin.dart
```diff
@@ -0,0 +1,53 @@
...
+    try {
+      handlers!.firstWhere((handler) => handler.call(setError));
+      return const None<MapEntry<Id, SetError>>();
+    } catch (e) {
+      logError('HandleSetErrorMixin::chainHandle(): [Exception] $e');
+      return Some<MapEntry<Id, SetError>>(setError);
```
sherlockvn, line 42: Can you explain this function ? What is Some ?

hoangdat: `Option`(Maybe) represent the absence of a value. Instead of returning one object when a function succeeds and null when it fails, your function should instead return an instance of an `Option`. 
```dart
class Some<A> extends Option<A> {
  final A _a;
  const Some(this._a);
  A get value => _a;
  @override B fold<B>(B ifNone(), B ifSome(A a)) => ifSome(_a);
  @override bool operator ==(other) => other is Some && other._a == _a;
  @override int get hashCode => _a.hashCode;
}

class None<A> extends Option<A> {
  const None();
  @override B fold<B>(B ifNone(), B ifSome(A a)) => ifNone();
  @override bool operator ==(other) => other is None;
  @override int get hashCode => 0;
}
```
`Some`: if value is available
`None`: if value not available.
For this case, `_handleInChain` will handle an error in chain, if error is handle successfully by one of the handler, no error will be remained -> return None. Otherwise, Some will be returned when no handler can handle the error. Using Option make us better to work in map. We don't need to care about it is null or not (avoid null-check boilerplate).

dab246 (approved):

PR #1219, opened 2022-11-18
Added compose file for quick local setup

tk-nguyen:
Also simplified the Dockerfile and moved `gzip` to a seperate `RUN` so the container can be restarted.

This compose file uses the `linagora/tmail-backend:memory-branch-master`image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,23 +23,7 @@ WORKDIR /app
 COPY . .
 
 # Precompile tmail flutter
-RUN cd core \
-  && flutter pub get \
-  && cd ../model \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd ../contact \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd ../rule_filter \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd ../forward \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd ../fcm \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd .. \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && flutter pub get && flutter pub run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart \
-  && flutter pub get && flutter pub run intl_generator:generate_from_arb --output-dir=lib/l10n --no-use-deferred-loading lib/main/localizations/app_localizations.dart lib/l10n/intl*.arb \
-  && flutter build web --profile
+RUN bash prebuild.sh && flutter build web --profile
 
 # Stage 2 - Create the run-time image
 FROM nginx:mainline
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -23,21 +23,7 @@ WORKDIR /app
...
-  && cd ../rule_filter \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd ../forward \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && cd .. \
-  && flutter pub get && flutter pub run build_runner build --delete-conflicting-outputs \
-  && flutter pub get && flutter pub run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart \
-  && flutter pub get && flutter pub run intl_generator:generate_from_arb --output-dir=lib/l10n --no-use-deferred-loading lib/main/localizations/app_localizations.dart lib/l10n/intl*.arb \
-  && flutter build web --profile
+RUN bash prebuild.sh && flutter build web --profile
```
chibenwa, line 26: Can't see prebuild.sh. Is that normal?

Dockerfile
```diff
@@ -49,4 +35,5 @@ COPY --from=build-env /app/build/web /usr/share/nginx/html
 EXPOSE 80
 
 # Before stating NGinx, re-zip all the content to ensure customizations are propagated
-CMD gzip -k -r /usr/share/nginx/html/ && nginx -g 'daemon off;'
+RUN gzip -k -f -r /usr/share/nginx/html/ 
+CMD ["nginx", "-g", "daemon off;"]
```
chibenwa, line 39: No this was intendeed to rezip assets upon container startup. So that you can override whatever you want. The corresponding zip would be the right one.

docker-compose.yaml
```diff
@@ -0,0 +1,28 @@
+version: "3"
+
+services:
+  tmail-frontend:
+    image: linagora/tmail-web:master
```
chibenwa, line 5:
```suggestion
    image: linagora/tmail-web:release
```

chibenwa (commented):
Nice!

docker-compose.yaml
tk-nguyen: I think this is just for testing, so using the master branch is ok. I used the master branch for `tmail-backend` image too

Dockerfile
tk-nguyen: I'm not sure what you meant? The prebuild.sh file is in the repo [here](https://github.com/linagora/tmail-flutter/blob/master/prebuild.sh)

Dockerfile
chibenwa: No I mean **don't change this line**.

tk-nguyen: If you run gzip every time the container starts, it will fail if the container restarts cause gzip will ask if you want to overwrite the existing files.

chibenwa: Ok, true. Then please modify the gzip command to overwrite the existing zipped fail.

But keep the gzip on the CMD line.

tk-nguyen: We can't do that either, cause the `-f` flag will recompress the already compressed files.

chibenwa: It's higher priority to me to get the right zipped content.

Not being able to restart the container is less impacting to me.

So please either you make a constructive proposal addressing my concerns, or you revert to the previous value.

chibenwa: How about putting things in one line and swallowing the errors upon gzip step?

tk-nguyen: I'll just revert to the previous CMD then.

hoangdat (approved):

chibenwa (approved):

Issue #437, opened 2022-03-31
[Dev] Handle `StateChange` of mailbox


hoangdat:
### Desc:
- receive `StateChange`
- request `Mailbox/changes` with debounce
- update the cache
- refresh the UI

### DoD:
- UI is updated in the foreground
- No notification needed
- in the case of background, cache is up-to-date with new changes

chibenwa:
Please add a debounce on `request Mailbox/changes` namely:

 - if another state change event was received less than 2 seconds ago then wait for up to those 2 seconds before the `request Mailbox/changes`
 
 IE we should not try to resynchronize ourselves more than once every 2 seconds.

chibenwa:
- Register that UI should be updated on next load when we are in background

hoangdat:
### DoD:
- [ ] Handle Mailbox `/changes` in foreground
- [ ] Handle Mailbox `/changes` in background/terminated

hoangdat:
2HD

PR #1255, opened 2022-11-30
TF-437 Handle `StateChange` of mailbox

dab246:
#437

### Resolved

https://user-images.githubusercontent.com/80730648/204715943-a02523d6-a28b-4593-b06b-f408fbfc8593.mp4


lib/features/push_notification/presentation/controller/fcm_controller.dart
```diff
@@ -126,6 +140,10 @@ class FcmController extends BaseController {
       if (!isForeground) {
         return PushNotificationAction(typeName, newState, accountId);
       }
+    } else if (typeName == TypeName.mailboxType) {
+      if (isForeground) {
+        return SynchronizeMailboxOnForegroundAction(typeName, newState, accountId);
+      }
```
hoangdat, line 148: How about the background case? MailboxView will be refreshed?

dab246: We need to `store state change of mailbox` when we receive message from `fcm`. When the user reopens the app, we start checking for `stored state change` and then do `refresh change`

hoangdat (approved):

hoangdat (approved):

Issue #284, opened 2022-02-24
[Story] As a user, I want to open email on Tmail app from notification on mobile or tablet 


dieptran88:
# Summary

* [Related EPIC](#related-epic)
* [Definition](#definition)
* [Screenshots](#screenshots)
* [Misc](#misc)

## Related EPIC

* TODO: {link}

## Definition

- Given that I am Tmail user and logged-in to the app
- Given that the notification permission is turned on for Tmail app and the app is running in background
- There is a new email and user can see push notification about this new message.
- If user click on notification in Lock screen, then the authentication screen will be opened
   - If I input correct password, the device is unlocked and I am redirected to the Email detail screen and I can see the content of email.
- If user is not in lock-screen , user can see floating/banner notification. User click on this notification and the email content screen will be opened 
- User can also swipe down to see the notification bar, then click on the new email notification, the email content screen will be opened. 
- When an email is opened from notification, that notification  is also cleared on notification center 

[Back to Summary](#summary)

## Screenshots

None

[Back to Summary](#summary)

## Misc

None

[Back to Summary](#summary)

hoangdat:
### DoD:

Open EmailView for mobile/tablet when click in notification

PR #1259, opened 2022-12-01
TF-284 Open email from local notification

sherlockvn:
#284

https://user-images.githubusercontent.com/43041967/205111563-7c0c9d81-455e-4e8b-bf60-fba63eef33b2.mp4


lib/features/push_notification/presentation/notification/local_notification_manager.dart
```diff
@@ -139,4 +145,12 @@ class LocalNotificationManager {
       );
     }
   }
+
+  void _closeStream() {
+    localNotificationsController.close();
+  }
+
+  void dispose() {
```
dab246, line 153: Should call it when the application is destroyed

dab246 (commented):
When clicking on the notification should navigate directly to `EmailView` should not show `ThreadView` to avoid users pressing miscellaneous at this screen. You can do it in the following way:
- Update `routePath` as `EmailDetailView` when clicking on notification
- Pass `EmailId` directly to `EmailView` and then perform `PresentationEmail` object via `EmailId` (You can refer to how to navigate when clicking the `OpenNewTab` event on the web version)

PR #1290, opened 2022-12-09
TF-1231 When remove Subject in AdvancedSearch, filter is not updated because it is empty

hoangdat:
Resolved:

https://user-images.githubusercontent.com/6462404/206617394-154eef1d-5a9d-4375-93bd-47d21b9b238c.mov

### DIFF
--- a/lib/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart
+++ b/lib/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart
@@ -1,3 +1,4 @@
+import 'package:dartz/dartz.dart';
 import 'package:equatable/equatable.dart';
 import 'package:jmap_dart_client/jmap/core/filter/filter.dart';
 import 'package:jmap_dart_client/jmap/core/filter/filter_operator.dart';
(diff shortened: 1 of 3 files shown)


lib/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart
```diff
@@ -70,6 +71,14 @@ class SearchEmailFilter with EquatableMixin {
     );
   }
 
+  T? _getOptionParam<T>(Option<T?>? option, T? defaultValue) {
```
dab246, line 74: IMO, should make this function public so it can be reused. You can put it in a class in main. WDYT?

dab246 (approved):
lgtm

lib/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart
hoangdat: it is very specific for this kind of function `copyWith` so I did not make it public

dab246: What if I want to use it elsewhere in the future?

Issue #1317, opened 2022-12-15
[iOS] notification can not show when new message is coming


hoangdat:
#### Reproduce:
- App in background
- new message is comming
- no notification 
- open the app again 
- notification show

#### Env:
- iphone X
- iOS 16
### Expected:
- [ ] notification must be showed when app in background/terminated

hoangdat:
![image](https://user-images.githubusercontent.com/6462404/209354845-d124a466-aae3-4316-8787-70b57867a555.png)

![image](https://user-images.githubusercontent.com/6462404/209354875-fc443a42-b966-48c9-bc9a-927150908cb8.png)

https://firebase.flutter.dev/docs/messaging/usage

hoangdat:
- must get `accountId` in storage
- Show common message: You have {%d} messages
- for iOS, when click in notification -> open Inbox

hoangdat:
- `You have new messages`

dab246:
2HD

dab246:
1 point

PR #1386, opened 2023-01-09
TF-1317 Fix notification not show when new message is coming

dab246:
#1317


hoangdat:
@haftus98 we are reviewing for the new way of notification is handle in iOS (Android still as normal: detail message in notification)
#### Root cause: 
we can not request the detail information of notification in background in iOS
#### Solution:
- when new messages come, general notifications will be showed: `You have new messages`.
- After that, if user click on the notification, app will open the `Inbox` only.

model/lib/account/account.dart
```diff
@@ -1,13 +1,30 @@
 import 'package:equatable/equatable.dart';
+import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:model/account/authentication_type.dart';
 
 class Account with EquatableMixin {
   final String id;
   final AuthenticationType authenticationType;
   final bool isSelected;
+  final AccountId? accountId;
+  final String? apiUrl;
```
hoangdat, line 10: IMO, `jmapBaseUrl`, but I remembered we store it in `credentialRepository`. Any duplication?

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -1422,18 +1420,47 @@ class MailboxDashBoardController extends ReloadableController {
...
+    final platformOs = tuplePayload.value1;
+
+    if (platformOs == FcmUtils.platformIOS) {
+      final newState = tuplePayload.value2;
+      log('MailboxDashBoardController::_handleClickLocalNotificationOnForeground():newState: $newState');
+      if (newState.isNotEmpty == true) {
+        if (onForeground) {
+          popAllRouteIfHave();
+          dispatchRoute(DashboardRoutes.thread);
+        }
```
hoangdat, line 1443: can you separate in method?

model/lib/account/account.dart
dab246: IMO,
- `jmapBaseUrl` should be user-secure. So I don't think it should be saved on `SharePreferences`.
- It will be duplicated with the `baseUrl` stored in `CredentialRepository`

WDYT? @hoangdat

nqhhdev (dismissed):

sherlockvn (dismissed):

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -585,7 +585,7 @@ class ThreadController extends BaseController with EmailActionController {
   }
 
   void setSelectAllEmailAction() {
-    emailList.value = emailList.map((email) => email.toSelectedEmail(selectMode: SelectMode.ACTIVE)).toList();
```
hoangdat, line 588: please find the way to remove `emailList`

lib/features/login/domain/usecases/refresh_token_oidc_interactor.dart
```diff
@@ -1,41 +0,0 @@
-
-import 'package:core/core.dart';
-import 'package:dartz/dartz.dart';
-import 'package:model/model.dart';
-import 'package:tmail_ui_user/features/login/domain/extensions/oidc_configuration_extensions.dart';
-import 'package:tmail_ui_user/features/login/domain/repository/account_repository.dart';
-import 'package:tmail_ui_user/features/login/domain/repository/authentication_oidc_repository.dart';
-import 'package:tmail_ui_user/features/login/domain/state/refresh_token_oidc_state.dart';
-
-class RefreshTokenOIDCInteractor {
```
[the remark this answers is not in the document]
dab246: It not used

hoangdat (approved):

Issue #1296, opened 2022-12-12
Auto suggestion is still slow in response


hoangdat:
### Desc:
- for all paltforms
### DoD:
- [ ] better in response for user easier in select recipients

hoangdat:
- reduce debounce time? 150ms?
- change from debounce to throttle?

hoangdat:
preflight is eliminated in staging

PR #1397, opened 2023-01-16
TF-1296 Improve auto suggestion is still slow response

dab246:
#1296 

### Resolved

https://user-images.githubusercontent.com/80730648/212636119-412ea41e-5319-4a05-8dee-77fafee69a7b.mov


hoangdat (approved):

chibenwa:
While we are at it can we call the backend from 3 chars or more?

hoangdat:
> While we are at it can we call the backend from 3 chars or more?

Still 3 char, only reduce debounce and adjust some animation UI

nqhhdev (approved):
LGTM

PR #1415, opened 2023-01-30
TF-1156: [Presentation] Auto scroll mailbox list while dragging and drop

nqhhdev:
TF-1156: [Presentation] Auto scroll mailbox list while dragging and drop

Demo :

https://user-images.githubusercontent.com/99852347/215420738-7fcf7286-dab2-4288-a16f-f77649aa968f.mov

### DIFF
--- a/lib/features/mailbox/presentation/mailbox_view_web.dart
+++ b/lib/features/mailbox/presentation/mailbox_view_web.dart
@@ -150,50 +150,72 @@ class MailboxView extends GetWidget<MailboxController> with AppLoaderMixin, Popu
   }
 
   Widget _buildListMailbox(BuildContext context) {
-    return SingleChildScrollView(
-        controller: controller.mailboxListScrollController,
-        key: const PageStorageKey('mailbox_list'),
-        physics: const ClampingScrollPhysics(),
-        padding: EdgeInsets.only(right: _responsiveUtils.isDesktop(context) ? 16 : 0),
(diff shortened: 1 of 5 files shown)


dab246 (approved):
lgtm

lib/features/mailbox/presentation/mailbox_view_web.dart
```diff
@@ -150,50 +150,72 @@ class MailboxView extends GetWidget<MailboxController> with AppLoaderMixin, Popu
...
+                : const SizedBox.shrink()),
+          ])
+        ),
+        Obx(() => controller.mailboxDashBoardController.isDraggingMailbox
+            ? Align(
+                alignment: Alignment.topCenter,
+                child: InkWell(
+                  onTap: () {},
+                  onHover: (value) => value ? controller.autoScrollTop() : controller.stopAutoScroll(),
+                  child: Container(
```
sherlockvn, line 206: Why dont you use Sizedbox ?

sherlockvn (approved):

hoangdat (approved):

lib/features/mailbox/presentation/mailbox_view_web.dart
nqhhdev: When the size box tapped, the ripple effect should happen on top of the widget. But it does not appear. The sizedbox appears, but ripple effect is not.

PR #1271, opened 2022-12-05
TF-1260: auto-complete dont input the final email

ManhNTX:
https://user-images.githubusercontent.com/107173849/205687424-5b46d074-5796-4105-a4ef-5fc7f84c6de1.mov

PR #1454, opened 2023-02-15
Fix search and auto-complete don't input the final email

nqhhdev:
Fix search and auto-complete don't input the final email

Issue :
GIVEN I type the name 
AND select the email address suggestion
WHEN I launch the search and sent it as text, not an email address
THEN there is no result

Solution :
- After choosing an email address, the suggestion should be sent to the email address
Demo 

https://user-images.githubusercontent.com/99852347/218955251-1cc6eccc-7872-4f78-ab5e-9242fad0f484.mov

### DIFF
--- a/lib/features/manage_account/presentation/mailbox_visibility/mailbox_visibility_view.dart
+++ b/lib/features/manage_account/presentation/mailbox_visibility/mailbox_visibility_view.dart
@@ -1,4 +1,3 @@
-import 'package:core/core.dart';
 import 'package:core/presentation/extensions/color_extension.dart';
 import 'package:core/presentation/resources/image_paths.dart';
 import 'package:core/presentation/state/success.dart';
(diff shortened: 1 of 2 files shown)


lib/features/manage_account/presentation/mailbox_visibility/mailbox_visibility_view.dart
```diff
@@ -1,10 +1,4 @@
 import 'package:core/core.dart';
-import 'package:core/presentation/extensions/color_extension.dart';
```
dab246, line 2: why remove it?

nqhhdev: I didn't look closely, I should have removed `import 'package:core/core.dart';` instead of deleting the details

dab246 (approved):

nqhhdev:
https://github.com/linagora/tmail-flutter/pull/1271

It's same my PR

PR #1465, opened 2023-02-16
[Team-Mailbox] Support drag-drop email to team-mailbox on web

nqhhdev:
- Clean email cache before refresh email when list email isEmpty

- [x] Support move from team-mailbox to personal mailbox

https://user-images.githubusercontent.com/99852347/219425801-be2e8dfe-65e1-408c-a211-06034052be5e.mov

- [x] Support move from team-mailbox to team-mailbox 

https://user-images.githubusercontent.com/99852347/219426069-0bbfa3b3-2810-4090-ae57-0092ba364b49.mov

- [x] Support move from personal mailbox to team-mailbox

https://user-images.githubusercontent.com/99852347/219426118-6cc717c8-c810-4670-8300-268c246cb491.mov

### DIFF
--- a/lib/features/thread/presentation/thread_controller.dart
+++ b/lib/features/thread/presentation/thread_controller.dart
@@ -440,7 +440,7 @@ class ThreadController extends BaseController with EmailActionController {
     }
   }
 
-  void refreshAllEmail() {
+  void refreshAllEmail() async {
     dispatchState(Right(LoadingState()));
     canLoadMore = true;
     cancelSelectEmail();
@@ -449,6 +449,7 @@ class ThreadController extends BaseController with EmailActionController {
       searchController.searchEmailFilter.value = _searchEmailFilter.clearBeforeDate();
       _searchEmail(limit: limitEmailFetched);
     } else {
+      await _cachingManager.cleanEmailCache();
       _getAllEmail();
     }
   }


hoangdat:
@chibenwa how about if we move email to the root of team mailbox? is it ok?

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -449,6 +449,7 @@ class ThreadController extends BaseController with EmailActionController {
       searchController.searchEmailFilter.value = _searchEmailFilter.clearBeforeDate();
       _searchEmail(limit: limitEmailFetched);
     } else {
+      await _cachingManager.cleanEmailCache();
```
hoangdat, line 452: what is the problem here? why need to clean the cache here?

nqhhdev: If I don't clean the cache when the list email has one element. it will call back the cache and yield that cache

Pls watch my demo


https://user-images.githubusercontent.com/99852347/219533434-7b7a777a-429f-4368-830e-68b47db5a2f1.mov

dieptran88:
> @chibenwa @dieptran88 how about if we move email to the root of team mailbox? Is it ok? Or when does that, it means move to the Inbox of team mailbox?

Yes I think it is ok to move to Inbox of team mailbox

Issue #1528, opened 2023-03-03
Focus of email body is not display well at the bottom line of composer


hoangdat:
### Reproduce:

https://user-images.githubusercontent.com/6462404/222662437-4d96a65f-1807-4c58-be1b-4723c4d8409e.mov

### DoD:
- new line at the bottom should be display well

PR #1583, opened 2023-03-10
TF1528: Auto down one line when typing

nqhhdev:
Issue: 
#1528 

When we use func `setFullScreenEditor ` it sets the html editor widget to a fixed height and then it cannot scroll automatically. The library already supports this so we shouldn't use func `setFullScreenEditor ` to set height during initialization.

- RESOLVE

[https://pub.dev/packages/html_editor_enhanced](url)
```
shouldEnsureVisible
Default value: false

This option parameter will scroll the editor container into view whenever the webview is focused or text is typed into the editor.

You can only use this parameter if your HtmlEditor is inside a Scrollview, otherwise it does nothing.

This is useful in cases where the page is a SingleChildScrollView or something similar with multiple widgets (eg a form). When the user is going through the different fields, it will pop the webview into view, just like a TextField would scroll into in view if text is being typed inside it.
```

`New composer`

https://user-images.githubusercontent.com/99852347/224885666-e7491c77-3c7e-4a8c-bdba-a76cdf819720.mov

`REPLY AND FORWARD`

https://user-images.githubusercontent.com/99852347/224885747-cd296220-6c58-4af8-b70a-1cecaa184c3c.mov

`Vacation`

https://user-images.githubusercontent.com/99852347/224925145-0978d158-d6e5-48ff-9f2f-d5e51928bf5c.mov

`Identity`

https://user-images.githubusercontent.com/99852347/224925185-6a62d4ba-5fb9-4249-a189-2c69a55102e3.mov

### DIFF
--- a/lib/features/composer/presentation/composer_view_web.dart
+++ b/lib/features/composer/presentation/composer_view_web.dart
@@ -792,7 +792,6 @@ class ComposerView extends GetWidget<ComposerController>
             otherOptions: const OtherOptions(height: 550),
             callbacks: Callbacks(onBeforeCommand: (currentHtml) {
               log('ComposerView::_buildHtmlEditor(): onBeforeCommand : $currentHtml');
-              controller.richTextWebController.setFullScreenEditor();
               controller.setTextEditorWeb(currentHtml);
             }, onChangeContent: (changed) {
               log('ComposerView::_buildHtmlEditor(): onChangeContent : $changed');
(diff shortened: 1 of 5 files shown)


dab246:
Please check `reply` or `forward` email

ManhNTX (approved):
LGTM

dab246:
Screenshot 2023-03-13 at 09 24 31

Please remove line bottom bar. 
U can reproduce it by changing composer visibility (fullscreen <-> normal <-> minimize)

nqhhdev:
> Screenshot 2023-03-13 at 09 24 31
> 
> Please remove line bottom bar. U can reproduce it by changing composer visibility (fullscreen <-> normal <-> minimize)

I updated it at PR : https://github.com/linagora/html-editor-enhanced/pull/6
Plss check it

dab246 (approved):
lgtm

PR #1595, opened 2023-03-13
Added Docker image build to CI

tk-nguyen:
We should add our Docker Hub credentials to secrets if we are using this. The secret keys should be `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`.

### DIFF
--- a/.github/workflows/image.yaml
+++ b/.github/workflows/image.yaml
@@ -1,0 +1,98 @@
+on:
+  push:
+    branches:
+      - "master"
+    tags:
+      - "v*.*.*"
+
+name: Build Docker images
+
+jobs:
+  build-dev-image:
+    name: Build development image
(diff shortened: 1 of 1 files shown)


hoangdat (approved):

.github/workflows/image.yaml
```diff
@@ -0,0 +1,65 @@
+on:
+  push:
+    branches:
+      - "master"
```
hoangdat, line 4: how about `release` branch?

tk-nguyen: That's handled by the tags trigger below.

hoangdat:
please make a move and remove WIP, Can we merge?

tk-nguyen:
> please make a move and remove WIP, Can we merge?

Not yet, until the Docker Hub situation is resolved, we can move forward with a dedicated account for CI.

Issue #1440, opened 2023-02-13
[Signin] User can not sign in for special case which following step in description


dab246:
Reproduce step:
1, Access link: https://tmail.upn.integration-open-paas.org/#/login
2, Click [Single Sign-on] button
3, Click back icon to back to previous page
4, Click [Single Sign-on] button again

Actual: User can't go login page
Expected: User can go login page

https://user-images.githubusercontent.com/124866146/218359037-eb0fb0d5-9c0a-4357-9420-521601dd4b93.mp4

PR #1649, opened 2023-03-24
TF-1440 Fix user can not sign again after click back browser button in OIDC

dab246:
### Issue

#1440 

### Root cause

- `auth_destination_url`: Is our current page path (Here is `LoginView` with path: `https://domain.com#/login`)
- Because `auth_destination_url` already exists in `SessionStorage`. The authentication method always checks if `auth_destination_url` is present on storage. If so, you cannot continue.

Screenshot 2023-03-24 at 16 33 25

### Resolved

https://user-images.githubusercontent.com/80730648/227429967-edc46c4a-cc61-48c3-9056-d51defb129db.mov


nqhhdev (approved):
Lgtm

PR #1659, opened 2023-03-26
Single logout 

hoangdat:
https://user-images.githubusercontent.com/6462404/227988875-79496b19-a2f1-4ae3-9000-379339bb465c.mov


dab246:
Does it affect mobile?

Issue #1657, opened 2023-03-25
Infinite scroll is broken


chibenwa:
### Description

It is not possible to load emails older than 16/03 in my INBOX.

### Expected result

 -> As this is a regression please find the change that introduced the regression.

I expect to be able to load more mails.

### Current behavior

https://user-images.githubusercontent.com/6928740/227723727-0a9d322a-4102-4146-8789-be667f7c5f48.mp4

As we can see no more emails are loaded.

Looking at the JMAP queries, "load more" misses the `mailboxId` property.

```
{
	"using": ["urn:ietf:params:jmap:core", "urn:ietf:params:jmap:mail", "urn:apache:james:params:jmap:mail:shares"],
	"methodCalls": [
		["Email/query", {
			"accountId": "50fb9073ba109901291988b0d78e8a602a6fcd96fbde033eb46ca308779f8fac",
			"filter": {
				"before": "2023-03-16T13:58:32.000Z"
			},
			"sort": [{
				"isAscending": false,
				"property": "receivedAt"
			}],
			"limit": 20
		}, "c0"],
		["Email/get", {
			"accountId": "50fb9073ba109901291988b0d78e8a602a6fcd96fbde033eb46ca308779f8fac",
			"#ids": {
				"resultOf": "c0",
				"name": "Email/query",
				"path": "/ids/*"
			},
			"properties": ["id", "subject", "from", "to", "cc", "bcc", "keywords", "size", "receivedAt", "sentAt", "preview", "hasAttachment", "replyTo", "mailboxIds"]
		}, "c1"]
	]
}
```

What happens is that I classify my messages and all the 20 messages older than 16/03/2023 are in different folders.

Thus they are discarded upon scrolling, which eventually becomes a NOOP.

### Context

tmail.linagora.com

Firefox

v 0.7.5

PR #1660, opened 2023-03-27
TF-1657 Fix infinite scroll is broken

dab246:
### Issue

#1657 

### Root cause

- Because `Load More` misses the `mailboxId` property.


nqhhdev (approved):
Lgtm

PR #1673, opened 2023-03-29
1643-Story-user-can-create-filter-rule-with-label

dieptran88:


chibenwa:
I expect `label as` to be a sub action of the action drop down menu.

Please change the image accordingly.

chibenwa (approved):

hoangdat (approved):

Issue #1545, opened 2023-03-06
[UX] [Text Background Color] Choose color popup is display not following one style which make user feel uncomfortable


QuangHoang1210:
OS: Window

OS: Andoird / Ios
Device: Mobile
TMail version: 0.7.1

*Reproduce Step:
1, User click on [Compose] button
2, User click on Text Background Color icon
3, User select Primary/Accent/Wheel type of choosing color screen

Actual: The choose color popup is changed once user select type.
Expected: We should fix the popup size which not changed once  user select type.

https://user-images.githubusercontent.com/124866146/223077321-2e2ab689-dabf-468a-8984-18f661878ee6.mp4

dab246:
Please suggest a specific design that has this problem.

chibenwa:
@dieptran88 ?

dab246:
### DoD

- Apply as design
  + Desktop/Tablet: Color picker in the middle of the screen
  + Mobile: Color Picker in the bottom of the screen. 
- Tittle dialog: `Select a color`
- Add 2 button: `Cancel` and `Select` in bottom dialog

![Image](https://user-images.githubusercontent.com/80730648/228712268-662ded71-907d-4124-9f6f-827ce300b443.png)

PR #1690, opened 2023-03-30
TF-1545: Update pick color dialog in rich text composer

nqhhdev:
Issue: #1545 

`Resolved` 

I've contributed with a PR to update this, keep an eye on it and wait for their response

- https://github.com/rydmike/flex_color_picker/pull/65
 

https://user-images.githubusercontent.com/99852347/229022199-74081c4b-ee52-4394-b021-bc1580f6640a.mov


chibenwa:
IMO we should not allow the color picker to go too wide...

## Good

On small size screen colors are logically alligned making it easier to pick one:

Screenshot from 2023-03-31 08-32-58

## Bad

On wider screens it just becomes a mess...

Screenshot from 2023-03-31 08-33-12

## Suggestion

Do not allow the color picker to grow too wide...

dab246:
- Should center the palette.
- Padding=16 has all (top, botto, lef, right)

nqhhdev:
> * Should center the palette.
> * Padding=16 has all (top, botto, lef, right)
> 
> Screenshot 2023-03-31 at 11 39 42

Screenshot 2023-03-31 at 13 13 02
I updated it

pubspec.yaml
```diff
@@ -54,7 +54,7 @@ dependencies:
   html_editor_enhanced:
     git:
       url: https://github.com/linagora/html-editor-enhanced.git
-      ref: email_supported
+      ref: change-flex-color-picker-lib
```
dab246: why updated this library?

dab246 (approved):

pubspec.yaml
chibenwa: Should be `master`

nqhhdev: -`html-editor-enhanced=`This library also uses `html-editor-enhanced`. So I switched to my branch for testing before being merged into master. I updated it again

Plss check my PR: https://github.com/linagora/html-editor-enhanced/pull/9

Issue #1688, opened 2023-03-30
[Identity] [Crashed] Tmail UI is broken after user create identity successfully by html style


QuangHoang1210:
OS: Window
Device: Laptop
TMail version: 0.7.6

*Precondition: User created identity with html style in signature
*Data test: 
- User firstname31.surname31@upn.integration-open-paas.org
- html source: 
`<!DOCTYPE html>
<html>
<body>

<h2 title="I'm a header">The title Attribute</h2>

<p title="I'm a tooltip">Mouse over this paragraph, to display the title attribute as a tooltip.</p>

</body>
</html>`

### *Reproduce Step:
1, User click on Manage Account
2, User click on Profile
3, User click on [+Create Identity] button
4, User copy an html into signature
5, User click [Create] button
6, User click Team Mail logo to back to homepage 

### **Actual:** 
UI is broken. The content from html signature is still display overlaps other element.

![Image](https://user-images.githubusercontent.com/124866146/228814154-3489400a-00c8-472b-bf1f-b5d0904c1814.png)

https://user-images.githubusercontent.com/124866146/228814221-10e3d373-abc2-4c15-b064-b45f67fe06c8.mp4

### **Expected:** 
User can go back to homepage smoothly. UI is not broken

chibenwa:
Please specify the browser

PR #1701, opened 2023-04-03
TF-1688 Fix Tmail UI is broken after user create identity successfully by html style

dab246:
### Issue

#1688 

### Root causes

- Flutter was unable to create enough overlay surfaces. This is usually caused by too many platform views being displayed at once.
-  This issue is still under discussion and there is no official fix from `flutter` side. See details here: https://github.com/flutter/flutter/issues/118159 or https://github.com/flutter/flutter/issues/104970

229408191-806b760a-5b9f-4ec5-8bb3-263b9b110743

### Workaround 

- Use `pointer_interceptor` version `0.9.1`

### Resolved

https://user-images.githubusercontent.com/80730648/229415507-18d0c007-3897-47fb-94d1-c58f9b09c495.mov

### DIFF
--- a/contact/pubspec.lock
+++ b/contact/pubspec.lock
@@ -276,10 +276,10 @@ packages:
     dependency: transitive
     description:
       name: flex_color_picker
-      sha256: "607c9fdb26be84d4a5a0931ab42a7eda725372e4f5ebaa2526ab6b22ead752f9"
+      sha256: f0e0db8e3e47435cfbe9aa15c71b898fa218be0fc4ae409e1e42d5d5266b2c90
       url: "https://pub.dev"
     source: hosted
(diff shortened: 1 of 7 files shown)


hoangdat:
how about to use `dependencies_override`?

dab246:
> how about to use `dependencies_override`?

Oh no, You only use it to override its own libraries. But cannot override the libraries inside it.

nqhhdev (approved):

hoangdat:
But please do an ADR to track why we need to downgrade dependencies

dab246:
> But please do an ADR to track why we need to downgrade dependencies

Updated.

nqhhdev (approved):

Issue #1698, opened 2023-04-02
[Mailbox] System not redirect correct url when user double click many times to view email in mailbox


QuangHoang1210:
OS: Window
Device: Laptop
TMail version: 0.7.6

*Precondition: 
- User already login successfully
*Test data: 
- Using account firstname31.surname31@upn.integration-open-paas.org
- Mailbox url:
```
- human-resources/INBOX : https://tmail.upn.integration-open-paas.org/#/dashboard?type=normal&context=e3f9e750-64b4-11ed-88c0-338bebbdf582
- hiring@upn.integration-open-paas.org/INBOX : https://tmail.upn.integration-open-paas.org/#/dashboard?type=normal&context=e996d2e0-64b4-11ed-88c0-338bebbdf582
- sales/INBOX : https://tmail.upn.integration-open-paas.org/#/dashboard?type=normal&context=ca2cd4e0-64b4-11ed-88c0-338bebbdf582
```

### *Reproduce Step:
1, User click on search icon at mailbox module
2, User input 'in' to search textbox
3, User click to view each inbox in searching result
4, User double click on human-resources/INBOX more than one time
5, User double click on hiring@upn.integration-open-paas.org/INBOX more than one time
5, User click on sales/INBOX

### **Actual:** 
System redirect to wrong url: https://tmail.upn.integration-open-paas.org/#/dashboard?type=normal&context=e3f9e750-64b4-11ed-88c0-338bebbdf582

### **Expected:** 
System should redirect to correct url: https://tmail.upn.integration-open-paas.org/#/dashboard?type=normal&context=ca2cd4e0-64b4-11ed-88c0-338bebbdf582

https://user-images.githubusercontent.com/124866146/229363362-0878e4da-8381-4763-a70a-5d31fb611f15.mp4

PR #1754, opened 2023-04-10
TF-1698 Fix system not redirect correct url when user double click many times to view email in mailbox

dab246:
### Issue

#1698 

### Root cause

- Since the `dashBoardAction` variable does not refresh the stream, the controller cannot listen.
- Using `throttleTime` within 800ms so it is not possible to get the latest value when changing continuously in a short period of time.

### Resolved

https://user-images.githubusercontent.com/80730648/230828425-52983bf8-15b9-48d5-bbbd-5e015a7541b9.mov


nqhhdev (approved):
Lgtm

hoangdat (approved):

Issue #1779, opened 2023-04-20
[Compose] [Impovement] [UX] Logic of replacing dot in long email


QuangHoang1210:
OS: IOS/ANDROID
Device: Mobile
TMail version: 0.7.7 

*Precondition: 
- User already created signature in profile as human-resources email
*Data test: User firstname31.surname31

### *Reproduce Step:
1, User click [Compose]
2, User select identity human-resources

### **Actual:** 
It replace dot after "-" character  then the email is displayed not long as others one.

![Image](https://user-images.githubusercontent.com/124866146/233253185-daa49929-270b-4741-991b-7cbce8a60646.png)

### **Expected:** 
It should be displayed longer email same as others.

Issue #1778, opened 2023-04-20
[Signature] System not display signature in email which has been sent


QuangHoang1210:
OS: IOS/ANDROID
Device: Mobile
TMail version: 0.7.7 

*Precondition: 
- User already created signature in profile as default
*Data test: User firstname31.surname31

### *Reproduce Step:
1, User click [Compose]
2, User select identity
3, User input required field to send an email
4, User go to check email in sent box

### **Actual:** 
There is missing signature in email which has been sent

![Image](https://user-images.githubusercontent.com/124866146/233252577-32b223e1-5498-433f-a33a-41f40f3f7d2c.png)

### **Expected:** 
Sending email should be included signature in body

https://user-images.githubusercontent.com/124866146/233252604-d8c8bf87-eede-4c14-bfb0-6962c8b2555c.mp4

PR #1783, opened 2023-04-21
TF-1778,1779 System not display signature in email after sent on mobile

dab246:
### Issue

#1778 , #1779 

### Root cause

- `#1778`: Due to a syntax error in javascript. In `string` contains the characters `'` and `"`
- `#1779`:  When we use the `overflow=TextOverflow.ellipsis` property for the `Text` widget for long texts, it will result in incorrect string breaks. Since string contains some characters that are supposed to be word breaks in the string. The characters ` ` and `-`

### Dependent

- Need merged https://github.com/linagora/enough_html_editor/pull/20

### Resolved

https://user-images.githubusercontent.com/80730648/233637320-83e20587-ec2c-4104-9408-3dfaa75d81ec.mov

### DIFF
--- a/core/lib/presentation/extensions/string_extension.dart
+++ b/core/lib/presentation/extensions/string_extension.dart
@@ -1,4 +1,5 @@
 import 'package:core/utils/app_logger.dart';
+import 'package:flutter/material.dart';
 
 extension StringExtension on String {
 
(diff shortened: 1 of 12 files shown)


hoangdat:
thanks for this kind of issue, please write an ADR also

hoangdat (approved):

nqhhdev (approved):

Issue #1743, opened 2023-04-06
Horizontal scroll bar in Email detail view


chibenwa:
### Description

![Screenshot from 2023-04-06 13-53-15](https://user-images.githubusercontent.com/6928740/230294975-a3e07e48-8eb5-4967-b55f-2665080abb99.png)

chibenwa:
[email.txt](https://github.com/linagora/tmail-flutter/files/11167222/email.txt)

PR #1794, opened 2023-04-24
TF-1743: Hide Hide Horizontal scroll bar in Email detail view

nqhhdev:
- Issue: #1743 

- Root cause: 
The horizontal scrollbar appears when the content inside the body element is wider than the width of the viewport. By default, the overflow property is set to "visible", which means that the content can overflow the element's boundaries.

- Resolved:

`Mac OS`

https://user-images.githubusercontent.com/99852347/234059455-a4a71800-2423-4ab3-8928-03dc8cfe988b.mov

`Ubuntu (Firefox, chromium)`

https://user-images.githubusercontent.com/99852347/234156703-8c3b23b6-17a8-4ed9-bd07-55424047e6f5.mov

### DIFF
--- a/core/lib/presentation/utils/html_transformer/html_template.dart
+++ b/core/lib/presentation/utils/html_transformer/html_template.dart
@@ -53,7 +53,7 @@ String generateHtml(String content, {
     </style>
     ${javaScripts ?? ''}
     </head>
-    <body>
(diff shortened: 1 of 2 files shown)


dab246 (approved):
lgtm

sherlockvn (approved):

Issue #1806, opened 2023-05-02
[Bug]: Impossible to disable used fonts in email composer


tprudentova:
•	Environment(version): tmail.upn.integration-open-paas.org, v 0.7.8, Google Chrome
•	Date of occurrence: 28.04.2023
•	Steps to reproduce:
1.	Type any text in the email composer
2.	Apply different fonts to different parts of it (bold, italic etc)
3.	Go to the next line
4.	Remove all fonts 
5.	Type some text 
6.	The fonts are still applied to it
•	Expectation: When I disable all fonts, the text is typed in in regular font
•	Image/Video envidence:

https://user-images.githubusercontent.com/97684890/235603693-39657f22-fec7-4952-866a-5320993bb3f0.mov

chibenwa:
Reproduced (tmail.linagora.com on v0.7.6, chromium)

PR #1819, opened 2023-05-05
TF-1806 Fix impossible to disable used fonts in email composer

dab246:
### Issue

#1806 

### Root cause

- Due the `onChangeSelection` function is not called every time the editor changes the style. `onChangeSelection` is called only when `ToolbarWidgetState != null`

### Resolved

https://user-images.githubusercontent.com/80730648/236453232-c7679c46-771c-401b-a48a-6d3ef1a1f50a.mov


dab246:
> At time 0:44, you disable `U`, but the editor still apply `U`. What happen?

Oh, IMO, it still belongs to the above tag so it still contains the `U` style. Tested again and it works fine.

nqhhdev (approved):

Issue #1811, opened 2023-05-04
[Dev] Store email when receive in background


dab246:
### Issue

- #1702 - US 2

### DoD

- [ ] Store email when receive in background (Android)

### Desc

- Keep 10 email recent

QuangHoang1210:
Not Passed  0.7.11 -> https://github.com/linagora/tmail-flutter/issues/1923

PR #1837, opened 2023-05-11
TF-1811 Store email when receive in background for Android

dab246:
### Issue

#1811 

### Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/019befec-7735-4799-9347-4872e5813635

### DIFF
--- a/lib/features/push_notification/presentation/listener/email_change_listener.dart
+++ b/lib/features/push_notification/presentation/listener/email_change_listener.dart
@@ -107,6 +107,10 @@ class EmailChangeListener extends ChangeListener {
         }
       } else if (action is PushNotificationAction) {
         _pushNotificationAction(action.newState, action.accountId, action.userName, action.session);
+
+        if (FcmUtils.instance.isMobileAndroid) {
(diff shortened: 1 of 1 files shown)


nqhhdev (approved):

lib/features/push_notification/presentation/listener/email_change_listener.dart
```diff
@@ -107,6 +107,10 @@ class EmailChangeListener extends ChangeListener {
         }
       } else if (action is PushNotificationAction) {
         _pushNotificationAction(action.newState, action.accountId, action.userName, action.session);
+
+        if (FcmUtils.instance.isMobileAndroid) {
+          _getNewReceiveEmailFromNotificationAction(action.session, action.accountId, action.newState);
+        }
```
hoangdat, line 113: Can we do these actions in parallel?

dab246: Processing inside the two functions is done, it doesn't have to wait for each other. It's just that it's being called in a synchronous function.

hoangdat (approved):

PR #1848, opened 2023-05-16
Create TF-1847-user-can-cancel-recovery-process.md

dieptran88:


hoangdat:
@dieptran88 I propose to add a `Status` section in a story to easier to follow the status of this story (in the case of dev only)
```
## Status
Pending/In progress/Done
```

Issue #1862, opened 2023-05-19
Right click on email item in email list view


chibenwa:
### Description

Firefox:

```
GIVEN I am in email list view
WHEN I right click on an item
THEN the right clicks menu open
AND the move dialog is shown

WHEN I close the right click dialog 
THEN the move dialog is still here
```

Chome :

```
GIVEN I am in email list view
WHEN I right click on an item
THEN the right clicks menu open
AND the move dialog is shown

WHEN I close the right click dialog 
THEN the move dialog is still here
```

### Expected result

Do not show the move dialog at all when I right click...

### Current behavior

Firefox:

https://github.com/linagora/tmail-flutter/assets/6928740/ae55073d-367e-4769-85ef-7ea99d636f1b

Chrome:

https://github.com/linagora/tmail-flutter/assets/6928740/977862a6-d2b5-47b0-9b70-b117a77d0a04

### Context

https://oncommit-tmail.linagora.com

QuangHoang1210:
Verified on 0.8.2 - 

Firefox: Looking well
Chrome: The move dialog still displayed  in 1 second when user right click on email rows.

Expected: The move dialog only display when user drag the email.



https://github.com/linagora/tmail-flutter/assets/124866146/b09a7440-c549-42be-9d8d-33cb0e4365b9

dab246:
@QuangHoang1210 Please verify help us. Thanks

PR #1874, opened 2023-05-24
TF-1862: Fix show  widget of Draggable when click right on email

nqhhdev:
- Issue: #1862 

- The childWhenDragging widget will still be draggable, including the ability to trigger the right-click event. This is because the childWhenDragging widget is a part of the draggable behavior and inherits the drag-related properties and behaviors from the Draggable widget.The Draggable widget allows you to customize the appearance and behavior of the dragged item during the dragging process by providing the feedback and childWhenDragging properties. The feedback widget is used to display a visual representation of the dragged item, while the childWhenDragging widget is used to replace the original item when it is being dragged.

- Resolved :

Wrap it with a GestureDetector and handle the right-click event as mentioned in the previous answer. This way, you can have separate behavior for the original item and the item being dragged.

- `Chome` 

https://github.com/linagora/tmail-flutter/assets/99852347/ea7b3261-781d-4195-a1d8-7d51eb50c6a3

### DIFF
--- a/lib/features/thread/presentation/thread_view.dart
+++ b/lib/features/thread/presentation/thread_view.dart
@@ -360,10 +360,7 @@ class ThreadView extends GetWidget<ThreadController>
   Widget _buildEmailItemDraggable(BuildContext context, PresentationEmail presentationEmail) {
     return GestureDetector(
       behavior: HitTestBehavior.translucent,
-      onSecondaryTapDown: (details) {
-        // 1. Use empty callback to disable D&D on mouse right button
-        // 2. Call `showMenu` to show context menu
-      },
+      onSecondaryTapDown: (_) {},
       child: Draggable<List<PresentationEmail>>(
         data: controller.listEmailDrag,
         feedback: _buildFeedBackWidget(context),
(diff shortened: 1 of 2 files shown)


lib/features/thread/presentation/thread_view.dart
```diff
@@ -358,18 +358,25 @@ class ThreadView extends GetWidget<ThreadController>
...
-      childWhenDragging: _buildEmailItemWhenDragging(context, presentationEmail),
-      dragAnchorStrategy: pointerDragAnchorStrategy,
-      onDragStarted: () {
-        controller.calculateDragValue(presentationEmail);
-        controller.onDragMailBox(true);
+    return GestureDetector(
+      behavior: HitTestBehavior.translucent,
+      onSecondaryTapDown: (details) {
+        // 1. Use empty callback to disable D&D on mouse right button
+        // 2. Call `showMenu` to show context menu
```
hoangdat, line 365: IMO, it is deserved to have an ADR and comment the ADR here.

dab246:
Please squash & rebase

dab246 (approved):

Issue #1812, opened 2023-05-04
[Dev] Store sending emails in Sending Queue mailbox


dab246:
### Issue

- #1702 - US 1.a

### DoD

- [ ] Store sending emails when offline
- [ ] Add task to worker manager

### Desc

- Create object `SendingEmailHiveCache` 

```
class SendingEmailHiveCache {

  final String id;

  final Map<String, dynamic> emailData;

  final String emailActionType;

  final String? sentMailboxId;

  final String? emailIdDestroyed;

  final String? emailIdAnsweredOrForwarded;

  final String? identityId;

  final String? mailboxNameRequest;

  final String? creationIdRequest;
}
```

- Store sending email to cache after click button send while no connection and toast message `Your email has been saved to the sending queue`

- Create task from SendingEmailCache to add inside worker manager

PR #1880, opened 2023-05-26
TF-1812  Store sending emails in Sending Queue

dab246:
### Issue

#1812 

### Resolved

**1. Android:** 

Screenshot_1684988078

**2. iOS**

- Task is scheduled but time is determined by iOS operating system (within 48h)
- Unable to pass data `InputData` to background task


nqhhdev (approved):

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -1703,12 +1722,58 @@ class MailboxDashBoardController extends ReloadableController {
...
+        consumeState(_sendEmailInteractor.execute(
+          session,
+          accountId,
+          emailRequest,
+          mailboxRequest: mailboxRequest
+        ));
+      } else {
+        final sendingEmail = emailRequest.toSendingEmail(_uuid.v1(), mailboxRequest: mailboxRequest);
+        _storeSendingEmail(accountId, session.username, sendingEmail);
+      }
```
hoangdat, line 1743: how about if all the case of network, we will call `_storeSendingEmai`?

lib/features/offline_mode/observer/sending_email_observer.dart
```diff
@@ -14,15 +11,14 @@ class SendingEmailObserver extends WorkObserver<Email> {
...
   @override
-  void handleSuccessViewState(Success success) {
-  }
-
-  @override
-  WorkStatus observe(Email value) {
-    return WorkStatus.failure;
+  Future<void> bindDI() async {
+    log('SendingEmailObserver::bindDI(): ');
+    return Future.value();
```
hoangdat, line 22: please make sure not put instance duplicated

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
dab246: why call when there is network? If so, will we be sent the same email again when the network is changed!

lib/features/offline_mode/observer/sending_email_observer.dart
dab246: The task is run in the background. So it's always a unique instance

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
hoangdat: I think no, when network is connected, this action will be triggered after delay initial, so that, we remove from the sending queue is ok :))

hoangdat: but if the task run will raise a notification, I think no need to call this in the case network connected

hoangdat (approved):

PR #1904, opened 2023-06-02
Json-File-Arabic-Language

sakriwassim:
In this PR , I add the file Json Arabic I think you need it to the Arabic version


hoangdat:
why you do not it in weblate?

Issue #1923, opened 2023-06-12
[Offline Mode] [Background] User received notification of new emails but user cannot view that email content by click on the notification in Offline Mode


QuangHoang1210:
OS:  IOS / Android
Device: Mobile
TMail version: 0.8.0

*Precondition: 
- The TMAIL application of user B running in background

### *Reproduce Step:
1, User B open another application
2, User A sent 12 new emails to user B
3, User B switch offline
4, User B click on each notification of new emails

### **Actual:** 
User cannot open to view some emails <following the status bellow>

Email 1 -- **Cannot view offline**
Email 2 -- **Cannot view offline**
Email 3 -- Can view offline
Email 4 -- **Cannot view offline**
Email 5 -- Can view offline
Email 6 -- Can view offline
Email 7 -- Can view offline
Email 8 -- Can view offline
Email 9 -- Can view offline
Email 10 -- **Cannot view offline**
Email 11 -- Can view offline
Email 12 -- **Cannot view offline**

### **Expected:** 
User can view offline 10 most recent emails <following the status bellow>

Email 1 -- **Cannot view offline**
Email 2 -- **Cannot view offline**
Email 3 -- Can view offline
Email 4 -- Can view offline
Email 5 -- Can view offline
Email 6 -- Can view offline
Email 7 -- Can view offline
Email 8 -- Can view offline
Email 9 -- Can view offline
Email 10 -- Can view offline
Email 11 -- Can view offline
Email 12 -- Can view offline

https://github.com/linagora/tmail-flutter/assets/124866146/167d2140-808b-49af-8806-9a96db0f7913

PR #1953, opened 2023-06-21
TF-1923 Fix user cannot view that email content by click on the notification in Offline Mode

dab246:
### Issue

#1923 

### Root cause

- New message list is retrieved duplicate when the status has not been updated.
- Use system time to sort new saved messages.
- Some logic about store new email is incorrect

### Solution

- Use received time of email `receivedAt` to cache and organize emails.
- Only save new email when it does not exist in the cache.
- Each `repository` should only support a single job. To better manage `exception` at the `interactor` layer

### Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/9b2eef1c-2dc8-422d-a23e-7ced27ef26b1


nqhhdev (approved):

lib/features/email/domain/state/get_detailed_email_by_id_state.dart
```diff
@@ -9,15 +9,14 @@ class GetDetailedEmailByIdLoading extends UIState {}
...
 class GetDetailedEmailByIdSuccess extends UIState {
 
-  final Email email;
-  final DetailedEmail detailedEmail;
+  final Map<Email, DetailedEmail> mapDetailedEmail;
   final AccountId accountId;
   final Session session;
 
-  GetDetailedEmailByIdSuccess(this.email, this.detailedEmail, this.accountId, this.session);
+  GetDetailedEmailByIdSuccess(this.mapDetailedEmail, this.accountId, this.session);
```
sherlockvn, line 16: why don't you use required here, because 3 parameters is pretty long

lib/features/email/domain/usecases/get_email_content_interactor.dart
```diff
@@ -31,13 +30,13 @@ class GetEmailContentInteractor {
       yield Right<Failure, Success>(GetEmailContentLoading());
 
       if (PlatformInfo.isMobile) {
-        yield* _tryToGetOpenedEmailCache(session, accountId, emailId, baseDownloadUrl, composeEmail: composeEmail, draftsEmail: draftsEmail);
+        yield* _getStoredOpenedEmail(session, accountId, emailId, baseDownloadUrl, composeEmail: composeEmail, draftsEmail: draftsEmail);
       } else {
         yield* _getContentEmailFromServer(session, accountId, emailId, baseDownloadUrl, composeEmail: composeEmail, draftsEmail: draftsEmail);
```
sherlockvn, line 35: Wow, Stream in another Stream in another Stream, look so cool ! and clean

lib/features/email/domain/state/get_detailed_email_by_id_state.dart
dab246: We should only use `required` when the parameters are ambiguous and it is required. IMO, it's not so important while variables are already too obvious for developers.

hoangdat (approved):

Issue #1932, opened 2023-06-14
[Sending Queue] System displayed notification of emails which sent successfully from Sending Queue once user reconnected


QuangHoang1210:
OS:  Android
Device: Mobile
TMail version: 0.8.0

*Precondition: 
- User A is offline 

### *Reproduce Step:
1, User A send / foward / reply an email in offline mode
2, User A click [Proceed] button in popup Sending Queue
3, User A reconnect network

### **Actual:** 

- Emails which in Sending Queue is sent. 
- System displayed notification on the mobile notification center but it redirect user to TMail application INBOX once user click on it

### **Expected:** 

- Emails which in Sending Queue is sent successfully.
- No notification will be displayed once email sent.
- Only email which have address as User A will be displayed as notification.

https://github.com/linagora/tmail-flutter/assets/124866146/6c89125e-958c-418b-a70e-11a389316d03

PR #1954, opened 2023-06-21
TF-1932 Turn off notifications after emailing

dab246:
### Issue

#1932 

### Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/3fbf8aa3-e437-4088-8d3b-47a676eac8a0

### DIFF
--- a/lib/features/offline_mode/observer/sending_email_observer.dart
+++ b/lib/features/offline_mode/observer/sending_email_observer.dart
@@ -5,7 +5,6 @@ import 'package:core/presentation/state/success.dart';
 import 'package:core/utils/app_logger.dart';
 import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/session/session.dart';
-import 'package:model/extensions/list_email_content_extension.dart';
 import 'package:model/extensions/session_extension.dart';
 import 'package:model/oidc/token_oidc.dart';
 import 'package:tmail_ui_user/features/caching/config/hive_cache_config.dart';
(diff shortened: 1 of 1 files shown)


hoangdat:
@dieptran88 we need to show notification after sending queue item success or not?

dieptran88:
> @dieptran88 we need to show notification after sending queue item success or not?

I think yes, we agreed to have the noti after sending queue successfully because it is not sent right after the user compose it, so need to inform user.

nqhhdev (approved):

dieptran88:
> > @dieptran88 we need to show notification after sending queue item success or not?
> 
> I think yes, we agreed to have the noti after sending queue successfully because it is not sent right after the user compose it, so need to inform user.

Hi @dab246 @QuangHoang1210 @hoangdat , after comparing with other mail clients and get consulting from designer, we think sending notification in this case is not very necessary and may show too much notifications in a time if there are multiple emails in sending queue

hoangdat:
Ok, totally agree with Diep about that. 
- No notification in Notification center of Android/iOS
- Only noti by toast inside app (only when app in foreground, sure).

How about it? @dab246 @QuangHoang1210

dab246:
> 



> Ok, totally agree with Diep about that.
> 
> * No notification in Notification center of Android/iOS
> * Only noti by toast inside app (only when app in foreground, sure).
> 
> How about it? @dab246 @QuangHoang1210

If so, then I don't think it is necessary to notify in the app in the foreground because if there are many messages in the queue to send. We will have to display quite a few toasts at the same time. Re-rendering will cause flickering.

QuangHoang1210:
Yeah. IMO we do not need to have notification this case.
The email was sent will be moved to Sent mailboxes.

The email is pending will be stay in sending queue. So no need to show notification for this one. User can check it by moving mailboxes from Sent -> Sending queue.

Note: We also have email sending status in Sending Queue.

dab246:
@hoangdat We need to finalize a solution to this problem.

hoangdat:
ok with Dat Vu, keep the error in the list of sending queue is enough, no need more for success case. Thanks

PR #1964, opened 2023-06-26
TF-1947 Swipe left in unread item of ThreadView

sakriwassim:
@hoangdat 
@nqhhdev
@chibenwa
make mail as read and make mail as unread are working know 

https://github.com/linagora/tmail-flutter/assets/62904371/e2993e92-16e8-46fe-bebc-80a24b97b9ca

https://github.com/linagora/tmail-flutter/assets/62904371/99c66724-265a-479a-b567-788a5403f095


lib/features/thread/presentation/mixin/base_email_item_tile.dart
```diff
@@ -55,6 +55,43 @@ mixin BaseEmailItemTile {
     }
   }
 
+  Widget buildIconAvatarSwipe(
```
hoangdat, line 58: please do it as a `stateless widget`

lib/features/thread/presentation/widgets/dismissible_widget.dart
```diff
@@ -0,0 +1,77 @@
+import 'package:core/presentation/extensions/color_extension.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter_slidable/flutter_slidable.dart';
+import 'package:tmail_ui_user/features/thread/presentation/mixin/base_email_item_tile.dart';
+
+class DismissibleWidget<T> extends StatelessWidget with BaseEmailItemTile {
```
hoangdat, line 6: why we need this widget?

sakriwassim: this widget is the second tile if we slide right or left it take the two
icon and item and the function of Dismissed I separate in a new file to make the code cleaner

lib/features/thread/presentation/mixin/base_email_item_tile.dart
sakriwassim: should is make it inside new file as a stateless widget ?

Issue #1933, opened 2023-06-14
[Sending Queue] [UI] inbox and sending queue is highlight once some emails in sending queue was sent in the same time


QuangHoang1210:
OS:  IOS / Android
Device: Mobile
TMail version: 0.8.0

*Precondition: 
- At least 6 emails are existed in Sending Queue
- The issue only happen after the emails in Sending Queue was sent

### *Reproduce Step:
1, User A is offline
2, User A open Sending Queue
3, User A reconnect the network
4, User A click on back to view the mailboxes list after emails in sending queue was sent

### **Actual:** 

![Image](https://user-images.githubusercontent.com/124866146/245910023-d492f457-a5ed-48d4-ae66-477a25af8d9d.png)

### **Expected:** 

Only Sending Queue will be highlight. 
System will highlight other mailboxes once user tap to select other mailboxes

https://user-images.githubusercontent.com/124866146/245910233-8ce7db68-547a-4e79-a2a1-c1a954cd3e06.mp4

PR #1969, opened 2023-06-27
TF-1933 Fix inbox and sending queue is highlight once some emails in sending queue was sent in the same time

dab246:
### Issue

#1933 

### Root cause

- Because the workers in the `workmanager` are executed concurrently. Therefore, more than one task in `isolate` accessing the `hive` database at a time leads to incorrect data synchronization.

### Solution

- We complete the api-related tasks on the server at workmanager isolates. Update data of `hive` database on main isolate.

### Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/b4816b0e-7e8a-4e0e-9995-c5724701f11c


lib/features/offline_mode/work_manager/sending_email_worker.dart
```diff
@@ -5,69 +5,68 @@ import 'package:core/presentation/state/success.dart';
...
   DynamicUrlInterceptors? _dynamicUrlInterceptors;
   AuthorizationInterceptors? _authorizationInterceptors;
   GetSessionInteractor? _getSessionInteractor;
-  DeleteSendingEmailInteractor? _deleteSendingEmailInteractor;
   SendingQueueIsolateManager? _sendingQueueIsolateManager;
-  UpdateSendingEmailInteractor? _updateSendingEmailInteractor;
   SendingEmailCacheManager? _sendingEmailCacheManager;
 
-  Completer<bool>? _completer;
+  late Completer<bool> _completer;
```
sherlockvn, line 47: Don't you think use `late` here is dangerous ? IMO, use nullable and then when you surely assign to it, use with !

dab246: In this case it is not dangerous at all. If you read carefully you can see that `Completer` used here must be a non-nullable variable. I can initialize it directly at declaration but I use `late` so that it is only assigned value during runtime, not compilation

nqhhdev (approved):
LGTM

lib/features/session/domain/usecases/get_session_interactor.dart
```diff
@@ -11,9 +13,35 @@ class GetSessionInteractor {
 
   Stream<Either<Failure, Success>> execute() async* {
     try {
+      yield Right<Failure, Success>(GetSessionLoading());
+      if (PlatformInfo.isMobile) {
+        yield* _getStoredSession();
```
hoangdat, line 18: ahh, only mobile support `offlineMode`. OK.
But I think the first logic is better, if can not get session from server, please get session from store, because session maybe change time to time, when somebody have new `limit`, `quota`, bla bla

hoangdat:
keep it postpone or merge to another branch

hoangdat (approved):

Issue #1974, opened 2023-06-30
RefreshToken with OIDC on` jmap.linagora.com/oidc`


dab246:

hoangdat:
hope you have more information:

Crashed logs collected from Store:
```
Exception java.lang.IllegalArgumentException:
  at net.openid.appauth.Preconditions.checkArgument (Preconditions.java)
  at net.openid.appauth.Preconditions.checkNotEmpty (Preconditions.java)
  at net.openid.appauth.TokenRequest$Builder.setRefreshToken (TokenRequest.java)
  at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.performTokenRequest (FlutterAppauthPlugin.java)
  at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin.access$600 (FlutterAppauthPlugin.java)
  at io.crossingthestreams.flutterappauth.FlutterAppauthPlugin$3.onFetchConfigurationCompleted (FlutterAppauthPlugin.java)
  at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute (AuthorizationServiceConfiguration.java)
  at net.openid.appauth.AuthorizationServiceConfiguration$ConfigurationRetrievalAsyncTask.onPostExecute (AuthorizationServiceConfiguration.java)
  at android.os.AsyncTask.finish (AsyncTask.java:755)
  at android.os.AsyncTask.access$900 (AsyncTask.java:192)
  at android.os.AsyncTask$InternalHandler.handleMessage (AsyncTask.java:772)
  at android.os.Handler.dispatchMessage (Handler.java:107)
  at android.os.Looper.loop (Looper.java:224)
  at android.app.ActivityThread.main (ActivityThread.java:7561)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:539)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:995)

```

chibenwa:
This exception is raised because the refresh token string is empty.

Lemon do return a refresh token. 

![Screenshot from 2023-07-03 20-16-24](https://github.com/linagora/tmail-flutter/assets/6928740/a02d276e-1d20-4da6-8c7b-cb3679038cde)


Do TMail store the refresh token?

dab246:
> This exception is raised because the refresh token string is empty.
> 
[...]

PR #1978, opened 2023-07-04
TF-1974 RefreshToken with OIDC on `jmap.linagora.com/oidc`

dab246:
### Issue

#1974 

### Root cause

- I have hard-coded for `RefreshToken` as null. Before calling `appAuth.token()` and getting the same error as store.

Screenshot 2023-07-04 at 00 46 59

- After that investigation, we concluded that `refreshToken` was empty leading to the application crash.  Details in the native library [openid](https://github.com/openid/AppAuth-Android/blob/d73ced59d07a4b93d9039ab3d193942fc7197959/library/java/net/openid/appauth/TokenRequest.java#L399)

Screenshot 2023-07-04 at 00 51 45

- It only happens on Android devices

### Solution

- Check for non-empty `refreshToken` before being used.

### Related 

- Same issue raised on `Flutter AppAuth` library [Issues#396](https://github.com/MaikuB/flutter_appauth/issues/396)

### DIFF
--- a/lib/features/email/domain/usecases/download_attachments_interactor.dart
+++ b/lib/features/email/domain/usecases/download_attachments_interactor.dart
@@ -72,7 +72,8 @@ class DownloadAttachmentsInteractor {
       }
     } catch (exception) {
       log('DownloadAttachmentsInteractor::execute(): $exception');
-      if (exception is DownloadAttachmentHasTokenExpiredException) {
+      if (exception is DownloadAttachmentHasTokenExpiredException &&
+          exception.refreshToken.isNotEmpty) {
         yield* _retryDownloadAttachments(
             accountId,
             baseDownloadUrl,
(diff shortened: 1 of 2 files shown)


hoangdat:
OK for the case of crashed. How about the work with `lemon`, why we can not exchange to get the new token when it is expired?

dab246:
> OK for the case of crashed. How about the work with `lemon`, why we can not exchange to get the new token when it is expired?

It was a matter of the `Lemon` side. Should we check why it returns `refreshToken` as empty along with `accessToken`?

nqhhdev (approved):

chibenwa:
> Should we check why it returns refreshToken as empty along with accessToken?

Yes of course! I'll ping @guimard

guimard:
> > Should we check why it returns refreshToken as empty along with accessToken?
> 
> Yes of course! I'll ping @guimard

Hi,

I just fixed LNG configuration

sherlockvn (approved):

Issue #2057, opened 2023-07-26
[Dev] Display calendar detail widget in email view


dab246:
## Description

```
GIVEN an email with the X-MEETING-UID header
WHEN I open it
THEN the calendar widget includes event detail
```

![Image](https://user-images.githubusercontent.com/80730648/256217644-4dce2202-a918-4619-adb9-9e2da3d1e2c4.png)

## Technical steps

This can be obtained via the `EventCalendar/parse`

PR #2059, opened 2023-07-28
TF-2057 Display calendar detail widget in email view

dab246:
## Issue

#2057 

## Resolved

- Mobile

https://github.com/linagora/tmail-flutter/assets/80730648/5df1c476-1127-419f-b1d5-8843ba09236a

- Tablet

https://github.com/linagora/tmail-flutter/assets/80730648/d2385f89-a629-4d0c-a5e0-6df6b8e232ff

- Web

https://github.com/linagora/tmail-flutter/assets/80730648/9cd42875-8110-43fc-8cf7-3521a2bfc9ed


lib/features/base/styles/hyper_link_widget_styles.dart
```diff
@@ -0,0 +1,8 @@
+
```
nqhhdev, line 1: Remove line

lib/features/email/presentation/email_view.dart
```diff
@@ -281,54 +283,55 @@ class EmailView extends GetWidget<SingleEmailController> with AppLoaderMixin {
...
+            } else {
+              return const SizedBox.shrink();
+            }
           } else {
-            return const SizedBox.shrink();
+            if (success is GetEmailContentLoading) {
+              return loadingWidget;
+            } else {
+              return const SizedBox.shrink();
+            }
```
sherlockvn, line 334: i really want to reduce if-else nested, but don't know how. wdyt ?

lib/features/email/presentation/extensions/calendar_event_extension.dart
```diff
@@ -269,4 +269,20 @@ extension CalendarEventExtension on CalendarEvent {
...
     }
   }
+
+  List<String> get videoConferences {
+    if (extensionFields != null && extensionFields!.mapFields.isNotEmpty) {
+      final videoConferences = extensionFields!.mapFields['X-OPENPAAS-VIDEOCONFERENCE'];
+      if (videoConferences != null) {
+        final videoConferencesNotEmpty = videoConferences
+          .whereNotNull()
+          .where((link) => link.isNotEmpty)
```
sherlockvn, line 279: do you think `.where((link) => link != null && link.isNotEmpty)` faster ?

dab246: It's not faster. Above all, the result you get is a list of objects that can be `nullable` during compile time.

lib/features/email/presentation/email_view.dart
dab246: In many cases, we need to separate the if-else condition clearly so that it is easy to understand to avoid the case. To make it easier to understand we just need to separate it into functions. But the if-else condition is still the same

sherlockvn (approved):

PR #2104, opened 2023-08-22
TF-1974 Fix refresh token OIDC failed

dab246:
## Issue

#1974 

## Root cause

- We get this error when we call refreshToken: `PlatformException(token_failed, Concurrent operations detected: token, token, null, null)` . As implied by the error, this would mean your application is making concurrent calls to the plugin where it's not waiting for an operation to finish before starting another one.

## Solution

- Use `QueuedInterceptorsWrapper`.  If there are multiple concurrent requests, the request is added to a queue before
entering the interceptor. Only one request at a time enters the interceptor, and after that request is processed by the interceptor, the next request will enter the interceptor.

## Resolved

- Work fine  on device Android

### DIFF
--- a/lib/features/login/data/network/config/authorization_interceptors.dart
+++ b/lib/features/login/data/network/config/authorization_interceptors.dart
@@ -13,7 +13,7 @@ import 'package:tmail_ui_user/features/login/data/local/account_cache_manager.da
 import 'package:tmail_ui_user/features/login/data/local/token_oidc_cache_manager.dart';
 import 'package:tmail_ui_user/features/login/data/network/authentication_client/authentication_client_base.dart';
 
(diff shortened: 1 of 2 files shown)


lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -146,6 +139,16 @@ class AuthorizationInterceptors extends InterceptorsWrapper {
 
   bool _isRefreshTokenNotEmpty() => _token != null && _token!.refreshToken.isNotEmpty;
 
+  bool _isAuthenticationError(DioError dioError) {
```
hoangdat, line 142:
```suggestion
  bool validateToRefreshToken(DioError dioError) {
```

hoangdat (approved):

PR #2111, opened 2023-08-28
[HOT-FIX] Fix the login screen freezes when pasting the link in the browser address bar

dab246:
## Issue
- [x] Fix the login screen freezes when pasting the link in the browser address bar
- [x] Fix TextEditingController was used after being disposed in LoginView
- [x] Make the beautify logging tool
- [x] Replace all png icon to svg icon

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/22ceee53-abdf-4511-be26-7b45e060fd62


lib/features/login/presentation/login_bindings.dart
```diff
@@ -54,7 +54,7 @@ class LoginBindings extends BaseBindings {
 
   @override
   void bindingsController() {
-    Get.lazyPut(() => LoginController(
+    Get.create(() => LoginController(
```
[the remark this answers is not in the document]
dab246: We use `create` instead of `lazyPut` in order to always create a new `instance` every time we move to the `Login` screen. If you use `lazyPut` it has only 1 instance, so in case when you do the `Login -> Session -> Login` screen move. If this is done very quickly, it will result that LoginView2 will use the controller of LoginView1 while LoginController1 has been disposed.

core/lib/utils/app_logger.dart
```diff
@@ -3,17 +3,38 @@ import 'dart:async';
...
+
+  String logsStr = value ?? '';
+  logHistory.value = '$logsStr\n${logHistory.value}';
+
+  switch (level) {
+    case Level.wtf:
+      logsStr = '\x1B[31m!!!CRITICAL!!! $logsStr\x1B[0m';
+      break;
+    case Level.error:
+      logsStr = '\x1B[31m$logsStr\x1B[0m';
```
sherlockvn, line 19: But how developer know it's an error, a warning, ...

dab246: IMO, It's clearly marked by color, enum value and `TeamMail` key.

hoangdat (approved):

Issue #2129, opened 2023-09-07
[Attach image] I cannot attach image


QuangHoang1210:
OS: Window
Device: IOS/ANDROID
TMail version: 0.9.1

*Precondition: 
- Mobile (IOS/Android) 0.9.1
- Tmail.linagora.com production v0.9.0

### *Reproduce Step:
1, User click on [Compose] button
2, User attach an image

### **Actual:** 
Error msg: Cannot upload this image
User cannot attach an image

### **Expected:** 
User can attach an image.

https://user-images.githubusercontent.com/124866146/266187352-64d19600-4ea1-41f8-8689-a4ffcb2644e2.mp4

QuangHoang1210:
passed on 0.9.3 canary

PR #2132, opened 2023-09-07
TF-2129 Fix cannot attach file on mobile

dab246:
## Issue

#2129 

## Root cause

- Impact from upgrading version flutter `3.10.6` and dart `3.0.0`. Dart SDK have tightened the rules for which objects can be sent across isolates (see https://api.dart.dev/stable/3.0.0/dart-isolate/SendPort/send.html and https://github.com/dart-lang/sdk/commit/67683c39)

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/b836a564-b032-4a01-9cbb-b1b4716bcebb

### DIFF
--- a/lib/features/upload/domain/model/mobile_file_upload.dart
+++ b/lib/features/upload/domain/model/mobile_file_upload.dart
@@ -1,0 +1,24 @@
+
+import 'package:equatable/equatable.dart';
+
+class MobileFileUpload with EquatableMixin {
+  final String fileName;
+  final String filePath;
+  final int fileSize;
+  final String mimeType;
+
+  MobileFileUpload(
+    this.fileName,
+    this.filePath,
+    this.fileSize,
+    this.mimeType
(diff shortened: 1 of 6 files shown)


lib/features/upload/domain/model/mobile_file_upload.dart
```diff
@@ -0,0 +1,24 @@
+
+import 'package:equatable/equatable.dart';
+
+class MobileFileUpload with EquatableMixin {
+  final String fileName;
+  final String filePath;
+  final int fileSize;
```
sherlockvn, line 7: should have unit also

dab246: IMO, unnecessary, because it's clear

nqhhdev (approved):

hoangdat (approved):

PR #2133, opened 2023-09-08
[HOT-FIX] Fix refresh token with oidc on `upn`

dab246:
## Issue

#1974 

## Root cause

- When executing tasks concurrently, they are pushed into the queue along with the old `header` values (the old `authentication` is retained). So when the first request receives a `401` error and tries to get a new token, it will be updated with the new `authentication-header` value.

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/fc883e99-a417-49f9-93eb-e6ad9db28cb1


hoangdat:
Can you write it and the implementation of `queue interceptor` in ADR. Thanks

hoangdat (approved):

Issue #2121, opened 2023-09-04
[BENOIT's MUM] Drag and drop text from outside the composer inside the composer


chibenwa:
### Description

When I drag and drop text inside the composer, line breaks are lost.

### Expected result

When I drag end drop text inside the composer, line break should not be lost.

Remarks: When I copy and paste text from outside the composer inside the composer, line breaks are not lost.

![Screenshot from 2023-09-04 20-44-13](https://github.com/linagora/tmail-flutter/assets/6928740/e627e857-334a-4afe-80b0-c375b4e97b11)
![Screenshot from 2023-09-04 20-44-43](https://github.com/linagora/tmail-flutter/assets/6928740/05046d43-1b28-4b90-b4df-5cfe3ae01034)

### Current behavior
When I drag and drop text from outside of the app inside the composer, line break are lost.

![Screenshot from 2023-09-04 20-45-03](https://github.com/linagora/tmail-flutter/assets/6928740/cdeffe63-8659-4059-8a3e-9ab18fe02d9e)

![Screenshot from 2023-09-04 20-45-09](https://github.com/linagora/tmail-flutter/assets/6928740/8073da16-be29-4c22-9323-4726ba24baba)

![Screenshot from 2023-09-04 20-45-35](https://github.com/linagora/tmail-flutter/assets/6928740/e44085a1-fc4f-4f88-9f70-e0c0b0d3ea93)

PR #2058, opened 2023-07-26
TF-2042 Display calendar information widget in email view

dab246:
## Issue

#2042 

## Resolved

- Mobile:

https://github.com/linagora/tmail-flutter/assets/80730648/de54389f-2098-4dcd-80a6-f0d66864cfc1

- Tablet: 

https://github.com/linagora/tmail-flutter/assets/80730648/20dd8928-3437-429a-993c-d72186720e08

PR #2150, opened 2023-09-15
TF-2121 Fix drag and drop text outside the composer inside the composer

dab246:
## Issue

#2121 

## Root cause

- Because the `drag and drop` handling function of the `summernote` library is incorrect.
- Fixed here: https://github.com/linagora/html-editor-enhanced/pull/22

## Dependent

- Need merged: https://github.com/linagora/html-editor-enhanced/pull/22

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/0a0b30fa-5891-4514-a9fb-ba838c17e2c9

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -913,8 +913,8 @@ packages:
     dependency: "direct main"
     description:
       path: "."
-      ref: email_supported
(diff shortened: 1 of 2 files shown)


hoangdat (approved):

dab246:
Fix the same library #2058

Issue #2160, opened 2023-09-20
[Mobile] A part of the bottom of email content has been cut off -- only happen with long email html template


QuangHoang1210:
OS: IOS / Android
Device: Mobile
TMail version: 0.9.2 

### *ENV:
- jmap.linagora.com
- jmap.linagora.com/oidc

### *Reproduce Step:
1, User A send an email to user B <following email file bellow>
2, User A, B view email details and scroll down to view all of email

### **Actual:** 
A part of the bottom of email content has been cut off. Cannot load more

### **Expected:** 
A, B should view all email content

https://user-images.githubusercontent.com/124866146/269163570-944ee61a-0042-43df-bc4d-491a0347c349.mp4

QuangHoang1210:
_**For the email in inbox -> Passed

But the email content has been cut off in case (Foward / Reply)**_

### ENV: Mobile -- IOS /ANDROID -- Tmail v.0.10.1 

### Test data: 

[⛳ Tous les coups comptent !  😄.zip](https://github.com/linagora/tmail-flutter/files/12920693/Tous.les.coups.comptent.zip)


### Reproduce step:
1, User open email above
2, User click on Reply/Foward
3, User scroll down to view the email content

### Actual:
- A part of email content is missing

### Expected:
- User can view full email content. Because sometime user want to edit the email content during Reply/Foward before sending it

https://github.com/linagora/tmail-flutter/assets/124866146/cf9617d8-f719-4496-b020-243fb9ef0d93

PR #2173, opened 2023-09-26
TF-2160 Fix email content is cut off

dab246:
## Issue

#2160 

## Root cause

- Because some `img` tags do not have a defined height. So when we use `lazy loading` image, the calculation of the content height (`body.scrollHeight`) will not be accurate because the image appears in the newly downloaded frame.

## Solution

- We will listen for size changes in `_onContentSizeChanged`, and only update `actualHeight` when we find a height greater than the current height.

## Resolved

- Mobile

https://github.com/linagora/tmail-flutter/assets/80730648/7c771e07-6a99-4948-adc6-bd34484de3ac

- Web:

https://github.com/linagora/tmail-flutter/assets/80730648/f6e4343d-0351-4d69-bdfc-cc92e51db32f


hoangdat:
please add an ADR on the reason while switch from loading in html to dio. Thanks

dab246:
> please add an ADR on the reason while switch from loading in html to dio. Thanks


No, we won't be loading images with `dio` to get the size anymore. Reason:
- Because many html content contains a lot of images, the loading process also takes quite a while,
- Download size is the actual size of the image. It is not suitable for display sizes on mobile devices.

So I used another solution to fix it. Please see the description at the top of the PR

hoangdat (approved):

Issue #2189, opened 2023-09-29
[Offline] while app in offline, After click on send button in composer, dialog show, but when click outside the dialog, all disappear


hoangdat:
OS:  Android / IOS
Device: Mobile
TMail version: 0.9.3

### *Reproduce Step:
1, User A turn off network
2, User A click on [Composer] button
3, User A input correct all input fields
4, User click [Send] button
5, User click outside the Sending offline dialog

### **Actual:** 
After step 4
- System redirect user to previous screen 
- Sending offline dialog is displayed
After step5
- Dialog is disappeared
- The sending email is removed
- Sending email is not go in drafts or anywhere

### **Expected:** 
After step 4
- System keep displaying Composer screen
- Sending offline dialog is displayed
After step5
- Dialog is disappeared
- System keep displaying composer screen where user can continue writing email

https://github.com/linagora/tmail-flutter/assets/124866146/890cfac2-02d7-4422-a515-230458ceaadf

dab246:
It seems you are misunderstanding something here. If you do not perform any action on dialog. Looks like you've ignored it, and we haven't done anything about it yet.

QuangHoang1210:
@dab246  This is UX supportive. We should let user keep continue the writing email actions in case closing dialog once click outisde.

Orther way we can prevent user click outside the dialog.

hoangdat:
- [ ] keep composer available before click on proceed
- [ ] `cancel`- `close`

QuangHoang1210:
**-> Passed on Android 

-> Failed on IOS**

### ENV: IOS - Tmail 0.10.1

### Precondition: Turn off Wifi

### Reproduce step:
1, User send an email in offline state

### Actual: 
- System not showing dialog but "Email being sent" toast will be displayed

### Expected:
- System show dialog Sending Offline Process normally and same Android behaviour

https://github.com/linagora/tmail-flutter/assets/124866146/866d50ac-bbbc-4b0a-b481-0317a406bf97

hoangdat:
Only in iOS:

- Dialog the same with offline but the `Proceed` button replace by `Save to Draft`
- Title: `You are in offline mode`
- Message: `The message will be Save to Draft. You can try again when being online.`
- Buttons: `Cancel` `Save to Draft`.

PR #2274, opened 2023-10-19
TF-2189 Fix sending queue on iOS

dab246:
## Issue

- #2189 

## Dependent 

- Need merged: https://github.com/linagora/enough_html_editor/pull/23

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/b9a3662b-b035-4c36-a083-5ff056e14964

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -54,6 +54,7 @@ import 'package:tmail_ui_user/features/composer/presentation/model/draggable_ema
 import 'package:tmail_ui_user/features/composer/presentation/model/image_source.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/inline_image.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/prefix_recipient_state.dart';
+import 'package:tmail_ui_user/features/composer/presentation/model/save_to_draft_arguments.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/save_to_draft_view_event.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/screen_display_mode.dart';
 import 'package:tmail_ui_user/features/composer/presentation/styles/composer_style.dart';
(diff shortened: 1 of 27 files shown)


lib/features/composer/presentation/composer_controller.dart
```diff
@@ -862,134 +865,128 @@ class ComposerController extends BaseController {
...
+    if (composerArguments.value == null ||
+        mailboxDashBoardController.userProfile.value == null ||
+        mailboxDashBoardController.sessionCurrent == null ||
+        mailboxDashBoardController.accountId.value == null
+    ) {
+      logError('ComposerController::isParamUserNotNull: Param is NULL');
+      return true;
+    }
+    return false;
+  }
```
hoangdat, line 878:
```suggestion
  void _validateUserParams {
    if (composerArguments.value == null ||
        mailboxDashBoardController.userProfile.value == null ||
        mailboxDashBoardController.sessionCurrent == null ||
        mailboxDashBoardController.accountId.value == null
    ) {
      throw `ABC Exception`;
    }
    return;
  }
```
then handle exception in `final`

dab246: We don't take any action on it. IMO, it is not necessary to do so on the presentation layer

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1294,21 +1269,14 @@ class ComposerController extends BaseController {
...
-    if (event.emailIdEditing == null) {
-      mailboxDashBoardController.saveEmailToDraft(
-        session: event.session,
-        accountId: event.accountId,
-        email: newEmail
-      );
-    } else {
-      mailboxDashBoardController.updateEmailToDraft(
+    mailboxDashBoardController.saveEmailToDraft(
+      arguments: SaveToDraftArguments(
```
hoangdat, line 1281: why use `save to draft` for this case? it is also save in local db?

pubspec.yaml
```diff
@@ -239,6 +239,11 @@ dependency_overrides:
 
   pointer_interceptor: 0.9.1
 
+  enough_html_editor:
+    git:
+      url: https://github.com/linagora/enough_html_editor.git
+      ref: improvement/support-get-content-signature
```
hoangdat, line 245: please use `email_supported`

hoangdat (commented):
only one question: why we need to change `save to draft` argument?

dab246:
> only one question: why we need to change `save to draft` argument?

We will handle it on the dashboard controller, avoiding the case of the composer controller being canceled

hoangdat (approved):

Issue #2177, opened 2023-09-27
[Attachment] Download attachment file is not working


hoangdat:
### Desc:
- Max Nguyen: download in Chrome, Window 11 -> no download request, v0.9.3

dab246:
Work fine on Mac.
@QuangHoang1210  please verify it on windows/linux

chibenwa:
I encouter the same on Chromium with 0.9.0

Since the switch to 0.9.3 no issue

QuangHoang1210:
@dab246  @hoangdat  Still reproduce but rarely

PR #2283, opened 2023-10-23
TF-2177 Download attachment file is not working

dab246:
## Issue

#2177 

## Root cause

- Due Token not found in cache, multiple accounts exist at the same time

Screenshot 2023-10-23 at 10 54 50

Screenshot 2023-10-23 at 11 32 05

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/899542ba-c256-49e2-a84a-f0adf170e3a3

### DIFF
--- a/lib/features/login/data/extensions/list_account_cache_extensions.dart
+++ b/lib/features/login/data/extensions/list_account_cache_extensions.dart
@@ -1,0 +1,23 @@
+import 'package:collection/collection.dart';
+import 'package:core/utils/app_logger.dart';
+import 'package:tmail_ui_user/features/login/data/extensions/account_cache_extensions.dart';
+import 'package:tmail_ui_user/features/login/data/model/account_cache.dart';
+
+extension ListAccountCacheExtension on List<AccountCache> {
+  List<AccountCache> unselected() => map((account) => account.unselected()).toList();
+
+  List<AccountCache> removeDuplicated() {
+    final listAccountId = map((account) => account.accountId).whereNotNull().toSet();
+    log('ListAccountCacheExtension::removeDuplicated:listAccountId: $listAccountId');
+    retainWhere((account) => listAccountId.remove(account.accountId));
+    log('ListAccountCacheExtension::removeDuplicated:listAccount: $this');
+    return this;
+  }
+
+  Map<String, AccountCache> toMap() {
+    return {
+      for (var account in this)
+        account.id: account
+    };
+  }
+}
--- a/lib/features/login/data/local/account_cache_manager.dart
+++ b/lib/features/login/data/local/account_cache_manager.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:core/utils/app_logger.dart';
 import 'package:model/account/personal_account.dart';
 import 'package:tmail_ui_user/features/caching/clients/account_cache_client.dart';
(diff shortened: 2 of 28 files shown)


lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -94,20 +91,20 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         _token!.refreshToken
       );
 
-      final accountCurrent = await _accountCacheManager.getSelectedAccount();
+      final currentAccount = await _accountCacheManager.getCurrentAccount();
 
-      await _accountCacheManager.deleteSelectedAccount(_token!.tokenIdHash);
```
hoangdat, line 99: As I understand, we failed to delete account?

lib/features/login/domain/repository/account_repository.dart
```diff
@@ -6,5 +6,5 @@ abstract class AccountRepository {
 
   Future<void> setCurrentAccount(PersonalAccount newCurrentAccount);
 
-  Future<void> deleteCurrentAccount(String accountId);
+  Future<void> deleteCurrentAccount(String hashId);
```
hoangdat, line 9: IMO, no problem with `accountId`. `hashId` make more confused

dab246: accountId easily confused with AccountId in Session. hashId is just an identifier for a cached account. It can be tokenHashId or UserName

lib/features/login/data/network/config/authorization_interceptors.dart
dab246: It doesn't mean it can't be erased. But we mistakenly delete it when deleting according to the token variable stored in memory

lib/features/login/data/local/account_cache_manager.dart
```diff
@@ -10,24 +11,29 @@ class AccountCacheManager {
...
     }
   }
 
-  Future<void> setSelectedAccount(PersonalAccount account) {
-    log('AccountCacheManager::setSelectedAccount(): $_accountCacheClient');
+  Future<void> setCurrentAccount(PersonalAccount account) async {
+    log('AccountCacheManager::setCurrentAccount(): $account');
+    final accountCacheExist = await _accountCacheClient.isExistTable();
+    if (accountCacheExist) {
+      await _accountCacheClient.clearAllData();
```
hoangdat, line 30: - Move it to migration script.
- set all other account to false in `isSelected`
- delete all duplicated account ()

lib/features/login/data/extensions/list_account_cache_extensions.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:collection/collection.dart';
+import 'package:core/utils/app_logger.dart';
+import 'package:tmail_ui_user/features/login/data/extensions/account_cache_extensions.dart';
+import 'package:tmail_ui_user/features/login/data/model/account_cache.dart';
+
+extension ListAccountCacheExtension on List<AccountCache> {
+  List<AccountCache> unselected() => map((account) => account.unselected()).toList();
+
+  List<AccountCache> removeDuplicated() {
+    final listAccountId = map((account) => account.accountId).whereNotNull().toSet();
```
hoangdat, line 10: What happens for this case: id is different but all other fields are the same

dab246: It will delete duplicates. Please see unit test

hoangdat (approved):

Issue #2245, opened 2023-10-10
[Attachment] Cannot upload attachment or Image on Canary v0.10.0


QuangHoang1210:
OS: WINDOW
Device: Mobile, LAPTOP
TMail version: 0.10.0 https://canary-tmail.linagora.com/

### *Reproduce Step:
1, User click on composer
2, User click on Attachment icon
3, User select any files

### **Actual:** 
It's uploading failed

### **Expected:** 
Upload successfully

### *Noted: It happens same with  upload  image

**Attachment:**

https://github.com/linagora/tmail-flutter/assets/124866146/1b1ec228-112b-4dd1-a8cb-6080c6bfbf21

**Upload image:**

https://github.com/linagora/tmail-flutter/assets/124866146/730cdee4-8749-4ec9-b447-d5462ac8c959

dab246:
Please compare with version `0.9.3` on tmail.linagora.com

hoangdat:
- [ ] add refresh token flow

hoangdat:
- can upload with new token, but no response in a long time 
- try to use new token with postman -> 401

hoangdat:
verify on 0.10.4

QuangHoang1210:
Passed on 0.10.4 canary and 0.10.3 production


https://github.com/linagora/tmail-flutter/assets/124866146/ff174168-7d75-4a8b-a2c8-17879efbad86

PR #2285, opened 2023-10-24
TF-2245 Cannot upload attachment or image when token expired

dab246:
## Issue

#2245

## Root cause

- Because data is uploaded in Stream format. So every time we call the request again we have to re-initialize it.

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/9b58be00-ef78-4c74-b916-c45c6b5cabdf

### DIFF
--- a/lib/features/login/data/network/config/authorization_interceptors.dart
+++ b/lib/features/login/data/network/config/authorization_interceptors.dart
@@ -128,7 +128,9 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
         final response = await _dio.request(
           requestOptions.path,
-          data: BodyBytesStream.fromBytes(uploadExtra[FileUploader.bytesExtraKey]),
+          data: uploadExtra[FileUploader.platformExtraKey] == 'web'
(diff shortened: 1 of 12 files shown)


lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -113,16 +114,38 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+        );
+
+        final response = await _dio.request(
+          requestOptions.path,
+          data: uploadExtra[FileUploader.platformExtraKey] == 'web'
+            ? BodyBytesStream.fromBytes(uploadExtra[FileUploader.bytesExtraKey])
+            : File(uploadExtra[FileUploader.filePathExtraKey]).openRead(),
+          queryParameters: requestOptions.queryParameters,
+          options: newOptions,
+        );
```
hoangdat, line 136: can we separate the to method?

dab246: oke, i will update it

hoangdat (approved):

PR #2286, opened 2023-10-24
Feat: Fix sending queue on mobile

dab246:
## Issues

- [x] Remove config work manager on iOS
- [x] Fix sending queue return failure on Android


lib/features/offline_mode/work_manager/sending_email_worker.dart
```diff
@@ -76,9 +76,11 @@ class SendingEmailWorker extends Worker {
       HiveCacheConfig().setUp()
     ]);
 
-    SendingQueueInteractorBindings().dependencies();
-    SendEmailInteractorBindings().dependencies();
-    SendingQueueBindings().dependencies();
+    await Future.sync(() {
```
hoangdat, line 79: srr, but why it can fix the issue?

dab246: SendingQueueController uses MailboxDashboardController so we need to bind it.

hoangdat (approved):

PR #2293, opened 2023-10-27
Rel v0.10.2

hoangdat:
### Added
- Change font size in composer
- New attachment view
- Scroll to top on web

### Fixed
- Message content is cutted
- Signature button is included in email content
- Download error when token refreshed
- Upload failed when token expired


.github/workflows/gh-pages.yaml
```diff
@@ -5,6 +5,9 @@ on:
 
 name: Deploy PR on Github Pages
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+
```
dab246, line 10: what is this?

dab246 (approved):

PR #2303, opened 2023-11-02
Added 404 page to handle redirect

tk-nguyen:
This should fix `gh-pages` redirect after logging in through SSO


dab246:
@tk-nguyen Does it work?

tk-nguyen:
> @tk-nguyen Does it work?

Doesn't appear to be :(

Issue #2311, opened 2023-11-06
User has been log out on Tmail version 0.10.4 once the network connection is not stable


QuangHoang1210:
OS: IOS / Android
Device: Mobile
TMail version: 0.10.4

### *ENV:
- jmap.linagora.com/oidc

### *Reproduce Step:
1, User login successfully
2, User kill app TMAIL
3, User turn off wifi
4, User open TMAIL
5, User turn on Wifi
6, User open TMAIL again

### **Actual:** 
User is logout, tmail is loading forever
(sometime it's loading forever, sometime it's loading sucessfully but user log out)

### **Expected:** 
User keep login

https://github.com/linagora/tmail-flutter/assets/124866146/e2dcb61c-ec94-4175-99ed-6cbdc3db04b5

PR #2319, opened 2023-11-13
TF-2311 Fix user has been log out on Tmail once the network connection is not stable

dab246:
## Issue

#2311 

## Root cause

Because getting session from server failed leading to logout. Sometimes the process of making a request to get session `/.well-known/jmap` times out indefinitely.

## Resolved

https://github.com/linagora/tmail-flutter/assets/80730648/12558971-29f4-4a61-8bc9-135ef13afb7b


lib/features/base/base_controller.dart
```diff
@@ -168,6 +160,12 @@ abstract class BaseController extends GetxController
 
   void handleExceptionAction({Failure? failure, Exception? exception}) {
     logError('BaseController::handleExceptionAction():failure: $failure | exception: $exception');
+    if (!authorizationInterceptors.isAppRunning) {
+      return;
+    }
+    if (exception is BadCredentialsException || exception is ConnectionError) {
```
hoangdat, line 166: Why need to log out when we have `ConnectionError`?

lib/features/session/presentation/session_controller.dart
```diff
@@ -53,9 +53,12 @@ class SessionController extends ReloadableController {
 
   @override
   void handleFailureViewState(Failure failure) {
-    super.handleFailureViewState(failure);
     if (failure is GetSessionFailure) {
       _handleSessionFailure(failure);
+    } if (failure is GetStoredSessionFailure) {
+      _handleSessionFailure(failure);
```
hoangdat, line 59: I worry about the recursive call to this function.

lib/features/session/presentation/session_controller.dart
```diff
@@ -85,39 +89,41 @@ class SessionController extends ReloadableController {
...
-      if (errorMessage.isNotEmpty && currentOverlayContext != null) {
-        _appToast.showToastErrorMessage(currentOverlayContext!, errorMessage);
-      }
+    logError('SessionController::_handleSessionFailure():errorMessage: $errorMessage');
+    if (errorMessage.isNotEmpty && currentOverlayContext != null) {
+      _appToast.showToastErrorMessage(currentOverlayContext!, errorMessage);
+    }
+
+    if (PlatformInfo.isMobile) {
+      _handleGetStoredSession();
```
hoangdat, line 124: Recursive point

hoangdat: why not handle it directly in `interceptor`?

dab246: > why not handle it directly in `interceptor`?

Why interceptor? IMO it is interactor?

lib/features/base/base_controller.dart
dab246: It is one of the cases we tested before. When response is null

lib/features/login/data/network/config/authorization_interceptors.dart
```diff
@@ -74,6 +78,10 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
       case AuthenticationType.none:
         break;
     }
+    if (PlatformInfo.isMobile && _validateToGetSession(options.uri)) {
+      options.connectTimeout = _sessionConnectTimeout;
+      options.receiveTimeout = _sessionReceiveTimeout;
+    }
```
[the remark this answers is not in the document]
dab246: IMO, it's not very necessary. But it's okay to separate.

lib/features/session/presentation/session_controller.dart
hoangdat: I mean, `presentation` no need to care it is come from `server` or `local`. Only give `presentation` the result. Keep the logic of get in somewhere (maybe the interceptor, when error in `server`, get from `local`....)

dab246: Before, we only wanted to call it when we lost network connection and on mobile so we handled it at the presentation layer. Furthermore, we always want to get the latest session from the server because it can change at any time. But it leads to many error cases like this issue, so we will handle it directly in the interactor

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -377,6 +373,14 @@ class ThreadController extends BaseController with EmailActionController {
...
       }
       _getAllEmailAction();
+    } else if (error is MethodLevelErrors) {
+      if (currentOverlayContext != null && error.message?.isNotEmpty == true) {
+        appToast.showToastErrorMessage(
+          currentOverlayContext!,
+          error.message!
+        );
+      }
+      clearState();
```
hoangdat, line 383: in what case we have this?

dab246: When calling jmap request `Email/query` we get errors from jmap `MethodLevelErrors`. (During testing I encountered it several times on the preprod server). So it always shows the loading bar on the ThreadView screen so we should show the error message to the user and reset `ViewState` to hide the loading bar.

lib/features/base/base_controller.dart
```diff
@@ -169,7 +160,7 @@ abstract class BaseController extends GetxController
     if (!authorizationInterceptors.isAppRunning) {
       return;
     }
-    if (exception is BadCredentialsException || exception is ConnectionError) {
```
hoangdat: please revert this

Issue #1718, opened 2023-04-04
Helper to easily unsubscribe from email / mailing lists


chibenwa:
# Summary

Commercial email and email from mailing lists often include information to easily unsubscribe.

EG:

```
List-Unsubscribe: <mailto:topdevvn+unsubscribe_637edc2ee899a039346076@gmail.com>,
 <https://hr.topdev.vn/email/unsubscribe/637edc2ee899a039346076>
```

In `Email detailed view` when I open such email, add an option to easily unsubscribe from this email.

https links should be prefered, but mailto links, if they are the only available, should be handled too.

# Story
### Description

- Given that I am TMail user
- After log-in successfully, I can see the list of emails of Inbox folder 
- If I signed up on a site that sends lots of emails, like promotions or newsletters, and the sender provides in their email content an unsubscribe link, then TMail will display a more convenient way to unsubscribe from the mailing list:
- On email detail screen, I can see a hyperlink "Unsubscribe" next to the sender's name, When I click on this link, I can see a popup asking:"Are you sure you'd like to stop receiving similar messages from [Sender's name]?
-  I click button Unsubscribe in the popup to confirm:
   - If the unsubscribe is a https link, then the system will send the unsubscribe-confirmation to the sender through this link without opening it 
   - If the unsubscribe is a mailto link, then the system will auto-generate a message with pre-filled fields as included in the mailto link then send to the sender to unsubscribe. 
- After unsubscribe successfully, there will be a toast message: "Unsubscribed from this mailing list"
- Now When I opened the message that I just unsubscribe before, the hyperlink:"Unsubscribe" next to sender name is not displayed anymore, and there will be a text under saying :"You unsubscribe from [Sender's email address]".

### Screenshots
Web: 

![Group 158](https://github.com/linagora/tmail-flutter/assets/68209176/9e8115c7-9fa7-48ee-b411-73c05dfbf1ed)

![Group 159](https://github.com/linagora/tmail-flutter/assets/68209176/876c8904-040f-4247-b05f-b02edec0a90b)

![Group 160](https://github.com/linagora/tmail-flutter/assets/68209176/cc96142f-1257-4ab6-9323-3bb09114d9eb)

Mobile:

[...]

PR #2330, opened 2023-11-16
TF-1718 Helper to easily unsubscribe from email / mailing lists

dab246:
## Issue

#1718 

### 1. Web

https://github.com/linagora/tmail-flutter/assets/80730648/aac58535-ce07-4ef6-84c7-ba8ead7826d0

### 2. Mobile

https://github.com/linagora/tmail-flutter/assets/80730648/2673a49c-6f85-47a3-b9f3-629c518d2001


hoangdat:
how about the case `unsubcribe` only inside of email content?

dab246:
> how about the case `unsubcribe` only inside of email content?

We don't care about it. Because it belongs to a 3rd party. I have checked some clients like Gmail and it also cannot handle such cases.

hoangdat (approved):

Issue #2363, opened 2023-11-27
Native browser navigation


chibenwa:
Browser arrow do not yield expected results...

 - WHEN I click back and I am in an open email, I do not get back onto the mailbox
 - When I click back after switching into another mailbox I am not back to INBOX
 
Uploading tmail-native-browser-navigation.mp4…

Browser should remember the URL visited and add it to its history. I should also allow to switch back and forth within that history

Issue #1831, opened 2023-05-11
System not response correctly view belong to URL once user open email details view then click back button of browser


QuangHoang1210:
OS: Window
Device: Laptop
TMail version: 0.7.8

### *Reproduce Step:
1, User open one email to view email details
2, User click on back button of browser

### **Actual:** 
- System redirect URL: Tmail.linagora.com/#
- System not redirect user to correctly view belong to URL. It keeping the current email details view.

### **Expected:** 
- System redirect URL: Tmail.linagora.com/#
- System should redirect user to correctly view belong to URL.

https://user-images.githubusercontent.com/124866146/237561134-a6657039-34c9-4049-9fd6-0647debe9729.mp4

PR #2397, opened 2023-12-10
TF-1831 System not response correctly view when click back button

dab246:
## Issues

- [x] #1831 
- [x] #2363 

## Resolved

- Demo Web:

https://github.com/linagora/tmail-flutter/assets/80730648/46b0e9ec-2ed0-4354-857d-f6f2464194a8

- Demo Mobile:

https://github.com/linagora/tmail-flutter/assets/80730648/a3abb751-e482-4db6-bf8a-366049481651

### DIFF
--- a/lib/features/base/base_mailbox_controller.dart
+++ b/lib/features/base/base_mailbox_controller.dart
@@ -338,36 +338,34 @@ abstract class BaseMailboxController extends BaseController {
             )))
       ).show(context);
     } else {
-      showDialog(
-        context: context,
+      Get.dialog(
+        PointerInterceptor(child: (EditTextDialogBuilder()
+          ..key(const Key('rename_mailbox_dialog'))
+          ..title(AppLocalizations.of(context).renameFolder)
+          ..cancelText(AppLocalizations.of(context).cancel)
+          ..setErrorString((value) {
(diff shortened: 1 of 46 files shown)


lib/features/base/base_mailbox_controller.dart
```diff
@@ -418,10 +416,8 @@ abstract class BaseMailboxController extends BaseController {
         ..onConfirmAction(AppLocalizations.of(context).delete, () => onDeleteMailboxAction(presentationMailbox))
       ).show();
     } else {
-      showDialog(
-        context: context,
-        barrierColor: AppColor.colorDefaultCupertinoActionSheet,
-        builder: (context) => PointerInterceptor(
+      Get.dialog(
```
hoangdat, line 419: what is the different? no context? it is safe to manage dialog?

dab246: - Removing that Context is a special feature of GetX, it will not be affected, allowing us to create it in one place in the app.
- We use GetX's dialog to conveniently check if the dialog is displayed. Because GetX supports the `Get.isDialogOpen` function

hoangdat (approved):

Issue #2384, opened 2023-12-06
Improve notification for iOS to get full details


hoangdat:
### Desc:

Refs: https://github.com/vector-im/element-ios?search=1
Apple: https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications
Demo Twake:  https://github.com/linagora/twake-on-matrix/pull/1078
FCM in Twake: https://pub.dev/packages/fcm_shared_isolate

### DoD:

hoangdat:
- [ ] NSE work: change fcm_shared_isolated, listen noti, trigger receive noti
- [ ] Share access_token & refresh_token
- [ ] use token to get plain notificaiton

hoangdat:
### Part 2
- [ ] alert in APN payload
- [ ] from stateId -> many local notifications 

### Part 3
- [ ] open email when click on notification
- [ ] refresh list when open notification from background

### Part 4
- [ ] hasMoreChange in state in push notification

hoangdat:
### part 5
- [ ] support offline

PR #2408, opened 2023-12-18
TF-2384 [PART-2] Improvement notification iOS

dab246:
## Issue

#2384 

## Task

- [x] Show notification with list emails
- [x] Handle refresh token when token expired

### DIFF
--- a/ios/TwakeCore/Network/Interceptor/AuthenticationInterceptor.swift
+++ b/ios/TwakeCore/Network/Interceptor/AuthenticationInterceptor.swift
@@ -1,0 +1,81 @@
+import Alamofire
+import Foundation
+
+class AuthenticationInterceptor: RequestInterceptor {
+    var authentication: Authentication?
+    let accountId: String
+    var tokenRefreshManager: TokenRefreshManager?
+
+    private lazy var keychainController = KeychainController(service: .sessions,
+                                                             accessGroup: InfoPlistReader.main.keychainAccessGroupIdentifier)
+
+    init(authentication: Authentication?, accountId: String, tokenRefreshManager: TokenRefreshManager?) {
+        self.authentication = authentication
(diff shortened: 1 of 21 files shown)


ios/TwakeMailNSE/NotificationService.swift
```diff
@@ -1,30 +1,33 @@
 import UserNotifications
 
 class NotificationService: UNNotificationServiceExtension {
-    
+
+    private let timeIntervalNotificationTrigger: Int = 2
```
hoangdat, line 5: please specific the time unit: `timeIntervalNotificationTriggerInSecond`

ios/TwakeCore/Network/Interceptor/AuthenticationInterceptor.swift
```diff
@@ -0,0 +1,81 @@
...
+
+            self.keychainController.updateTokenOidc(
+                accountId: self.accountId,
+                newTokenOidc: TokenOidc(
+                    token: accessToken,
+                    tokenId: tokenResponse.tokenId,
+                    expiredTime: "\(tokenResponse.expiresTime ?? 0)",
+                    refreshToken: refreshToken
+                )
+            )
```
hoangdat, line 59: Flutter side can use this updated token or not?

hoangdat, line 59: What happens if `NSE refreshed token` then `Flutter side` also refresh token?

dab246: We will handle it in the interceptor. When a 401 error response is returned, we will use the newly updated `token` in the keychain to continue making the request. If not, a `refreshToken` request will be made

hoangdat (approved):

Issue #2420, opened 2024-01-03
[MULTIPLE-ACCOUNT] Use TupleKey for all boxes in hive database


dab246:
### Local Database
- Account: Id (tokenId in case of OIDC - username in case of BasicAuth) - AccountCache 
  - [ ] combine with `AuthenticationInfo` and `TokenOIDCClien`
  - [ ] add Authority in TokenOIDC
- AuthenticationInfo (BasicAuth infomartion) -> combine to `Account`
  - username - AuthenticationInfoCache
- TokenOIDCCacheClient  -> combine to `account `
- ✅ CacheVersionClient
- ✅ EmailCacheClient: TupleKey - EmailCache
- ✅ EncryptionKeyCache
- ✅ FcmCacheClient: store `State` of Push
- ✅ MailboxCacheClient: TupleKey - MailboxCache
- ✅ NewEmailHiveCacheClient: [Offline-mobile] store incoming email when receive push, always 10 latest
- ✅ OpenedEmailHiveCacheClient: [Offline-mobile] store opened email, 30 latest 
- ✅ Recent LoginUrl/LoginUsername
- ✅ SendingEmailHiveCacheClient
- SessionHiveCacheClient
  - [ ] TupleKey(AccountId, Username) - SessionCache
- ✅ StateCacheClient: tupleKey
- FCMSubscriptionCacheClient
  - [ ] TupleKey(AccountId, Username) - FCMSubscriotionCache

PR #2429, opened 2024-01-04
[MA-PART-1] TF-2420 Use tuple key for boxes in hive database

dab246:
## Issue

#2420

## Task

- [x] Combine PersonalAccount with AuthenticationInfo and TokenOidc
- [x] SessionCache
- [x] FirebaseRegistrationCache

## Demo

[demo.webm](https://github.com/linagora/tmail-flutter/assets/80730648/c207b2ae-89ef-4115-98e5-5209ee1e4ffc)


lib/features/login/domain/usecases/logout_interactor.dart
```diff
@@ -0,0 +1,43 @@
...
+import 'package:core/presentation/state/success.dart';
+import 'package:dartz/dartz.dart';
+import 'package:model/account/authentication_type.dart';
+import 'package:tmail_ui_user/features/login/domain/exceptions/authentication_exception.dart';
+import 'package:tmail_ui_user/features/login/domain/repository/account_repository.dart';
+import 'package:tmail_ui_user/features/login/domain/state/logout_state.dart';
+import 'package:tmail_ui_user/features/login/domain/usecases/logout_basic_auth_interactor.dart';
+import 'package:tmail_ui_user/features/login/domain/usecases/logout_oidc_interactor.dart';
+
+class LogoutInteractor {
```
hoangdat, line 11: IMO, `LogoutCurrentAccountInteractor`

lib/features/push_notification/presentation/controller/fcm_token_controller.dart
```diff
@@ -165,7 +177,14 @@ class FcmTokenController extends FcmBaseController {
...
   void handleSuccessViewState(Success success) {
     log('FcmTokenController::_handleSuccessViewState(): $success');
-    if (success is GetFirebaseRegistrationByDeviceIdSuccess &&
+    if (success is GetAuthenticatedAccountSuccess &&
+        success.account.accountId != null &&
+        success.account.userName != null) {
+      _getStoredFirebaseRegistrationFromCache(
+        success.account.accountId!,
+        success.account.userName!
+      );
```
hoangdat, line 186: How about if we combine it in one interactor `getActiveAccount + `_getStoredFirebaseRegistrationFromCache`?

lib/features/caching/caching_manager.dart
```diff
@@ -113,9 +113,12 @@ class CachingManager {
...
   Future<void> onUpgradeCache(int oldVersion, int newVersion) async {
     log('CachingManager::onUpgradeCache():oldVersion $oldVersion | newVersion: $newVersion');
-    await clearData();
-    if (oldVersion > 0 && oldVersion < newVersion && newVersion == 7) {
+    if (oldVersion > 0 &&
+        oldVersion < newVersion &&
+        (newVersion == 7 || newVersion == 10)) {
       await clearAll();
+    } else {
+      await clearData();
```
hoangdat, line 121: we should refactor it:
- Create abstract class `MigrateSteps`
- Extend this class when we need, base on the condition. Fx: `MigrateV7`. `MigrateV10`.
- All the needed conditions, please move inside each implemented class.
Then we wil have a clear code here
```
Future<void> onUpgradeCache(int oldVersion, int newVersion) async {
    log('CachingManager::onUpgradeCache():oldVersion $oldVersion | newVersion: $newVersion');
    MigrateV7().execute();
    MigrateV10().execute();
    // ... 
}
```
WDYT? @dab246 
cc @nqhhdev @sherlockvn

lib/features/base/base_controller.dart
```diff
@@ -184,9 +179,21 @@ abstract class BaseController extends GetxController
 
   void handleFailureViewState(Failure failure) async {
     logError('BaseController::handleFailureViewState(): ${failure.runtimeType}');
-    if (failure is LogoutOidcFailure) {
-      if (_isFcmEnabled) {
-        _getStoredFirebaseRegistrationFromCache();
+    if (failure is LogoutOidcFailure ||
```
hoangdat, line 182: all here related to logout, can we separate in a method.

lib/features/push_notification/presentation/controller/fcm_token_controller.dart
dab246: We shouldn't do that, because we won't call it when `FCM` is disabled.

lib/features/caching/caching_manager.dart
dab246: IMO, this separation is quite reasonable. But it should be named differently. Because we will have many types of databases later. So the version will be different.
- abstraction: `UpgradeDatabaseSteps`.
- Hivedatabase: `UpgradeHiveDatabaseStepsV10`

dab246:
@hoangdat  Please review again it

hoangdat (approved):

PR #2485, opened 2024-01-18
TF-2464 Fix plain notification iOS

dab246:
## Issue

#2464 

## Root cause

- Due to the process of updating `Keychain` in NSE in the form of `JSON String`. But when parsing it is `Data`

## Demo

IMG_2607

https://github.com/linagora/tmail-flutter/assets/80730648/835b7119-30b8-4844-aefc-49dead949e83

### DIFF
--- a/ios/TwakeCore/Jmap/JmapClient.swift
+++ b/ios/TwakeCore/Jmap/JmapClient.swift
@@ -83,8 +83,10 @@ class JmapClient {
             interceptor: authenticationInterceptor,
             onSuccess: { (data: JmapResponseObject<Email>) in
                 if let response = data.parsing(methodName: JmapConstants.EMAIL_GET_METHOD_NAME, methodCallId: "c1") {
-                    if let listEmail = response.list, !listEmail.isEmpty {
-                        self.totalListEmails.append(contentsOf: listEmail)
+                    if let listEmail = response.list, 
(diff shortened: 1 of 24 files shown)


ios/TwakeCore/Jmap/JmapClient.swift
```diff
@@ -83,8 +83,17 @@ class JmapClient {
...
-                        self.totalListEmails.append(contentsOf: listEmail)
+                    if let listEmail = response.list, 
+                       !listEmail.isEmpty {
+                        let sortedListEmails = listEmail.sorted(by: { (email1, email2) -> Bool in
+                            if let date1 = email1.receivedAt?.convertUTCDateToLocalDate(),
+                               let date2 = email2.receivedAt?.convertUTCDateToLocalDate() {
+                                return date1 < date2
+                            }
+                            return false
+                        })
```
hoangdat, line 94: separate to a single method

hoangdat (commented):
can you make also a improvement: fist line of notification should be a `sender name`.

ios/TwakeCore/Jmap/JmapClient.swift
dab246: done

hoangdat (approved):

PR #2493, opened 2024-01-18
Fix push notification on Android

dab246:
## Issue

- [x] #2462
- [x] #2463
- [x] Cannot show input login form when get stored token failed
- [x] Cannot auto refresh email list when switch `Background` -> `Foreground` mode
- [x] Build failed on XCode 15.1
- [x] Cannot request notification permission when reload app on Android

## Root causes

- `ISSUE-2462`  - Cannot be reproduced but the problem is due to the problem of state not being synchronized in hive when changing the same in `Background` and `Foreground` mode
- `ISSUE-2463`  - Due to notification permission cannot request again

## Demo

https://github.com/linagora/tmail-flutter/assets/80730648/a74fe04e-df8e-4c02-b57b-709945e05b48

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -66,9 +66,9 @@ PODS:
     - FirebaseCoreInternal (~> 10.0)
     - GoogleUtilities/Environment (~> 7.8)
     - GoogleUtilities/Logger (~> 7.8)
-  - FirebaseCoreInternal (10.5.0):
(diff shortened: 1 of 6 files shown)


hoangdat (commented):
I found some changes here were duplicated with in #2485

dab246:
> I found some changes here were duplicated with in #2485

Yes. We close this PR. Every changed updated at #2485

Issue #2461, opened 2024-01-16
[CUSTOMER] Impossible to refresh an expired JWT claim


chibenwa:
### Description

WHEN using an expired JWT claim
Upon `401` TwakeMail blocks forever

### Expected result

Once again, I expect correct error management:

1. TwakeMail can audit the token, realize it is expired and discard it.
2. TwakeMail when faced with a 401 error should discard current credentials and redirect to login page (when using OIDC, of the OIDC provider)

### Current behavior

https://github.com/linagora/tmail-flutter/assets/6928740/3425d1b2-eb1c-4b94-b1cc-29f0b33c3c27

Apisix log:

```
2024/01/16 14:52:22 [error] 49#49: *65226939 [lua] openid-connect.lua:319: phase_func(): OIDC introspection failed: jwt signature verification failed: 'exp' claim expired at Wed, 10 Jan 2024 23:18:23 GMT, client: 10.233.66.36, server: _, request: "GET /.well-known/jmap HTTP/1.1", host: "jmap.avocat.fr", referrer: "https://e-mail.avocat.fr/"
2024/01/16 14:52:22 [warn] 49#49: *65226939 [lua] plugin.lua:1098: run_plugin(): openid-connect exits with http status code 401, client: 10.233.66.36, server: _, request: "GET /.well-known/jmap HTTP/1.1", host: "jmap.avocat.fr", referrer: "https://e-mail.avocat.fr/"
10.233.66.36 - - [16/Jan/2024:14:52:22 +0000] jmap.avocat.fr "GET /.well-known/jmap HTTP/1.1" 401 653 0.010 "https://e-mail.avocat.fr/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36" - - - "http://jmap.avocat.fr"
```

Cc @guimard

hoangdat:
Can I verify it on preprod? It also has the same configuration?

hoangdat:
Maybe we forgot scope in configuration in `env.file` of front-end 
```
OIDC_SCOPES=openid,profile,email,offline_access
```

chibenwa:
True.

```
https://ci.linagora.com/linagora/lrs/saas/tools/helm-charts/tmail-frontend/-/blob/main/tmail-frontend/values.yaml#L109
```

holds the following value:

```
oidcScopes: openid,profile,email
```

So it should rather be:

```
[...]

PR #2503, opened 2024-01-23
TF-2461 Impossible to refresh an expired JWT claim

dab246:
## Issue

#2461

## Root cause

- Retry on `onError` is blocked by queue interceptor, the current one was not resolved but still try to get the new one and the new one also have the error too.


lib/features/login/data/network/config/time_out_interceptors.dart
```diff
@@ -1,22 +0,0 @@
-import 'package:core/utils/app_logger.dart';
-import 'package:core/utils/platform_info.dart';
-import 'package:dio/dio.dart';
-import 'package:tmail_ui_user/features/login/data/network/endpoint.dart';
-
-class TimeOutInterceptors extends InterceptorsWrapper {
```
hoangdat, line 6: why we need to remove it?

dab246: IMO, it's not necessary

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -0,0 +1,117 @@
...
+            requestOptions: RequestOptions(path: '/jmap')));
+
+        authorizationInterceptors.setTokenAndAuthorityOidc(
+          newToken: OIDCFixtures.tokenOidcExpired,
+          newConfig: OIDCFixtures.oidcConfiguration);
+
+        expect(authorizationInterceptors.validateToRefreshToken(dioError401), true);
+      });
+
+      test('validateToRefreshToken should return false when conditions are not met', () async {
```
hoangdat, line 72: we found 4 conditions in checking refresh token, each condition should have a testcase

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -0,0 +1,117 @@
...
+      tokenOidcCacheManager,
+      accountCacheManager,
+      iosSharingManager);
+
+    dio.interceptors.add(authorizationInterceptors);
+
+    dioAdapter = DioAdapter(dio: dio);
+  });
+
+  group('AuthorizationInterceptor test', () {
```
hoangdat, line 56: How about to add more test-case:
- related to `FileUploader.uploadAttachmentExtraKey`
- make sure we should not be blocked anymore

dab246: - In the case of uploading it depends too much on external objects, it is difficult to write test logic here. We have to separate it into separate functions to make it easier to write unit tests.
- We've got a `refreshingTokensOIDC` test case that only calls once when the token expires.

hoangdat: - ok for the case of `FileUploader.uploadAttachmentExtraKey`
- but, it is `QueueInterceptor`, so I think we must find the way to test the integration of interceptors, make sure it not be blocked in onError anymore

dab246:
We close this PR to reopen new PR (#2508)  to maintain for `v0.11.3`.

Issue #2513, opened 2024-01-26
Mobile/Tablet Web app can not load


hoangdat:
### Description

```
Login TwakeMail on web (Safari/Chrome) of mobile/tablet
```

![293825523-b0502140-c3d4-43cb-8241-170a36227ac0](https://github.com/linagora/tmail-flutter/assets/80730648/de92966a-ef53-4e62-b05a-5f0f1d8f75cf)

![293825495-7492256f-822b-4fde-ba09-2183dcf3f33b](https://github.com/linagora/tmail-flutter/assets/80730648/14ec1289-1d48-482c-a998-1bbabfe7390a)

### Expected result

- Run as Desktop Browser

PR #2521, opened 2024-01-29
TF-2513 Mobile/Tablet web app can not loaded

dab246:
## Issue 

#2513 

## Root cause

Because [Web renderers](https://docs.flutter.dev/platform-integration/web/renderers) are incorrect: 

- When running and building apps for the web. The point was that by default it uses different renderers (`CanvasKit for desktop ` and `HTML for mobile`). And since the app used things that required `CanvasKit`, the application in mobile browsers was displayed incorrectly.

## Solutions

- Use the correct `svg` file
- Use `TextOverflow.ellipsis`

## Demo

- Tablet web

https://github.com/linagora/tmail-flutter/assets/80730648/5b7b4cc6-de0a-4a78-9b71-31154f46e366

- Desktop web

https://github.com/linagora/tmail-flutter/assets/80730648/9f8effeb-6161-4846-868b-24c5531663cb


hoangdat:
How we can detect it depend on the browser user use to open app?

web/index.html
```diff
@@ -59,75 +66,36 @@
...
-      if (scriptLoaded) {
-        return;
-      }
-      scriptLoaded = true;
-      var scriptTag = document.createElement('script');
-      scriptTag.src = 'main.dart.js';
-      scriptTag.type = 'application/javascript';
-      document.body.append(scriptTag);
-    }
+    let useHtml = false;
```
hoangdat, line 69: why we need this, when we dont use `html` anymore?

hoangdat (commented):
Can we find some of the components we must use `Canvaskit`? And then write the ADR. Thanks

web/index.html
dab246: IMO, we should leave a flag like that so we can edit it later. If someone wants to update the `web-renderer` during runtime, they can add `javascript` to handle it here.

dab246:
> Can we find some of the components we must use `Canvaskit`? And then write the ADR. Thanks

@hoangdat IMO, we don't need to force the web build to a specific `web-renderer`. We should leave `auto` to be suitable and optimized for specific platforms. Therefore, we will fix the issue of `widget/font/style` failing on platforms where `canvaskit` renderer is not supported.

core/lib/presentation/resources/image_paths.dart
```diff
@@ -12,7 +12,7 @@ class ImagePaths {
   String get icDownload => _getImagePath('ic_download.svg');
   String get icMore => _getImagePath('ic_more.svg');
   String get icPhotoLibrary => _getImagePath('ic_photo_library.svg');
-  String get icEmptyEmail => _getImagePath('ic_empty_email.svg');
+  String get icEmptyEmail => _getImagePath('ic_empty_email.png');
```
hoangdat, line 15: why we need  to change it to PNG?

core/lib/presentation/resources/image_paths.dart
```diff
@@ -158,7 +158,7 @@ class ImagePaths {
   String get icQuotasWarning => _getImagePath('ic_quotas_warning.svg');
   String get icQuotasOutOfStorage => _getImagePath('ic_quotas_out_of_storage.svg');
   String get icQuickCreatingRule => _getImagePath('ic_quick_creating_rule.svg');
-  String get icPageNotFound => _getImagePath('ic_page_not_found.svg');
+  String get icPageNotFound => _getImagePath('ic_page_not_found.png');
```
hoangdat, line 161: idem

core/lib/presentation/utils/style_utils.dart
```diff
@@ -10,9 +9,6 @@ class CommonTextStyle {
     fontWeight: FontWeight.normal,
   );
 
-  static const defaultTextOverFlow = PlatformInfo.isWeb
-      ? TextOverflow.fade
-      : TextOverflow.ellipsis;
-
-  static const defaultSoftWrap = PlatformInfo.isWeb ? false : true;
+  static const defaultTextOverFlow = TextOverflow.ellipsis;
```
hoangdat, line 12: I dont remember why we should use fade, please carefully

core/lib/presentation/resources/image_paths.dart
hoangdat: PNG increase app too much, can we get support from design team for it?

dab246: SVG image has an error when exported

core/lib/presentation/utils/style_utils.dart
dab246: We use it because it crashes the font `Inter` on the web. But we shouldn't change it at this time, so I flagged checking `PlatformInfo.isWebMobile`

dab246: Avoid `Inter` font error when using `TextOverflow.ellipsis` on earlier versions of flutter 2.10. But now on 3.16 it is working fine.

hoangdat (approved):

Issue #2610, opened 2024-02-17
Mail sent from iphone: missing attachments


chibenwa:
![Screenshot from 2024-02-17 15-07-38](https://github.com/linagora/tmail-flutter/assets/6928740/c54d0fc9-8705-4b86-8c90-b9bc95324886)

![Screenshot from 2024-02-17 15-07-58](https://github.com/linagora/tmail-flutter/assets/6928740/a01b4e2e-a391-4cef-99e4-e5d15cc8b38c)

![Screenshot from 2024-02-17 15-08-47](https://github.com/linagora/tmail-flutter/assets/6928740/bf9898e8-0479-427c-8567-b290fd7ff50c)

QuangHoang1210:
passed on 0.11.3

<img width="1260" alt="Screenshot 2024-02-25 at 11 11 29 PM" src="https://github.com/linagora/tmail-flutter/assets/124866146/4275719a-a0b0-4422-85a9-da7160c6df48">

PR #2617, opened 2024-02-20
TF-2610 Fix missing attachments when mail sent from iphone

dab246:
## Issue

#2610

## Root cause

- The attachment display logic is incorrect. In this case, we don't display when `disposition = inline` and `cid is null`

Screenshot 2024-02-20 at 15 52 54

## Solution

- Change attachment display logic.
- Write `adr` for attachments display logic


dab246:
@chibenwa  @hoangdat Please help me review the `adr` document about `attachments` display logic

chibenwa (approved):

hoangdat (approved):

Issue #2591, opened 2024-02-08
Copy and paste rich text in composer leads to overlap


chibenwa:
### Description

I copy and pasted text from an ODT document.

I noticed the styling makes no line break, leading to text overlap which is not beautiful.

![image](https://github.com/linagora/tmail-flutter/assets/6928740/f9ab2b05-f810-4ee2-b4d2-0fc9c317e507)

### Expected result

I expect no overlap.

### Current behavior

![Screenshot from 2024-02-08 22-20-11](https://github.com/linagora/tmail-flutter/assets/6928740/7300b651-33b0-4e37-8288-57427bef76da)

![Screenshot from 2024-02-08 22-19-59](https://github.com/linagora/tmail-flutter/assets/6928740/951bbd15-910d-4d6c-b629-e193a0171a99)

#### Reproduction Steps

[bug.eml.txt](https://github.com/linagora/tmail-flutter/files/14214995/bug.eml.txt)

### Context

tmail.linagora.com v0.11.3-patch4-v2

chibenwa:
The display of the email generated by tmail in thunderbird is also fucked up

![image](https://github.com/linagora/tmail-flutter/assets/6928740/b20fa28c-a20f-4bcf-afa6-1b2515721a3d)

So do not change HTML viewing

However if I do the same operation in tb the email generated is actually correctly folded

![image](https://github.com/linagora/tmail-flutter/assets/6928740/9b26a34c-0dd0-440e-9df9-1e2c7f3c7b0b)

So please look in the direction of "copy and pasted text formated by other tools"

If need be I'd even agree to unformat text pasted from external stuff.

dab246:
## Propose

- Change style of html text with property `line-height: 1px;` to `line-height: 100%;`

dab246:
Only on Chrome

tddang-linagora:
### Related issues
https://github.com/summernote/summernote/issues/3983
https://github.com/summernote/summernote/issues/3638

### Solution
Both of the above issues suggest changing the line height to 100% should be the fix. I will be using the suggested method in https://github.com/tneotia/html-editor-enhanced/issues/257 to implement it.

PR #2659, opened 2024-02-29
TF-2591 Fix text overlap after pasting rich text

tddang-linagora:
### Issue
#2591 
### Solution
Run a JS script to update text-height for all elements to 100% on content pasted & drag-dropped to the editor
### Demo
- On paste

https://github.com/linagora/tmail-flutter/assets/160106668/e10fc7eb-9d52-4e56-9327-51ae593d5c66

- On drop

https://github.com/linagora/tmail-flutter/assets/160106668/4d91c4df-45c8-4c0b-aba0-c7702ccf5483

### DIFF
--- a/core/lib/utils/html/html_utils.dart
+++ b/core/lib/utils/html/html_utils.dart
@@ -16,6 +16,22 @@ class HtmlUtils {
         });''',
     name: 'lineHeight100Percent');
 
+  static const registerDropListener = (
+    script: '''
+      document.querySelector(".note-editable").addEventListener(
+        "drop",
+        (event) => window.parent.postMessage(
+          JSON.stringify({"name": "registerDropListener"})))''',
+    name: 'registerDropListener');
+
+  static const unregisterDropListener = (
+    script: '''
+      console.log("unregisterDropListener");
+      const editor = document.querySelector(".note-editable");
+      const newEditor = editor.cloneNode(true);
+      editor.parentNode.replaceChild(newEditor, editor);''',
+    name: 'unregisterDropListener');
+
   static String customCssStyleHtmlEditor({TextDirection direction = TextDirection.ltr}) {
     if (PlatformInfo.isWeb) {
       return '''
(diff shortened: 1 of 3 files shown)


core/lib/presentation/utils/html_transformer/html_utils.dart
```diff
@@ -97,6 +97,12 @@ class HtmlUtils {
     </script>
   ''';
 
+  static const lineHeight100Percent = (
+    script: '''
+      document.querySelectorAll("*")
+        .forEach((element) => element.style.lineHeight = "100%");''',
+    name: 'lineHeight100Percent');
```
dab246, line 104: 1. We should only query tags with style `lineHeight`.
2. It will affect other tags when it wants to use `lineHeight` of a certain value. So you should only change it for the case `copy & paste` or `drag & drop` text

dab246:
Please also demo on other browser (firefox, safari)

core/lib/presentation/utils/html_transformer/html_utils.dart
```diff
@@ -97,6 +97,31 @@ class HtmlUtils {
     </script>
   ''';
 
+  static const lineHeight100Percent = (
+    script: '''
+      document.querySelectorAll("*")
+        .forEach((element) => {
+          if (element.style.lineHeight !== "normal")
+            element.style.lineHeight = "100%";
```
dab246, line 105: IMO, we will be affected when some tags have a specific `lineHeight` set (like set via rich text button).
### Propose

Way 1: Keep the content of the text intact when `copy & paste` or `drag & drop` into the editor.
Way 2: Only change `lineHeight` for content to be `dragged` or `copied` in the editor

tddang-linagora: Will implement option 2

tddang-linagora: I can't seem to find a way to get the `dragged` or `copied` content. Suggesting to keep it this way & fix bug later if it ever occurs

dab246 (approved):
lgtm

PR #2751, opened 2024-03-28
Fix mobile app force logout when notification enable in OIDC

dab246:
## Issue

- App more than 2 minutes (expired time)
- When enable notification + OIDC, app sometimes have forced logout

## Root cause

- Because multiple requests are called at the same time. Because of using `QueueInterceptor`, requests are completed according to the FIFO principle. But because they are all called at the same time, the data in the request is still the old data (AuthorizationHeader). Therefore, when the first request gets the new token and uses it, the following requests will not be able to use this new token, leading to errors.

## Solution

Perform verification for requests by checking:
- AuthorizationHeader
- Token Expired

### DIFF
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -49,11 +49,11 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
     _token = newToken;
     _configOIDC = newConfig;
     _authenticationType = AuthenticationType.oidc;
-    log('AuthorizationInterceptors::setTokenAndAuthorityOidc: TOKEN_INITIAL = $newToken');
+    log('AuthorizationInterceptors::setTokenAndAuthorityOidc: INITIAL_TOKEN = ${newToken?.token} | EXPIRED_TIME = ${newToken?.expiredTime}');
   }
 
   void _updateNewToken(TokenOIDC newToken) {
-    log('AuthorizationInterceptors::_updateNewToken: NEW_TOKEN = $newToken');
+    log('AuthorizationInterceptors::_updateNewToken: NEW_TOKEN = ${newToken.token} | EXPIRED_TIME = ${newToken.expiredTime}');
     _token = newToken;
   }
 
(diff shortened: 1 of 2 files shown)


tddang-linagora (approved):

chibenwa:
What is the user impact of this work?

hoangdat:
> What is the user impact of this work?

Mobile user need to go to `login screen`.

- Right now, still not get any feedback on it, dev team found this

chibenwa:
Because on CNB users complains that after some time, need to reinstall the application.
I strongly suspect a refresh token handling issue on mobile...

dab246:
> Because on CNB users complains that after some time, need to reinstall the application. I strongly suspect a refresh token handling issue on mobile...

`reinstall the application` ? I think it should be login again, right? If so, IMO, it's related to the issue being fixed in this PR

hoangdat:
> I was not able to reproduce that myself but apparently they were not redirected back to the OIDC provider...

you test on linagora or customer platform?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -49,11 +49,11 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
     _token = newToken;
     _configOIDC = newConfig;
     _authenticationType = AuthenticationType.oidc;
-    log('AuthorizationInterceptors::setTokenAndAuthorityOidc: TOKEN_INITIAL = $newToken');
+    log('AuthorizationInterceptors::setTokenAndAuthorityOidc: INITIAL_TOKEN = ${newToken?.token} | EXPIRED_TIME = ${newToken?.expiredTime}');
   }
 
   void _updateNewToken(TokenOIDC newToken) {
-    log('AuthorizationInterceptors::_updateNewToken: NEW_TOKEN = $newToken');
+    log('AuthorizationInterceptors::_updateNewToken: NEW_TOKEN = ${newToken.token} | EXPIRED_TIME = ${newToken.expiredTime}');
```
guimard, line 56: Why line 52 contains `${newToken?.expiredTime}` _(with `?`)_ and this line contains `${newToken.expiredTime}`. Error risk ?

guimard:
Does this affects CNB too ?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
hoangdat: `newToken` is non-null parameter. We pre-check it in line 101-105, moreover, we will throw exception in case we can not execute `refreshToken` success with idP (line 97-99)

chibenwa:
Well this might be the bug i encoutered friday...

dab246 (commented):
LGTM

Issue #2758, opened 2024-03-31
Blue bar content not support HTML


hoangdat:
### Description
Blue bar content not support HTML

<img width="946" alt="Screenshot 2024-04-01 at 05 55 41" src="https://github.com/linagora/tmail-flutter/assets/6462404/3cc36499-58e1-4fbf-b71e-bd6bfc78808c">

PR #2762, opened 2024-04-01
TF-2758 Email event description in html

tddang-linagora:
### Issue
- #2758 

### Before
Screenshot 2024-04-01 at 16 12 43
Screenshot 2024-04-01 at 16 13 01
### After
Screenshot 2024-04-01 at 16 13 21
Screenshot 2024-04-12 at 11 25 40

### DIFF
--- a/lib/features/email/presentation/widgets/calendar_event/calendar_event_detail_widget.dart
+++ b/lib/features/email/presentation/widgets/calendar_event/calendar_event_detail_widget.dart
@@ -8,7 +8,6 @@ import 'package:tmail_ui_user/features/email/presentation/styles/calendar_event_
 import 'package:tmail_ui_user/features/email/presentation/widgets/calendar_event/calendar_event_action_button_widget.dart';
 import 'package:tmail_ui_user/features/email/presentation/widgets/calendar_event/event_attendee_detail_widget.dart';
 import 'package:tmail_ui_user/features/email/presentation/widgets/calendar_event/event_body_content_widget.dart';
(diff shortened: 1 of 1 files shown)


chibenwa:
image

That's HTML in the event.

image

And that's HTML in the JMAP call too.

Please directly render HTML (after sanitizing) rather than passing through a plain text escaped representation.

chibenwa (changes requested):
CF earlier comment

core/lib/presentation/extensions/string_extension.dart
```diff
@@ -46,4 +46,12 @@ extension StringExtension on String {
...
       .toString();
   }
+
+
+  String get decodeHtml {
+    return replaceAll('&lt;', '<')
+      .replaceAll('&gt;', '>')
+      .replaceAll('&quot;', '"')
+      .replaceAll('&amp;', '&');
+  }
```
hoangdat, line 56: did we have any HTML transformer?

hoangdat, line 56: did we have any HTML transformer?

tddang-linagora:
> image
> 
> That's HTML in the event.
> 
> image
> 
> And that's HTML in the JMAP call too.
> 
> Please directly render HTML (after sanitizing) rather than passing through a plain text escaped representation.

Done

core/lib/presentation/extensions/string_extension.dart
hoangdat: how about using `TransformConfiguration` to handle html? WDYT @dab246

dab246 (approved):
lgtm

hoangdat (approved):

Issue #2533, opened 2024-01-31
Memory footprint


chibenwa:
Chromium summarizes the footprint of a web page

![Screenshot from 2024-01-31 22-14-31](https://github.com/linagora/tmail-flutter/assets/6928740/16ce106d-c8fb-40b9-a127-515cc437ad83)

![Screenshot from 2024-01-31 22-15-53](https://github.com/linagora/tmail-flutter/assets/6928740/02f4ece3-c383-4fd8-91a3-a407939312dd)

![Screenshot from 2024-01-31 22-16-13](https://github.com/linagora/tmail-flutter/assets/6928740/9a4e65d6-2606-460c-a8ce-919f8af5f384)

It could be interesting to get a review of where that memory is used and see if we get some easy quick wins ;-)

hoangdat:
- [ ] compare with other Flutter web app too
- [ ] investigating tool for memory consumption in browser

hoangdat:
https://github.com/linagora/tmail-flutter/assets/6462404/4db4f77d-28a6-475b-a1f5-7b2fac02b3cf

tddang-linagora:
### Relevant issues:
- https://github.com/flutter/flutter/issues/96869

hoangdat:
hi @tddang-linagora , please take a look also for https://github.com/dart-lang/leak_tracker/blob/main/doc/leak_tracking/OVERVIEW.md

tddang-linagora:
> hi @tddang-linagora , please take a look also for https://github.com/dart-lang/leak_tracker/blob/main/doc/leak_tracking/OVERVIEW.md

Greate package. However, it doesn't support Web platform.

chibenwa:
Okish now

PR #2783, opened 2024-04-05
TF-2533 [PART-3] Composer memory leak

tddang-linagora:
### Issue
- #2533 

### Before

https://github.com/linagora/tmail-flutter/assets/160106668/8d79f169-b592-4965-8533-99224e91cee8

### After

https://github.com/linagora/tmail-flutter/assets/160106668/e1ee69ce-b62f-4294-9fa8-53034e9500c8

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -33,6 +33,7 @@ import 'package:receive_sharing_intent/receive_sharing_intent.dart';
 import 'package:rich_text_composer/rich_text_composer.dart';
 import 'package:super_tag_editor/tag_editor.dart';
 import 'package:tmail_ui_user/features/base/base_controller.dart';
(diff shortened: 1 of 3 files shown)


chibenwa (approved):

hoangdat (approved):

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -216,6 +217,8 @@ class ComposerController extends BaseController {
   @override
   void onClose() {
     _initTextEditor = null;
+    _textEditorWeb = null;
+    viewState.value = Right(UIClosedState());
```
dab246, line 221: Please test additional cases of `reply/forward` with long email, large inline image. If I remember correctly we have some similar variables to store the email content. Please clean it up when composer closes.

tddang-linagora: Will demo on part 2 due to the open-email-memory-leak is not resolved in this PR

PR #2789, opened 2024-04-09
TF-2533 Create ADR to put memory footprint task on hold

tddang-linagora:
### Related issue
- #2533

### DIFF
--- a/docs/adr/0044-memory-leak-at-current-state.md
+++ b/docs/adr/0044-memory-leak-at-current-state.md
@@ -1,0 +1,34 @@
+# 44. Memory leak at current state
+
+Date: 2024-04-09
+
+## Status
+
+- Issues: 
+  - [Memory footprint #2533](https://github.com/linagora/tmail-flutter/issues/2533)
+- PRs:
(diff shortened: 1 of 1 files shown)


docs/adr/0044-memory-leak-at-current-state.md
```diff
@@ -0,0 +1,34 @@
...
+  - Email content view:
+    - Web: Resolved
+    - Mobile: Remained until [InAppWebView memory leak on Android #1023](https://github.com/pichillilorenzo/flutter_inappwebview/issues/1023) is fixed
+  - Identity create/edit:
+    - Web: Resolved
+    - Mobile: No problem detected, nothing changed.
+
+## Decision
+
+- This work will be on hold until further notice
```
chibenwa, line 30: On which conditions would work on this be resumed?

tddang-linagora: On @hoangdat command

chibenwa: I suggest: if significant memory leaks are reportedeg consuming twice the regular amount of memory within one day usage.

tddang-linagora: @chibenwa Done

hoangdat (approved):

Issue #2769, opened 2024-04-03
[IOS/WEB PC] Weird error msg when user upload too many attachment which almost reach the maximum range


QuangHoang1210:
**ENV:** 
- IOS app
- Tmail v0.11.3411

**Precondition:**
- User already upload many files which mostly reach maximum 20mb

**Step:**
1- User open composer
2- User upload many attachment
3- User continue pick attachment which included one file more than maximum mb accepted

**Actual:**
- Attachment is uploading but a weird red toast displaying

**Expected:**
- No red toast displaying

![Image](https://github.com/linagora/tmail-flutter/assets/124866146/0b8fc6fb-86d7-4436-a302-f5ea27f7fb34)

https://github.com/linagora/tmail-flutter/assets/124866146/d33b389a-3e48-4fd5-b5cc-72eec2f02237

PR #2792, opened 2024-04-10
TF-2769 Fix isolate illegal data

tddang-linagora:
### Issue
- #2769 

### Related documents
- https://api.flutter.dev/flutter/dart-isolate/SendPort/send.html
- https://github.com/machinescream/worker_manager/issues/86

### Root cause
We are using `lib/features/upload/data/network/file_uploader.dart:FileUploader:uploadAttachment:_handleUploadAttachmentAction` function to handle upload file in mobile, with the help of `Dio`, inside a separate isolate with app's main isolate. When this `Dio` throw `DioError` exception, by any reason, this exception in turns get sent from this separate isolate to the main isolate. Inside this `DioError`, the `RequestOptions:data` is sent back out of this separate isolate, with type `Stream`, exactly the type we passed into this isolate by `File(argsUpload.mobileFileUpload.filePath).openRead()`. This type is not supported by `send()` function of `Isolate` (see above document), so this isolate throw its own exception, telling us that the type we're letting that isolate call `send()` is "illegal". Also, we're showing the whole stacktrace of the error if this case happens.

**Log**
```
flutter: *** Request ***
flutter: uri: https://jmap.linagora.com/upload/8203ca9793c9c26875c5c1463ad74ec448f7bc416439afce48fc1210e9f1b77a
flutter: method: POST
flutter: responseType: ResponseType.json
flutter: followRedirects: true
flutter: persistentConnection: true
flutter: connectTimeout: null
flutter: sendTimeout: null
flutter: receiveTimeout: null
flutter: receiveDataWhenStatusError: true
flutter: extra: {upload-attachment: {platform: mobile, path: /private/var/mobile/Containers/Data/Application/84C6E2FF-DF6E-42CF-BC42-95435D8C9421/tmp/com.linagora.ios.teammail-Inbox/IMG_1632.PNG, type: image/png, size: 1050986}}

flutter: *** DioError ***:
flutter: uri: https://jmap.linagora.com/upload/8203ca9793c9c26875c5c1463ad74ec448f7bc416439afce48fc1210e9f1b77a
flutter: DioError [bad response]: The request returned an invalid status code of 502.
flutter: uri: https://jmap.linagora.com/upload/8203ca9793c9c26875c5c1463ad74ec448f7bc416439afce48fc1210e9f1b77a
flutter: statusCode: 502
flutter: headers:
flutter:  connection: keep-alive
flutter:  content-type: text/html; charset=utf-8
flutter:  date: Fri, 12 Apr 2024 08:49:28 GMT
flutter:  content-length: 229
flutter:  server: nginx
flutter:  x-apisix-upstream-status: 502
```

### Scenario
The error showing in the demos below is on purpose, to show how the error is handled. In real cases, the attachments will be uploaded as normal.

**Before**

https://github.com/linagora/tmail-flutter/assets/160106668/387c4a21-4f2e-4432-b859-198cadc3af55

**After**

https://github.com/linagora/tmail-flutter/assets/160106668/b3b7ea6a-46c3-4d09-b8e2-91a6ef8d77d4

### DIFF
--- a/lib/features/upload/data/network/file_uploader.dart
+++ b/lib/features/upload/data/network/file_uploader.dart
@@ -81,43 +81,54 @@ class FileUploader {
       UploadFileArguments argsUpload,
       worker.TypeSendPort sendPort
   ) async {
-    final rootIsolateToken = argsUpload.isolateToken;
-    BackgroundIsolateBinaryMessenger.ensureInitialized(rootIsolateToken);
-    await HiveCacheConfig.instance.setUp();
+    try {
+      final rootIsolateToken = argsUpload.isolateToken;
+      BackgroundIsolateBinaryMessenger.ensureInitialized(rootIsolateToken);
+      await HiveCacheConfig.instance.setUp();
 
-    final headerParam = argsUpload.dioClient.getHeaders();
-    headerParam[HttpHeaders.contentTypeHeader] = argsUpload.mobileFileUpload.mimeType;
-    headerParam[HttpHeaders.contentLengthHeader] = argsUpload.mobileFileUpload.fileSize;
+      final headerParam = argsUpload.dioClient.getHeaders();
+      headerParam[HttpHeaders.contentTypeHeader] = argsUpload.mobileFileUpload.mimeType;
+      headerParam[HttpHeaders.contentLengthHeader] = argsUpload.mobileFileUpload.fileSize;
 
-    final mapExtra = <String, dynamic>{
-      uploadAttachmentExtraKey: {
-        filePathExtraKey: argsUpload.mobileFileUpload.filePath,
-      }
-    };
+      final mapExtra = <String, dynamic>{
+        uploadAttachmentExtraKey: {
+          filePathExtraKey: argsUpload.mobileFileUpload.filePath,
+        }
+      };
 
-    final resultJson = await argsUpload.dioClient.post(
-      Uri.decodeFull(argsUpload.uploadUri.toString()),
-      options: Options(
-        headers: headerParam,
-        extra: mapExtra
-      ),
-      data: File(argsUpload.mobileFileUpload.filePath).openRead(),
-      onSendProgress: (count, total) {
-        log('FileUploader::_handleUploadAttachmentAction():onSendProgress: FILE[${argsUpload.uploadId.id}] : { PROGRESS = $count | TOTAL = $total}');
-        sendPort.send(
-          UploadingAttachmentUploadState(
-            argsUpload.uploadId,
-            count,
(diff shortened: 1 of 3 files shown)


chibenwa:
Cool however can you explain why we get the error in the first place?
Network issue?

tddang-linagora:
> Cool however can you explain why we get the error in the first place? Network issue?

I would need to reproduce this bug to understand what went wrong. Will do.

dab246:
IMO, This is not the way to solve this problem. We need to find out why it failed to upload.

chibenwa:
> IMO, This is not the way to solve this problem. We need to find out why it failed to upload.

WRONG

We need to fix BOTH

chibenwa:
(sure I agree with you @dab246 that no error in the first place is non -gegociable however correct error management IS a must.)

dab246:
> > IMO, This is not the way to solve this problem. We need to find out why it failed to upload.
> 
> WRONG
> 
> We need to fix BOTH



That's natural. Maybe you don't understand what I mean. WE NEED TO FIND THE CAUSE THAT CAUSED IT TO FAILED. As for the reason for displaying such a long error content in `toast message`, the above solution has been resolved. BUT EXPECTED IS NOT LIKE THAT.

chibenwa:
Sure I 100% that's why I commented in the first place: more investigations on the root cause are also needed.

tddang-linagora:
> > IMO, This is not the way to solve this problem. We need to find out why it failed to upload.
> 
> WRONG
> 
> We need to fix BOTH

@chibenwa Added log. This might not be able to resolve from the app side if this is what happened to our tester.

lib/features/upload/domain/state/attachment_upload_state.dart
```diff
@@ -54,10 +54,7 @@ class SuccessAttachmentUploadState extends Success {
 class ErrorAttachmentUploadState extends FeatureFailure {
   final UploadTaskId uploadId;
 
-  ErrorAttachmentUploadState({
-    required this.uploadId,
-    dynamic exception
-  }) : super(exception: exception);
+  ErrorAttachmentUploadState({required this.uploadId});
```
dab246, line 57: It should not be deleted. We may need to show toast for each exception

tddang-linagora: Done

dab246 (approved):
- [ ] Squash & rebase

dab246:
Also fixed at #3129

dab246 (approved):

hoangdat (approved):

PR #2816, opened 2024-04-19
TF-2810 Handle crashed when no browser available for OIDC

hoangdat:
### Root cause: 

https://github.com/MaikuB/flutter_appauth/pull/465

No browser available for OIDC

### Solution: 
- parse native exception to business exception `NoSuitableBrowserForOIDCException`
- show error to user

#### Case1: in dnslookup form 
[case_dnslookup.webm](https://github.com/linagora/tmail-flutter/assets/6462404/1e72921a-0941-4009-ab66-0b74caa66fff)

#### Case 2: in baseUrl form
[case_baseUrl.webm](https://github.com/linagora/tmail-flutter/assets/6462404/2859f21c-9a8a-4ffa-9ce7-57a6366870f5)

### DIFF
--- a/lib/features/login/domain/exceptions/login_exception.dart
+++ b/lib/features/login/domain/exceptions/login_exception.dart
@@ -1,4 +1,21 @@
 
+import 'package:flutter/services.dart';
+
 class NotFoundDataResourceRecordException implements Exception {}
 
 class NotFoundUrlException implements Exception {}
+
+class NoSuitableBrowserForOIDCException implements Exception {
+
+  static const noBrowserAvailableCode = 'no_browser_available';
+
+  static bool verifyException(dynamic exception) {
+    if (exception is PlatformException) {
+      if (exception.code == noBrowserAvailableCode) {
+        return true;
+      }
+    }
+    return false;
+  }
+}
+
--- a/lib/features/login/domain/usecases/get_token_oidc_interactor.dart
+++ b/lib/features/login/domain/usecases/get_token_oidc_interactor.dart
@@ -3,10 +3,12 @@ import 'package:core/presentation/state/failure.dart';
 import 'package:core/presentation/state/success.dart';
 import 'package:core/utils/app_logger.dart';
 import 'package:dartz/dartz.dart';
+import 'package:flutter/services.dart';
 import 'package:model/account/authentication_type.dart';
 import 'package:model/account/personal_account.dart';
 import 'package:model/oidc/oidc_configuration.dart';
 import 'package:model/oidc/token_oidc.dart';
+import 'package:tmail_ui_user/features/login/domain/exceptions/login_exception.dart';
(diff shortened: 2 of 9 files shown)


lib/features/login/domain/exceptions/login_exception.dart
```diff
@@ -1,4 +1,21 @@
...
+  static const noBrowserAvailableCode = 'no_browser_available';
+
+  static bool verifyNoSuitableBrowserOIDC(dynamic exception) {
+    if (exception is PlatformException) {
+      if (exception.code == noBrowserAvailableCode) {
+        return true;
+      }
+    }
+    return false;
+  }
```
dab246, line 19: I don't like writing like that. Each exception class should only serve a single error. We shouldn't deal with logic here.

hoangdat: quite not get your point, it is exception from platform, we should convert it back to business exception

lib/features/login/domain/usecases/get_token_oidc_interactor.dart
```diff
@@ -46,6 +47,10 @@ class GetTokenOIDCInteractor {
       yield Right<Failure, Success>(GetTokenOIDCSuccess(tokenOIDC, config));
     } catch (e) {
       logError('GetTokenOIDCInteractor::execute(): $e');
+      if (NoSuitableBrowserForOIDCException.verifyNoSuitableBrowserOIDC(e)) {
+        yield Left<Failure, Success>(GetTokenOIDCFailure(NoSuitableBrowserForOIDCException()));
+        return;
+      }
       yield Left<Failure, Success>(GetTokenOIDCFailure(e));
     }
```
tddang-linagora, line 55:
```suggestion
    } on PlatformException catch (e) {
      logError('GetTokenOIDCInteractor::execute():PlatformException: $e');
      if (exception.code == noBrowserAvailableCode) {
        yield Left<Failure, Success>(GetTokenOIDCFailure(NoSuitableBrowserForOIDCException()));
      } else {
        yield Left<Failure, Success>(GetTokenOIDCFailure(e));
      }
    } catch (e) {
      logError('GetTokenOIDCInteractor::execute(): $e');
      yield Left<Failure, Success>(GetTokenOIDCFailure(e));
    }
```

tddang-linagora: IMO exception should not contain business logic

hoangdat: no, it is not a business logic, it is a mapping from external exception to business exception

dab246 (approved):

tddang-linagora (approved):

dab246:
- [ ] Use `catches with on clauses` 
```
try {
 somethingRisky()
}
on Exception catch(e) {
  doSomething(e);
}
```

PR #2851, opened 2024-05-06
Fix notification flood on android

dab246:
## Issues

- [x] Sounds and vibrates multiple times when there are multiple notifications at the same time
- [x] Notification unstable when terminated
- [x] ADR for noti `silent`
- [ ] Unit test notification??

## Solution

- Set 
  - `silent=true` for all notifications with id `EmailId` 
  - `silent=false` for notifications with id `GroupdId`.

## Demo

1. Notifications when  `disconnected` & `reconnected`:

https://github.com/linagora/tmail-flutter/assets/80730648/c44195cf-753d-496b-94b8-fd9ab0c45953

2. Notification unstable when terminated:

- Before: 

https://github.com/linagora/tmail-flutter/assets/80730648/d759fd90-2013-4632-b51d-aa149a6d30b2

- After: 

https://github.com/linagora/tmail-flutter/assets/80730648/a35fad3b-ee5f-43c3-b8e0-d3bd51d1c74a

### DIFF
--- a/lib/features/push_notification/presentation/controller/fcm_message_controller.dart
+++ b/lib/features/push_notification/presentation/controller/fcm_message_controller.dart
@@ -13,6 +13,7 @@ import 'package:jmap_dart_client/jmap/core/session/session.dart';
 import 'package:jmap_dart_client/jmap/core/state.dart' as jmap;
 import 'package:jmap_dart_client/jmap/core/user_name.dart';
 import 'package:jmap_dart_client/jmap/push/state_change.dart';
+import 'package:model/model.dart';
 import 'package:rxdart/rxdart.dart';
 import 'package:tmail_ui_user/features/base/action/ui_action.dart';
 import 'package:tmail_ui_user/features/caching/config/hive_cache_config.dart';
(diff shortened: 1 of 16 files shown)


tddang-linagora:
Is it possible to demo the fix?

dab246:
### Note

- [ ] Delete `group_id` notification after logout

lib/features/push_notification/presentation/controller/fcm_message_controller.dart
```diff
@@ -227,76 +213,98 @@ class FcmMessageController extends FcmBaseController {
...
   void handleFailureViewState(Failure failure) {
     log('FcmMessageController::_handleFailureViewState(): $failure');
-    _clearPayloadData();
   }
 
   @override
   void handleSuccessViewState(Success success) {
     log('FcmMessageController::_handleSuccessViewState(): $success');
-    if (success is GetAuthenticatedAccountSuccess) {
-      _handleGetAuthenticatedAccountSuccess(success);
```
hoangdat, line 298: why don't need to `getAuthenticatedAccount`?

dab246: We just need to get its value to process on the interactor side so we don't need to check it on the controller side anymore.

hoangdat (approved):
- work well
- no side effect in web

tddang-linagora (approved):

PR #2852, opened 2024-05-06
TW-2835 Add right exception thrower for database operation of AuthenticationOIDCDatasource

hoangdat:
### Root cause:
- use `remoteExceptionThrower` for cache operation

### Demo:
Screenshot 2024-05-06 at 17 52 04


tddang-linagora (approved):

dab246:
Please also check other datasources to see if `ExceptionThrow` is used correctly?

hoangdat:
> Please also check other datasources to see if `ExceptionThrow` is used correctly?

No more datasource have wrong thrower

Issue #2878, opened 2024-05-21
Handle notification toggle in Flutter for Android


tddang-linagora:
### Description
Handle toggle push notification in Flutter side for Android, with the UI created in #2877. The only different between push notification on and off state is no push notification will be shown. Everything else up until that point is kept as is.

### Definitions of done
- [ ] Check the status of app's push notification setting state before calling local push notification

hoangdat:
please define DoD

tddang-linagora:
Close due to changes in implementation of notification toggle. There will be no need for platform specific implement anymore.

PR #2888, opened 2024-05-22
[Part-2] TF-2878 toggle android push notification

tddang-linagora:
### Issue
- #2878


lib/features/push_notification/presentation/listener/email_change_listener.dart
```diff
@@ -263,6 +267,25 @@ class EmailChangeListener extends ChangeListener {
     required List<PresentationEmail> emailList
   }) async {
     log('EmailChangeListener::_handleLocalPushNotification(): EMAIL_LENGTH = ${emailList.length}');
+
+    if (PlatformInfo.isAndroid) {
+      final notificationSettingEnabled = (await _getAppNotificationSettingCacheInteractor?.execute().last)
```
dab246, line 272: We should check it once in the constructor, instead of multiple times every time we show a push notification.

tddang-linagora: What if the permission is changed after the class is initiated? `EmailChangeListener` is a Singleton.

dab246: We still have to request it but not in the push notification handler function. We should handle it in inject dependencies constructors.

tddang-linagora: Please clarify

tddang-linagora: Will update

tddang-linagora:
Close due to changes in implementation of notification toggle. There will be no need for platform specific implement anymore.

PR #2942, opened 2024-06-14
[TEST] Support store log to file to tracking on mobile

dab246:
## Purpose

- To easily monitor `log` on mobile devices in `release/profile` mode

## Demo

Screenshot 2024-06-14 at 16 58 31

### DIFF
--- a/core/lib/utils/log_tracking.dart
+++ b/core/lib/utils/log_tracking.dart
@@ -1,0 +1,200 @@
+import 'dart:async';
+import 'dart:collection';
+import 'dart:io';
+
+import 'package:core/domain/exceptions/download_file_exception.dart';
+import 'package:core/utils/app_logger.dart';
+import 'package:core/utils/file_utils.dart';
+import 'package:core/utils/platform_info.dart';
+import 'package:dartz/dartz.dart';
+import 'package:equatable/equatable.dart';
+import 'package:flutter/foundation.dart';
+import 'package:intl/intl.dart';
+import 'package:path_provider/path_provider.dart';
+
+class LogTracking {
+  static const String logFolder = 'TraceLogs';
+  static const String logFileNameDatePattern = 'yyyy-MM-dd';
+  static const String logMessageDatePattern = 'yyyy-MM-dd, HH:mm:ss';
+
+  LogTracking._();
+
+  factory LogTracking() => _instance ??= LogTracking._();
+
+  static LogTracking? _instance;
+
+  bool enableTraceLog = PlatformInfo.isMobile;
+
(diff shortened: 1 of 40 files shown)


core/lib/utils/log_tracking.dart
```diff
@@ -0,0 +1,192 @@
...
+
+  Future _executeTraceLog() async {
+    while (true) {
+      if (_messagesQueue.isEmpty) {
+        _isScheduled = false;
+        return;
+      }
+
+      final message = _messagesQueue.removeFirst();
+      await saveLog(message: message);
```
hoangdat, line 48: can we need try-catch here?

core/test/utils/log_tracking_test.dart
```diff
@@ -0,0 +1,86 @@
...
+        folderPath: LogTracking.logFolder);
+
+      final listMessage= content
+        .split('\n')
+        .map((message) => message.trim())
+        .where((message) => message.isNotEmpty);
+
+      expect(listMessage.length, equals(9));
+      expect(listMessage.first, contains('hello 1'));
+      expect(listMessage.last, contains('hello 8'));
```
hoangdat, line 53: should add more `expect`
```
      expect(listMessage, containsAllInOrder(list));
```

core/lib/utils/log_tracking.dart
dab246: IMO, We don't need to, because the possibility of errors here is almost non-existent. If there is only in the `saveLog` function, we have `try/catch` it.

core/test/utils/log_tracking_test.dart
dab246: > should add more `expect`
> 
> ```
>       expect(listMessage, containsAllInOrder(list));
> ```

Because of the message we print it with the time. so we need to fix the function so it takes quite a while. On the other hand, you will see a file automatically generated when running test `core/applicationDocumentsPath/TraceLog`

core/lib/utils/log_tracking.dart
dab246: try/catch added

core/lib/utils/log_tracking.dart
```diff
@@ -38,15 +38,17 @@ class LogTracking {
   }
 
   Future _executeTraceLog() async {
-    while (true) {
-      if (_messagesQueue.isEmpty) {
-        _isScheduled = false;
-        return;
+    try {
```
hoangdat, line 41: why not `try-catch` inside `while`, we still good in log

dab246: done

dab246:
Duplicated with #3022

PR #2956, opened 2024-06-21
Fix duplicated selected AccountCache

hoangdat:


tddang-linagora (approved):

lib/features/login/data/local/account_cache_manager.dart
```diff
@@ -35,8 +35,8 @@ class AccountCacheManager {
         .removeDuplicated()
         .whereNot((account) => account.accountId == newAccountCache.accountId)
         .toList();
+      await _accountCacheClient.clearAllData();
```
dab246, line 38: why do that?

hoangdat: In case of newAllAccount empty, you not update the cache.

dab246: So what effect does it have on the current cache, if we follow the old flow? Please describe more clearly the root cause of this problem.

dab246 (approved):
LGTM

Issue #2926, opened 2024-06-13
Memory leak in composer controller


hoangdat:
### DoD:
- [x] detect also in mobile/web
- [x] why `dispose` of `composer_controller` is not invoked?
- [x] verify the root cause relate to text input
- [x] `dashboardViewStateWorker` is not init in `composer_controller`

tddang-linagora:
>why dispose of composer_controller is not invoked?

According to 
- https://github.com/jonataslaw/getx/issues/3051#issuecomment-1987601565
>Dispose does nothing.

dab246:
> > why dispose of composer_controller is not invoked?
> 
> According to
> 
> * [GetxController dispose is not working jonataslaw/getx#3051 (comment)](https://github.com/jonataslaw/getx/issues/3051#issuecomment-1987601565)
> 
> > Dispose does nothing.

So, we should create PR to move actions in `dispose` of GetxController to `onClose`

tddang-linagora:
>verify the root cause relate to text input

**Similar issue**
- https://github.com/flutter/flutter/issues/148291

**Fixes**

- https://github.com/flutter/flutter/pull/150661

tddang-linagora:
**Before**

https://github.com/linagora/tmail-flutter/assets/160106668/53e37f48-a6d3-4088-9809-c9b769e9e252

**After**

https://github.com/linagora/tmail-flutter/assets/160106668/2d71cc84-01e7-43e0-81c3-4893f8a1491a

dab246:
> > verify the root cause relate to text input
> 
> **Similar issue**
> 
> * [UndoManager does not release UndoManagerClient causing memory leaks flutter/flutter#148291](https://github.com/flutter/flutter/issues/148291)
> 
> **Fixes**
> 
> * [Fix: Memory leak in UndoHistory widget because it never de-registered itself as global UndoManager client (Resolves #148291) flutter/flutter#150661](https://github.com/flutter/flutter/pull/150661)

## Root cause

- `TextField` widget use `EditableText` widget. `EditableText` widget use `UndoHistory` widget. UndoHistory leak UndoManagerClient

chibenwa:
Should be handled in Definition of done of release. No release without a quick memory leak audit ;-)

PR #2957, opened 2024-06-24
Trigger dispose before onClose in BaseController

tddang-linagora:
**Original issue**
- #2926 

**Why does this PR exist?** 
- https://github.com/jonataslaw/getx/issues/3051#issuecomment-1987601565

### DIFF
--- a/lib/features/base/base_controller.dart
+++ b/lib/features/base/base_controller.dart
@@ -375,4 +375,10 @@ abstract class BaseController extends GetxController
     authorizationInterceptors.clear();
     await cachingManager.closeHive();
   }
+
+  @override
+  void onClose() {
+    super.onClose();
+    dispose();
+  }
 }
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -182,7 +182,6 @@ class ComposerController extends BaseController with DragDropFileMixin {
   ButtonState _sendButtonState = ButtonState.enabled;
 
   late Worker uploadInlineImageWorker;
-  late Worker dashboardViewStateWorker;
   late bool _isEmailBodyLoaded;
 
   ComposerController(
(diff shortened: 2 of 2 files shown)


lib/features/base/base_controller.dart
```diff
@@ -375,4 +375,10 @@ abstract class BaseController extends GetxController
     authorizationInterceptors.clear();
     await cachingManager.closeHive();
   }
+
+  @override
+  void onClose() {
+    super.onClose();
+    dispose();
```
dab246, line 382: Please test again on web

tddang-linagora: The only place we have concern about `dispose()` callback is `ComposerController`, so it's fine

dab246: Many controllers inherit from it. Are you sure other controllers are not affected?. Please test the features on the website: DestinationPicker, Signature, Forward Email, Vacation, etc.

hoangdat:
hi @tddang-linagora any update on it

lib/features/base/base_controller.dart
tddang-linagora: We only care about the one overriding `dispose`. If originally `dispose` "does nothing" (as its author said), why do we have to care if its not modified?

dab246: > We only care about the one overriding `dispose`. If originally `dispose` "does nothing" (as its author said), why do we have to care if its not modified?

The problem here is that we handle many things in the `dispose` function. In the current version, if `dispose` is not called, the handlers will not be executed. And now if you call it in `onClose` maybe some of those actions will throw an error, we don't know. So we need to check that `controllers` inherit from `BaseController`.

dab246: Please ensure that you have tested most of the app's features and that they are working properly with this change.

PR #2964, opened 2024-06-28
[POC] Integration test for Flutter web app

hoangdat:
To run it:
```
flutter drive --driver=integration_test/integration_test_driver.dart --target=integration_test/app_test.dart --debug --browser-name=chrome --web-port=8080 
```

`Blocker`: No way to interact with redirected web site (SSO stie)


hoangdat:
not use. Moved with semantic and Playwright

PR #2969, opened 2024-07-01
Use `junk` role for `Spam/Junk` folder

### LINKED ISSUE #2894: IMAP Mailbox Name Attributes for Spam folder is Junk
### Description

JMAP spec states that mailboxes role must be one of [IMAP Mailbox Name Attributes](https://www.iana.org/assignments/imap-mailbox-name-attributes/imap-mailbox-name-attributes.xhtml)

According to this link, the attribute name for Junk/Spam folder is Junk and not Spam.

As a consequence, the mailbox used by the server is ignored, and a Spam folder is created instead.

This can be reproduced by connecting to a Stalwart server.

jip149:
This is the role that should be used according to the JMAP specification. Closes #2894

In order not to break compatibility with current JAMES implementation (that uses incorrect `spam` role), this should be merged when `jmap-dart-client` [PR 89](https://github.com/linagora/jmap-dart-client/pull/89) has been merged.

### DIFF
--- a/contact/pubspec.lock
+++ b/contact/pubspec.lock
@@ -564,7 +564,7 @@ packages:
     description:
       path: "."
       ref: master
-      resolved-ref: "21d15bc1a6a75e048ee1e3cd751dd0815492ce20"
+      resolved-ref: "1e7d1e2aaaca2a079b997aa2188e8e7039e2ad72"
       url: "https://github.com/linagora/jmap-dart-client.git"
     source: git
     version: "0.0.1"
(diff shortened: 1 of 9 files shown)


chibenwa (approved):

jip149:
The failed check is caused by permission issues. Is there something I can do to help?

hoangdat:
> The failed check is caused by permission issues. Is there something I can do to help?

Dont worry, we only want to testing more with James. I will let you know when we are done

hoangdat:
hi @jip149 , I push a commit to update dedicated `jmap_dart_client` with `junk`.

dab246 (approved):
LGTM

PR #2986, opened 2024-07-08
Hot-fix composer composer

dab246:


hoangdat:
- [ ] inline image of reply message is not showed well 
image

dab246:
> * [x]  inline image of original message is not showed well when reply email also have inline image
> 
> image

Fixed. @hoangdat  Please test again. Thanks.

https://github.com/linagora/tmail-flutter/assets/80730648/5fc82fc8-f4b4-451b-98a5-dd9d5763e871

dab246:
Handle in #3014

Issue #2949, opened 2024-06-20
Remove duplicated information in email of calendar event 


dieptran88:
### Desc: 
- Currently in a calendar event email, the same information is displayed 3 times:

![image](https://github.com/linagora/tmail-flutter/assets/68209176/615552ac-f224-4779-9a61-fa214a96ae11)

![image](https://github.com/linagora/tmail-flutter/assets/68209176/462a0cb0-8c64-4f97-8e63-30d132898b57)
 
### DoD:
- [ ] We should eliminate the duplicated content 
- [ ] make sure precise information in each blue bar

PR #3020, opened 2024-07-19
TF-2949 Fix remove duplicated information in email of calendar event

dab246:
## Issue

#2949

## Resolved

https://github.com/user-attachments/assets/3142fb15-9f4f-4162-8fe5-a8fcecd6d7ad


hoangdat:
- [ ] how about for the case of many participants?

dab246:
> * [x]  how about for the case of many participants?

Please watch the demo video from second `1:04`

Screenshot 2024-07-23 at 09 04 54

tddang-linagora (approved):

hoangdat:
- [ ] how about to expand content (red square) as much as possible 
Screenshot 2024-07-26 at 15 01 39 3

dab246:
> * [ ]  how about to expand content (red square) as much as possible
> 
> Screenshot 2024-07-26 at 15 01 39 3

We have to set its size. Currently its maximum height is `250dp`. It contains a webview so it can't auto-size to the content length because it's wrapped around a scrollview.

hoangdat:
- [ ] we have the way to expand it but dont have the way to collapse it
image

dab246:
> * [ ]  we have the way to expand it but dont have the way to collapse it
> 
> image

Should we do that? The design doesn't have it yet.

hoangdat:
> We have to set its size. Currently its maximum height is 250dp. It contains a webview so it can't auto-size to the content length because it's wrapped around a scrollview.

But Mobile can show whole content. Please do the same

dab246:
> * [x]  we have the way to expand it but dont have the way to collapse it
> 
> image

Added.

https://github.com/user-attachments/assets/e3bf36a6-ac86-41ec-997e-86c26a24a379

hoangdat (approved):

Issue #2517, opened 2024-01-26
[Bug] When I rename a folder with special characters, it does not show error message 


dieptran88:
### Context

- Web v0.11.3
- https://tmail.linagora.com/

### Reproduction Steps
Case 1: Rename a folder 
1. Click Rename a folder
2. In rename popup, enter some special characters => Save
3. The folder name is not changed but there is no error message 

=> Expectation: When the folder name is not changed, there should be an error message to explain 

Case 2: Create a new folder
1. Click create a new folder
2. Input folder name with some special characters => Save
3. There is an error message displayed but the message content should be improved

### Additional information
[screen-capture (66).webm](https://github.com/linagora/tmail-flutter/assets/68209176/f710076e-b2b5-4e0d-a2f4-d99bb583fdc0)

dab246:
Related #444

PR #3056, opened 2024-08-05
TF-2517 Fix rename a folder with special characters, it does not show error message.

dab246:
## Issue 

#2517 

## Mailbox special characters validation

- Does not contain one of the forbidden characters `%*\r\n` or start with `#`

## Resolved

https://github.com/user-attachments/assets/17852f72-6a88-411d-990e-9998f97ade50

### DIFF
--- a/lib/features/mailbox_creator/domain/extensions/mailbox_name_special_character_validator_extension.dart
+++ b/lib/features/mailbox_creator/domain/extensions/mailbox_name_special_character_validator_extension.dart
@@ -1,0 +1,15 @@
+
+extension MailboxNameSpecialCharacterValidatorExtension on String {
+  bool get isValid {
+    if (startsWith('#')) {
+      return false;
+    }
+
+    final forbiddenChars = RegExp(r'[%*\r\n]');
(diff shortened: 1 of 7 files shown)


lib/features/mailbox_creator/domain/extensions/mailbox_name_special_character_validator_extension.dart
```diff
@@ -0,0 +1,15 @@
+
+extension MailboxNameSpecialCharacterValidatorExtension on String {
+  bool get isValid {
+    if (startsWith('#')) {
+      return false;
+    }
+
+    final forbiddenChars = RegExp(r'[%*\r\n]');
```
hoangdat, line 8: why you have this list of chars?

dab246: That is from jmap server response in `Mailbox/set`

Screenshot 2024-08-12 at 16 03 54

dab246 (commented):
Approved

hoangdat (approved):

Issue #2938, opened 2024-06-13
Request read receipt is not saved onto drafts


chibenwa:
### Description

GIVEN bob write an email
AND bob asks for read receipts
AND saves the email
WHEN bob opens the draft back
THEN the information about read receipts is lost

We would expect the information not to disapear.

Technical details:

 - We already create the draft with `header:Disposition-Notification-To:asText: "btellier@linagora.com"` -> GOOD
 - But the composer opening from the draft do NOT take this header into account and thus looses the information -> Not good and likely easy to fix

PR #3059, opened 2024-08-05
TF-2938 Fix request read receipt is not saved onto drafts

dab246:
## Issue

#2938 

## Resolved

https://github.com/user-attachments/assets/7c3954e7-f15d-42a9-8c00-6384c2aab4ef


tddang-linagora (approved):

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -688,7 +689,7 @@ class ComposerController extends BaseController with DragDropFileMixin implement
       if (composerArguments.value?.emailActionType == EmailActionType.reopenComposerBrowser) {
         log('ComposerController::_initEmail: hasRequestReadReceipt = ${arguments.hasRequestReadReceipt}');
         hasRequestReadReceipt.value = arguments.hasRequestReadReceipt ?? false;
-      } else {
+      } else if (composerArguments.value?.emailActionType != EmailActionType.editDraft) {
```
hoangdat, line 692: why we need to getSettings in this case? IMO, we only need to read from the draft and add it to composer (dont care about setting here)

tddang-linagora:
Worked on web

tddang-linagora (approved):

PR #3063, opened 2024-08-05
[Hotfix] Email list auto reload again when perform some actions email (Mark as read/star,..)

dab246:
## Issue

- Email list auto reload again when perform some actions email (Mark as read/star,..)

https://github.com/user-attachments/assets/8aabd7ac-0c9d-4b59-b5b5-0d70d7056c16

## Root cause

- Because every time we perform an action with email, we will refresh the mailbox list. On the other hand, we always reset the value for the currently selected mailbox, which leads to it always calling the `getAllEmail` function.

Screenshot 2024-08-05 at 19 13 52

## Solution

We will check if the current mailbox is selected then only update that mailbox but not call the function to get the email list again. Switch the `SelectedMailbox` listener to `EmailUIAction` to perform the call to `getAllEmail`

## Resolved

https://github.com/user-attachments/assets/1884dc29-0ce2-41ae-85d1-6247e021dde4

### DIFF
--- a/lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
+++ b/lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
@@ -588,6 +588,9 @@ class MailboxDashBoardController extends ReloadableController with UserSettingPo
 
   void setSelectedMailbox(PresentationMailbox? newPresentationMailbox) {
     log('MailboxDashBoardController::setSelectedMailbox: SELECTED_MAILBOX_ID = ${newPresentationMailbox?.id.asString} |  SELECTED_MAILBOX_NAME = ${newPresentationMailbox?.name?.name} | ');
+    if (selectedMailbox.value?.id != newPresentationMailbox?.id) {
+      dispatchEmailUIAction(GetAllEmailAction(newPresentationMailbox));
+    }
     selectedMailbox.value = newPresentationMailbox;
   }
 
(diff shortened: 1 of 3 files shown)


tddang-linagora (approved):
Worked on Web

hoangdat:
- [ ] add integration_test for web to this PR

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -588,6 +588,9 @@ class MailboxDashBoardController extends ReloadableController with UserSettingPo
 
   void setSelectedMailbox(PresentationMailbox? newPresentationMailbox) {
     log('MailboxDashBoardController::setSelectedMailbox: SELECTED_MAILBOX_ID = ${newPresentationMailbox?.id.asString} |  SELECTED_MAILBOX_NAME = ${newPresentationMailbox?.name?.name} | ');
+    if (selectedMailbox.value?.id != newPresentationMailbox?.id) {
+      dispatchEmailUIAction(GetAllEmailAction(newPresentationMailbox));
+    }
```
hoangdat, line 593: IMO, it is not clean. select function should only select, dont need to do other thing. Get all email should move back to listener of this change.

hoangdat:
can I close it @dab246 ?

dab246:
> can I close it @dab246 ?

Yes. Migrated to #3073

PR #3097, opened 2024-08-19
[HOT-FIX] Fix cannot copy & paste base64 image as inline image in composer web

dab246:
## Issue

- Cannot copy & paste base64 image as inline image in composer web

## Dependent

https://github.com/linagora/html-editor-enhanced/pull/44

## Resolved

- Single image:

https://github.com/user-attachments/assets/8d9d9e63-05fb-478d-a937-a909006d95b8

- Multiple Image:

https://github.com/user-attachments/assets/96469ac8-40fe-4cea-a2c9-8498188a78ca

### DIFF
--- a/lib/features/composer/presentation/extensions/file_upload_extension.dart
+++ b/lib/features/composer/presentation/extensions/file_upload_extension.dart
@@ -1,7 +1,4 @@
 import 'dart:convert' as convert;
-import 'dart:typed_data' as type_data;
-
-import 'package:flutter/foundation.dart';
 import 'package:html_editor_enhanced/utils/file_upload_model.dart';
 import 'package:model/upload/file_info.dart';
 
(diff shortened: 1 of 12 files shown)


tddang-linagora (approved):
Tested on web

pubspec.yaml
```diff
@@ -57,10 +57,12 @@ dependencies:
       url: https://github.com/linagora/rich-text-composer.git
       ref: main
 
+  # TODO: We will change it when the PR in upstream repository will be merged
+  # https://github.com/linagora/html-editor-enhanced/pull/44
   html_editor_enhanced:
     git:
       url: https://github.com/linagora/html-editor-enhanced.git
-      ref: main
+      ref: support-copy-and-paste-multiple-image
```
hoangdat, line 65: update it

lib/features/composer/presentation/extensions/file_upload_extension.dart
```diff
@@ -23,22 +20,17 @@ extension FileUploadExtension on FileUpload {
     return base64;
   }
 
-  Future<FileInfo?> toFileInfo() async {
+  FileInfo? toFileInfo() {
     if (base64Data != null) {
-      final bytes = await compute(convertBase64ToBytes, base64Data!);
```
hoangdat, line 28: is it ok in performance aspect?

dab246: Still ok, because `compute` only improves performance on mobile, not on web.

pubspec.yaml
dab246: done

lib/features/composer/presentation/extensions/file_upload_extension.dart
hoangdat: so why not keep it for mobile?

dab246: 1. We haven't used it anywhere
2. FileUpload is an object of the `html_editor_enhanced` library. We only use `html_editor_enhanced` for web. So we can't write `compute` in the extension of `FileUpload`
3. If we want to keep that code, we should copy it to another `Utils` class,

hoangdat (approved):

Issue #3095, opened 2024-08-16
[PublicAsset] Handle pasting base64 image to identity editor


tddang-linagora:
## Description
With PublicAsset feature up and running, we should stop our support for base64 image type in signature. For any user using Twake Mail that has PublicAsset capability supported, Twake Mail will stop showing any base64 image that is pasted to its signature editor.

## DoDs

- [ ] ~Write a function to verify if the image html's string is of http/https source or of base64 source~
- [ ] Trigger the transform event on paste
- [ ] Bonus: Trigger the transform event on drop text

chibenwa:
We need to:
 - Still show well base64 in existing emails
 - When creating email, pasting base64 we need to upload an inlined attachment
 - When creating signature, pasting base64 we need to upload a public asset and reference it.

PR #3098, opened 2024-08-20
TF-3095 public asset handle base64 on paste

tddang-linagora:
## Issue
- #3095 

## Demo

https://github.com/user-attachments/assets/adca3267-dc5f-45e9-8d7e-e68777b7e9ca

### DIFF
--- a/lib/features/identity_creator/presentation/identity_creator_controller.dart
+++ b/lib/features/identity_creator/presentation/identity_creator_controller.dart
@@ -17,6 +17,7 @@ import 'package:flutter/cupertino.dart';
 import 'package:flutter/scheduler.dart';
 import 'package:flutter_image_compress/flutter_image_compress.dart';
 import 'package:get/get.dart';
+import 'package:html_editor_enhanced/html_editor.dart';
 import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/capability/capability_identifier.dart';
 import 'package:jmap_dart_client/jmap/core/capability/core_capability.dart';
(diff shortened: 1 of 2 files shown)


lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -782,15 +785,50 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin {
...
+      maxWidth: maxWidth);
+  }
+
+  void onPasteImageFailure(
+    BuildContext context,
+    List<FileUpload>? listFileUpload,
+    {String? base64,
+    required UploadError uploadError}
+  ) {
+    logError('$runtimeType::onPasteImageFailure: $uploadError');
```
dab246, line 805:
```suggestion
    logError('IdentityCreatorController::onPasteImageFailure: $uploadError');
```

lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -782,15 +785,50 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin {
...
     required BuildContext context,
     required DropDoneDetails details,
     required double maxWidth
   }) async {
-    try {
-      clearFocusEditor(context);
+    clearFocusEditor(context);
+
+    final listFileInfo = await onDragDone(context: context, details: details);
+    if (!context.mounted) return;
```
dab246, line 819: We should not block upload when context is not mounted. Because upload process is not bound by context. Context only affects view and here is toast

lib/features/identity_creator/presentation/identity_creator_controller.dart
tddang-linagora: It's the same either way?

lib/features/identity_creator/presentation/identity_creator_controller.dart
tddang-linagora: Done

hoangdat (approved):

lib/features/identity_creator/presentation/identity_creator_controller.dart
dab246: It's the same result, but we should put it correctly in each case. Here `IdentityCreatorController` is not inherited by any class so we just need to write a specific name for it. anw, it's okay to write like that, but be careful next time.

dab246 (approved):
lgtm

PR #3109, opened 2024-08-23
[HOT-FIX] `Mark as Spam` action is not displayed when edit rule filter

dab246:
## Issue

`Mark as Spam` action is not displayed when edit rule filter

## Root cause

Spam mailbox id not found do get all mailbox not completed

## Resolved

https://github.com/user-attachments/assets/e83f99db-0bf8-478f-9775-be981f42b281

### DIFF
--- a/lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
+++ b/lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
@@ -9,7 +9,6 @@ import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/session/session.dart';
 import 'package:jmap_dart_client/jmap/mail/email/email_address.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
-import 'package:model/mailbox/presentation_mailbox.dart';
 import 'package:model/model.dart';
 import 'package:rule_filter/rule_filter/rule_action.dart';
 import 'package:rule_filter/rule_filter/rule_append_in.dart';
@@ -42,7 +41,6 @@ import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 import 'package:tmail_ui_user/main/routes/app_routes.dart';
 import 'package:tmail_ui_user/main/routes/dialog_router.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
-
 
 class RulesFilterCreatorController extends BaseMailboxController {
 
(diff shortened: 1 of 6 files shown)


lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
```diff
@@ -145,7 +135,18 @@ class RulesFilterCreatorController extends BaseMailboxController {
     }
   }
 
+  @override
+  void onDone() {
+    viewState.value.fold((failure) => null, (success) {
+      if (success is GetAllMailboxSuccess) {
+        _setUpRuleFilterActions();
```
tddang-linagora, line 142: Why can't this method be called in `handleSuccessViewState`?

lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
```diff
@@ -206,90 +214,72 @@ class RulesFilterCreatorController extends BaseMailboxController {
...
+          );
         }
+      } else {
+        mailboxSelected.value = PresentationMailbox.unifiedMailbox;
+        listEmailRuleFilterActionSelected.add(
+          MoveMessageActionArguments(mailbox: PresentationMailbox.unifiedMailbox)
+        );
       }
-      listEmailRuleFilterActionSelected.refresh();
     }
```
tddang-linagora, line 278:
```suggestion
    if (_currentTMailRule!.action.appendIn.mailboxIds.isNotEmpty != true) return;
    
    final mailboxNode = findMailboxNodeById(_currentTMailRule!.action.appendIn.mailboxIds.first);
    if (mailboxNode == null) {
      mailboxSelected.value = PresentationMailbox.unifiedMailbox;
      listEmailRuleFilterActionSelected.add(
        MoveMessageActionArguments(mailbox: PresentationMailbox.unifiedMailbox)
      );
    } else if (mailboxNode.item.isSpam) {
      listEmailRuleFilterActionSelected.add(MarkAsSpamActionArguments());
    } else {
      mailboxSelected.value = mailboxNode.item;
      listEmailRuleFilterActionSelected.add(
        MoveMessageActionArguments(mailbox: mailboxNode.item)
      );
    }
```

Avoid nesting `if` statements

lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
dab246: Because it will process the incoming data multiple times. Here we only want to process the last data received.

lib/features/rules_filter_creator/presentation/rules_filter_creator_controller.dart
dab246: Your suggestion will make us lose the processing behind it. There should be no return call here. We can separate the function to handle it.

tddang-linagora (approved):

hoangdat (approved):

Issue #3123, opened 2024-08-30
[Bug] Irrelevant "Load more" when filtering emails by "Unread"


tprudentova:
0.13.0, staging, tested on Firefox, Safari, Opera

1. Have a few unread letters in inbox 
2. Select the "Unread" filter 
3. Scroll to the bottom 
4. "Load more" that doesn't do anything, even though you've scrolled to the actual last unread letter 

https://github.com/user-attachments/assets/44eef5f9-14fe-4b5e-9710-789cb2b2e547

PR #3130, opened 2024-09-04
TF-3123 Fix load more irrelevant when filtering emails by unread

dab246:
## Issue

#3123 

## Reproduced

https://github.com/user-attachments/assets/e81f19ca-c5da-49ca-8e7e-d21cd05856b9

## Root cause

- Because the list returned when loading more always has the `last element` of the current list

## Resolved

https://github.com/user-attachments/assets/25695b8d-da62-44c5-a224-7af2d094ff9d


tddang-linagora (approved):

dab246:
> * [X]  cannot load more when app run in big screen
> 
>  Screen.Recording.2024-09-13.at.18.43.44.mov

Fixed. Please see below video demo:


https://github.com/user-attachments/assets/be95f282-f866-4c6e-ae63-fe2016df86fa

tddang-linagora (approved):

PR #3139, opened 2024-09-11
Replace "!is" with "is!"

leveretka:
In Dart, there is no operator `!is`. The intention was most likely to use `is!`.

In Dart, construction `!is` will perform a non-null assertion on the left operand (`!`) and then execute the `is` operator. So the code almost never behaves as expected.

At Sonar, we've implemented a rule for such issues: https://rules.sonarsource.com/dart/RSPEC-7054/

This is especially painful when you have Flutter & Kotlin in one project because Kotlin has `!is`.

Let me know if my suggestion makes sense to you.

You can also try to analyze your project on SonarCloud following this announcement: https://community.sonarsource.com/t/sonarcloud-supports-dart/124883

### DIFF
--- a/lib/features/base/reloadable/reloadable_controller.dart
+++ b/lib/features/base/reloadable/reloadable_controller.dart
@@ -89,7 +89,7 @@ abstract class ReloadableController extends BaseController {
   }
 
   void _handleGetSessionFailure(dynamic exception) {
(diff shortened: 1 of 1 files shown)


dab246:
@leveretka  this issue fixed at #3140 for `refactor` branch. Can we close this?

PR #3140, opened 2024-09-11
[HOT-FIX] operator not supported in dart

dab246:
## Issue

- https://github.com/linagora/tmail-flutter/pull/3139

## Root cause

- In Dart, there is no operator `!is`. Construction `!is` will perform a non-null assertion on the left operand (!) and then execute the is operator. So the code almost never behaves as expected.

## Documents 

- https://dart.dev/language/operators#type-test-operators

### DIFF
--- a/test/main/utils/type_operator_test.dart
+++ b/test/main/utils/type_operator_test.dart
@@ -1,0 +1,88 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:tmail_ui_user/main/exceptions/remote_exception.dart';
+
+bool isNotBadCredentialsExceptionUseIsNotOperator(dynamic exception) {
+  return exception is! BadCredentialsException;
+}
+
+bool isBadCredentialsExceptionUseIsOperator(dynamic exception) {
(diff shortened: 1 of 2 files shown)


tddang-linagora (approved):

test/main/utils/type_operator_test.dart
```diff
@@ -0,0 +1,88 @@
...
+    });
+  });
+
+  group('!is operator unit test', () {
+    test('should return true when exception is a BadCredentialsException', () {
+      // Arrange
+      const badCredentialsException = BadCredentialsException();
+
+      // Act
+      final result = isBadCredentialsExceptionUseIsOperator(badCredentialsException);
```
leveretka, line 71: I assume you wanted to use the function `isBadCredentialsExceptionUseNotIsOperator` here. Also on the line 82

dab246: Oh, that was my mistake. Fixed.

leveretka (approved):

hoangdat (approved):

Issue #3025, opened 2024-07-23
`Mark as read/unread/important/unimportant` action refresh email view, reset selection in Search


tddang-linagora:
### Description

https://github.com/user-attachments/assets/d1598a97-5a84-4bed-b052-3cc6c2cb35ab

### Expected result
Selection is kept, email view is not refreshed

### Current behavior
Selection is lost, email view is refreshed

#### Reproduction Steps
- Open tmail.linagora.com, search anything
- From the result email, select few of them
- Mark as read/unread/important/unimportant any email

### Context

Firefox, MacOS

PR #3187, opened 2024-10-02
TF-3025 [SEARCH] Search result list auto refresh when perform some actions `read/star/...`

dab246:
## Issue

#3025 

## Resolved

- Search result list auto refresh:

https://github.com/user-attachments/assets/45e91ff6-beb7-4891-9b9d-4b0b69a74677

- Actions for selection list incorrectly:

https://github.com/user-attachments/assets/e06074a6-0877-4389-893a-f903af70f332

### DIFF
--- a/lib/features/thread/domain/usecases/search_email_interactor.dart
+++ b/lib/features/thread/domain/usecases/search_email_interactor.dart
@@ -1,5 +1,6 @@
 
-import 'package:core/core.dart';
+import 'package:core/presentation/state/failure.dart';
+import 'package:core/presentation/state/success.dart';
 import 'package:dartz/dartz.dart';
 import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/filter/filter.dart';
@@ -26,11 +27,11 @@ class SearchEmailInteractor {
       Set<Comparator>? sort,
       Filter? filter,
       Properties? properties,
-      bool isRefreshChange = false,
+      bool needRefreshSearchState = false,
     }
   ) async* {
     try {
-      if (isRefreshChange) {
+      if (needRefreshSearchState) {
         yield Right(RefreshingSearchState());
       } else {
         yield Right(SearchingState());
--- a/lib/features/thread/domain/usecases/search_email_interactor.dart
+++ b/lib/features/thread/domain/usecases/search_email_interactor.dart
@@ -26,10 +26,15 @@ class SearchEmailInteractor {
       Set<Comparator>? sort,
       Filter? filter,
       Properties? properties,
+      bool isRefreshChange = false,
     }
   ) async* {
     try {
-      yield Right(SearchingState());
+      if (isRefreshChange) {
+        yield Right(RefreshingSearchState());
+      } else {
+        yield Right(SearchingState());
+      }
 
       final emailList = await threadRepository.searchEmails(
         session,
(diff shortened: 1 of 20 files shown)


tddang-linagora (approved):

chibenwa:
If I do not search for unread emails then we do not need to update the search when the user reads a mail.

With the current implementation, would we refresh it?

dab246:
> If I do not search for unread emails then we do not need to update the search when the user reads a mail.
> 
> With the current implementation, would we refresh it?

Yes, we will still refresh it. Please see below demo:


https://github.com/user-attachments/assets/1c0a4e2f-1947-4c97-93ae-e0cb3ec7f2d6

chibenwa:
> Yes, we will still refresh it.

Ok so by doing this we just x10 the count of search requests to the server, correct?

dab246:
> > Yes, we will still refresh it.
> 
> Ok so by doing this we just x10 the count of search requests to the server, correct?

No, I think I misunderstood you. 
1. If email is unread 
```
WHEN user open this email or click `blue dot` icon 
THEN refresh email list & email mark as read
```
2. If email is read 

```
WHEN user open this email
THEN not refresh email list
```

hoangdat:
- [ ] when app has realtime update, search result also reload. We should not reload it

https://github.com/user-attachments/assets/b514c761-6b0a-4736-bf8d-5d62fb1d4c9c

dab246:
> * [x]  when app has realtime update, search result also reload. We should not reload it
> 
>  Screen.Recording.2024-10-09.at.09.13.38.mov

It still persists the list when real time update. Please check again.



https://github.com/user-attachments/assets/086a81da-6732-4f82-8858-781f09a49ac7

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -810,12 +814,14 @@ class ThreadController extends BaseController with EmailActionController {
     searchController.clearTextSearch();
   }
 
-  void _searchEmail({UnsignedInt? limit}) {
+  void _searchEmail({UnsignedInt? limit, bool isRefreshChange = false}) {
```
hoangdat, line 817:
```suggestion
  void _searchEmail({UnsignedInt? limit, bool needRefreshSearchState = false}) {
```
WDYT?

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -585,7 +589,7 @@ class ThreadController extends BaseController with EmailActionController {
   void _refreshEmailChanges({jmap.State? currentEmailState}) {
     log('ThreadController::_refreshEmailChanges(): currentEmailState: $currentEmailState');
     if (searchController.isSearchEmailRunning) {
-      _searchEmail(limit: limitEmailFetched);
+      _searchEmail(limit: limitEmailFetched, isRefreshChange: true);
```
hoangdat, line 592: we can write the test:
```
WHEN thread controller in searching, 
AND `MarkAsStarEmailSuccess` is comming,
THEN `_searchEmailInteractor` is invoked with proper arguments
AND `listEmailController` should not jumped
AND `mailboxDashBoardController.emailsInCurrentMailbox` should not be cleared
```

lib/features/thread/presentation/thread_controller.dart
dab246: it doesn't make sense in this case

hoangdat:
Please test mobile UI too 
- [ ] selection appbar is not display well 
IMG_6435

lib/features/thread/domain/usecases/search_email_interactor.dart
```diff
@@ -26,10 +26,15 @@ class SearchEmailInteractor {
       Set<Comparator>? sort,
       Filter? filter,
       Properties? properties,
+      bool isRefreshChange = false,
```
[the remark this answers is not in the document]
dab246: done

lib/features/thread/presentation/thread_controller.dart
dab246: done

tddang-linagora (approved):

hoangdat (approved):

PR #3195, opened 2024-10-08
Simplify the prebuild script

tk-nguyen:
Detect existence of `pubspec.yaml` instead of hardcoding module names.

### DIFF
--- a/scripts/prebuild.sh
+++ b/scripts/prebuild.sh
@@ -4,36 +4,19 @@ set -e
 # debug log
 set -x
 
-cd core
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
+# Add additional modules to the end of this, seperated by space
+modules=("core" "model" "contact" "forward" "rule_filter" "fcm" "email_recovery" "server_settings")
 
-## Install necessary pods
-# cd ../ios
-# flutter pub get && pod install
+for mod in "${modules[@]}"; do
+    (
+        cd "$mod"
+        flutter pub get
+        dart run build_runner build --delete-conflicting-outputs
+    )
+done
 
-cd ../model
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
-
-cd ../contact
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
-
-cd ../forward
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
-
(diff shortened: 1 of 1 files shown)


tddang-linagora (approved):

scripts/prebuild.sh
```diff
@@ -4,36 +4,11 @@ set -e
...
-
-cd ../server_settings
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
-
-cd ..
-flutter pub get \
-    && dart run build_runner build --delete-conflicting-outputs \
-    && dart run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart \
-    && dart run intl_generator:generate_from_arb --output-dir=lib/l10n --no-use-deferred-loading lib/main/localizations/app_localizations.dart lib/l10n/intl*.arb
+flutter pub get
```
dab246, line 7: `flutter pub get` for each module is necessary. Running it in the main module will likely not update other modules.

tk-nguyen: Readded.

scripts/prebuild.sh
```diff
@@ -4,36 +4,10 @@ set -e
 # debug log
 set -x
 
-cd core
-flutter pub get && dart run build_runner build --delete-conflicting-outputs
+find . -name "pubspec.yaml" -exec bash -c '
+  (cd $(dirname "$0") && flutter pub get && dart run build_runner build -d)
+' {} \;
```
hoangdat, line 9: please keep the order of execute build-runner

tk-nguyen: Can't, only way to do this is hardcoding

dab246: Screenshot 2024-10-09 at 10 01 45

IMO, we should specify the modules then run the loop.

tk-nguyen: > Screenshot 2024-10-09 at 10 01 45
> 
> IMO, we should specify the modules then run the loop.

Yes, this is my original plan

dab246:
- [ ] Squash it to one commit

tk-nguyen:
> * [ ]  Squash it to one commit

Done.

hoangdat (approved):

PR #3213, opened 2024-10-19
Add german to web app startup - de.json

chrismaster:
Add german to web app startup


hoangdat (approved):
FYI, this banner only available in web responsive screen: mobile, table. We encourage user should download app from store instead of using mobile web app. Thanks

dab246 (approved):

hoangdat:
hi @chrismaster , please use `git rebase` instead of `git merge`. Thank you

Issue #3200, opened 2024-10-15
[Improvement] cleanup the options menu for folders


florentos17:
when clicking the three dots next to a personal folder:
![image](https://github.com/user-attachments/assets/cd4e0a99-cd9f-4130-990c-e8ff36dcf066)

we get this menu:
![image](https://github.com/user-attachments/assets/ce61029d-e41b-4b7c-ad60-8044de4fedc6)

i suggest the following improvements:
1. the first entry capitalizes every word and the others do not. it would be good to make things consistent ;
2. i would move the "delete folder" option to the end of the list. i feel like it makes more sense, plus it seems weird to have a red option in the middle.

dab246:
> 1. the first entry capitalizes every word and the others do not. it would be good to make things consistent ;
>

IMO, In `Personal` folder, folders are named by user, we should not interfere with it. Avoid confusing users and affecting UX.

florentos17:
> IMO, In `Personal` folder the folders are named by user specified, we should not interfere with it. Avoid confusing users affecting UX.

sorry i was not clear ; i was talking about the first entry of the menu: `Open in New Tab` there are upper case letters for `New` and `Tab`, whereas for the other entries there are not: `New subfolder` with lower case `subfolder`

hoangdat:
+1. Thank you, @florentos17

PR #3215, opened 2024-10-21
TF-3200 updated the menu options

florentos17:
i am tackling [this issue](https://github.com/linagora/tmail-flutter/issues/3200)

1. i modified `Open in New Tab` to `Open in new tab`
2. the order of the options is now changed to include `Delete folder` as the last option

### DIFF
--- a/lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
+++ b/lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
@@ -60,11 +60,11 @@ mixin MailboxWidgetMixin {
         MailboxActions.markAsRead,
       MailboxActions.move,
       MailboxActions.rename,
-      MailboxActions.delete,
       if (mailbox.isSubscribedMailbox)
         MailboxActions.disableMailbox
       else
-        MailboxActions.enableMailbox
+        MailboxActions.enableMailbox,
+      MailboxActions.delete
     ];
   }
 
--- a/lib/l10n/intl_messages.arb
+++ b/lib/l10n/intl_messages.arb
@@ -1,5 +1,5 @@
 {
-  "@@last_modified": "2024-10-21T09:30:30.005573",
+  "@@last_modified": "2024-10-25T09:31:37.466558",
   "initializing_data": "Initializing data...",
   "@initializing_data": {
     "type": "text",
@@ -2518,7 +2518,7 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "openInNewTab": "Open in New Tab",
+  "openInNewTab": "Open in new tab",
   "@openInNewTab": {
     "type": "text",
     "placeholders_order": [],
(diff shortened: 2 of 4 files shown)


hoangdat:
Hi @chrismaster, please 
- run `dart run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart` after changing something in app_localization
- please also change it in Fr too

Thank you

lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
```diff
@@ -60,11 +60,11 @@ mixin MailboxWidgetMixin {
         MailboxActions.markAsRead,
       MailboxActions.move,
       MailboxActions.rename,
-      MailboxActions.delete,
       if (mailbox.isSubscribedMailbox)
         MailboxActions.disableMailbox
       else
-        MailboxActions.enableMailbox
+        MailboxActions.enableMailbox,
+      MailboxActions.delete
```
chibenwa, line 67:
```suggestion
        MailboxActions.enableMailbox,
        MailboxActions.delete
```

chibenwa (approved):

lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
dab246: The `enableMailbox` action is in an `else` condition so it needs to be indented. The `delete` action does not need to be indented.

florentos17:
> Hi @chrismaster, please
> 
>     * run `dart run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart` after changing something in app_localization
> 
>     * please also change it in `Fr, En, It, Ru` too
> 
> 
> Thank you

did you mean chrismaster or me ?
i had already run the command, the commit includes the resulting code
as for the translations, i don't think any is needed since it is already all in lowercase letters for `Fr`, `It` and `Ru` (only the english version was featuring uppercase letters and needed change)

dab246:
> * as for the translations, i don't think any is needed since it is already all in lowercase letters for `Fr`, `It` and `Ru` (only the english version was featuring uppercase letters and needed change)

Also change it in the `intl_en` file

florentos17:
one more approval is required so that we could merge and close this work :)

dab246:
> one more approval is required so that we could merge and close this work :)

@florentos17 Please rebase it

lib/l10n/intl_messages.arb
```diff
@@ -1226,12 +1226,6 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "nameToBeDisplayed": "Name to be displayed to recipients",
-  "@nameToBeDisplayed": {
-    "type": "text",
-    "placeholders_order": [],
-    "placeholders": {}
-  },
```
dab246, line 1234: why do we need delete it?

florentos17: oh you are right, i rebased this file wrong...

dab246:
@florentos17  

- [ ] Please rebase again

Issue #3222, opened 2024-10-22
Event button: remove yes-no-maybe for accepted-tentative-


chibenwa:
### Description

![Screenshot from 2024-10-22 22-49-59](https://github.com/user-attachments/assets/03fdc4b1-9713-40d6-aa14-9c5926eabfd8)

Clicking those buttons -> ERROR

![Screenshot from 2024-10-22 22-50-06](https://github.com/user-attachments/assets/70a16c7d-dc3a-4a2e-a87c-b617c90ba522)

 -> remove those buttons for participent status update email

PR #3224, opened 2024-10-23
TF-3222 Hide reply calendar event action buttons if email is event reply

tddang-linagora:
## Issue
- #3222 

## Reference
- https://www.rfc-editor.org/rfc/rfc5546.html#section-3.2

## Resolved

https://github.com/user-attachments/assets/da2e68ed-3f0a-45e1-8941-906aeaab457e

### DIFF
--- a/lib/features/email/presentation/extensions/calendar_event_extension.dart
+++ b/lib/features/email/presentation/extensions/calendar_event_extension.dart
@@ -356,7 +356,12 @@ extension CalendarEventExtension on CalendarEvent {
   }
 
   bool get isDisplayedEventReplyAction => method != null
-    && method != EventMethod.reply
+    && _methodIsRepliable
     && organizer != null
     && participants?.isNotEmpty == true;
+
+  bool get _methodIsRepliable => 
+    method == EventMethod.request ||
+    method == EventMethod.add ||
+    method == EventMethod.counter;
 }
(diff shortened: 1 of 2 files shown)


lib/features/email/presentation/extensions/calendar_event_extension.dart
```diff
@@ -356,6 +356,7 @@ extension CalendarEventExtension on CalendarEvent {
   }
 
   bool get isDisplayedEventReplyAction => method != null
+    && method != EventMethod.reply
```
chibenwa, line 359: While we are at it, what are other possible values here and for which others does it make sense to hide the buttuns?

tddang-linagora: Other possible values

Screenshot 2024-10-24 at 09 16 13

tddang-linagora: IMO

Publish: Hide
Request: Show
Reply: Hide
Add: Show
Cancel: Hide
Refresh: Hide
Counter: Show
DeclineCounter: Hide

@chibenwa How about you?

chibenwa: Agree

dab246 (approved):
LGTM

hoangdat (approved):

Issue #3226, opened 2024-10-24
Auto complete should rely minimum length to auto completable  on the capability object


quantranhong1999:
Currently, it seems frontend hardcode at least 3 input chars to start fire JMAP autocomplete request.

That minimum length should be configurable.

Frontend can rely on backend capability object. Example:
```json
"com:linagora:params:jmap:contact:autocomplete": {
  "minInputLength": 2
}
```

cf: https://github.com/linagora/tmail-backend/issues/1209

(backend change is implemented and should be included in the next release > 0.11.0)

PR #3233, opened 2024-10-25
TF-3226 Auto complete should rely minimum length to auto completable on the capability object

dab246:
## Issue

#3226


contact/lib/contact/model/autocomplete_capability.dart
tddang-linagora, line 1: Check github action warning

contact/lib/contact/model/autocomplete_capability.dart
```diff
@@ -0,0 +1,31 @@
...
+part 'autocomplete_capability.g.dart';
+
+@JsonSerializable(
+  explicitToJson: true,
+  includeIfNull: false,
+  converters: [
+    UnsignedIntNullableConverter()
+  ]
+)
+class AutocompleteCapability extends CapabilityProperties {
```
tddang-linagora, line 15: Should this be defined in `jmap-dart-client` instead?

dab246: No, it is an extension so we have already created a module to handle it.

contact/lib/contact/model/autocomplete_capability.dart
dab246: No problem, only name is not spell

tddang-linagora (dismissed):

lib/features/base/base_controller.dart
```diff
@@ -471,4 +472,12 @@ abstract class BaseController extends GetxController
...
     await cachingManager.closeHive();
   }
+
+  int getMinInputLengthAutocomplete({
+    required Session session,
+    required AccountId accountId,
+  }) {
+    final minInputLength = session.getMinInputLengthAutocomplete(accountId);
+    return minInputLength?.value.toInt() ?? AppConfig.defaultMinInputLengthAutocomplete;
+  }
```
hoangdat, line 482: Why need to put it in controller? How about to use extension or util class?

dab246: We already had handle `getMinInputLengthAutocomplete` in session extension and wirte unit test for it. This function to get value default `minInputLengthAutocomplete` when minInputLength from session is null. Additional, we only use it on all controller.

test/features/base/base_controller_test.dart
```diff
@@ -201,4 +211,85 @@ void main() {
...
   });
+
+  group('BaseController::getMinInputLengthAutocomplete::test', () {
+    late MockBaseController mockBaseController;
+
+    setUp(() {
+      mockBaseController = MockBaseController();
+    });
+
+    test('SHOULD return session min input length WHEN AutocompleteCapability available', () {
```
hoangdat, line 222: should one more test case:
```
test('SHOULD return default min input length WHEN AutocompleteCapability available, but no minInputLenght', () {
```

dab246: done

hoangdat (approved):

tddang-linagora (approved):

Issue #3243, opened 2024-10-29
Email without content


chibenwa:
### Description

[Livraison de votre colis ALUMINOR PREDICT.eml.txt](https://github.com/user-attachments/files/17555188/Livraison.de.votre.colis.ALUMINOR.PREDICT.eml.txt)

![image](https://github.com/user-attachments/assets/7d29bc72-ef7e-4384-99a4-b0cc215b3260)

PR #3245, opened 2024-10-29
TF-3243 Fix email without content

dab246:
## Issue 

#3243

## Root cause

- Because the function `Uri.decodeFull(inputText)` throws an exception `FormatException: Invalid UTF-8 byte (at offset 11)` during handle `3D Links ` in DOM.
- This error only occurs on the web

## Solution

- Add `try/catch` blocks to all email `transformer` DOM handler  functions

## Resolved

- Web:

https://github.com/user-attachments/assets/de975aff-0f1d-454e-b41a-af26412e8aa8

- Mobile:

[demo-mobile.webm](https://github.com/user-attachments/assets/06d394df-f359-4271-94fa-fddaa04a436e)

### DIFF
--- a/core/lib/presentation/utils/html_transformer/dom/add_lazy_loading_for_background_image_transformers.dart
+++ b/core/lib/presentation/utils/html_transformer/dom/add_lazy_loading_for_background_image_transformers.dart
@@ -1,5 +1,6 @@
 import 'package:core/data/network/dio_client.dart';
 import 'package:core/presentation/utils/html_transformer/base/dom_transformer.dart';
+import 'package:core/utils/app_logger.dart';
 import 'package:html/dom.dart';
 
 class AddLazyLoadingForBackgroundImageTransformer extends DomTransformer {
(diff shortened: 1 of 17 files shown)


chibenwa:
> Add try/catch blocks to all email transformer and sanitize handler functions

Does this means that **if I send a broken base64 link** then HTML sanitizing is bypassed and thus I could pass an XSS payload *after* the broken base64 link?

dab246:
> > Add try/catch blocks to all email transformer and sanitize handler functions
> 
> Does this means that **if I send a broken base64 link** then HTML sanitizing is bypassed and thus I could pass an XSS payload _after_ the broken base64 link?

No. We do HTML sanitization before we do URL sanitization in the a tag. So we always block XSS from the start.

tddang-linagora (dismissed):

hoangdat (dismissed):

hoangdat:
please rebase @dab246

dab246:
> please rebase @dab246

Done. @hoangdat

hoangdat (approved):

Issue #3247, opened 2024-10-29
[SEARCH] Cc is not displayed when clicking quick results


chibenwa:
### Description

https://github.com/user-attachments/assets/0cb895f2-a0bd-49eb-a867-ff655c167a38

### Extected result

When clicking "live search results" I do expect to have same ccs in the email

Bug is reported by Alex.

Cc @hoangdat

PR #3248, opened 2024-10-30
TF-3247 [SEARCH] Fix Cc is displayed when clicking quick search results

dab246:
## Issue

#3247 

## Root cause

- Due to missing `Cc`property when passing in `Email/get` method

## Resolved

https://github.com/user-attachments/assets/3e46d7fb-eff6-4b6d-827c-7de443613979


tddang-linagora (approved):

hoangdat:
Hi @dab246 
Quick search will be change frequently by user. From my point of view, please don't use FULL properties of GET method. Only add metadata properties. It will be good for performance.

WDYT? @chibenwa @Arsnael

dab246:
> Hi @dab246 Quick search will be change frequently by user. From my point of view, please don't use FULL properties of GET method. Only add metadata properties. It will be good for performance.
> 
> WDYT? @chibenwa @Arsnael

Yes, we are doing that now, we only get the necessary and basic `properties` for opening an email. We don't get other properties like `attachments, bodyValue, html....` here.

hoangdat:
Ah, I misunderstand. srr.

hoangdat (approved):

Issue #3253, opened 2024-11-04
Body content is lost 


hoangdat:
### Description

LinShare send back email content, but file information is not display

### Expected result

Should display file information of LinShare

### Current behavior

<img width="1132" alt="image" src="https://github.com/user-attachments/assets/f92a2eb5-56fa-4286-a3cb-605155f9b129">

Cc @chibenwa

chibenwa:
/home/hp/Downloads/Vous avez partagé un fichier.eml.txt

PR #3254, opened 2024-11-04
TF-3253 Fix body content is lost

dab246:
## Issue

#3253 

## Root cause

- The `section` tag of the html content is removed when sanitizing the html

## Solution

- Add the `section` tag to the list of tags allowed to appear in html

## Resolved

https://github.com/user-attachments/assets/a4b6bd07-81ed-4cc2-82a2-2b6a9ed8c821


tddang-linagora (dismissed):

hoangdat:
I found some `event handler` for section in cheat sheet. Can you add more test to make sure it is also sanitized well

image

dab246:
> I found some `event handler` for section in cheat sheet. Can you add more test to make sure it is also sanitized well
> 


done

tddang-linagora (approved):

hoangdat (approved):

PR #3270, opened 2024-11-11
TF-3269 Update login robot with starting page

### LINKED ISSUE #3269: [Patrol Integration test] Tests failed
### Description
Integration tests are broken

### Root cause
The new "Starting page" was added into our app, but was not accounted for in the tests

### Additional information

https://github.com/linagora/tmail-flutter/actions/runs/11770077298/job/32782005459

tddang-linagora:
## Issue
- #3269 

## Resolved

[starting-page-patrol.webm](https://github.com/user-attachments/assets/7ed525b3-c796-4173-b200-cde30c0370b5)

### DIFF
--- a/integration_test/robots/login_robot.dart
+++ b/integration_test/robots/login_robot.dart
@@ -14,7 +14,7 @@ class LoginRobot extends CoreRobot {
   Future<void> expectWelcomeViewVisible() => ensureViewVisible($(TwakeWelcomeView));
 
   Future<void> tapOnUseCompanyServer() async {
-    await $.pump(const Duration(seconds: 1));
+    await $.pumpAndSettle();
     await $(AppLocalizations().useCompanyServer).tap();
   }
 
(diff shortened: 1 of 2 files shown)


integration_test/robots/login_robot.dart
```diff
@@ -3,12 +3,21 @@ import 'package:flutter/material.dart';
...
 
 import '../base/core_robot.dart';
 
 class LoginRobot extends CoreRobot {
   LoginRobot(super.$);
 
+  Future<void> expectWelcomeViewVisible() => ensureViewVisible($(TwakeWelcomeView));
+
+  Future<void> tapOnUseCompanyServer() async {
+    await $.pump(const Duration(seconds: 1));
```
dab246, line 17: why do we need wait 1 seconds?

tddang-linagora: The "Interactors" and their dependencies of "LoginBindings" will not load in time if we take the action immediately after `TwakeWelcomeView` is loaded

dab246: So, we need to use another handle to make sure that the bindings are injected successfully. It's not a good idea to have random wait times like that. Should try more

dab246 (approved):
lgtm

hoangdat (approved):

Issue #3274, opened 2024-11-14
Deleted message vault: take into horizon


chibenwa:
CF 

The back exposes the restoration horizon eg 15 days

-> Add a little warning banner `Only messages deleted since 15 days may be restored.`
-> Time range selection may not expose options older than the horizon

![signal-2024-11-13-123626](https://github.com/user-attachments/assets/a158a287-8317-4a01-9cf1-76109b292eaa)

chibenwa:
Cc @guimard

Issue #3219, opened 2024-10-22
Display Limits of of Email Recovery


florentos17:
the JMAP session advertises two limits through its `MessageVaultCapability` object:
- the maximum number of mail that a user can recover at once, eg with one query: `maxEmailRecoveryPerRequest`
- the period during which a user can restore a mail alone after having deleted it: `restorationHorizon`

we would like to **have a yellow banner indicating these limits** in the email recovery window

chibenwa:
Can we link corresponding tmail-backend ticket please?

florentos17:
it is related to [this backend issue](https://github.com/linagora/tmail-backend/issues/1201)
and the corresponding backend pull requests are [this one](https://github.com/linagora/tmail-backend/pull/1224) and [this one](https://github.com/linagora/tmail-backend/pull/1248)

hoangdat:
- [ ] Must have verifying step after user select `Deletion date`. In this case, I choose the time range for the last 1 year.

<img width="717" alt="image" src="https://github.com/user-attachments/assets/ea49862c-0a78-4e24-9595-3419979b5790">

PR #3281, opened 2024-11-18
Display limits of email recovery

florentos17:
tackling [the issue 3219](https://github.com/linagora/tmail-flutter/issues/3219) (duplicated with [the issue 3274](https://github.com/linagora/tmail-flutter/issues/3274))

### DIFF
--- a/lib/features/email_recovery/presentation/email_recovery_controller.dart
+++ b/lib/features/email_recovery/presentation/email_recovery_controller.dart
@@ -8,9 +8,7 @@ import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/session/session.dart';
 import 'package:jmap_dart_client/jmap/core/utc_date.dart';
 import 'package:jmap_dart_client/jmap/mail/email/email_address.dart';
-import 'package:model/autocomplete/auto_complete_pattern.dart';
-import 'package:model/extensions/email_address_extension.dart';
-import 'package:model/mailbox/expand_mode.dart';
+import 'package:model/model.dart';
 import 'package:permission_handler/permission_handler.dart';
 import 'package:super_tag_editor/tag_editor.dart';
 import 'package:tmail_ui_user/features/base/base_controller.dart';
@@ -25,6 +23,7 @@ import 'package:tmail_ui_user/features/email_recovery/presentation/controller/in
 import 'package:tmail_ui_user/features/email_recovery/presentation/model/email_recovery_arguments.dart';
 import 'package:tmail_ui_user/features/email_recovery/presentation/model/email_recovery_field.dart';
 import 'package:tmail_ui_user/features/email_recovery/presentation/model/email_recovery_time_type.dart';
+import 'package:tmail_ui_user/features/email_recovery/presentation/model/session_extension.dart';
 import 'package:tmail_ui_user/features/manage_account/presentation/extensions/datetime_extension.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/app_config.dart';
@@ -34,7 +33,7 @@ class EmailRecoveryController extends BaseController with DateRangePickerMixin {
   GetAutoCompleteInteractor? _getAutoCompleteInteractor;
   GetDeviceContactSuggestionsInteractor? _getDeviceContactSuggestionsInteractor;
 
-  final deletionDateFieldSelected = EmailRecoveryTimeType.last1Year.obs;
+  final deletionDateFieldSelected = EmailRecoveryTimeType.last7Days.obs;
   final receptionDateFieldSelected = EmailRecoveryTimeType.allTime.obs;
   final startDeletionDate = Rxn<DateTime>();
   final endDeletionDate = Rxn<DateTime>();
@@ -347,6 +346,10 @@ class EmailRecoveryController extends BaseController with DateRangePickerMixin {
     KeyboardUtils.hideKeyboard(context);
     popBack();
   }
+
+  String getRestorationHorizonAsString() => arguments!.session.getRestorationHorizonAsString();
+
+  DateTime getRestorationHorizonAsDateTime() => arguments!.session.getRestorationHorizonAsDateTime();
 
   @override
   void dispose() {
--- a/lib/features/email_recovery/presentation/model/email_recovery_field.dart
+++ b/lib/features/email_recovery/presentation/model/email_recovery_field.dart
@@ -28,7 +28,7 @@ enum EmailRecoveryField {
   String getHintText(BuildContext context) {
     switch (this) {
       case EmailRecoveryField.deletionDate:
-        return AppLocalizations.of(context).last1Year;
+        return AppLocalizations.of(context).last7Days;
       case EmailRecoveryField.receptionDate:
         return AppLocalizations.of(context).allTime;
       case EmailRecoveryField.subject:
@@ -40,16 +40,20 @@ enum EmailRecoveryField {
     }
   }
 
-  List<EmailRecoveryTimeType> getSupportedTimeTypes() {
+  List<EmailRecoveryTimeType> getSupportedTimeTypes(DateTime restorationHorizon) {
     switch (this) {
       case EmailRecoveryField.deletionDate:
-        return [
-          EmailRecoveryTimeType.last1Year,
+        final supportedTypes = [
           EmailRecoveryTimeType.last7Days,
+          EmailRecoveryTimeType.last15Days,
           EmailRecoveryTimeType.last30Days,
           EmailRecoveryTimeType.last6Months,
-          EmailRecoveryTimeType.customRange,
-        ];
+          EmailRecoveryTimeType.last1Year,
+        ].where((type) => restorationHorizon
+                                .subtract(const Duration(seconds: 2)) // to allow "15 days" if restorationHorizon is exactly 15
+                                .isBefore(type.toOldestUTCDate()!.value)).toList();
+
+        return [...supportedTypes, EmailRecoveryTimeType.customRange];
       case EmailRecoveryField.receptionDate:
         return [
           EmailRecoveryTimeType.allTime,
--- a/lib/features/email_recovery/presentation/model/session_extension.dart
+++ b/lib/features/email_recovery/presentation/model/session_extension.dart
@@ -1,0 +1,42 @@
+
+import 'package:duration/duration.dart';
+import 'package:email_recovery/email_recovery/capability_deleted_messages_vault.dart';
+import 'package:jmap_dart_client/jmap/core/session/session.dart';
+import 'package:model/extensions/session_extension.dart';
+
+extension SessionExtension on Session {
+
+  String getRestorationHorizonAsString() {
+    String restorationHorizon = 'restorationHorizon';
+    String defaultHorizon = "15 days";
+
+    return (getCapabilityProperties(accountId, capabilityDeletedMessagesVault)
+        ?.props[0] as Map<String, dynamic>?)
+        ?[restorationHorizon]
+        ?? defaultHorizon;
+  }
+
+  Duration getRestorationHorizonAsDuration() {
+    try {
+      String horizonWithCorrectFormat = getRestorationHorizonAsString()
+          .replaceAll(" days", "d")
+          .replaceAll(" day", "d")
+          .replaceAll(" hours", "h")
+          .replaceAll(" hour", "h")
+          .replaceAll(" minutes", "m")
+          .replaceAll(" minute", "m")
+          .replaceAll(" seconds", "s")
+          .replaceAll(" second", "s")
+          .replaceAll(" milliseconds", "ms")
+          .replaceAll(" millisecond", "ms");
+
+      return parseDuration(horizonWithCorrectFormat, separator: ' ');
+    } catch (e) {
+      return const Duration(days: 15);
+    }
+  }
+
+  DateTime getRestorationHorizonAsDateTime() {
+    return DateTime.now().subtract(getRestorationHorizonAsDuration());
+  }
+}
--- a/lib/features/email_recovery/presentation/widgets/limits_banner.dart
+++ b/lib/features/email_recovery/presentation/widgets/limits_banner.dart
@@ -1,0 +1,33 @@
+import 'package:flutter/material.dart';
+
+class LimitsBanner extends StatelessWidget {
+  final String bannerContent;
+  
+  const LimitsBanner({
+    super.key,
+    required this.bannerContent,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      width: double.infinity,
+      padding: const EdgeInsets.symmetric(vertical: 12.0, horizontal: 16.0),
+      decoration: BoxDecoration(
+        color: Colors.yellow[400],
+        borderRadius: const BorderRadius.all(Radius.circular(8.0)),
+      ),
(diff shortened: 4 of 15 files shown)


florentos17:
This curious thing is happening: when accessing the recover deleted messages panel, at first an error as thrown, but when i trigger a hot reload, then everything is fine:


https://github.com/user-attachments/assets/d41fe78c-c16f-4ab2-be0a-bab5888f3445

florentos17:
the issue i had the other day is now resolved, everything is working fine ! it looks like this
Screenshot from 2024-11-20 16-05-37

in the web version, the options in the dropdown are filtered out so that "30 days" and more do not show if the horizon is 15 days for example. but in the mobile version, the period is selected with this widget (same as the one we get by clicking the little calendar icon in the web version) and i did not want to change the suggested options because the widget is also used for other features.
Screenshot from 2024-11-20 16-05-37

lib/features/email_recovery/presentation/email_recovery_controller.dart
```diff
@@ -338,6 +342,26 @@ class EmailRecoveryController extends BaseController with DateRangePickerMixin {
...
 
+  Duration getRestorationHorizon() {
+    String restorationHorizon = ((arguments!.session
+        .props[0] as Map<CapabilityIdentifier, CapabilityProperties>?)
+        ?[capabilityDeletedMessagesVault]
+        ?.props[0] as Map<String, dynamic>?)
+        ?['restorationHorizon']
+        ?? "15 days";
+
+    String restorationHorizonValidFormat = "15d";
```
florentos17, line 353: currently the server advertizes the horizon like "15 days". i will change it to "15d" because it is what is needed for `parseDuration`

florentos17: i found it made more sense to keep the advertisement that way since we need the human legible string anyway for the content of the banner

chibenwa:
@hoangdat is this work ok?

@florentos17 could you rebase this?

lib/features/email_recovery/presentation/email_recovery_controller.dart
```diff
@@ -338,6 +342,35 @@ class EmailRecoveryController extends BaseController with DateRangePickerMixin {
...
+        .replaceAll(" hours", "h")
+        .replaceAll(" hour", "h")
+        .replaceAll(" minutes", "m")
+        .replaceAll(" minute", "m")
+        .replaceAll(" seconds", "s")
+        .replaceAll(" second", "s")
+        .replaceAll(" milliseconds", "ms")
+        .replaceAll(" millisecond", "ms");
+
+    return parseDuration(horizonWithCorrectFormat, separator: ' ');
```
tddang-linagora, line 367: If the whole purpose of `duration` package is to use `parseDuration`, please write your own `parse` function and remove the package from `pubspec.yaml`. Also unit test that function.

florentos17: i copied the `parseDuration` function from https://github.com/desktop-dart/duration/blob/master/lib/src/parse/parse.dart#L6, should i include the license ?

florentos17:
if it's all good for you then i'll squash and rebase

tddang-linagora (approved):

hoangdat:
- [ ] Must have verifying step after user select `Deletion date`. In this case, I choose the time range for the last 1 year.

image

core/lib/utils/duration_utils.dart
```diff
@@ -0,0 +1,89 @@
+class DurationUtils {
+
+  // this class is a local copy of the one found at
+  // https://github.com/desktop-dart/duration/blob/master/lib/src/parse/parse.dart#L6
+  static Duration parseDuration(String input, {String separator = ','}) {
```
hoangdat, line 5: why we need this? how about using dependency directly?

chibenwa: I remembered it was asked by @tddang-linagora

tddang-linagora: We should minimize the using of third party library. If we only need a small function, let's copy it over.

florentos17:
> * [x]  Must have verifying step after user select `Deletion date`. In this case, I choose the time range for the last 1 year.

already done on the backend side !

florentos17:
> * [ ]  Must have verifying step after user select `Deletion date`. In this case, I choose the time range for the last 1 year.

it is already done on the backend side (the request is modified to be sure it complies with the calendar limit if not already).
but you are absolutely right, we could add a check here on the frontend as well: circle the date in red if invalid and/or refuse to send the request if invalid. 

but do you think it’s necessary ? the user is already warned with the yellow banner, and it’s a tiny and hidden feature, so maybe it could be ok to leave it like that ?

core/lib/utils/duration_utils.dart
hoangdat: from my point of view, we should use 3rd party lib to get the support from the community, don't need effort in maintaining it.

hoangdat:
> but do you think it’s necessary ? the user is already warned with the yellow banner, and it’s a tiny and hidden feature, so maybe it could be ok to leave it like that ?

- the best solution: prevent user to select the date range out of the limitation, but for it, huge effort in refactoring the data picker.
- the acceptable solution: warning for unaccepted picked date.

Btw, when I push the recovery request with date time out of limitation, back-end still response the completed status with 0 completion, 0 .... We should have better response for this.

chibenwa:
> the acceptable solution: warning for unaccepted picked date.

Technically with the yellow banner that's already the case.

The option is quite hidden (who knows where deleted message menu is) and is niche. 1% of users will open dead letter menu item. 1% will manually select a date. So we are speaking of user experience for 0.01%.

Are we lying to the user? No, backend will transparently filter data, which is what the banner says. Acceptable.

Can we please move on?

core/lib/utils/duration_utils.dart
chibenwa: Tend to agree with @hoangdat .

Why maintain this code if we have it from a healthy package (size, community, etc..)

florentos17: i'll drop the commit `removed package duration usages` when squashing then (if ok for you @tddang-linagora)

lib/l10n/intl_messages.arb
```diff
@@ -4041,35 +4041,5 @@
...
-  "@sigInSaasFailed": {
-    "type": "text",
-    "placeholders_order": [],
-    "placeholders": {}
-  },
-  "createTwakeIdFailed": "Create Twake Id failed. Please check again.",
-  "@createTwakeIdFailed": {
-    "type": "text",
-    "placeholders_order": [],
-    "placeholders": {}
```
hoangdat, line 4073: please dont remove it

lib/features/email_recovery/presentation/email_recovery_controller.dart
```diff
@@ -348,6 +352,35 @@ class EmailRecoveryController extends BaseController with DateRangePickerMixin {
...
```
hoangdat, line 377: write test for this method.

hoangdat (commented):
Please attach video demo for this.

hoangdat:
- [ ] big concerned: as you said, backend will filter unaccepted date range, but why not show the error or other response. It is like a success request but nothing was restored @florentos17.

### Reproduce step
1. open Trash
2. recover 
3. choose `Last 1 year`

https://github.com/user-attachments/assets/6c24203e-9de1-4b63-8d11-661b8a20ae64

lib/features/email_recovery/presentation/email_recovery_controller.dart
dab246, line 377: We should write it in extension of Session

lib/features/email_recovery/presentation/widgets/limits_banner.dart
```diff
@@ -0,0 +1,33 @@
...
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      width: double.infinity,
+      padding: const EdgeInsets.symmetric(vertical: 12.0, horizontal: 16.0),
+      decoration: BoxDecoration(
+        color: Colors.yellow[400],
+        borderRadius: BorderRadius.circular(8.0),
```
dab246, line 18:
```
        borderRadius: BorderRadius.all(Radius.circular(8.0)),
```

florentos17:
@dab246 i took you comments into account

florentos17:
> * [ ]  big concerned: as you said, backend will filter unaccepted date range, but why not show the error or other response. It is like a success request but nothing was restored @florentos17.

technically, it IS a success request. the user was warned that no mail older than 15 days would be restored, and it is exactly what happens. the banner acts as the warning. i agree with benoit that we could leave it like that.

model/pubspec.yaml
```diff
@@ -56,6 +56,11 @@ dependencies:
 
   dartz: 0.10.1
 
+  duration: 4.0.3
+
+  email_recovery:
+    path: ../email_recovery
```
hoangdat, line 62: IMO, `model` no need to be aware about `email_recovery`. Why not move related stuffs back to `email_recovery`

florentos17: as you want ! @dab246 is it ok with you ?

dab246: I agree with @hoangdat opinion

dab246: > as you want ! @dab246 is it ok with you ?

You can create a `session_extension` in `email_recovery` without having to use `session_extension` in the model.

florentos17: you're right ! i did that

lib/features/email_recovery/presentation/widgets/limits_banner.dart
florentos17: done !

lib/features/email_recovery/presentation/model/email_recovery_field.dart
```diff
@@ -40,16 +40,20 @@ enum EmailRecoveryField {
...
+        final supportedTypes = [
           EmailRecoveryTimeType.last7Days,
+          EmailRecoveryTimeType.last15Days,
           EmailRecoveryTimeType.last30Days,
           EmailRecoveryTimeType.last6Months,
-          EmailRecoveryTimeType.customRange,
-        ];
+          EmailRecoveryTimeType.last1Year,
+        ].where((type) => restorationHorizon
+                                .subtract(const Duration(seconds:2)) // to allow "15 days" if restorationHorizon is exactly 15 days
```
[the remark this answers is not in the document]
florentos17: done !

lib/features/email_recovery/presentation/email_recovery_controller.dart
florentos17: done !

florentos17: done !

lib/l10n/intl_messages.arb
florentos17: done !

core/lib/utils/duration_utils.dart
florentos17: done !

florentos17:
i rebased again to fix the last conflict.
i think all of the comments were addressed. does it look good to you ?

lib/features/email_recovery/presentation/model/email_recovery_field.dart
```diff
@@ -40,16 +40,20 @@ enum EmailRecoveryField {
...
+        final supportedTypes = [
           EmailRecoveryTimeType.last7Days,
+          EmailRecoveryTimeType.last15Days,
           EmailRecoveryTimeType.last30Days,
           EmailRecoveryTimeType.last6Months,
-          EmailRecoveryTimeType.customRange,
-        ];
+          EmailRecoveryTimeType.last1Year,
+        ].where((type) => restorationHorizon
+                                .subtract(const Duration(seconds: 2)) // to allow "15 days" if restorationHorizon is exactly 15
```
dab246, line 53: Why subtract `2s` and not some other number?

lib/l10n/intl_messages.arb
```diff
@@ -186,12 +186,6 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "reply_to_email_address_prefix": "Reply to",
-  "@reply_to_email_address_prefix": {
-    "type": "text",
-    "placeholders_order": [],
-    "placeholders": {}
-  },
```
dab246, line 194: why remove it?

lib/features/email_recovery/presentation/model/email_recovery_field.dart
florentos17: i figured that it was small enough to not significantly interfere with the horizon (smallest reasonable horizon would be around 1 minute, so 2 seconds does not change much), while making sure the desired option is considered valid.
but it is not an absolute choice, it can be changed

lib/l10n/intl_messages.arb
florentos17: yes i just saw, it was automatically rebased that way but it is an error, i am fixing it.

florentos17: i have a question though: why adding the line   "@@last_modified": "2024-11-25T10:57:37.546143" at the beginning of the file ? this is the source of almost every conflict i encounter...! ^^

hoangdat: Good question. And it is the answer from Copilot :). please remove it and verify the build locally: web and app.

lib/features/email_recovery/presentation/model/session_extension.dart
```diff
@@ -0,0 +1,37 @@
...
+
+extension SessionExtension on Session {
+
+  String getRestorationHorizonAsString() {
+    return ((props[0] as Map<CapabilityIdentifier, CapabilityProperties>?)
+        ?[capabilityDeletedMessagesVault]
+        ?.props[0] as Map<String, dynamic>?)
+        ?['restorationHorizon']
+        ?? "15 days";
+  }
```
hoangdat, line 16: Can you do use the function `getCapabilityProperties`?

lib/features/email_recovery/presentation/model/session_extension.dart
```diff
@@ -0,0 +1,37 @@
...
+        .replaceAll(" hours", "h")
+        .replaceAll(" hour", "h")
+        .replaceAll(" minutes", "m")
+        .replaceAll(" minute", "m")
+        .replaceAll(" seconds", "s")
+        .replaceAll(" second", "s")
+        .replaceAll(" milliseconds", "ms")
+        .replaceAll(" millisecond", "ms");
+
+    return parseDuration(horizonWithCorrectFormat, separator: ' ');
```
hoangdat, line 31: should have try catch and have fallback value here to prevent wrong format in session and parseDuration error

hoangdat: and more test cases

lib/l10n/intl_messages.arb
florentos17: if i remove it it automatically gets re-added when running `dart run intl_generator:extract_to_arb --output-dir=./lib/l10n lib/main/localizations/app_localizations.dart` so everyone will have to pay attention to re-remove it before committing. should i do that ?

hoangdat (approved):

lib/l10n/intl_messages.arb
hoangdat: no, keep current status

tddang-linagora (approved):

florentos17:
i squashed the fixup commits, it should be ready to merge :)

PR #3289, opened 2024-11-20
[WIP] Delegate cache control from Flutter to browser

tddang-linagora:


chibenwa:
Super nice.

@hoangdat @tddang-linagora when might we get it?

web/index.html
```diff
@@ -104,15 +104,33 @@
...
+          });
+        });
+    } else {
+      _flutter.loader.load({
+        onEntrypointLoaded: async function(engineInitializer) {
+          const appRunner = await engineInitializer.initializeEngine();
+          await appRunner.runApp();
+        }
+      });
+    }
```
hoangdat, line 133: IMO, it is deserved an ADR

tddang-linagora: Added

web/index.html
```diff
@@ -104,15 +104,33 @@
...
-    });
+    if ('serviceWorker' in navigator) {
+      navigator
+        .serviceWorker
+        .getRegistrations()
+        .then(async function(registrations) {
+          try {
+            await Promise.all(registrations.map(function(registration) {
+              return registration.unregister();
+            }));
```
dab246, line 115: If we do so will it cause side effects. If there are some service workers not spawned by flutter will we still destroy it?

tddang-linagora: Yes. So far, only `flutter_service_worker` is registered. Removing all service workers also means if there is any unexpected service worker that is installed, that goes too.

hoangdat: add this to `Consequences` in ADR

hoangdat (approved):

Issue #3290, opened 2024-11-21
[MU] Finner handling of reconnection


chibenwa:
### Description

Today,

```
GIVEN an expired OIDC token
THEN Tmail will prompt me to reconnect before refreshing the page
```

This "sure to reconnect" pop up was added to avoid data loss when composing an email.

While it serves this purpose VERY well it is not really needed and a bit bothering (users complains) when NOT editing an email.

I'd propose to:
 - Keep the current behaviour ONLY when the composer is opened
 - Improve the message: `Your session expired. We need to take you back to the login page in order to refresh it. You might want to save the email you are currently editing before we do so.` or something like that
 - If no composer is opened, let's just refresh.

chibenwa:
Thoughts @guimard  @hoangdat ?

hoangdat:
We added it from #2928.

IMO, it is 50-50 option. Still reasonable for me to aware what happen before going to login page. 

Please make decision, we will change.

chibenwa:
Decision is made.

See above.

Please proceed.

PR #3313, opened 2024-11-27
TF-3290 Only display the reconnect confirmation dialog when the editor is open

dab246:
## Issue

#3290 

### Composer is open

- Click `Save & Refresh ` button:

https://github.com/user-attachments/assets/162c11bc-dd97-4675-947d-d4ddfe024005

- Click `No` button:

https://github.com/user-attachments/assets/3fdf2f25-e3ba-4b42-b608-81d950c2a087

- Click `Cancel` button:

https://github.com/user-attachments/assets/91adcc8b-c91b-4e72-964e-efb785052a7f

### Composer is closed

https://github.com/user-attachments/assets/f6c3a4c1-27d7-4624-bbbd-7ef95ad73514

### DIFF
--- a/lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
+++ b/lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
@@ -413,6 +413,7 @@ class MailboxDashBoardController extends ReloadableController
     } else if (success is GetAllIdentitiesSuccess) {
       _handleGetAllIdentitiesSuccess(success);
     } else if (success is GetComposerCacheSuccess) {
+      _removeComposerCacheOnWeb();
       goToComposer(ComposerArguments.fromSessionStorageBrowser(success.composerCache));
     } else if (success is GetIdentityCacheOnWebSuccess) {
       goToSettings();
@@ -1523,12 +1524,14 @@ class MailboxDashBoardController extends ReloadableController
     composerArguments = arguments;
     ComposerBindings().dependencies();
     composerOverlayState.value = ComposerOverlayState.active;
+    twakeAppManager.openComposerOnWeb();
   }
 
   void closeComposerOverlay({dynamic result}) async {
     composerArguments = null;
     ComposerBindings().dispose();
     composerOverlayState.value = ComposerOverlayState.inActive;
+    twakeAppManager.closeComposerOnWeb();
     if (result is SendingEmailArguments) {
       handleSendEmailAction(result);
     } else if (result is SendEmailSuccess ||
@@ -3035,6 +3038,7 @@ class MailboxDashBoardController extends ReloadableController
     sessionCurrent = null;
     mapMailboxById = {};
     mapDefaultMailboxIdByRole = {};
+    twakeAppManager.closeComposerOnWeb();
     super.onClose();
   }
 }
--- a/lib/l10n/intl_messages.arb
+++ b/lib/l10n/intl_messages.arb
@@ -1,5 +1,5 @@
 {
-  "@@last_modified": "2024-12-09T13:59:58.460897",
+  "@@last_modified": "2024-12-12T15:50:46.911821",
   "initializing_data": "Initializing data...",
   "@initializing_data": {
     "type": "text",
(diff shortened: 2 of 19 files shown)


lib/main.dart
```diff
@@ -38,9 +39,17 @@ Future<void> runTmail() async {
   runApp(const TMailApp());
 }
 
-class TMailApp extends StatelessWidget {
+class TMailApp extends StatefulWidget {
```
tddang-linagora, line 42: What is the purpose of `StatefulWidget` for `TMailApp`?

dab246: Manage object of TwakeAppManager

lib/main.dart
```diff
@@ -74,6 +83,13 @@ class TMailApp extends StatelessWidget {
...
       initialRoute: AppRoutes.home,
-      getPages: AppPages.pages);
+      getPages: AppPages.pages,
+    );
+  }
+
+  @override
+  void dispose() {
+    _twakeAppManager.dispose();
+    super.dispose();
```
tddang-linagora, line 93: `TMailApp` is equal to the whole app. If this method is called, doesn't it mean the app is terminated? Why does we need to dispose this specific dependency manually?

dab246: It is called before the application terminates. We must remove it to avoid unnecessary memory leaks when using the variable.

tddang-linagora (dismissed):

lib/l10n/intl_messages.arb
```diff
@@ -4071,5 +4065,11 @@
     "type": "text",
     "placeholders_order": [],
     "placeholders": {}
+  },
+  "messageWarningDialogWhenExpiredOIDCTokenAndReconnection": "Your session expired. We need to take you back to the login page in order to refresh it. You might want to save the email you are currently editing before we do so.",
```
hoangdat, line 4069: how about 
```suggestion
  "messageWarningDialogWhenExpiredOIDCTokenAndReconnection": "Your session has expired. We’ll try to save your draft before taking you to the login page, but please copy any important content to be safe.",
```

dab246: This is pretty good for user guidance. Updated it.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -401,6 +401,7 @@ class MailboxDashBoardController extends ReloadableController with UserSettingPo
     } else if (success is GetAllIdentitiesSuccess) {
       _handleGetAllIdentitiesSuccess(success);
     } else if (success is GetComposerCacheSuccess) {
+      _removeComposerCacheOnWeb();
```
hoangdat, line 404: Why we need to remove cache here?

lib/main/utils/twake_app_manager.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/model/composer_overlay_state.dart';
+
+class TwakeAppManager {
+  ComposerOverlayState? _composerOverlayState;
```
hoangdat, line 4: why we need this to manage composer state? why not only base on the cache storage to verify it?

dab246: > why we need this to manage composer state? why not only base on the cache storage to verify it?

When we open a new composer, we haven't saved it yet so we can't use cache storage to verify. We can only rely on a variable in memory to trigger it.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
dab246: IMO, we should delete it when successfully reopened from cache so that when the user re-enters the url from the browser address bar, the composer can be closed. Of course, the cases of `reload`, or `badCredential` still save the composer to cache.


https://github.com/user-attachments/assets/b1f25fd4-7e16-4dec-8d54-f8f936fba7e4

tddang-linagora (approved):

tddang-linagora (dismissed):

tddang-linagora (approved):

hoangdat (approved):

PR #3328, opened 2024-12-09
Fix patrol test on Firebase Test Lab

tddang-linagora:
## Root cause
- Current `Patrol` version is `3.11.1`, fixing the issue of white space in test name for androidx test orchestrator 1.5.0 ([Patrol changelogs](https://pub.dev/packages/patrol/changelog))
- Androidx test orchestrator released version 1.5.1, which is to fix the same issue
- With `Patrol: 3.11.1`, our test cases now crash with https://github.com/leancodepl/patrol/issues/1725

## Solution
Revert `Patrol` back 1 patch to `3.11.0`, reverting the change `Patrol` tried to fix with Androidx test orchestrator 1.5.0

## Before
Currently we have 6 test cases, but only 2 is ran. The test crashed so the other 4 is discarded.

Screenshot 2024-12-09 at 11 03 46

## After
All 6 test cases is ran.

Screenshot 2024-12-09 at 11 03 58

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -100,5 +100,5 @@ dependencies {
     implementation 'androidx.work:work-runtime-ktx:2.7.0'
     implementation 'com.android.support:multidex:1.0.3'
     implementation 'androidx.window:window:1.0.0'
-    androidTestUtil "androidx.test:orchestrator:1.5.0"
+    androidTestUtil "androidx.test:orchestrator:1.5.1"
 }
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1525,10 +1525,10 @@ packages:
     dependency: "direct dev"
     description:
       name: patrol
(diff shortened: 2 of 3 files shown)


android/app/build.gradle
```diff
@@ -100,5 +100,5 @@ dependencies {
     implementation 'androidx.work:work-runtime-ktx:2.7.0'
     implementation 'com.android.support:multidex:1.0.3'
     implementation 'androidx.window:window:1.0.0'
-    androidTestUtil "androidx.test:orchestrator:1.5.0"
+    androidTestUtil "androidx.test:orchestrator:1.5.1"
```
dab246, line 103: try use latest version

tddang-linagora: Could not resolve all files for configuration ':app:androidTestUtil'.
Could not find androidx.test:orchestrator:latest.

dab246 (approved):
lgtm

hoangdat (approved):

PR #3352, opened 2024-12-16
Disable language tool check for text field

dab246:
## Issue 

- Language tool spellcheck calls LanguageTool API too many times. When changing composer subject

## Solution

- Disable language tool spellcheck for text field

## Resolved

https://github.com/user-attachments/assets/817a4011-1e04-47cb-8b9b-4e512005c424

### DIFF
--- a/contact/pubspec.lock
+++ b/contact/pubspec.lock
@@ -688,15 +688,6 @@ packages:
       url: "https://pub.dev"
     source: hosted
     version: "6.6.1"
-  languagetool_textfield:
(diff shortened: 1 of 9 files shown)


hoangdat (approved):

chibenwa:
> Language tool spellcheck calls LanguageTool API too many times. When changing composer subject

Really, this is **not the problem here**.

THe problem is **leaking user dats unencrypted outside of our infra** to people with dubitous terms of service!

Issue #3349, opened 2024-12-13
[MU] Forwarded message disclaimer and from address


chibenwa:
### Description

![image](https://github.com/user-attachments/assets/19e857a3-6567-425f-87b8-08ce789d52d3)

Should be

```
----- Forwarded message ----
Subject: Abc Def
Date: xxx
From: Madina <madina@govmu.org>

...
```

(Always the addess in <>)

Note that it seems not to happen all the time

![image](https://github.com/user-attachments/assets/7603a1b8-e4e3-4c7d-b2ae-7907558467ee)

dab246:
## DoD
- [ ] Add `To` field
- [ ] Format address: `Name <emailaddress>`

hoangdat:
included in 0.14.4

PR #3354, opened 2024-12-16
TF-3349 Fix [MU] forwarded message disclaimer and from address

dab246:
## Issue

#3349

## Root cause

- We use the characters `<` and `>` in the html content so the browser treats them as HTML tags.

Screenshot 2024-12-16 at 15 24 45

## Solution

- Use `&lt;` and `&gt;` for the angle brackets to prevent browsers from interpreting them as HTML tags.

## Resolved

https://github.com/user-attachments/assets/f6dd69f2-04fd-4ed2-91d8-2d35625d5e66


tddang-linagora (dismissed):

model/lib/extensions/email_address_extension.dart
```diff
@@ -27,6 +27,17 @@ extension EmailAddressExtension on EmailAddress {
     return '';
   }
 
+  String asFullHtmlString() {
```
hoangdat, line 30: write unit test for this

hoangdat: Widget test to make sure HTML content contain enough: `Subject, From, To, Cc, Bcc, Reply to`

tddang-linagora (dismissed):

hoangdat:
- should we handle `HeaderEmailQuoted` with transforming email content?

model/lib/extensions/email_address_extension.dart
dab246: done

dab246:
> * should we handle `HeaderEmailQuoted` with transforming email content?

Done


https://github.com/user-attachments/assets/7174c037-445f-4baa-ab60-7d93302899ec

dab246:
> Rather no: the email subject displayed should be `<script>alert('abc def')</script> ghi klm` (correctly escaped)

Good catch. We will use escape html to fix this.



https://github.com/user-attachments/assets/ffa40bb1-9af2-4b9e-ac1a-5770b65d2059

dab246:
- [x] Also sanitize email subject when Print email

 

https://github.com/user-attachments/assets/f340fa58-afd6-4b04-81c1-7c1cb8f163cf

model/lib/extensions/list_email_address_extension.dart
```diff
@@ -27,6 +27,18 @@ extension SetEmailAddressExtension on Set<EmailAddress>? {
...
+    if (this?.isNotEmpty != true) return '';
+
+    final listEmail = this
+      !.map((emailAddress) => emailAddress.asFullHtmlString())
+      .toList();
+
+    return listEmail.isNotEmpty ? listEmail.join(separator) : '';
+  }
+
+  String asFullHtmlStringUseCommaSeparator() => asFullHtmlString(', ');
```
hoangdat, line 40: if list have length = 1, we don't need the separator.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -578,169 +579,184 @@ class ComposerController extends BaseController
...
-      composerArguments.value = arguments;
 
-      _initIdentities(arguments);
+    if (arguments is! ComposerArguments) return;
 
-      injectAutoCompleteBindings(
-        mailboxDashBoardController.sessionCurrent,
-        mailboxDashBoardController.accountId.value
+    if (arguments.emailActionType == EmailActionType.forward) {
+      final newEmailSubject = await sanitizeEmailSubject(
```
hoangdat, line 591: Asynchronous calling here still safe for us in building composer view?

hoangdat: I still prefer using synchronize calling `HtmlEcscape().convert(headerQuotedContent)`. Benefit:
- not touch to the mechanism of composer loading, especially long rich content, inline images, ...
- handle html escape for content of `reply`, `replyAll`, `forward`

Keep transformHtml with escapeTransformer handler for printEmail, and other stuffs.

lib/features/composer/presentation/composer_bindings.dart
tddang-linagora, line 1: Please resolve GitHub Actions analysis

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: If using `TransformHtmlEmailContentInteractor` is still desired, prefer wrapping `TransformHtmlEmailContentInteractor` in another interactor for `subject` alone, and use `handleSuccessViewState` and `handleFailureViewState` to handle result, rather than blocking the `_initEmail` call with `async-await`

model/lib/extensions/list_email_address_extension.dart
dab246: The `join` function takes care of this.

Screenshot 2024-12-18 at 02 02 00

lib/features/composer/presentation/composer_controller.dart
dab246: done

dab246: https://github.com/user-attachments/assets/7a45049b-ca1b-4f89-8290-d5a30c5d1c4b

lib/features/composer/presentation/composer_bindings.dart
dab246: No problems found.

hoangdat:
please also cherry pick to `maintenance-v0.11.4003`

dab246:
> please also cherry pick to `maintenance-v0.11.4003`

Implemented at https://github.com/linagora/tmail-flutter/pull/3360

tddang-linagora (approved):

hoangdat (approved):

Issue #3372, opened 2024-12-20
[MU] Emptying trash: Many unnecessary Email/query + Email/get requests


chibenwa:
### Description

WHEN I empty my trash with 180 messages in it

THEN Tmail...
 - Performs an initial email/query + email/get operation to catch the emails
 - Then Email/set destroy a batch of mails
 - Then resyncs too much on websocket notifications (not the issue here)
 - Then issue hundreds of Email/query + Email/get that actually returns empty payload <= that 's the issue here.
 
![Screenshot from 2024-12-20 11-32-17](https://github.com/user-attachments/assets/c2fcdae2-cadf-40e5-9770-07021fa66355)

Why are those requests made?

### Expected result

Avoid needless requests when emptying trash...

chibenwa:
(is it possible that resynch on an empty mailbox triggers an aggressive Email/query+get ) ?

If so having a debounce on this behaviour would be nice!

hoangdat:
- [ ] Replace `Email/query + Email/get` + `Email/set` with `Email/query + Email/set`

chibenwa:
> Replace Email/query + Email/get + Email/set with Email/query + Email/set

+1 but notDestroyed setError results will need fine tuning...

If not found treat the message as deleted, otherwize do considère deletion failed.

chibenwa:
(this multiple entry issue is temporary and will disapear with multiple composer so is temporary.
A cache that handles not gracefully multiple entries is better than no cache at all...
Si please do not over engineer the multiple cache entry issue...)

Issue #3369, opened 2024-12-20
[MU] CID email not displayed for 3D link


chibenwa:
### Description

The following mail do not display well:

![image](https://github.com/user-attachments/assets/eec9891a-bc30-4a4d-9d23-20bfc2d10e18)

[Greeting Card.eml.txt](https://github.com/user-attachments/files/18208927/Greeting.Card.eml.txt)

### Expected result

Displaying the image

(Like in TB)

![image](https://github.com/user-attachments/assets/dd0518e6-a899-403e-97b9-d4e9700a4611)

Note that I fully agree keeping it as an attachment too, but we need to display the inlined image.

### Technical details

Email was sent using Outlook 15.0

Failing image:

```
<img =
width=3D606 height=3D426 id=3D"Picture_x0020_1" =
src=3D"cid:image001.png@01DB5155.249583D0">
```

The corresponding part have a CID but no content-disposition...

Issue #3347, opened 2024-12-13
Teams, Zoom, Google Classroom Online meeting links


hoangdat:
### Desc:

When a user is invited to join an online meeting, they do not receive the invitation link that allows them to easily click and access the meeting. Instead, the user is required to manually add the invitation link and password to join.

![image](https://github.com/user-attachments/assets/76d9fe80-d0a2-4f64-b24b-39bd8e8ffe1a)

- meeting link is not in good html
```
{
    "sessionState": "2c9f1b12-b35a-43e6-9af2-0106fb53a943",
    "methodResponses": [
        [
            "CalendarEvent/parse",
            {
                "accountId": "94db052675894262fb8644f9517aaf25e8246c5c2eae08b34da40b748f30ec0b",
                "parsed": {
                    "71c6cd30-b924-11ef-b3d3-277cfcdad3a2_4": [
                        {
                            "uid": "040000008200E00074C5B7101A82E00800000000217BC829314DDB01000000000000000010000000D1A864BBBDEDB14AB118071FC859B1A9",
                            "title": "test body of invitation",
                            "description": "\n\n.........................................................................................................................................\nJoin online meeting<https://join.skype.com/qs7qMduufFjY>\n.........................................................................................................................................\n",
                            "start": "2024-12-13T16:30:00",
                            "end": "2024-12-13T17:00:00",
                            "utcStart": "2024-12-13T09:30:00Z",
                            "utcEnd": "2024-12-13T10:00:00Z",
                            "timeZone": "Asia/Bangkok",
                            "location": "",
                            "method": "REQUEST",
                            "sequence": 0,
                            "priority": 5,
                            "freeBusyStatus": "busy",
                            "status": "confirmed",
                            "privacy": "public",
                            "organizer": {
                                "name": "Dat PHAM",
                                "mailto": "dphamhoang@outlook.com"
                            },
                            "participants": [
                                {
                                    "name": "dphamhoang@linagora.com",
                                    "mailto": "dphamhoang@linagora.com",
                                    "role": "REQ-PARTICIPANT",
                                    "participationStatus": "NEEDS-ACTION",
                                    "expectReply": true
                                },
                                {
                                    "name": "tdvu@linagora.com",
                                    "mailto": "tdvu@linagora.com",
                                    "role": "REQ-PARTICIPANT",
                                    "participationStatus": "NEEDS-ACTION",
                                    "expectReply": true
                                },
                                {
                                    "name": "tddang@linagora.com",
[...]

Issue #3341, opened 2024-12-11
Turn on Web Socket mobile


hoangdat:
### Desc:
- Currently, mobile app mostly base on FCM to update the app in both background and foreground
- We should apply the same mechanism for all platform: web socket for foreground
  with this, we will have some advantage:
  - maintain the same logic for all platform
  - reduce much JMAP methods  

### DoD:
- [x] make sure web socket in use when mobile app running in foreground. 
- [x] FCM still available in background/terminated 
- [x] integration test for web socket in case app in foreground: do an action in chat list, verify chat list update

PR #3381, opened 2024-12-26
Bump version to v0.14.5

hoangdat:
### Fixed 
- #3336 Make Echo ping of web socket optional
- #3347 Format Calendar event description
- #3341 Enable Web socket for mobile in foreground
- #3334 Remove all own resynch after any actions in mailbox, email
- #3333 Remove Mailbox/query for spam banner when reload app
- #3332 Remove resynch when switching mailbox
- #3372 Remove Mailbox resynch when Emptying trash
- #3369 Display CID without disposition as an attachment

### Added
- #3181 Contact support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,18 @@
-## [0.14.3] - 2024-12-18
+## [0.14.5] - 2024-12-26
+### Fixed 
+- #3336 Make Echo ping of web socket optional
+- #3347 Format Calendar event description
+- #3341 Enable Web socket for mobile in foreground
+- #3334 Remove all own resynch after any actions in mailbox, email
+- #3333 Remove Mailbox/query for spam banner when reload app
+- #3332 Remove resynch when switching mailbox
+- #3372 Remove Mailbox resynch when Emptying trash
+- #3369 Display CID without disposition as an attachment
+
+### Added
+- #3181 Contact support 
+
+## [0.14.4] - 2024-12-18
 ### Fixed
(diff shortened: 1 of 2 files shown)


chibenwa:
Is it ready for me deploying it on oncommit-tmail.linagora.com ?

If so, i will do it tonight.

chibenwa (approved):

hoangdat:
I still testing it, then build it. I will let you know when it is available

tddang-linagora:
Memory usage check:
- Move emails: OK

https://github.com/user-attachments/assets/99e04261-3619-465b-8d98-733bf59d89be

- Mark emails: OK

https://github.com/user-attachments/assets/9b6944c6-0bff-4d0e-baf7-dcc1eb4589c9

chibenwa:
As explained on the chat I am unhappy with the way it works.

We lost a loooot in user experience: the actions done by the users takes a LONG time to reflect 
(Only upon websocket notification)

If you drop the websocket connection
(eg when I did also upgrade the back...) 
Then the list not actualized

As it stands, this can't be shipped to a customer without pain.

We need to look into the Email/set RESULT

To see which emails where update it
And update the caches / view accordingly straight away
That way user actions are SYNCHRONOUS and are reflected as soon as the JMAP server answers.

Is it clear?

chibenwa:
Empty trash is completly broken on my Linagora account

 -> It fails deleting all mails
 -> I do not have a progess bar

😑

Issue #3396, opened 2025-01-06
Mobile app can not get the latest email when open app from terminated - 0.14.5


hoangdat:
### Desc:
- Mobile app can not get the latest email when open app from terminated
- Maybe web app also have the same problem

### DoD:
- when open app, app need to show up-to-date data

PR #3400, opened 2025-01-06
TF-3396 Fix mobile app can not get the latest email when open app from terminated

dab246:
## Issue

#3396 

## Root cause

- Since we skipped get change when switching between mailboxes, we didn't get the latest changes when the app was in terminated state.

## Solution

- To solve the above problem we have to call get change (mailbox and email) every time we enter the app. This get change is only called when the `state` of mailbox and email from push notification (`FCM, Websocket`) is saved in cache (it has been processed when we do push notification).

## Resolved

https://github.com/user-attachments/assets/6103d5cc-e424-4018-a8da-d7144e4b6c47


chibenwa:
@hoangdat any chance to get this timely?

tddang-linagora (approved):

hoangdat:
> @hoangdat any chance to get this timely?

ASAP, but not have much value for web app.

dab246:
> ASAP, but not have much value for web app.
>

I have updated the most effective solution, it works well on both mobile and web.

tddang-linagora (approved):

hoangdat (approved):

Issue #3323, opened 2024-12-05
Handle corner cases of web socket disconnected


hoangdat:
#### Case 1
1. User still keep the web app in browser's tab
2. User close the laptop
3. After more than 10 hours, maybe `refreshToken` still alive 
4. Open laptop -> open app
5. Socket connection of web app was closed, console log showed that it retried to exceed maximum counter but have CORS error
6. web app still refresh token when web app opened 

### Case 2
1. Web app open
2. Network disconnected
3. Web socket retried failed
4. Then network comeback
5. Web socket still not re connect 

### DoD:
- [ ] make sure that web socket connection can be reconnected in these case

tddang-linagora:
`tmail.linagora.com` log after computer sleeps for a day

![Screenshot 2024-12-05 at 09 04 56](https://github.com/user-attachments/assets/77cf7b34-e182-4da7-bdaf-4252216410c5)

hoangdat:
- test with mail.lin-saas.dev

PR #3417, opened 2025-01-09
TF-3323 Fix web socket still not reconnect when network disconnected

dab246:
## Issue

#3323 

https://github.com/user-attachments/assets/2c888ab8-64c5-4f4d-9f25-ec501bfeb3b7

## Root cause

- Because when we disconnect the network, the web socket does not receive any data to maintain the connection.
- If we enable `WS_ECHO_PING=true` then the web socket connection is still maintained continuously. But when we disable it, the web socket data is not maintained and synchronized to the client side.

## Solution

- To solve this problem, we will listen for changes in network connectivity and changes in the state of the application (`previously deployed`). Whenever we lose network connectivity, we close the clean up web socket and reopen the connection when the network is available.

## Resolved

https://github.com/user-attachments/assets/8e7c2c7f-0dde-4b46-a544-edf2ff59b0a2

### DIFF
--- a/lib/features/push_notification/presentation/controller/web_socket_controller.dart
+++ b/lib/features/push_notification/presentation/controller/web_socket_controller.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 
+import 'package:connectivity_plus/connectivity_plus.dart';
 import 'package:core/presentation/state/failure.dart';
 import 'package:core/presentation/state/success.dart';
 import 'package:core/utils/app_logger.dart';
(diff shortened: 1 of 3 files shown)


tddang-linagora (approved):

chibenwa (approved):

dab246:
- In case:

```
Open app on browser tab.
Keep it like that to automatically put the device to sleep or fold it in.
After about 15 minutes (maybe longer). We turn the machine back on.
Perform any action with email
See if there is any change from the web socket?
```

Tested on `lin-saas.dev` and web socket reconnects automatically. 
Everything is working fine with this PR.

lib/features/push_notification/presentation/controller/web_socket_controller.dart
```diff
@@ -239,4 +251,19 @@ class WebSocketController extends PushBaseController {
...
+    _networkConnectionController = getBinding<NetworkConnectionController>();
+    _connectivitySubscription = _networkConnectionController
+      ?.connectivity
+      .onConnectivityChanged.listen((status) {
+        if (status == ConnectivityResult.none) {
+          log('WebSocketController::_monitorNetwork:No network connection');
+          _cleanUpWebSocketResources();
+        } else {
+          log('WebSocketController::_monitorNetwork:Network connection restore');
+          _connectWebSocket();
```
hoangdat, line 265: Please verify memory leak after we turn off/on network n times

dab246: No memory leak when continuously turning on/off the network.


https://github.com/user-attachments/assets/34dc9781-3d7c-4c81-ba8e-7f4af545b45f

hoangdat (approved):

PR #3428, opened 2025-01-13
Small improvements

florentos17:
i suggest here two small improvements for tmail-frontend:

1. on the login page, sometimes i was getting a frustrating `unknown error occurred, please try again`. upon inspecting the logs, the error was actually known and had a helpful descriptive message. i think it would be better to display that message rather than leaving it in the logs ? i corrected that in one line.
2. even when working with a server that does not support the `recover deleted messages` feature (eg, the capability `deletedMessagesVault` `com:linagora:params:jmap:messages:vault`), the option still appears in the `trash` mailbox actions. i corrected the bug by adding a little `if` statement.

### DIFF
--- a/lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
+++ b/lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
@@ -13,6 +13,7 @@ import 'package:tmail_ui_user/features/mailbox/presentation/mailbox_controller.d
 import 'package:tmail_ui_user/features/mailbox/presentation/model/context_item_mailbox_action.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_actions.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_categories.dart';
+import 'package:tmail_ui_user/features/mailbox/presentation/utils/mailbox_utils.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/widgets/mailbox_bottom_sheet_action_tile_builder.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/widgets/app_dashboard/app_list_dashboard_item.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
@@ -27,7 +28,8 @@ mixin MailboxWidgetMixin {
 
   List<MailboxActions> _listActionForDefaultMailbox(
     PresentationMailbox mailbox,
-    bool spamReportEnabled
+    bool spamReportEnabled,
+    bool deletedMessageVaultSupported
   ) {
 
     return [
@@ -38,7 +40,8 @@ mixin MailboxWidgetMixin {
       if (mailbox.isTrash)
         ...[
           MailboxActions.emptyTrash,
-          MailboxActions.recoverDeletedMessages
+          if (deletedMessageVaultSupported)
+            MailboxActions.recoverDeletedMessages,
         ]
       else if (mailbox.isSpam)
         ...[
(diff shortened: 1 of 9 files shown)


lib/main/utils/toast_manager.dart
```diff
@@ -46,7 +47,7 @@ class ToastManager {
     } else if (exception is NoNetworkError) {
       return AppLocalizations.of(context).youAreOffline;
     } else {
-      return null;
+      return exception.message;
```
chibenwa, line 50: +1

Though the message won't be localized (french user will get a message in english I believe), it is better than nothing.

florentos17:
some checks were failing so i rebased on latest master

lib/main/utils/toast_manager.dart
hoangdat: Agree with Benoit. But please don't make our Toast too long, maximum 2 lines is ok. We should add action `Details` to make user can expand in this case.

lib/features/mailbox/presentation/mixin/mailbox_widget_mixin.dart
```diff
@@ -285,6 +300,13 @@ mixin MailboxWidgetMixin {
         ?? false;
   }
 
+  static bool isDeletedMessageVaultSupported(Session? session, AccountId? accountId) {
```
hoangdat, line 303: If you prefer `static` method, should move it to `*Utils` class.

florentos17: done

lib/main/utils/toast_manager.dart
florentos17: done !

florentos17:
@hoangdat do you have any more comments about this work ?

chibenwa (approved):

lib/main/utils/toast_manager.dart
```diff
@@ -45,8 +47,14 @@ class ToastManager {
...
       return AppLocalizations.of(context).youAreOffline;
+    } else if (exception is HandshakeException) {
+      return AppLocalizations.of(context).handshakeException(exception.osError?.message ?? '');
     } else {
-      return null;
+      try {
+        return AppLocalizations.of(context).unexpectedError(exception.message);
+      } catch(_) {
+        return AppLocalizations.of(context).unexpectedError(exception.toString());
+      }
```
florentos17, line 57: there was a mistake in my previous commit: `exception` being dynamic, it does not necessarily have a `message` attribute.
i had this with a `CanRetryOIDCException` that had no message and was "hiding" a `HandshakeException`.

florentos17:
following [this comment](https://github.com/linagora/tmail-flutter/discussions/3435#discussioncomment-12008196), i made the error handling a bit clearer:
- by providing `https://api.fastmail.com/jmap/session`, the error `Operation Unauthorized` is printed
- by providing `https://api.fastmail.com`, the error `checkOIDCIsAvailable(): Unexpected character in server response` is printed (HTML instead of the expected JSON)

florentos17:
i just cleaned the commit history.

to wrap up, this PR has fixes to:
- display an error message upon login instead of `Unknown error`, along with better exception handling ;
- hide some options and settings if not supported by the server

lib/features/mailbox/presentation/utils/mailbox_utils.dart
```diff
@@ -50,4 +57,25 @@ class MailboxUtils {
...
+      return false;
+    }
+    if (!CapabilityIdentifier.jmapTeamMailboxes.isSupported(session, accountId)) {
+      return false;
+    }
+
+    return (session.getCapabilityProperties(accountId, CapabilityIdentifier.jmapTeamMailboxes)
+        ?.props[0] as Map<String, dynamic>?)
+    ?[subaddressingSupported]
+        ?? false;
```
tddang-linagora, line 72: Indent

florentos17:
Is it ok if I force push here to clean the 37 commits that don't belong to this PR ?

Then we will see more clearly the changes from this PR, but every comment has been addressed so I think this work is pretty much done ?

tddang-linagora (approved):

dab246 (approved):
lgtm

hoangdat (approved):

Issue #3345, opened 2024-12-11
Sort order blinks when deleting emails in search


chibenwa:
tmail.linagora.com

https://github.com/user-attachments/assets/74091fab-7ab9-4d81-bce7-acb58fdc2dc6

Sort order blinks when deleting emails in search

chibenwa:
No longer the case

PR #3437, opened 2025-01-17
TF-3345 Fix sort order blinks when deleting emails in search

dab246:
## Issue

#3345

## Reproduce

https://github.com/user-attachments/assets/04455b15-7583-4741-9efc-ad74710a3557

## Root cause

- Because the `before` date time is null when sort order by sender every time user refresh change email

## Resolved

https://github.com/user-attachments/assets/05768aa7-95c9-4e52-90d0-85f61becf6ee

### DIFF
--- a/lib/features/thread/presentation/thread_controller.dart
+++ b/lib/features/thread/presentation/thread_controller.dart
@@ -666,7 +666,9 @@ class ThreadController extends BaseController with EmailActionController {
         _searchEmailFilter.sortOrderType.isScrollByPosition(),
         0,
       ),
-      beforeOption: const None(),
+      beforeOption: !_searchEmailFilter.sortOrderType.isScrollByPosition()
+          ? const None()
+          : null,
     );
     final searchViewState = await _searchEmailInteractor.execute(
       _session!,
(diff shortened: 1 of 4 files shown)


tddang-linagora (approved):

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -666,7 +666,9 @@ class ThreadController extends BaseController with EmailActionController {
         _searchEmailFilter.sortOrderType.isScrollByPosition(),
         0,
       ),
-      beforeOption: const None(),
+      beforeOption: !_searchEmailFilter.sortOrderType.isScrollByPosition()
+          ? const None()
```
hoangdat, line 670: not a good idea in case: 
- user search in a range of time, but when it refreshes, `before` is not applied

dab246: @hoangdat  Please attach error video here.

hoangdat: Can we verify `SearchEmailFilter` in some cases:
- load more: scroll by position and not by position (before filter?)
- refresh: scroll by position and not by position (before filter?)

hoangdat (approved):

Issue #2305, opened 2023-11-02
Support Cyrus server


chibenwa:
### Notes from make better email 2023

The cyrus test server repo is:

https://github.com/cyrusimap/cyrus-docker-test-server

There's a README in there with a bit more detail, but here's what I did to get tmail working today:

```
brong@elg:~$ cat ~/env.file
SERVER_URL=http://localhost:8080/     
```    

in one screen:

```
sudo docker run -it -p 8002:80 -v ~/env.file:/usr/share/nginx/html/assets/env.file chibenwa/tmail-web:fix-cyrus
```

and in another screen:

```
sudo docker run -it -p 8080:8080 -p 8110:8110 -p 8143:8143 -p 8001:8001 ghcr.io/cyrusimap/cyrus-docker-test-server:latest
```    

And then on Windows, I had to run Chrome with CORS disabled:

```
/"Program Files"/Google/Chrome/Application/chrome.exe --user-data-dir="C://chrome-dev-disabled-security" --disable-web-security --disable-site-isolation-trials
```

And then I could just log in at [http://localhost:8002/ ](http://localhost:8002/)and use any of the usernames user1 through user5 to log in!  Any password that doesn't contain the magic string bad is valid, e.g. 'x'.

If you want a user with a couple of emails in it (just the Fastmail welcome emails) you can run:

```
brong@elg:~/src/cyrus-docker-test-server$ curl -T examples/userdata.json http://localhost:8001/hello     
``` 

Replace "hello" with whatever username you want to create.

You can also download the JSON state from that same URL.  This is the weird format that Apple said they use for their tests!

You can log in via POP3 on port 8110 or IMAP on port 8143 and see the same data, or via JMAP on port 8080 of course!

NEXT STEPS:

 - fix and re-enable VacationResponse support
 - add Identity support in Cyrus
 - add EmailSubmission support (set up the docker image to allow sending emails between users)

Enjoy,

Bron.

chibenwa:
CF https://github.com/linagora/james-project-private/issues/943

chibenwa:
Supported ;-)

Issue #1606, opened 2023-03-15
Session: handle relative URL


chibenwa:
### Description

Needed for easy Cyrus support.

Cyrus exposes only relative URL into the JMAP session object.

![download](https://user-images.githubusercontent.com/6928740/225282337-a6aedfc1-4f84-47be-9998-0e2fe7f8fa0e.png)

### Definition of done

Given a relative path, eg apiUrl -> `/jmap/` then prepend the JMAP base URL (so given a JMAP URL `https://jmap.whatever.tld` the apiUrl would become `https://jmap.whatever.tld/jmap`.

Same thing for all URLs appearing in the session option (upload, download, websockets, eventSource, etc...)

This is needed by the OPS team for a POC to sell TeamMail over Cyrus :muscle: :muscle: :muscle: 

Cc @ducnm0711 @guimard

dab246:
Where can we test it?

dab246:
On which server can we test? `Cyrus` ?

chibenwa:
Yes

chibenwa:
@ducnm0711 share the docker compose here please

hoangdat:
Please dont forget the current support for Fastmail. We should handle the relative and absolute cases also.

PR #3457, opened 2025-01-30
Support cyrus server

florentos17:
tackling [issue 2305](https://github.com/linagora/tmail-flutter/issues/2305)

each commit has a self-explanatory message, but do not hesitate if you have any question !

i have tested the changes with both tmail-backend and cyrus to check that the expected behaviour happens, and it looks all good to me.

### DIFF
--- a/core/lib/presentation/extensions/uri_extension.dart
+++ b/core/lib/presentation/extensions/uri_extension.dart
@@ -5,23 +5,20 @@ extension URIExtension on Uri {
 
   Uri toQualifiedUrl({required Uri baseUrl}) {
     log('SessionUtils::toQualifiedUrl():baseUrl: $baseUrl | sourceUrl: $this');
-    if (toString().startsWith(baseUrl.toString())) {
-      final qualifiedUrl = toString().removeLastSlashOfUrl();
+    if (hasOrigin) {
+      final qualifiedUrl = toString();
       log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
       return Uri.parse(qualifiedUrl);
+    } else if (toString().isEmpty) {
+      log('SessionUtils::toQualifiedUrl():qualifiedUrl: $baseUrl');
+      return baseUrl;
     } else {
-      if (!hasOrigin) {
-        final baseUrlValid = baseUrl.toString().removeLastSlashOfUrl();
-        final sourceUrlValid = toString().addFirstSlashOfUrl().removeLastSlashOfUrl();
-        log('SessionUtils::toQualifiedUrl():baseUrlValid: $baseUrlValid | sourceUrlValid: $sourceUrlValid');
-        final qualifiedUrl = baseUrlValid + sourceUrlValid;
-        log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
-        return Uri.parse(qualifiedUrl);
-      } else {
-        final qualifiedUrl = toString().removeLastSlashOfUrl();
-        log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
-        return Uri.parse(qualifiedUrl);
-      }
+      final baseUrlValid = baseUrl.toString().removeLastSlashOfUrl();
+      final sourceUrlValid = toString().addFirstSlashOfUrl();
+      log('SessionUtils::toQualifiedUrl():baseUrlValid: $baseUrlValid | sourceUrlValid: $sourceUrlValid');
+      final qualifiedUrl = baseUrlValid + sourceUrlValid;
+      log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
+      return Uri.parse(qualifiedUrl);
     }
   }
 
--- a/core/test/presentation/extensions/uri_extension_test.dart
+++ b/core/test/presentation/extensions/uri_extension_test.dart
@@ -16,11 +16,21 @@ void main() {
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
     });
 
+    test('convertToQualifiedUrl() should return qualified url when baseUrl is `https://domain.com` and sourceUrl is `https://domain.com/jmap/`', () async {
+      final baseUrl = Uri.parse('https://domain.com');
+      final sourceUrl = Uri.parse('https://domain.com/jmap/');
+
+      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap/');
+      final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
+
+      expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
+    });
+
     test('convertToQualifiedUrl() should return qualified url when baseUrl is `https://domain.com` and sourceUrl is `/jmap/`', () async {
       final baseUrl = Uri.parse('https://domain.com');
       final sourceUrl = Uri.parse('/jmap/');
 
-      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap');
+      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap/');
       final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
 
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
@@ -50,7 +60,7 @@ void main() {
       final baseUrl = Uri.parse('https://domain.com/jmap');
       final sourceUrl = Uri.parse('/');
 
-      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap');
+      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap/');
       final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
 
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
@@ -70,7 +80,7 @@ void main() {
       final baseUrl = Uri.parse('https://domain.com:2000/jmap');
       final sourceUrl = Uri.parse('https://domain.com:2001/jmap/');
 
-      final qualifiedUrlExpected = Uri.parse('https://domain.com:2001/jmap');
+      final qualifiedUrlExpected = Uri.parse('https://domain.com:2001/jmap/');
       final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
 
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
@@ -86,11 +96,21 @@ void main() {
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
     });
 
-    test('convertToQualifiedUrl() should return qualified url when baseUrl is localhost and sourceUrl is `https://domain.com:2001/basiauth/jmap`', () async {
+    test('convertToQualifiedUrl() should return qualified url when baseUrl is localhost and sourceUrl is `https://domain.com:2001/basicauth/jmap`', () async {
       final baseUrl = Uri.parse('https://localhost:9080/basicauth');
       final sourceUrl = Uri.parse('https://domain.com:2001/basicauth/jmap');
 
       final qualifiedUrlExpected = Uri.parse('https://domain.com:2001/basicauth/jmap');
+      final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
+
+      expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
+    });
+
+    test('convertToQualifiedUrl() should return qualified url when baseUrl is localhost and sourceUrl is `https://domain.com:2001/basicauth/jmap/`', () async {
+      final baseUrl = Uri.parse('https://localhost:9080/basicauth');
+      final sourceUrl = Uri.parse('https://domain.com:2001/basicauth/jmap/');
+
+      final qualifiedUrlExpected = Uri.parse('https://domain.com:2001/basicauth/jmap/');
       final qualifiedUrlResult = sourceUrl.toQualifiedUrl(baseUrl: baseUrl);
 
       expect(qualifiedUrlResult, equals(qualifiedUrlExpected));
--- a/integration_test/utils/scenario_utils_mixin.dart
+++ b/integration_test/utils/scenario_utils_mixin.dart
@@ -2,6 +2,7 @@ import 'dart:async';
 import 'dart:io' hide HttpClient;
 
 import 'package:collection/collection.dart';
+import 'package:core/utils/app_logger.dart';
 import 'package:get/get.dart';
 import 'package:jmap_dart_client/http/http_client.dart';
 import 'package:jmap_dart_client/jmap/core/capability/capability_identifier.dart';
@@ -142,33 +143,39 @@ mixin ScenarioUtilsMixin {
         filePath: path,
         fileSize: await file.length(),
       );
-      final uploadUri = mailboxDashBoardController.sessionCurrent!.getUploadUri(
-        mailboxDashBoardController.accountId.value!,
-        jmapUrl: mailboxDashBoardController.dynamicUrlInterceptors.jmapUrl,
-      );
-      
-      final uploadAttachmentInteractor = Get.find<UploadAttachmentInteractor>();
-      final uploadAttachmentState = await uploadAttachmentInteractor
-        .execute(fileInfo, uploadUri)
-        .last;
-      final attachment = await uploadAttachmentState.fold(
-        (failure) => null,
-        (success) async {
-          if (success is UploadAttachmentSuccess) {
-            final uploadAttachment = await success.uploadAttachment.progressState.last;
-            return uploadAttachment.fold(
-              (failure) => null,
-              (success) {
-                if (success is! SuccessAttachmentUploadState) return null;
 
-                return success.attachment;
-              },
-            );
-          }
-        },
-      );
-      if (attachment != null) {
-        attachments.add(attachment);
+      try {
+        final uploadUri = mailboxDashBoardController.sessionCurrent!.getUploadUri(
+          mailboxDashBoardController.accountId.value!,
+          jmapUrl: mailboxDashBoardController.dynamicUrlInterceptors.jmapUrl,
+        );
+
+        final uploadAttachmentInteractor = Get.find<UploadAttachmentInteractor>();
+        final uploadAttachmentState = await uploadAttachmentInteractor
+          .execute(fileInfo, uploadUri)
+          .last;
+        final attachment = await uploadAttachmentState.fold(
+          (failure) => null,
+          (success) async {
+            if (success is UploadAttachmentSuccess) {
+              final uploadAttachment = await success.uploadAttachment.progressState.last;
+              return uploadAttachment.fold(
+                (failure) => null,
+                (success) {
+                  if (success is! SuccessAttachmentUploadState) return null;
+
+                  return success.attachment;
+                },
+              );
+            }
+          },
+        );
+        if (attachment != null) {
+          attachments.add(attachment);
+        }
+      } catch (e) {
+        logError('ScenarioUtilsMixin::uploadAttachments(): $e');
+        return attachments;
       }
     }
 
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -44,6 +44,7 @@ import 'package:tmail_ui_user/features/composer/domain/state/restore_email_inlin
 import 'package:tmail_ui_user/features/composer/domain/state/save_email_as_drafts_state.dart';
 import 'package:tmail_ui_user/features/composer/domain/state/send_email_state.dart';
 import 'package:tmail_ui_user/features/composer/domain/state/update_email_drafts_state.dart';
+import 'package:tmail_ui_user/features/composer/domain/state/upload_attachment_state.dart';
 import 'package:tmail_ui_user/features/composer/domain/usecases/create_new_and_save_email_to_drafts_interactor.dart';
 import 'package:tmail_ui_user/features/composer/domain/usecases/create_new_and_send_email_interactor.dart';
 import 'package:tmail_ui_user/features/composer/domain/usecases/download_image_as_base64_interactor.dart';
@@ -709,8 +710,15 @@ class ComposerController extends BaseController
             inlineImages: arguments.inlineImages);
 
           final accountId = mailboxDashBoardController.accountId.value;
-          final downloadUrl = mailboxDashBoardController.sessionCurrent
-            ?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+          dynamic downloadUrl;
+          try {
+            downloadUrl = mailboxDashBoardController.sessionCurrent
+              ?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+          } catch (e) {
+            logError('ComposerController::_initEmail(): $e');
+            downloadUrl = null;
+          }
+
           if (accountId == null || downloadUrl == null) return;
           _getEmailContentFromSessionStorageBrowser(
             htmlContent: arguments.emailContents ?? '',
@@ -1226,15 +1234,21 @@ class ComposerController extends BaseController
     );
   }
 
+
   void _uploadAttachmentsAction({required List<FileInfo> pickedFiles}) {
     final session = mailboxDashBoardController.sessionCurrent;
     final accountId = mailboxDashBoardController.accountId.value;
     if (session != null && accountId != null) {
-      final uploadUri = session.getUploadUri(accountId, jmapUrl: dynamicUrlInterceptors.jmapUrl);
-      uploadController.justUploadAttachmentsAction(
-        uploadFiles: pickedFiles,
-        uploadUri: uploadUri,
-      );
+      try {
+        final uploadUri = session.getUploadUri(accountId, jmapUrl: dynamicUrlInterceptors.jmapUrl);
+        uploadController.justUploadAttachmentsAction(
+          uploadFiles: pickedFiles,
+          uploadUri: uploadUri,
+        );
+      } catch (e) {
+        log('ComposerController::_uploadAttachmentsAction: $e');
+        uploadController.consumeState(Stream.value(Left(UploadAttachmentFailure(e, pickedFiles[0]))));
+      }
     } else {
       log('ComposerController::_uploadAttachmentsAction: SESSION OR ACCOUNT_ID is NULL');
     }
@@ -1407,15 +1421,20 @@ class ComposerController extends BaseController
       return;
     }
 
-    consumeState(_getEmailContentInteractor.execute(
-      session,
-      accountId,
-      emailId,
-      mailboxDashBoardController.baseDownloadUrl,
-      TransformConfiguration.forEditDraftsEmail(),
-      additionalProperties: Properties({
-        IndividualHeaderIdentifier.identityHeader.value}),
-    ));
+    try {
+      consumeState(_getEmailContentInteractor.execute(
+        session,
+        accountId,
+        emailId,
+        mailboxDashBoardController.baseDownloadUrl,
+        TransformConfiguration.forEditDraftsEmail(),
+        additionalProperties: Properties({
+          IndividualHeaderIdentifier.identityHeader.value}),
+      ));
+    } catch (e) {
+      logError('ComposerController::_handleUploadInlineSuccess(): $e');
+      consumeState(Stream.value(Left(GetEmailContentFailure(e))));
+    }
   }
 
   void _getEmailContentOffLineSuccess(GetEmailContentFromCacheSuccess success) {
@@ -1855,7 +1874,12 @@ class ComposerController extends BaseController
   void _handleUploadInlineSuccess(SuccessAttachmentUploadState uploadState) {
     uploadController.clearUploadInlineViewState();
 
-    final baseDownloadUrl = mailboxDashBoardController.sessionCurrent?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+    String? baseDownloadUrl;
+    try {
+      baseDownloadUrl = mailboxDashBoardController.sessionCurrent?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+    } catch (e) {
+      logError('ComposerController::_handleUploadInlineSuccess(): $e');
+    }
     final accountId = mailboxDashBoardController.accountId.value;
 
     if (baseDownloadUrl != null && accountId != null) {
@@ -1867,6 +1891,9 @@ class ComposerController extends BaseController
         uploadState.fileInfo,
         maxWidth: maxWithEditor,
       ));
+    } else {
+      log('ComposerController::_handleUploadInlineFailure(): baseDownloadUrl: $baseDownloadUrl, accountId: $accountId');
+      consumeState(Stream.value(Left(DownloadImageAsBase64Failure(e))));
     }
   }
 
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -832,15 +832,15 @@ class ComposerController extends BaseController
     required EmailActionType actionType,
     String? listPost,
   }) {
-    final userName = mailboxDashBoardController.sessionCurrent?.username.value;
+    final senderEmailAddress = mailboxDashBoardController.sessionCurrent?.getOwnEmailAddress();
     final isSender = presentationEmail.from
       .asList()
-      .any((element) => element.emailAddress.isNotEmpty && element.emailAddress == userName);
+      .any((element) => element.emailAddress.isNotEmpty && element.emailAddress == senderEmailAddress);
 
     final recipients = presentationEmail.generateRecipientsEmailAddressForComposer(
       emailActionType: actionType,
       isSender: isSender,
-      userName: userName,
+      userName: senderEmailAddress,
       listPost: listPost,
     );
 
@@ -1506,7 +1506,7 @@ class ComposerController extends BaseController
       if (arguments.emailActionType == EmailActionType.editDraft) {
         return arguments.presentationEmail?.firstEmailAddressInFrom ?? '';
       } else {
-        return mailboxDashBoardController.sessionCurrent?.username.value ?? '';
+        return mailboxDashBoardController.sessionCurrent?.getOwnEmailAddress() ?? '';
       }
     }
     return '';
--- a/lib/features/composer/presentation/extensions/create_email_request_extension.dart
+++ b/lib/features/composer/presentation/extensions/create_email_request_extension.dart
@@ -9,7 +9,7 @@ import 'package:jmap_dart_client/jmap/mail/email/keyword_identifier.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
 import 'package:model/email/email_action_type.dart';
 import 'package:model/extensions/email_address_extension.dart';
-import 'package:model/extensions/username_extension.dart';
+import 'package:model/extensions/session_extension.dart';
 import 'package:model/mailbox/presentation_mailbox.dart';
 import 'package:tmail_ui_user/features/composer/domain/model/email_request.dart';
 import 'package:tmail_ui_user/features/composer/presentation/extensions/identity_extension.dart';
@@ -26,7 +26,7 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
     if (identity?.email?.isNotEmpty == true) {
       return { identity!.toEmailAddress() };
     } else {
-      return { session.username.toEmailAddress() };
+      return { EmailAddress(null, session.getOwnEmailAddress()) };
     }
   }
 
@@ -34,7 +34,7 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
     if (emailActionType == EmailActionType.editDraft && fromSender.isNotEmpty) {
       return fromSender.first.emailAddress;
     } else {
-      return session.username.value;
+      return session.getOwnEmailAddress();
     }
   }
 
@@ -42,7 +42,7 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
     if (identity?.replyTo?.isNotEmpty == true) {
       return identity!.replyTo!.toSet();
     } else {
-      return { session.username.toEmailAddress() };
+      return { EmailAddress(null, session.getOwnEmailAddress()) };
     }
   }
 
--- a/lib/features/contact/presentation/contact_controller.dart
+++ b/lib/features/contact/presentation/contact_controller.dart
@@ -153,6 +153,9 @@ class ContactController extends BaseController with AutoCompleteResultMixin {
   Future<void> _searchContactByNameOrEmail(String query) async {
     log('ContactController::_searchContactByNameOrEmail(): query: $query');
     final listContact = await _getAutoCompleteSuggestion(query);
+    if (query.isEmail && !listContact.any((emailAddress) => emailAddress.email == query)) {
+      listContact.add(EmailAddress(null, query));
+    }
     searchedContactList.value = listContact;
   }
 
--- a/lib/features/email/presentation/controller/single_email_controller.dart
+++ b/lib/features/email/presentation/controller/single_email_controller.dart
@@ -66,6 +66,7 @@ import 'package:tmail_ui_user/features/email/domain/state/move_to_mailbox_state.
 import 'package:tmail_ui_user/features/email/domain/state/parse_calendar_event_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/parse_email_by_blob_id_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/preview_email_from_eml_file_state.dart';
+import 'package:tmail_ui_user/features/email/domain/state/preview_pdf_file_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/print_email_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/send_receipt_to_sender_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/store_event_attendance_status_state.dart';
@@ -314,6 +315,8 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
       _handlePreviewEmailFromEMLFileFailure(failure);
     } else if (failure is GetHtmlContentFromAttachmentFailure) {
       _handleGetHtmlContentFromAttachmentFailure(failure);
+    } else if (failure is PreviewPDFFileFailure) {
+      _handlePreviewPDFFileFailure(failure);
     }
   }
 
@@ -530,18 +533,23 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
       )));
     } else {
       if (session != null && accountId != null) {
-        final baseDownloadUrl = session!.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
-        TransformConfiguration transformConfiguration = PlatformInfo.isWeb
-          ? TransformConfiguration.forPreviewEmailOnWeb()
-          : TransformConfiguration.forPreviewEmail();
+        try {
+          final baseDownloadUrl = session!.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+          TransformConfiguration transformConfiguration = PlatformInfo.isWeb
+            ? TransformConfiguration.forPreviewEmailOnWeb()
+            : TransformConfiguration.forPreviewEmail();
 
-        consumeState(_getEmailContentInteractor.execute(
-          session!,
-          accountId!,
-          emailId,
-          baseDownloadUrl,
-          transformConfiguration
-        ));
+          consumeState(_getEmailContentInteractor.execute(
+            session!,
+            accountId!,
+            emailId,
+            baseDownloadUrl,
+            transformConfiguration
+          ));
+        } catch (e) {
+          logError('SingleEmailController::_getEmailContentAction(): $e');
+          consumeState(Stream.value(Left(GetEmailContentFailure(e))));
+        }
       }
     }
   }
@@ -767,14 +775,19 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
 
   void _downloadAttachmentsAction(List<Attachment> attachments) {
     if (accountId != null && session != null) {
-      final baseDownloadUrl = session!.getDownloadUrl(
-        jmapUrl: dynamicUrlInterceptors.jmapUrl,
-      );
-      consumeState(_downloadAttachmentsInteractor.execute(
-        attachments,
-        accountId!,
-        baseDownloadUrl,
-      ));
+      try {
+        final baseDownloadUrl = session!.getDownloadUrl(
+          jmapUrl: dynamicUrlInterceptors.jmapUrl,
+        );
+        consumeState(_downloadAttachmentsInteractor.execute(
+          attachments,
+          accountId!,
+          baseDownloadUrl,
+        ));
+      } catch (e) {
+        logError('SingleEmailController::_downloadAttachmentsAction(): $e');
+        consumeState(Stream.value(Left(DownloadAttachmentsFailure(e))));
+      }
     }
   }
 
@@ -821,15 +834,20 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
 
   void _exportAttachmentAction(Attachment attachment, CancelToken cancelToken) {
     if (accountId != null && session != null) {
-      final baseDownloadUrl = session!.getDownloadUrl(
-        jmapUrl: dynamicUrlInterceptors.jmapUrl,
-      );
-      consumeState(_exportAttachmentInteractor.execute(
-        attachment,
-        accountId!,
-        baseDownloadUrl,
-        cancelToken,
-      ));
+      try {
+        final baseDownloadUrl = session!.getDownloadUrl(
+          jmapUrl: dynamicUrlInterceptors.jmapUrl,
+        );
+        consumeState(_exportAttachmentInteractor.execute(
+          attachment,
+          accountId!,
+          baseDownloadUrl,
+          cancelToken,
+        ));
+      } catch (e) {
+        logError('SingleEmailController::_exportAttachmentAction(): $e');
+        consumeState(Stream.value(Left(ExportAttachmentFailure(e))));
+      }
     }
   }
 
@@ -873,16 +891,21 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
 
   void downloadAttachmentForWeb(Attachment attachment) {
     if (accountId != null && session != null) {
-      final baseDownloadUrl = session!.getDownloadUrl(
-        jmapUrl: dynamicUrlInterceptors.jmapUrl,
-      );
       final generateTaskId = DownloadTaskId(uuid.v4());
-      consumeState(_downloadAttachmentForWebInteractor.execute(
-          generateTaskId,
-          attachment,
-          accountId!,
-          baseDownloadUrl,
-          _downloadProgressStateController));
+      try {
+        final baseDownloadUrl = session!.getDownloadUrl(
+          jmapUrl: dynamicUrlInterceptors.jmapUrl,
+        );
+        consumeState(_downloadAttachmentForWebInteractor.execute(
+            generateTaskId,
+            attachment,
+            accountId!,
+            baseDownloadUrl,
+            _downloadProgressStateController));
+      } catch (e) {
+        logError('SingleEmailController::downloadAttachmentForWeb(): $e');
+        consumeState(Stream.value(Left(DownloadAttachmentForWebFailure(attachment: attachment, taskId: generateTaskId, exception: e))));
+      }
     } else {
       consumeState(Stream.value(
         Left(DownloadAttachmentForWebFailure(
@@ -2002,8 +2025,14 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
     Attachment attachment,
   ) {
     final accountId = mailboxDashBoardController.accountId.value;
-    final downloadUrl = mailboxDashBoardController.sessionCurrent
+    dynamic downloadUrl;
+    try {
+      downloadUrl = mailboxDashBoardController.sessionCurrent
         ?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+    } catch (e) {
+      logError('SingleEmailController::_getEmailContentAction(): $e');
+      downloadUrl = null;
+    }
     final blobId = attachment.blobId;
 
     if (accountId == null || downloadUrl == null || blobId == null) {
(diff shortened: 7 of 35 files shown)


florentos17:
the force push was to fixed the 'unused import' warnings and prefix each commit message with `TF-2305`

florentos17:
regarding the way `URIExtension::toQualifiedUrl` works: i did some adjustments to the unit tests because i feel like the wrong behaviour was expected: to me, if a trailing slash is advertized, then it should remain. I say this because it is the way cyrus works, but i dont know if this is universal so definitely tell me what you think about that.

with the current changes, tmail-flutter works well with tmail-backend and with cyrus. [this issue (`Session: handle relative URL`)](https://github.com/linagora/tmail-flutter/issues/1606) mentions compatibility with fastmail. i dont have a fastmail server at hand but if you have a URL for one i'd be happy to test !

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +52,28 @@ extension SessionExtension on Session {
...
   }
 
+  String? getEmailAddress() {
+    if(username.value.isEmail) {
+      return username.value;
+    } else {
+      try {
+        var principalsCapability = getCapabilityProperties<DefaultCapability>(
+            AccountId(Id(username.value)),
+            CapabilityIdentifier(Uri.parse('urn:ietf:params:jmap:principals')));
```
dab246, line 62: Please define this capability

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +52,28 @@ extension SessionExtension on Session {
...
+        String wrappedAddress = ((((principalsCapability?.properties
+            ?.values.last) as Map<String, dynamic>)
+            .values.last) as Map<String, dynamic>)
+            .values.toString();
+
+        String address = wrappedAddress.substring("(mailto:".length, wrappedAddress.length - ")".length);
+        return address.isEmail ? address : null;
+      } catch (_) {
+        return null;
+      }
```
dab246, line 73: We should separate function to handle it

core/lib/presentation/extensions/url_extension.dart
```diff
@@ -46,4 +46,9 @@ extension URLExtension on String {
       return '/$this';
     }
   }
+
+  bool hasTrailingSlash() {
```
dab246, line 50: Where did it use?

core/lib/presentation/extensions/uri_extension.dart
```diff
@@ -27,7 +32,7 @@ extension URIExtension on Uri {
 
   bool get hasOrigin {
     try {
-      return origin.isNotEmpty;
+      return !hasScheme ? false : origin.isNotEmpty;
```
dab246: IMO, We put it in a `try/catch` block so it should be enough to cover all errors. We don't need to test each unit like that.

lib/features/composer/presentation/extensions/create_email_request_extension.dart
```diff
@@ -44,7 +46,7 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
     } else if (identity?.replyTo?.isNotEmpty == true) {
       return identity!.replyTo!.toSet();
     } else {
-      return { session.username.toEmailAddress() };
+      return session.getEmailAddress() == null ? {} : { EmailAddress(null, session.getEmailAddress()) };
```
dab246, line 49:
```suggestion
      return session.getEmailAddress().isNotEmpty != true ? {} : { EmailAddress(null, session.getEmailAddress()) };
```

lib/features/home/domain/extensions/session_extensions.dart
```diff
@@ -53,8 +53,8 @@ extension SessionExtensions on Session {
 
   String get internalDomain {
     try {
-      if (GetUtils.isEmail(username.value)) {
-        return username.value.split('@').last;
+      if (getEmailAddress() != null) {
```
dab246, line 56:
```suggestion
      if (getEmailAddress().isNotEmpty == true) {
```

lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -375,8 +375,8 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin im
   void _setDefaultEmailAddressList() {
     listEmailAddressOfReplyTo.add(noneEmailAddress);
 
-    if (session?.username.value.isNotEmpty == true) {
-      final userEmailAddress = EmailAddress(null, session?.username.value);
+    if (session?.getEmailAddress() != null) {
```
dab246, line 383: idem

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -2748,6 +2752,28 @@ class MailboxDashBoardController extends ReloadableController with UserSettingPo
...
+        );
+        moveToMailbox(
+            sessionCurrent!,
+            accountId.value!,
+            moveToTrashMailboxRequest
+        );
+      }
+    }
+  }
+
```
dab246, line 2776: Please attach demo video for this fix.

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -1278,7 +1278,7 @@ class ThreadController extends BaseController with EmailActionController {
       ReadActions readActions = !email.hasRead ? ReadActions.markAsRead : ReadActions.markAsUnread;
       markAsEmailRead(email, readActions, MarkReadAction.swipeOnThread);
     } else if (direction == DismissDirection.endToStart) {
-      archiveMessage(context, email);
+      hasArchiveMailbox() ? archiveMessage(context, email) : moveMessageToTrash(context, email);
```
tddang-linagora, line 1281: Is this a confirmed behavior?

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
tddang-linagora, line 2776: Please remove this function. We already have a way to move email to trash. Look in to `ThreadController:handleEmailActionType`.

core/lib/presentation/extensions/url_extension.dart
florentos17: good catch, thank you, i removed the use in [commit 4fbba81](https://github.com/linagora/tmail-flutter/pull/3457/commits/4fbba813d3958ff70475f3379ccfeab877180d3f) and forgot to remove that as well.

core/lib/presentation/extensions/uri_extension.dart
florentos17: i agree with you and have already changed it back in [commit 4fbba81](https://github.com/linagora/tmail-flutter/pull/3457/commits/4fbba813d3958ff70475f3379ccfeab877180d3f)

lib/features/composer/presentation/extensions/create_email_request_extension.dart
florentos17: why not, but `getEmailAddress()` cannot return an empty string since there are `.isEmail` checks

lib/features/home/domain/extensions/session_extensions.dart
florentos17: idem

lib/features/identity_creator/presentation/identity_creator_controller.dart
florentos17: idem, `getEmailAddress` can only return `null` or a correct address

model/lib/extensions/session_extension.dart
```diff
@@ -2,22 +2,23 @@
...
 import 'package:uri/uri.dart';
 
 extension SessionExtension on Session {
 
-  String getDownloadUrl({String? jmapUrl}) {
-    final downloadUrlValid = jmapUrl != null
-      ? downloadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl))
-      : downloadUrl;
+  String getDownloadUrl(String jmapUrl) {
+    final downloadUrlValid = downloadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl));
```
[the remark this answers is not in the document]
florentos17: because
1. the `jmapUrl` attribute cannot be defaulted when using a cyrus server: it is not advertized in the session ;
2. it was already always provided anyway.

lib/features/composer/presentation/extensions/create_email_request_extension.dart
```diff
@@ -25,16 +25,18 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
   Set<EmailAddress> createSenders() {
     if (identity?.email?.isNotEmpty == true) {
       return { identity!.toEmailAddress() };
+    } else if (session.getEmailAddress() != null) {
+      return { EmailAddress(null, session.getEmailAddress()) };
     } else {
-      return { session.username.toEmailAddress() };
+      throw Exception("CreateEmailRequestExtension::createSenders() : unknown email address for current user");
```
dab246, line 31: Please return empty list. We should not throw exception here. Because it is not a mandatory condition when creating email.

lib/features/composer/presentation/extensions/create_email_request_extension.dart
dab246: OK, if you are sure to check the `isEmail` condition in the `getEmailAddress()` function then we can skip this change.

lib/features/manage_account/presentation/menu/settings/settings_first_level_view.dart
```diff
@@ -2,6 +2,8 @@ import 'package:core/presentation/extensions/color_extension.dart';
 import 'package:core/utils/platform_info.dart';
 import 'package:flutter/material.dart';
 import 'package:get/get.dart';
+import 'package:jmap_dart_client/jmap/core/user_name.dart';
+import 'package:model/model.dart';
```
dab246, line 6: Don't import the whole package, only what you need.

dab246:
@florentos17 Please confirm `Done` in the comments that you have fixed it. Thanks

core/test/presentation/extensions/uri_extension_test.dart
```diff
@@ -20,7 +20,7 @@ void main() {
       final baseUrl = Uri.parse('https://domain.com');
       final sourceUrl = Uri.parse('/jmap/');
 
-      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap');
+      final qualifiedUrlExpected = Uri.parse('https://domain.com/jmap/');
```
hoangdat, line 23: Why we need this? In what cases? 

IMO, it is our normalization, please keep it in uniform.

core/lib/presentation/extensions/uri_extension.dart
```diff
@@ -5,23 +5,20 @@ extension URIExtension on Uri {
...
-        final qualifiedUrl = toString().removeLastSlashOfUrl();
-        log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
-        return Uri.parse(qualifiedUrl);
-      }
+      final baseUrlValid = baseUrl.toString().removeLastSlashOfUrl();
+      final sourceUrlValid = toString().addFirstSlashOfUrl();
+      log('SessionUtils::toQualifiedUrl():baseUrlValid: $baseUrlValid | sourceUrlValid: $sourceUrlValid');
+      final qualifiedUrl = baseUrlValid + sourceUrlValid;
+      log('SessionUtils::toQualifiedUrl():qualifiedUrl: $qualifiedUrl');
+      return Uri.parse(qualifiedUrl);
```
hoangdat, line 21: What happens if `parse` failed? Can we use `tryParse` instead?

model/lib/extensions/session_extension.dart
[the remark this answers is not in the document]
hoangdat: > the jmapUrl attribute cannot be defaulted when using a cyrus server: it is not advertized in the session

Not really understand it.
Moreover, we also support other back-end: Stalwart, Cyrus, James. Keep it optional will keep flexible.

model/lib/extensions/session_extension.dart
```diff
@@ -26,10 +27,8 @@ extension SessionExtension on Session {
     return downloadUrlDecode;
   }
 
-  Uri getUploadUri(AccountId accountId, {String? jmapUrl}) {
-    final uploadUrlValid = jmapUrl != null
-      ? uploadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl))
-      : uploadUrl;
+  Uri getUploadUri(AccountId accountId, String jmapUrl) {
+    final uploadUrlValid = uploadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl));
```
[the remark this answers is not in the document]
hoangdat: we also support other back-end: Stalwart, Cyrus, James. Keep it optional will keep flexible.

model/lib/extensions/session_extension.dart
hoangdat: I did not find it in spec. It is Cyrus customization? Can you share session example of Cyrus here?

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +52,28 @@ extension SessionExtension on Session {
     }
   }
 
+  String? getEmailAddress() {
```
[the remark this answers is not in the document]
hoangdat: I still prefer `getUsername()` or `getOwnEmailAddress()`?

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +52,28 @@ extension SessionExtension on Session {
...
+    } else {
+      try {
+        var principalsCapability = getCapabilityProperties<DefaultCapability>(
+            AccountId(Id(username.value)),
+            CapabilityIdentifier(Uri.parse('urn:ietf:params:jmap:principals')));
+
+        String wrappedAddress = ((((principalsCapability?.properties
+            ?.values.last) as Map<String, dynamic>)
+            .values.last) as Map<String, dynamic>)
+            .values.toString();
```
hoangdat, line 67: not a fan of hard force casting and prefer strong typed for this

lib/features/thread/presentation/thread_controller.dart
hoangdat: to make it simple. No `Archive mailbox` no swipe to archive action (no move to Trash too).

lib/features/mailbox_dashboard/presentation/widgets/search_filters/search_filter_button.dart
```diff
@@ -31,7 +30,6 @@ class SearchFilterButton extends StatelessWidget {
   final DateTime? endDate;
   final Set<String>? listAddressOfFrom;
   final Set<String>? listAddressOfTo;
-  final UserName? userName;
```
hoangdat, line 34: why we remove it?

model/lib/extensions/session_extension.dart
florentos17: here is one of the `accountCapabilities` from the cyrus session:

```json
"urn:ietf:params:jmap:principals": {
  "currentUserPrincipalId": "bob",
  "urn:ietf:params:jmap:calendars": {
    "accountId": "bob",
    "account": null,
    "mayGetAvailability": true,
    "sendTo": {
      "imip": "mailto:bob@example.com"
    }
  }
},
```

it is the only place where the email address appears in the session, so benoit suggested in [this comment](https://github.com/linagora/james-project-private/issues/943#issuecomment-2607344537) to fetch the address from there if necessary

model/lib/extensions/session_extension.dart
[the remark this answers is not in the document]
florentos17: Previously the base URL `jmapUrl` attribute was optional and the default value was the `downloadUrl` provided in the session. For Tmail-backend, this is a full URL (so it's ok), but for Cyrus it is only the relative route:

```json
  "downloadUrl": "/jmap/download/{accountId}/{blobId}/{name}?accept={type}",
  "uploadUrl": "/jmap/upload/{accountId}/",
```

but I also made changes to the `toQualifiedUrl` method so that both cases are handled.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
florentos17: dat pham said we should remove it so no need anymore

lib/features/thread/presentation/thread_view.dart
```diff
@@ -551,34 +551,61 @@ class ThreadView extends GetWidget<ThreadController>
...
+                  const SizedBox(width: 11),
+                  Text(
+                    AppLocalizations.of(context).move_to_trash,
+                    style: const TextStyle(
+                      fontSize: 15,
+                      color: AppColor.primaryLightColor,
+                    ),
+                  ),
+                ],
+              ),
```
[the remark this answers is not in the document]
florentos17: i removed it entirely after [dat pham's comment](https://github.com/linagora/tmail-flutter/pull/3457#discussion_r1940666038)

lib/features/thread/presentation/thread_controller.dart
florentos17: ok, i removed the "move to trash" action and only kept the `hasArchiveMailbox` check, so that nothing shows if there is no archive mailbox

model/lib/extensions/session_extension.dart
florentos17: done

model/lib/extensions/session_extension.dart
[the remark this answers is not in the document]
florentos17: Unit tests done !

If we want to get the username, we should simply return `session.username`. This method is used to retrieve the full email address (sometimes used as the username, sometimes not). Do you want me to change `getEmailAddress()` to `getOwnEmailAddress()` ?

core/lib/presentation/extensions/url_extension.dart
florentos17: done

lib/features/composer/presentation/extensions/create_email_request_extension.dart
```diff
@@ -9,8 +9,8 @@ import 'package:jmap_dart_client/jmap/mail/email/keyword_identifier.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
 import 'package:model/email/email_action_type.dart';
 import 'package:model/extensions/email_address_extension.dart';
-import 'package:model/extensions/username_extension.dart';
 import 'package:model/mailbox/presentation_mailbox.dart';
+import 'package:model/model.dart';
```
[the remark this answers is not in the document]
florentos17: done

lib/features/mailbox_dashboard/presentation/mailbox_dashboard_view_web.dart
```diff
@@ -6,6 +6,7 @@ import 'package:flutter_svg/flutter_svg.dart';
 import 'package:get/get.dart';
 import 'package:model/extensions/presentation_mailbox_extension.dart';
 import 'package:model/extensions/username_extension.dart';
+import 'package:model/model.dart';
```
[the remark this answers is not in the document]
florentos17: done

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
florentos17: yes thank you, it's done

tddang-linagora (dismissed):

lib/features/composer/presentation/extensions/create_email_request_extension.dart
dab246: > Really ? I think an email cannot have an empty `From:` field, cf [RFC 5322](https://www.rfc-editor.org/rfc/rfc5322#section-3.6.2):
> 
> > The from field consists of the field name "From" and a comma-separated list of one or more mailbox specifications

`From` can take NULL value. Please read the details according to the [JMAP](https://jmap.io/spec-mail.html#properties-of-the-email-object) standard document. 

Screenshot 2025-02-04 at 17 40 31

lib/features/manage_account/presentation/menu/settings/settings_first_level_view.dart
florentos17: done

model/lib/extensions/session_extension.dart
florentos17: https://jmap.io/spec-sharing.html#principals

model/lib/extensions/session_extension.dart
florentos17: something like that ?

```dart
final sendTo = principalsCapability?.properties?['urn:ietf:params:jmap:calendars']?['sendTo'];
if (sendTo is Map<String, dynamic>) {
  final wrappedAddress = sendTo['imip'];
  if (wrappedAddress is String && wrappedAddress.startsWith('mailto:')) {
    String address = wrappedAddress.substring("mailto:".length);
    return address.isEmail ? address : null;
  }
}
```

lib/features/mailbox_dashboard/presentation/widgets/search_filters/search_filter_button.dart
florentos17: it was not used anywhere

hoangdat:
hi @florentos17, I rebased the branch from customer branch, so some files conflict. Please try to help us resolve it. 

This ticket is the first priority. please let us know when you are done in resolving conflict and comments. 

Thanks

florentos17:
@hoangdat when you forced pushed on master, it "added" 19 commits to this PR: commits that were previously on the master branch but are not anymore since the force push.

among these 19 commits:
- 12 are actually already on the new master (just a different hash), so i will force push on this branch to remove them from the current PR
- 7 are not on the new master at all... could that be an error ? here they are:
```
1a58542ab TF-3265 Fix double scrolling composer
...
47b842d25 TF-3189 composer now correctly encodes subaddresses
20c4720c3 TF-3189 new `reply to` field in the mail composer
d939ffb1c TF-3189 new confirmation popup when enabling subaddressing for a folder
0f61e3de8 TF-3189 subaddressing features only shown if supported by the server
810e304c1 TF-3189 new option to copy a folder's subaddress
ed0b14c35 TF-3189 new option to enable/disable subaddressing for a personal folderc
```

do you want me to leave them on the current PR so that they are reinstated ? or maybe cherry-pick them back yourself onto master ?

hoangdat:
Sub-addressing still has big conflict, we are in progress to work on it. Never forget. Don't worry.

core/test/presentation/extensions/uri_extension_test.dart
hoangdat: okie, in this case, make it uniform, keep `/` for all qualified url, and update all the test too.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -2755,7 +2755,7 @@ class MailboxDashBoardController extends ReloadableController
     dispatchAction(SelectionAllEmailAction());
   }
 
-  String get baseDownloadUrl => sessionCurrent?.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl) ?? '';
+  String get baseDownloadUrl => sessionCurrent?.getDownloadUrl(dynamicUrlInterceptors.jmapUrl!) ?? '';
```
hoangdat, line 2758: not a big fan of `!` without pre-check

model/lib/extensions/session_extension.dart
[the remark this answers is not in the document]
hoangdat: term `email address` is not specific for this. The reason for `usename` is the current account use the app, we need to distinguish with other email addresses in delegate mode or multiple account. But ok with `getOwnEmailAddress()`

lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -353,9 +353,14 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin im
   void _getAllIdentitiesSuccess(GetAllIdentitiesSuccess success) {
     if (success.identities?.isNotEmpty == true) {
       listEmailAddressDefault.value = success.identities!
+          .where((identity) => (identity.email != null && identity.email != ''))
```
tddang-linagora, line 356:
```suggestion
          .where((identity) => (identity.email != null && identity.email!.isNotEmpty))
```

model/lib/extensions/session_extension.dart
```diff
@@ -2,22 +2,24 @@
 import 'dart:ui';
 
 import 'package:core/presentation/extensions/uri_extension.dart';
+import 'package:get/get_utils/src/extensions/string_extensions.dart';
```
tddang-linagora, line 5: Why do we need this import?

lib/features/contact/presentation/contact_controller.dart
```diff
@@ -153,6 +153,9 @@ class ContactController extends BaseController with AutoCompleteResultMixin {
   Future<void> _searchContactByNameOrEmail(String query) async {
     log('ContactController::_searchContactByNameOrEmail(): query: $query');
     final listContact = await _getAutoCompleteSuggestion(query);
+    if (query.isEmail && !listContact.map((emailAddress) => emailAddress.email).contains(query)) {
+      listContact.add(EmailAddress(null, query));
+    }
```
hoangdat, line 158: what case we need it? can you show a demo?

lib/features/email/presentation/controller/single_email_controller.dart
```diff
@@ -530,7 +530,7 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
       )));
     } else {
       if (session != null && accountId != null) {
-        final baseDownloadUrl = session!.getDownloadUrl(jmapUrl: dynamicUrlInterceptors.jmapUrl);
+        final baseDownloadUrl = session!.getDownloadUrl(dynamicUrlInterceptors.jmapUrl!);
```
hoangdat, line 533: not safety: `!` without pre-check

lib/features/composer/presentation/extensions/create_email_request_extension.dart
dab246: @florentos17 Any update?

florentos17: Yes, I discussed it with @chibenwa and he said there was no point in making the app work if the user's email address is unknown.

Therefore he suggested to make `session.getOwnEmailAddress()` return an email-like `String` (currently nullable `String?`) and make it throw the exception if the email is unknown, in order to catch the error sooner and more explicitly.

This way, there would be no need for change in the current file. Would that be ok ?

dab246: Ok, so, please define specific this exception in exception class

core/test/presentation/extensions/uri_extension_test.dart
florentos17: But tmail-backend works the other way around: no `/` advertized, no `/` needed, and it does not work if the `/` is there...

So I feel like the solution is to keep the `/` if it is advertized, and not add it if not advertized ? But I agree that it is not uniform.

core/lib/presentation/extensions/uri_extension.dart
florentos17: What default behaviour would you expect if the parse fails and `null` is returned ?

The methods that call `toQualifiedUrl` (`getQualifiedApiUrl`, `getDownloadUrl`, `getUploadUrl`) mostly cannot work with `null`, so errors have to be thrown. But then it would be easier to just let `toQualifiedUrl` throw the error ?

model/lib/extensions/session_extension.dart
florentos17: is that ok @hoangdat ?

model/lib/extensions/session_extension.dart
florentos17: to have access to the getter `String::isEmail`

lib/features/contact/presentation/contact_controller.dart
florentos17: when the capability `com:linagora:params:jmap:contact:autocomplete` is absent, the `from` and `to` search filters do not work, because there are no suggestions to choose from:
[search-does-not-work.webm](https://github.com/user-attachments/assets/d9fc80b5-bc6b-4ec0-b8e8-c04b2062eca3)

with this change, if the query is an email address, it can be selected and the search can work:
[search-can-work.webm](https://github.com/user-attachments/assets/5369292f-d334-427b-97f9-201076921718)

model/lib/extensions/session_extension.dart
florentos17: @hoangdat what do you think ? Indeed I dont think there is any doc about fetching the email from there ; the idea comes from [@chibenwa 's comment](https://github.com/linagora/james-project-private/issues/943#issuecomment-2607344537)

model/lib/extensions/session_extension.dart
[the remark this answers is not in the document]
florentos17: Thinking back about it, you are right that keeping it optional will keep it flexible, it is better. 

On my latest commit you can see that I reverted it, but then I had to handle the case where the URL cannot be found at all by throwing an exception, is that ok ?

lib/features/composer/presentation/extensions/create_email_request_extension.dart
florentos17: done !

lib/features/identity_creator/presentation/identity_creator_controller.dart
florentos17: done

lib/features/email/presentation/controller/single_email_controller.dart
florentos17: done, i reverted the change

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +53,30 @@ extension SessionExtension on Session {
...
+  String? getEmailAddressFromPrincipalsCapability() {
+    try {
+      var principalsCapability = getCapabilityProperties<DefaultCapability>(AccountId(Id(username.value)), capabilityPrincipals);
+    
+      String wrappedAddress = ((((principalsCapability?.properties
+          ?.values.last) as Map<String, dynamic>)
+          .values.last) as Map<String, dynamic>)
+          .values.toString();
+    
+      String address = wrappedAddress.substring("(mailto:".length, wrappedAddress.length - ")".length);
```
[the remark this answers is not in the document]
florentos17: i answered on one of your earlier comments: https://github.com/linagora/tmail-flutter/pull/3457#discussion_r1944503393

model/test/extensions/session_extension_test.dart
```diff
@@ -288,7 +288,7 @@ void main() {
       );
 
       // act
-      final emailAddress = session.getEmailAddress();
+      final emailAddress = session.getOwnEmailAddress();
```
chibenwa, line 283: `session.myMailAddress` may even be more straightforward ?

lib/features/composer/domain/usecases/upload_attachment_interactor.dart
```diff
@@ -13,10 +14,11 @@ class UploadAttachmentInteractor {
 
   Stream<Either<Failure, Success>> execute(
     FileInfo fileInfo,
-    Uri uploadUri, {
+    Uri? uploadUri, {
     CancelToken? cancelToken,
   }) async* {
     try {
+      if (uploadUri == null) throw UnknownUriException();
```
dab246, line 21: We should not put the `uploadUri == null` check in this interactor. It is against the interactor's functionality. You should check it before calling this interactor. `uploadUri` is required in this interactor.

lib/features/contact/presentation/contact_controller.dart
dab246:
```suggestion
if (query.isEmail && !listContact.any((email) => email.email == query)) {
  listContact.add(EmailAddress(null, query));
}
```

model/lib/extensions/session_extension.dart
```diff
@@ -26,10 +38,16 @@ extension SessionExtension on Session {
...
-      ? uploadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl))
-      : uploadUrl;
+  Uri? getUploadUri(AccountId accountId, {String? jmapUrl}) {
+    final Uri uploadUrlValid;
+    if (jmapUrl != null) {
+      uploadUrlValid = uploadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl));
+    } else if (uploadUrl.hasOrigin) {
+      uploadUrlValid = uploadUrl;
+    } else {
+      return null;
```
dab246, line 48:
```suggestion
throw UnknownUriException();
```

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +71,42 @@ extension SessionExtension on Session {
     }
   }
 
+  String getOwnEmailAddress() {
+    if(username.value.isEmail) {
```
dab246, line 75:
```suggestion
    if (username.value.isEmail) {
```

lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -353,9 +353,12 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin im
   void _getAllIdentitiesSuccess(GetAllIdentitiesSuccess success) {
     if (success.identities?.isNotEmpty == true) {
       listEmailAddressDefault.value = success.identities!
+          .where((identity) => (identity.email != null && identity.email!.isNotEmpty))
```
dab246, line 356:
```suggestion
          .where((identity) => identity.email?.isNotEmpty == true)
```

lib/features/identity_creator/presentation/identity_creator_controller.dart
```diff
@@ -353,9 +353,12 @@ class IdentityCreatorController extends BaseController with DragDropFileMixin im
...
           .map((identity) => identity.toEmailAddressNoName())
           .toSet()
           .toList();
+
+      listEmailAddressDefault.add(EmailAddress(null, session?.getOwnEmailAddress()));
```
dab246, line 361: why do we need it?

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +71,42 @@ extension SessionExtension on Session {
...
+  String? getOwnEmailAddressFromPrincipalsCapability() {
+    try {
+      var principalsCapability = getCapabilityProperties<DefaultCapability>(AccountId(Id(username.value)), capabilityPrincipals);
+
+      String wrappedAddress = ((((principalsCapability?.properties
+          ?.values.last) as Map<String, dynamic>)
+          .values.last) as Map<String, dynamic>)
+          .values.toString();
+
+      String address = wrappedAddress.substring("(mailto:".length, wrappedAddress.length - ")".length);
```
dab246, line 103: IMO, Getting the email this way is inefficient, because we don't know exactly which field contains the email, the order of the keys of the returned json will be different on the servers, so truncating the string of the last element is redundant. So we should skip this.

lib/features/composer/domain/usecases/upload_attachment_interactor.dart
florentos17: This is what I did at first, but it means the interactor is not even called in case of a `null` `uploadUri`, so there is no `UploadAttachmentFailure`, and the red toast notification "upload attachment failed" does not show on the app.

I thought it would be better for the UX to have the red toast notification show up, but I understand the concern. Should I change it to put the check before ? Maybe I can throw a `GetUploadUriFailure` earlier to still have the toast notification ?

lib/features/identity_creator/presentation/identity_creator_controller.dart
florentos17: cyrus does not advertize the address like tmail does, so without the change the list of possible addresses is empty (clicking on the widget does nothing):
image

with the change, the default address appears:
image

and I checked, it does not duplicate the address if it was already there when using tmail

lib/features/identity_creator/presentation/identity_creator_controller.dart
florentos17: done

lib/features/contact/presentation/contact_controller.dart
florentos17: done, i applied the suggestion

lib/features/composer/domain/usecases/upload_attachment_interactor.dart
dab246: Yes, you check `uploadUri != null` before calling the interactor. And throw `UploadAttachmentFailure(UnknownUriException())` exception as soon as you check it for null

lib/features/identity_creator/presentation/identity_creator_controller.dart
dab246: So you should check if `listEmailAddressDefault.isEmpty` then add `getOwnEmailAddress`. Avoid duplicate emails.

florentos17: no, it is already good, there is already no duplication.
i can add a `listEmailAddressDefault.isEmpty` but it makes no difference

lib/features/composer/domain/usecases/upload_attachment_interactor.dart
florentos17: done !

model/lib/extensions/session_extension.dart
florentos17: done !

core/test/presentation/extensions/uri_extension_test.dart
hoangdat: > So I feel like the solution is to keep the / if it is advertized

yes. Update tests to include it

core/lib/presentation/extensions/uri_extension.dart
hoangdat: what happen then we throw error?

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -473,6 +475,9 @@ class MailboxDashBoardController extends ReloadableController
       toastManager.showMessageFailure(failure);
     } else if (failure is GetComposerCacheFailure) {
       _handleIdentityCache();
+    } else if (failure is UploadAttachmentFailure) {
+      final uploadController = Get.find<UploadController>();
+      uploadController.handleFailureViewState(failure);
```
tddang-linagora, line 480: Why handle `UploadAttachmentFailure` here? Where is `UploadController` is injected?

lib/features/mailbox_dashboard/presentation/model/search/quick_search_filter.dart
```diff
@@ -135,8 +135,9 @@ enum QuickSearchFilter {
         return searchFilter.emailReceiveTimeType == EmailReceiveTimeType.last7Days;
       case QuickSearchFilter.fromMe:
         return searchFilter.from.length == 1 &&
-          userName?.value.isNotEmpty == true &&
-          userName?.value == searchFilter.from.first;
+          currentUserEmail != null &&
+          currentUserEmail.isNotEmpty == true &&
```
tddang-linagora, line 139:
```suggestion
          currentUserEmail?.isNotEmpty == true &&
```

lib/features/public_asset/presentation/public_asset_controller.dart
```diff
@@ -126,6 +127,9 @@ class PublicAssetController extends BaseController {
       _handleCreatePublicAssetFailureState();
     } else if (failure is PublicAssetOverQuotaFailureState) {
       _handlePublicAssetOverQuotaFailureState(failure);
+    } else if (failure is UploadAttachmentFailure) {
+      final uploadController = Get.find<UploadController>();
+      uploadController.handleFailureViewState(failure);
```
tddang-linagora, line 132: Why handle `UploadAttachmentFailure` here? Where is `UploadController` is injected?

model/lib/extensions/session_extension.dart
hoangdat: what is your cyrus image and configuration? I did not find that in my running

integration_test/utils/scenario_utils_mixin.dart
```diff
@@ -142,34 +142,38 @@ mixin ScenarioUtilsMixin {
...
+                },
+              );
+            }
+          },
+        );
+        if (attachment != null) {
+          attachments.add(attachment);
+        }
+      } catch (e) {
+        mailboxDashBoardController.handleFailureViewState(UploadAttachmentFailure(e, fileInfo));}
```
hoangdat, line 176:
```suggestion
      } catch (e) {
        mailboxDashBoardController.handleFailureViewState(UploadAttachmentFailure(e, fileInfo));
      }
```

hoangdat, line 176: Note: it is an action in integration_test. So no need to handle failure case here.

Please log it for more information in catch clause.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1230,11 +1231,15 @@ class ComposerController extends BaseController
...
-        uploadUri: uploadUri,
-      );
+      try {
+        final uploadUri = session.getUploadUri(accountId, jmapUrl: dynamicUrlInterceptors.jmapUrl);
+        uploadController.justUploadAttachmentsAction(
+          uploadFiles: pickedFiles,
+          uploadUri: uploadUri,
+        );
+      } catch (e) {
+        mailboxDashBoardController.handleFailureViewState(UploadAttachmentFailure(e, pickedFiles[0]));
```
dab246, line 1241:
```suggestion
uploadController.consumeState(Stream.value(Left(UploadAttachmentFailure(e, pickedFiles[0]))));
```

hoangdat:
- [ ] still sending email failed because `Settings/get` method use unknown capability

```json
{
    "type": "urn:ietf:params:jmap:error:unknownCapability",
    "title": "JMAP request uses unsupported capabilities",
    "status": 400,
    "detail": "The Request object used capability 'com:linagora:params:jmap:settings', which is not supported by this server."
}
```

hoangdat:
- [ ] search did not work well when submit text in search bar

model/lib/extensions/session_extension.dart
hoangdat: still not find a good reason to link `principal::calendars` capabilities with current address.

lib/features/contact/presentation/contact_controller.dart
```diff
@@ -153,6 +153,9 @@ class ContactController extends BaseController with AutoCompleteResultMixin {
   Future<void> _searchContactByNameOrEmail(String query) async {
     log('ContactController::_searchContactByNameOrEmail(): query: $query');
     final listContact = await _getAutoCompleteSuggestion(query);
+    if (query.isEmail && !listContact.any((emailAddress) => emailAddress.email == query)) {
+      listContact.add(EmailAddress(null, query));
+    }
```
hoangdat, line 158: this logic was implemented inside `_getAutoCompleteSuggestion`. Why we need to add here?

dab246, line 158: We have actually solved this problem.

Screenshot 2025-02-10 at 11 46 57

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
hoangdat: not a good idea for handle this error here

integration_test/utils/scenario_utils_mixin.dart
dab246, line 176: This is a function that handles only the upload result to support email creation in `integration test`. So we don't need to handle the failure value. Please delete it.

lib/features/public_asset/presentation/public_asset_controller.dart
hoangdat: idem

lib/features/public_asset/presentation/public_asset_controller.dart
```diff
@@ -206,8 +210,12 @@ class PublicAssetController extends BaseController {
     if (session == null || accountId == null) return;
 
     final fileInfo = platformFile.toFileInfo();
-    final uploadUri = session!.getUploadUri(accountId!, jmapUrl: dynamicUrlInterceptors.jmapUrl);
-    consumeState(_uploadAttachmentInteractor.execute(fileInfo, uploadUri));
+    try {
+      final uploadUri = session!.getUploadUri(accountId!, jmapUrl: dynamicUrlInterceptors.jmapUrl);
+      consumeState(_uploadAttachmentInteractor.execute(fileInfo, uploadUri));
+    } catch (e) {
+      handleFailureViewState(UploadAttachmentFailure(e, fileInfo));
```
hoangdat, line 217: Please dispatch a failure state instead of calling viewState handler.

model/lib/extensions/session_extension.dart
florentos17: https://github.com/linagora/cyrus-postfix-docker

lib/features/public_asset/presentation/public_asset_controller.dart
florentos17: handling it there because `PublicAssetController::_uploadFileToBlobAction()` calls `session!.getUploadUri` which can throw an `UnknownUriException`

same comment for `UploadController`, I don't know where it is injected and I can directly copy-paste the method if it's better ?

lib/features/public_asset/presentation/public_asset_controller.dart
florentos17: ok, but how do I show the toast notification "could not upload attachment" without the viewState handler ?

lib/features/contact/presentation/contact_controller.dart
florentos17: cf my answer to the same comment: https://github.com/linagora/tmail-flutter/pull/3457#discussion_r1949070547

lib/features/public_asset/presentation/public_asset_controller.dart
florentos17: I handled it here to be able to show the toast notification "upload failed"
But you are right that calling the `UploadController` is not the best, maybe I can copy-paste the `_handleUploadAttachmentFailure` to access it directly from `PublicAssetController` ?

core/test/presentation/extensions/uri_extension_test.dart
florentos17: done !

lib/features/mailbox_dashboard/presentation/model/search/quick_search_filter.dart
florentos17: done !

model/lib/extensions/session_extension.dart
chibenwa: Hello @hoangdat 

This is the only place in Cyrus where we get a reliable mail address to use so as uggly as it it, this "hack" is necessary to make TMail usable with Cyrus.

Please refer to https://github.com/cyrusimap/cyrus-imapd/issues/5275

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
florentos17: yes, I moved it back to `ComposerController`

lib/features/composer/presentation/composer_controller.dart
florentos17: done !

integration_test/utils/scenario_utils_mixin.dart
florentos17: done !

florentos17: done !

lib/features/public_asset/presentation/public_asset_controller.dart
florentos17: done !

florentos17:
I think all of the comments have been addressed. Let me know if you have more or if I can clean the commit history ?

model/lib/extensions/session_extension.dart
dab246: Oke we can use it

lib/features/contact/presentation/contact_controller.dart
dab246: Oke, so make sure `listContact.add(EmailAddress(null, query));
    }` only was called when `listContact.isEmpty`

lib/features/identity_creator/presentation/identity_creator_controller.dart
dab246: Why is there no difference. If in case the server returns identity with email `EmailAddress(null, getOwnEmailAddress())` then your command will duplicate the email. You can write unit test to check it.

dab246: Why is there no difference. If in case the server returns identity with email `EmailAddress(null, getOwnEmailAddress())` then your command will duplicate the email. You can write unit test to check it.

model/lib/extensions/session_extension.dart
florentos17: done !

lib/features/contact/presentation/contact_controller.dart
florentos17: the current condition includes the check: `!listContact.any((emailAddress) => emailAddress.email == query)`
I think this is ok ?

lib/features/identity_creator/presentation/identity_creator_controller.dart
florentos17: When I check, there is no duplicate if I add the same `EmailAddress`. But I had another idea: if there is another `EmailAddress` with same `email` but different `name`, then it duplicates. 

So you are right, its better to put the check:
```dart
if (!listEmailAddressDefault.any((emailAddress) => emailAddress.email == session?.getOwnEmailAddress())) {
    listEmailAddressDefault.add(EmailAddress(null, session?.getOwnEmailAddress()));
}
```

is that ok ?

tddang-linagora (dismissed):

integration_test/utils/scenario_utils_mixin.dart
```diff
@@ -142,33 +143,39 @@ mixin ScenarioUtilsMixin {
...
+                },
+              );
+            }
+          },
+        );
+        if (attachment != null) {
+          attachments.add(attachment);
+        }
+      } catch (e) {
+        log('ScenarioUtilsMixin::uploadAttachments(): $e');
```
dab246, line 177:
```suggestion
        logError('ScenarioUtilsMixin::uploadAttachments(): $e');
```

lib/features/public_asset/presentation/public_asset_controller.dart
```diff
@@ -206,8 +224,13 @@ class PublicAssetController extends BaseController {
...
 
     final fileInfo = platformFile.toFileInfo();
-    final uploadUri = session!.getUploadUri(accountId!, jmapUrl: dynamicUrlInterceptors.jmapUrl);
-    consumeState(_uploadAttachmentInteractor.execute(fileInfo, uploadUri));
+    try {
+      final uploadUri = session!.getUploadUri(accountId!, jmapUrl: dynamicUrlInterceptors.jmapUrl);
+      consumeState(_uploadAttachmentInteractor.execute(fileInfo, uploadUri));
+    } catch (e) {
+      log('PublicAssetController::_uploadFileToBlobAction: $e');
+      consumeState(Stream.value(Left(UploadAttachmentFailure(e, fileInfo))));
```
hoangdat, line 232: idem

model/lib/extensions/session_extension.dart
```diff
@@ -2,22 +2,34 @@
...
-    final downloadUrlValid = jmapUrl != null
-      ? downloadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl))
-      : downloadUrl;
+    final Uri downloadUrlValid;
+    if (jmapUrl != null) {
+      downloadUrlValid = downloadUrl.toQualifiedUrl(baseUrl: Uri.parse(jmapUrl));
+    } else if (downloadUrl.hasOrigin) {
+      downloadUrlValid = downloadUrl;
+    } else {
+      throw UnknownUriException();
```
hoangdat, line 30: this function was use in many places without error catching, so IMO, please use `tryParse` instead

lib/features/public_asset/presentation/public_asset_controller.dart
hoangdat: Agree with it, but make sure all the handle for `UploadAttachmentFailure` still work well

integration_test/utils/scenario_utils_mixin.dart
florentos17: done !

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +71,43 @@ extension SessionExtension on Session {
...
+    } else if (getOwnEmailAddressFromPersonalAccount() != null) {
+      return getOwnEmailAddressFromPersonalAccount()!;
+    } else if (getOwnEmailAddressFromPrincipalsCapability() != null) {
+      return getOwnEmailAddressFromPrincipalsCapability()!;
+    } else {
+      throw UnknownAddressException();
+    }
+  }
+
+  String? getOwnEmailAddressFromPersonalAccount() {
```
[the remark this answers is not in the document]
florentos17: done !

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1226,15 +1229,35 @@ class ComposerController extends BaseController
...
+          failure.fileInfo.isInline == true
+              ? AppLocalizations.of(currentContext!).thisImageCannotBeAdded
+              : AppLocalizations.of(currentContext!).can_not_upload_this_file_as_attachments,
+          leadingSVGIconColor: Colors.white,
+          leadingSVGIcon: failure.fileInfo.isInline == true
+              ? imagePaths.icInsertImage
+              : imagePaths.icAttachment
+      );
+    }
+  }
```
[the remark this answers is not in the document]
florentos17: done !

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -381,6 +382,8 @@ class ComposerController extends BaseController
     } else if (failure is GetAlwaysReadReceiptSettingFailure) {
       hasRequestReadReceipt.value = false;
       _initEmailDraftHash();
+    } else if (failure is UploadAttachmentFailure) {
+      _handleUploadAttachmentFailure(failure);
```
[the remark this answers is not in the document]
florentos17: done !

model/lib/extensions/session_extension.dart
```diff
@@ -53,6 +71,43 @@ extension SessionExtension on Session {
...
+      return username.value;
+    } else if (getOwnEmailAddressFromPersonalAccount() != null) {
+      return getOwnEmailAddressFromPersonalAccount()!;
+    } else if (getOwnEmailAddressFromPrincipalsCapability() != null) {
+      return getOwnEmailAddressFromPrincipalsCapability()!;
+    } else {
+      throw UnknownAddressException();
+    }
+  }
+
```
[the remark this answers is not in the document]
florentos17: done !

hoangdat (approved):

hoangdat:
Good job. please rebase with `master` branch to handle conflict. @florentos17

hoangdat (approved):

dab246 (approved):
LGTM

Issue #3455, opened 2025-01-30
[COMPANY SERVER] Allow to authenticate with Authorization BEARER


chibenwa:
## Why

CF https://github.com/linagora/tmail-flutter/discussions/3435

> Hi, Our JMAP endpoint does not support basic auth, it only supports Bearer token authorization, either via OAuth (with pre-registered client id) or by manually creating an API token. We do not support OpenID Connect. Therefore it looks like there's no way to currently connect Twake-Mail to your Fastmail account, sorry. The issue of OAuth requiring pre-registration is well known in the email community, and we are working with other major vendors on a new autodiscovery mechanism and OAuth profile that will hopefully resolve this in the future. Kind regards, Neil.

THus, in order to support FASTMAIL users we shall add a way to pass application tokens....

## How

On the tmail login page when OIDC auto discovery failed add a `Advanced settings` foldable section.

If unfolded it displays a field `Applicative token` and a little explanation `Some JMAP services do not support login via password for third party apps but instead allow generate applicative tokens.` and a `Login` button.

![Image](https://github.com/user-attachments/assets/dca9d0f5-32cb-4baf-b7dd-c2ab7ca06014)

![Image](https://github.com/user-attachments/assets/ba4ee172-dc0d-463a-bd31-9251af1bd691)

John-Gee:
Hello,

Is there any hope for this? ;)

hoangdat:
We implemented it in https://github.com/linagora/tmail-flutter/pull/3527. But still need to confirm the UI design before push it in public

John-Gee:
Can I try a beta to see how it feels maybe?
I guess I could build against the PR to try.

Thank you!

PR #3459, opened 2025-01-31
handle parsing exception for `OIDCHttpClient::checkOIDCIsAvailable()`

florentos17:
here is a fix for a bug pointed out by [this user experience review](https://github.com/linagora/tmail-flutter/discussions/3435#discussioncomment-12008196).

- upon providing `https://api.fastmail.com/jmap/session`, the actual error is a `401 Unauthorized` and this case was handeled in [this pull request](https://github.com/linagora/tmail-flutter/pull/3428) ;
- upon providing the correct `https://api.fastmail.com`, another error occured because the server answers with HTML instead of JSON, eg, `FormatException`. the correct behaviour, which i fix with the current PR, is to ignore webfinger and fall back to the classic credential form.

### DIFF
--- a/lib/features/login/data/network/oidc_http_client.dart
+++ b/lib/features/login/data/network/oidc_http_client.dart
@@ -11,6 +11,7 @@ import 'package:model/oidc/response/oidc_response.dart';
 import 'package:tmail_ui_user/features/login/data/extensions/service_path_extension.dart';
 import 'package:tmail_ui_user/features/login/data/network/config/oidc_constant.dart';
 import 'package:tmail_ui_user/features/login/data/network/endpoint.dart';
+import 'package:tmail_ui_user/features/login/domain/exceptions/login_exception.dart';
 import 'package:tmail_ui_user/features/login/data/network/oidc_error.dart';
 import 'package:tmail_ui_user/main/utils/app_config.dart';
 import 'package:dio/dio.dart' show DioError;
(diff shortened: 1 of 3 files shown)


chibenwa:
Please record a little video accessing a fastmail account after this change, if possible.

tddang-linagora:
From [the same discussion](https://github.com/linagora/tmail-flutter/discussions/3435#discussioncomment-12003124):
>Our JMAP endpoint does not support basic auth

There's no point showing our own log in form. We need to add OAuth support in order for FastMail to work on Twake Mail.

lib/features/login/data/network/oidc_http_client.dart
```diff
@@ -38,6 +38,9 @@ class OIDCHttpClient {
       } else {
         return OIDCResponse.fromJson(jsonDecode(result));
       }
+    } on FormatException catch (exception) {
+      log('checkOIDCIsAvailable(): error while parsing server response (JSON expected): ${exception.message}');
+      rethrow;
     } on DioError catch (exception) {
```
dab246, line 44: Please be more specific about the error. `FormatException` is too general an error.

florentos17:
> There's no point showing our own log in form. We need to add OAuth support in order for FastMail to work on Twake Mail.

correct, [issue 3455](https://github.com/linagora/tmail-flutter/issues/3455) has already been open on the topic

lib/features/login/data/network/oidc_http_client.dart
florentos17: done, i created a `InvalidOIDCResponseException`

hoangdat:
first, please use git rebase to only pick your modification for this issue, then force push.

florentos17:
@hoangdat @tddang-linagora @dab246 I rebased on latest master and checked that the demo is still the same, it is all good.

Tell me if you have any last comment

tddang-linagora (approved):

hoangdat (approved):

dab246 (approved):
lgtm

Issue #3456, opened 2025-01-30
[COMPANY SERVER] Allow self signed cerificates


chibenwa:
## Why

This eases testing.

## WHat

Add the option when choosing company server URL

![Image](https://github.com/user-attachments/assets/be43ec4b-875a-406d-82c0-b7d6e8eb345b)

![Image](https://github.com/user-attachments/assets/8d1ccb3f-b28f-4775-b810-cedd4a3f14f5)

PR #3473, opened 2025-02-05
TF-3456 Checkbox to be able to accept self-signed certificates

florentos17:
tackling [issue #3456](https://github.com/linagora/tmail-flutter/issues/3456) 

the design can easily be changed, for now i kept it simple with an `ExpansionTile` and a `LabeledCheckbox`

here is a video demo:

[demo-checkbox-accept-ssc.webm](https://github.com/user-attachments/assets/a866860e-a34e-4842-89db-09b0e06b31e3)


florentos17:
I noticed a flaw in my previous work: with the current state, self-signed certificates can be accepted upon login (cf demo) but the choice is not remembered for later requests.
For example, if the users checks `accept`, he can login to a server using a self-signed certificate, but later when he wants to upload an attachment for an email, an error is thrown.

Here is what happens:
1. When the app starts, `Get.put()` registers a `DioClient` with default settings, meaning self-signed certificates are rejected ;
2. When the user logs in, `OIDCHttpClient` updates its own `DioClient` instance to accept or reject self-signed certificates based on the user’s choice ;
3. Later, whenever a `DioClient` is needed, `Get.find()` retrieves the initially registered instance, which still has the default settings (self-signed certificates are not accepted).

- One solution would be to cache the user’s choice and change `DioClient`'s constructor to apply the cached setting, if any.
- Or, maybe easier, once the user has made its choice, replace the existing `DioClient` in GetX with a new one that has the correct setting ? The problem with that is, GetX already registered many objects based on the initial `DioClient` (`OIDCHttpClient`, `DownloadClient`, `EmailAPI`...). It would be easier to just erase and reload every `NetworkBindings` dependency ? Would that be acceptable ?

Issue #3449, opened 2025-01-24
[MU] Handle mail priority header


chibenwa:
# What

As a mail expeditor I want to set mail as important:

![Image](https://github.com/user-attachments/assets/98ade778-8a61-4644-ba6e-ddd4ca5314ca)

It displays onto the receiving end as:

![Image](https://github.com/user-attachments/assets/bd260813-89ad-4e3b-82df-196e59d427ff)

# How

Within the composer if the option is selected set the following headers:

```
X-Priority: 1
Importance: high
Priority: urgent
```

If any of the header is set to this value then display the /!\ mark in email list

chibenwa:
Cc @Bobpodvalnyi

chibenwa:
https://www.figma.com/design/XqLqMINlZw09BdEHI8yLb9/Teammail---1.1?node-id=3877-3055&p=f&t=5Q88tflI1iASg4Nb-0

hoangdat:
2HD

hoangdat:
![Image](https://github.com/user-attachments/assets/201e6511-ed3c-41e8-9609-f97a49edaaef)

hi @chibenwa Can we set the setting in server or only in client side?

Arsnael:
You can use the settings jmap endpoint like for some other settings I guess

PR #3476, opened 2025-02-07
TF-3449 [MU] Handle mail priority header

dab246:
## Issue

#3449

## Dependent

- Need merged: https://github.com/linagora/jmap-dart-client/pull/106

## Resolved

- Web:

https://github.com/user-attachments/assets/d3fbeb15-63e4-4c38-8530-7b3809846597

- Mobile:

[demo-mobile.webm](https://github.com/user-attachments/assets/48ab1eff-e48b-483d-a404-11ddb3faf80f)

- Setup in Setting View

https://github.com/user-attachments/assets/99dfbc63-a6b1-4f38-ba91-f6a49d38ede4


model/lib/extensions/presentation_email_extension.dart
tddang-linagora, line 1: This looks like maintenance hell. Should we update all these using `copyWith` method of `PresentationEmail`?

dab246: It is not necessary that we explicitly state what the function does.

model/lib/extensions/email_extension.dart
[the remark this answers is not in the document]
dab246: idem

model/lib/extensions/presentation_email_extension.dart
tddang-linagora: We can still keep the functions, only update the `return` statements with `copyWith` instead of explicitly declare every properties, so that in the future if we ever need to add more properties to the `PresentationEmail`, we don't have to trace every call instance of this class and manually add properties there anymore.

dab246: OK, done

tddang-linagora (approved):

dab246:
- [x] Setup `Sender-set inportant flag` in Setting View

model/lib/email/presentation_email.dart
```diff
@@ -135,6 +142,22 @@ class PresentationEmail with EquatableMixin, SearchSnippetMixin {
...
+    final importance = importanceHeader?[IndividualHeaderIdentifier.importanceHeader]
+      ?.trim()
+      .toLowerCase();
+    final priority = priorityHeader?[IndividualHeaderIdentifier.priorityHeader]
+      ?.trim()
+      .toLowerCase();
+
+    return xPriority == MailPriorityHeader.firstXPriority &&
+      importance == MailPriorityHeader.highImportance &&
+      priority == MailPriorityHeader.urgentPriority;
```
hoangdat, line 158:
```suggestion
    return xPriority == MailPriorityHeader.firstXPriority ||
      importance == MailPriorityHeader.highImportance ||
      priority == MailPriorityHeader.urgentPriority;
```

hoangdat:
- [ ] IMO, for web app, we should add a column for `Priority`, it should stay beside `Star` and only display whenever user turn-on the preference.

model/lib/email/presentation_email.dart
dab246: Where did you read that you should use `or`?. As far as I know it should be `and`

dab246:
> * [x]  IMO, for web app, we should add a column for `Priority`, it should stay beside `Star` and only display whenever user turn-on the preference.

We can't do that because the `update` function of `Email/set` currently doesn't support updating `headers`

chibenwa:
>> IMO, for web app, we should add a column for Priority, it should stay beside Star and only display whenever user turn-on the preference.

> We can't do that because the update function of Email/set currently doesn't support updating headers

@hoangdat it sould be a display question only, nothin related to cache/network data model in general. I'd gladly discuss more to understand what are the blockers here if any.

hoangdat:
> I'd gladly discuss more to understand what are the blockers here if any.

no problem, only my concern about the position of icon in email item in list. But I think, design are good.

image

model/lib/email/presentation_email.dart
dab246: Updated

tddang-linagora (approved):

hoangdat:
was merged in #3532

Issue #3478, opened 2025-02-10
[Multiple Composer] [Dev] Create composer manager & convert singleton to multiton


dab246:
## DoD

- [x] Convert all class singleton in `ComposerBindings` to multion by use  `tag` of GetX
- [x] Create `ComposerManager` to manage all composer were created.

PR #3486, opened 2025-02-12
TF-3478 [Multiple Composer] Create composer manager

dab246:
## Issue

#3478 

## Resolved

- Opening a composer still works fine

https://github.com/user-attachments/assets/12556868-7ebd-456f-b246-87c384f11598

### DIFF
--- a/lib/features/composer/presentation/composer_bindings.dart
+++ b/lib/features/composer/presentation/composer_bindings.dart
@@ -78,6 +78,10 @@ import 'package:worker_manager/worker_manager.dart';
 
 class ComposerBindings extends BaseBindings {
 
+  final String? composerId;
+
+  ComposerBindings({this.composerId});
+
   @override
   void dependencies() {
     _bindingsUtils();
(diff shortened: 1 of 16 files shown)


lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -0,0 +1,35 @@
...
+      composers.remove(id);
+      log('ComposerManager::removeComposer:Success');
+    }
+  }
+
+  bool get hasComposer => composers.isNotEmpty;
+
+  List<String> get composerIds => composers.keys.toList();
+
+  String get singleComposerId => composerIds.first;
```
tddang-linagora, line 34: Is this temporary or would it be used in the future?

dab246: It will be used in the future, and may change.

tddang-linagora (approved):

lib/features/composer/presentation/composer_bindings.dart
```diff
@@ -89,43 +93,52 @@ class ComposerBindings extends BaseBindings {
...
       Get.find<FileUtils>(),
-    ));
+    ), tag: composerId);
   }
 
   @override
   void bindingsDataSourceImpl() {
     Get.lazyPut(() => AttachmentUploadDataSourceImpl(
-      Get.find<FileUploader>(),
+      Get.find<FileUploader>(tag: composerId),
```
hoangdat, line 107: Why we need a specific instance of FileUploader for each composer?

hoangdat: And not found the put for this.

dab246: > Why we need a specific instance of FileUploader for each composer?

Since it is only used in composer, we want to free it when not in use.

> And not found the put for this.

Please see `bindingUtils` method on the above

hoangdat:
- [ ] sometimes can not input in body of composer (maybe loading identity, because not see signature button)
image

dab246:
> * [x]  sometimes can not input in body of composer (maybe loading identity, because not see signature button)
>   

Cannot reproduce. IMO, your device is lagging. If possible, please describe the steps to reproduce.

hoangdat (approved):

Issue #3479, opened 2025-02-10
[Multiple Composer] [Dev] Display multiple composer view on screen at time


dab246:
## DoD

- [ ] Create multiple composer on screen
- [ ] Limit maximum composer view displayed (Fullscreen, Normal, Minimize).

PR #3489, opened 2025-02-13
TF-3479 [Multiple Composer] Display multiple composer view on screen at time 

dab246:
## Issue

#3479 

## Resolved 

https://github.com/user-attachments/assets/2dfb2e00-f34b-4981-b43c-7145cda4d36f


lib/features/composer/presentation/view/web/composer_overlay_view.dart
```diff
@@ -0,0 +1,81 @@
...
+  @override
+  void initState() {
+    super.initState();
+    _composerManager = Get.find<ComposerManager>();
+    _responsiveUtils = Get.find<ResponsiveUtils>();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    _composerManager.syncComposerStateWhenResponsiveChanged(
```
tddang-linagora, line 30: This method call is very expensive, as it run 2 `for` loops everytime `build` method is called. Should we debounce it?

dab246: > This method call is very expensive, as it run 2 `for` loops everytime `build` method is called. Should we debounce it?

Even though 2 `for` loops are used in the function, its complexity is only O(n), so IMO, it is not too expensive. Using debouce here we will consider its efficiency when using it, is it more efficient or not?

tddang-linagora: How about we let Flutter's stream help us optimize the performance with [distinct](https://api.flutter.dev/flutter/dart-async/Stream/distinct.html)?
- Move all the logic of `syncComposerStateWhenResponsiveChanged` to a stream, with `double screenWidth` is input, and a collection of composer controllers' state is output.
- Use the `distinct` method so that if the output of the previous state is the same as current state, the stream will not emit anything
- If there is new output from the stream, THEN we will update the composer controllers' state in the composer manager.
- Since the stream logic is handled asynchronously else where, it does not block the UI thread.

Let me know what you think.

dab246: > How about we let Flutter's stream help us optimize the performance with [distinct](https://api.flutter.dev/flutter/dart-async/Stream/distinct.html)?
> 
> * Move all the logic of `syncComposerStateWhenResponsiveChanged` to a stream, with `double screenWidth` is input, and a collection of composer controllers' state is output.
> * Use the `distinct` method so that if the output of the previous state is the same as current state, the stream will not emit anything
> * If there is new output from the stream, THEN we will update the composer controllers' state in the composer manager.
> * Since the stream logic is handled asynchronously else where, it does not block the UI thread.
> 
> Let me know what you think.

We don't have to do it too complicated, using `distinct` is not necessary, we just need to use `LayoutBuilder` to easily check if the width has changed or not? On the other hand, we need to wait for the calculation of the correct number of composers before re-rendering it, to avoid the case of overflow width size.

tddang-linagora: >we just need to use LayoutBuilder

Sure let's apply this change.

chibenwa:
I love the demo video!

It loks awseome, koodo @dab246

lib/features/composer/presentation/utils/composer_utils.dart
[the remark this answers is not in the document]
dab246: Removed

lib/features/composer/presentation/view/web/composer_overlay_view.dart
dab246: Done

tddang-linagora (approved):

dab246:
- [ ] Memory leak `IdentityInteractorsBindings`

tddang-linagora (approved):

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+      if (!controller.isHiddenScreen) {
+        countDisplayedComposer++;
+      }
+      composerControllers[id] = controller;
+      totalWidth += composerWidth;
+    }
+
+    final availableWidth = screenWidth
+      - ComposerStyle.composerExpandMoreButtonMaxWidth
+      - ComposerStyle.padding * 3
```
hoangdat, line 94: Why need `x * 3`? can you constant it?

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+      .toList()
+      .reversed
+      .firstWhereOrNull((id) => getComposerView(id).controller.isNormalScreen);
+  }
+
+  (
+    double availableWidth,
+    double totalWidth,
+    Map<String, ComposerController> composerControllers
+  ) _setUpDataToSyncComposerQueue({required double screenWidth}) {
```
hoangdat, line 77:
```suggestion
  (
    double availableScreenWidth,
    double totalOpenedComposersWidth,
    Map<String, ComposerController> composerControllers
  ) _setUpDataToSyncComposerQueue({required double screenWidth}) {
```

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+          controller.setScreenDisplayMode(ScreenDisplayMode.hidden);
+          return;
+        }
+
+        totalWidth = newTotalWidth;
+      }
+    }
+  }
+
+  void syncComposerStateWhenResponsiveChanged({required double screenWidth}) =>
```
hoangdat, line 203:
```suggestion
  void arrangeComposerWhenResponsiveChanged({required double screenWidth}) =>
```

hoangdat:
- [ ] in this case, composer manager cover all email actions: Reply, Forward, Reply All.  
Screenshot 2025-03-03 at 09 20 49

Do we have any idea for this?

lib/features/composer/presentation/manager/composer_manager.dart
dab246: Oh, it should be `x * 2`. that is horizontal padding. We just need to const the padding value on one side.

dab246: Update to `ComposerStyle.padding * 2`

lib/features/composer/presentation/manager/composer_manager.dart
dab246: Updated

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+      - (countDisplayedComposer - 1) * ComposerStyle.space;
+
+    return (
+      availableWidth,
+      totalWidth,
+      composerControllers,
+    );
+  }
+
+  void syncComposerStateWhenComposerQueueChanged({required double screenWidth}) {
```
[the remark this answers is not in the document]
dab246: Done

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+          controller.setScreenDisplayMode(ScreenDisplayMode.hidden);
+          totalWidth = newTotalWidth;
+        }
+
+        if (totalWidth <= availableWidth) break;
+      }
+    }
+  }
+
+  void syncComposerStateWhenComposerDisplayModeChanged({
```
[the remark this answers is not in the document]
dab246: Done

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+      if (currentHiddenComposerIds.isEmpty) return;
+
+      for (var index = currentHiddenComposerIds.length - 1; index >= 0; index--) {
+        final newTotalWidth = totalWidth + ComposerStyle.minimizeWidth;
+        if (newTotalWidth > availableWidth) break;
+
+        final id = currentHiddenComposerIds.elementAt(index);
+        composerControllers[id]!.setScreenDisplayMode(ScreenDisplayMode.minimize);
+        totalWidth = newTotalWidth;
+      }
```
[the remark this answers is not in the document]
dab246: Done

lib/features/composer/presentation/manager/composer_manager.dart
```diff
@@ -1,35 +1,280 @@
...
+          return;
+        }
+
+        newTotalWidth = totalWidth - ComposerStyle.normalWidth;
+        if (newTotalWidth <= availableWidth) {
+          controller.setScreenDisplayMode(ScreenDisplayMode.hidden);
+          return;
+        }
+
+        totalWidth = newTotalWidth;
```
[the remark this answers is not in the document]
dab246: Done

lib/features/composer/presentation/manager/composer_manager.dart
dab246: Done

dab246:
> * [x]  in this case, composer manager cover all email actions: Reply, Forward, Reply All.

> Do we have any idea for this?

IMO, we will add a "Hide" icon to the composers so that they can be hidden. Watch the demo video below. @hoangdat WDYT?



https://github.com/user-attachments/assets/a8bae3b5-ba07-4fba-9078-175675d82a5d

dab246:
- [ ] Verify logic hide -> open composer from list


https://github.com/user-attachments/assets/a1a7c389-24e5-4d86-92a8-ef7da86de1a0

dab246:
> * [x]  Verify logic hide -> open composer from list
> 

Fixed

hoangdat (approved):

PR #3496, opened 2025-02-19
Initial commit add sievescript_dart

aikuviet:


viethoang99:
ok

chibenwa:
Hello @viethoang99 

It looks very interesting to add sieve script support for twake mail if supported by the server.

The contribution would be gadly welcomed.

Would you be willing to contribute this?

chibenwa:
Second remark: would you be willing to contribute JMAP and blob support to https://github.com/linagora/jmap-dart-client ?

I do not see the point of an isolated JMAPClient solely for managing sieve script and blobs.

I checked license alignmenet: both are MIT! It make sense of having only one, stronger library for supporting JMAP in dart!

Issue #3472, opened 2025-02-05
[MU] Improve reading mail on small screen with attachments


chibenwa:
https://github.com/user-attachments/assets/b1380eac-a0a7-4081-8770-76b7c1de8573

Resolution

![Image](https://github.com/user-attachments/assets/dad0b864-8592-4010-aabb-0b2446153a3d)

Having two rows for attachment wastes space on those constrained screens.

We should make the display only one row

![Image](https://github.com/user-attachments/assets/7df9cc57-5430-4e41-87cf-1274f4aad204)

hoangdat:
@chibenwa 
How about add one more button `hide all` to collapse attachment viewer to single line header `11 attachments` with `Show all` button ?

hoangdat:
Confirmed by @Bobpodvalnyi 

Keep 1 line of attachment as a workaround solution

PR #3530, opened 2025-03-04
TF-3472 Prevent attachments from wrapping 2nd line

tddang-linagora:
## Issue
- #3472 

## Demo

https://github.com/user-attachments/assets/270b0aa9-23e7-48a1-8f6e-a37504cfdb6e

### DIFF
--- a/lib/features/email/presentation/styles/attachment/attachment_item_widget_style.dart
+++ b/lib/features/email/presentation/styles/attachment/attachment_item_widget_style.dart
@@ -31,4 +31,23 @@ class AttachmentItemWidgetStyle {
     color: AppColor.attachmentFileSizeColor,
     fontWeight: FontWeight.normal
   );
+
+  static double getMaxWidthItem({
+    required bool platformIsMobile,
+    required bool responsiveIsMobile,
+    required bool responsiveIsTablet,
+    required bool responsiveIsTabletLarge,
+  }) {
+    if (platformIsMobile) {
+      return responsiveIsMobile ? maxWidthMobile : maxWidthTablet;
+    } else {
+      if (responsiveIsTabletLarge) {
+        return maxWidthTabletLarge;
+      } else if (responsiveIsTablet) {
+        return maxWidthTablet;
+      } else {
+        return maxWidthMobile;
+      }
+    }
+  }
 }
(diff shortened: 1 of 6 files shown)


core/lib/presentation/utils/responsive_utils.dart
```diff
@@ -109,4 +110,48 @@ class ResponsiveUtils {
...
+    ).attachToRenderTree(buildOwner);
+    
+    buildOwner.buildScope(element);
+    buildOwner.finalizeTree();
+    
+    pipelineOwner.flushLayout();
+    pipelineOwner.flushCompositingBits();
+    pipelineOwner.flushPaint();
+    
+    return renderBox.size;
```
dab246, line 155: Using `PipelineOwner` is too expensive in terms of performance and memory, in this case we don't need to use it, we have many other more efficient ways of calculating. Suggestion: Try with `GlobalKey` or another size calculation.

tddang-linagora: This calculate method is called before the widget is laid out, hence `GlobalKey` solution won't work. I improved the calculate method performance.

dab246: Use set maxWidth for `More` button as const

tddang-linagora: done

core/lib/presentation/utils/responsive_utils.dart
```diff
@@ -109,4 +110,20 @@ class ResponsiveUtils {
...
+    } else {
+      if (isTabletLarge(context)) {
+        return 200;
+      } else if (isTablet(context)) {
+        return 240;
+      } else {
+        return 260;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
tddang-linagora: Done

dab246 (approved):
- [x] Rebase

hoangdat:
- [ ] still prefer at least one attachment for mobile 

Screenshot_2025-03-07-19-09-58-755_com linagora android teammail

dab246:
> * [x]  still prefer at least one attachment for mobile
> 
 Fixed



https://github.com/user-attachments/assets/92f90aeb-955e-4696-8a4b-c20484f66574

hoangdat (approved):

PR #3593, opened 2025-03-24
Hot fix after some reload minimized composer should be lost body

dab246:
## Issue

The `Save draft` dialog does not appear when reloading multiple `Minimize` and `Hide` composers and then clicking `X` to close composer

## Root cause

Because in `minimize` and `hide` state the value `_isEmailBodyLoaded = false` because `WebView` is not displayed. So we closed composer without doing anything.

## Solution

Update logic when close composer

## Resolved

https://github.com/user-attachments/assets/5404e70f-d156-40c4-8c87-f02639efb022

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1127,6 +1127,13 @@ class ComposerController extends BaseController
     emailContent = await _composerRepository.removeCollapsedExpandedSignatureEffect(
       emailContent: emailContent,
     );
+    if (emailIdEditing != null &&
+        savedActionType == EmailActionType.compose &&
+        currentEmailActionType == EmailActionType.reopenComposerBrowser) {
+      emailContent = await _composerRepository.removeStyleLazyLoadDisplayInlineImages(
+        emailContent: emailContent,
+      );
+    }
 
     final savedEmailDraft = SavedEmailDraft(
       subject: subjectEmail.value ?? '',
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -2,7 +2,6 @@
 import 'dart:async';
 import 'dart:math';
 
-import 'package:collection/collection.dart';
 import 'package:core/core.dart';
 import 'package:custom_pop_up_menu/custom_pop_up_menu.dart';
 import 'package:dartz/dartz.dart';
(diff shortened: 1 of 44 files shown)


tddang-linagora (dismissed):

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -2235,6 +2235,18 @@ class ComposerController extends BaseController
     }
   }
 
+  bool _validateCloseComposerWithoutSave() {
+    if (composerArguments.value == null) return true;
+
+    if (PlatformInfo.isWeb &&
+        !_isEmailBodyLoaded &&
+        !screenDisplayMode.value.isNotContentVisible()) return true;
```
hoangdat, line 2243: with it, save will be called all the time. please make sure, user dont need to `save` if nothing change

dab246: It's not a condition to save, just a condition to be able to close immediately or not. The condition you mentioned is below.

hoangdat: https://github.com/user-attachments/assets/d1a5ee94-6919-4db8-809f-bea4b43f39d5

dab246: Fixed


https://github.com/user-attachments/assets/57082302-cfb0-4f98-81ff-55bb0e7d4779

tddang-linagora (dismissed):

hoangdat:
- [ ] identity was not get well after open from draft

https://github.com/user-attachments/assets/88cc48ea-6630-4580-b362-f019333aa587

hoangdat:
- [ ] name with XSS script have problem when forwarding

Uploading Screen Recording 2025-03-31 at 15.03.17.mov…

hoangdat:
- [ ] missing `to`, `reply-to` information when forwarding 
Screenshot 2025-03-31 at 15 10 50

tddang-linagora:
https://github.com/user-attachments/assets/4177611c-c7aa-4934-9300-c464caafc8e7

Broken inline images when reloading hidden composers.

https://github.com/user-attachments/assets/426a2e54-a8ca-418d-a15d-5f73ab7b5f5b

Also happens with minimized composers.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -647,206 +596,47 @@ class ComposerController extends BaseController
...
-      } else if (composerArguments.value?.emailActionType != EmailActionType.editDraft) {
-        _getServerSetting();
-      }
+    if (screenDisplayMode.value.isNotContentVisible()) {
+      await setupSelectedIdentityWithoutApplySignature();
     }
   }
 
-  void _initSubjectEmail({
+  void initSubjectEmail({
```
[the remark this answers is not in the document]
dab246: Removed

dab246:
> * [x]  name with XSS script have problem when forwarding
> 

Fixed



https://github.com/user-attachments/assets/eec7aa83-a3b3-43a5-ab40-8c1adc746592

lib/features/email/data/local/html_analyzer.dart
```diff
@@ -237,4 +237,30 @@ class HtmlAnalyzer {
...
   }
+
+  Future<String> removeStyleLazyLoadDisplayInlineImages({required String emailContent}) async {
+    try {
+      final document = parse(emailContent);
+      final imgElements = document.querySelectorAll('img[style], img[loading]');
+      await Future.wait(imgElements.map((img) async {
+        String? style = img.attributes['style'];
+        if (style != null) {
+          style = style.replaceAll(RegExp(r'display\s*:\s*inline;?'), '').trim();
```
tddang-linagora, line 248: Please create a variable somewhere for this `RegExp` with name explaining what does it find.

dab246: IMO, we don't need to create it because it is already specified in the handler function name and it is specific to this function only.

dab246: We don't need to reuse it, creating variables also takes up more memory

tddang-linagora (approved):

hoangdat:
- [ ] missing BCC in reading email and replying email too

https://github.com/user-attachments/assets/54aaa7a4-9247-4cda-a67f-1a46a7f9d9c9

dab246:
> * [x]  missing BCC in reading email and replying email too
> 
>

After detect has discovered, we sent `Bcc` when calling `Email/set`, but when `Email/get` retrieved (second time) it did not have `Bcc`. Also tried on version `0.14.15` also got the same error. When performing cache clearing to retrieve `Email/get`, the email also did not have `Bcc`. Therefore, it is not a side effect error from this PR, we should create PR to fix it.


Screenshot 2025-04-01 at 11 15 04


Screenshot 2025-04-01 at 11 15 21

tddang-linagora (approved):

hoangdat (approved):

Issue #3507, opened 2025-02-24
Offline cache prevents updating EmailQueryView cache


chibenwa:
### The issue

```
GIVEN I logged in the app
AND I forget it in background mode for a while
AND I receive a few messages that I read / sort / move via Webmail
WHEN I resync the mobile app
THEN some emails I managed are still displayed
```

It  happens only on mobile and not on other platforms hence I suspect it to be linked with offline cache.

I reproduce this on a daily basis.

### Definition of done

Offline cache should not prevent updating EmailQueryView cache

chibenwa:
Cc @hoangdat could be GREAT to include this in next frontend release...

tddang-linagora:
>WHEN I resync the mobile app

Does this mean you pull to refresh the email view or you simply reopen the app from background?

dab246:
## Analysis

- When we perform too many actions on multiple emails `(Mark as read, mark as star, move, delete, etc.)`, `Email/get` from the value of `Email/changes` will throw an error.

<details>
<summary>Email/changes response</summary>
{
    "sessionState": "2c9f1b12-b35a-43e6-9af2-0106fb53a943",
    "methodResponses": [
        [
            "Email/changes",
            {
                "accountId": "694c6e72a66a8406190dac6e963019569cfee60931c113bf2e1515583a8cc988",
                "oldState": "093a9280-0f71-11f0-bbfa-5bbe2cd2bc40",
                "newState": "48f2de90-0f72-11f0-bbfa-5bbe2cd2bc40",
                "hasMoreChanges": true,
                "created": [
                    "3c479290-0f71-11f0-bbfa-5bbe2cd2bc40",
                    "379538a0-0f72-11f0-bbfa-5bbe2cd2bc40",
                    "3611a5e0-0f72-11f0-bbfa-5bbe2cd2bc40",
                    "cf68ac30-0f71-11f0-bbfa-5bbe2cd2bc40",
                    "7ea837c0-0f71-11f0-bbfa-5bbe2cd2bc40",
                    "85686da0-0f71-11f0-bbfa-5bbe2cd2bc40",
                    "ce0376e0-0f71-11f0-bbfa-5bbe2cd2bc40"
                ],
                "updated": [
                    "cef717a0-8cf9-11ef-baaf-9db063fcbfec",
                    "a4f59390-0a59-11f0-bbfa-5bbe2cd2bc40",
                    "211132f0-0a1d-11f0-bbfa-5bbe2cd2bc40",
[...]

PR #3615, opened 2025-04-03
TF-3507 Fix offline cache prevents updating email query view cache

dab246:
## Issue

#3507

## Root cause

- When we perform too many actions on multiple emails `(Mark as read, mark as star, move, delete, etc.)`, `Email/get` from the value of `Email/changes` will throw an error.

```json
 [
            "error",
            {
                "type": "requestTooLarge",
                "description": "Too many items in an email read at level FULL. Got 183 items instead of maximum 100."
            },
            "c2"
        ]
```

- Some emails in the `update` of the `Email/change` response were not found in `Email/get`

Image

- But we still update `newState` of email 

Image

## Solution

- Perform 2 actions when `Pull to refresh`:
  - `Refresh` : Call get all emails
  - `Deep refresh`: Clean email cache & call get all emails

## Resolved

[demo.webm](https://github.com/user-attachments/assets/5fb39309-fddd-4768-8127-fffb28ae87e5)

### DIFF
--- a/core/lib/presentation/views/pull_to_refresh/pull_to_refresh_widget.dart
+++ b/core/lib/presentation/views/pull_to_refresh/pull_to_refresh_widget.dart
@@ -13,11 +13,9 @@ class PullToRefreshWidget extends StatefulWidget {
   final String normalRefreshText;
   final String deepRefreshText;
   final String pullDownToRefreshText;
-  final String refreshingText;
-  final String pullingText;
   final String releaseToText;
-  final String pullFurtherForText;
   final String pullHarderForText;
+  final Icon? pullDownToRefreshIcon;
   final Icon? normalRefreshIcon;
   final Icon? deepRefreshIcon;
 
@@ -26,15 +24,13 @@ class PullToRefreshWidget extends StatefulWidget {
     required this.child,
     required this.onNormalRefresh,
     required this.onDeepRefresh,
-    this.deepRefreshThreshold = 200.0,
+    this.deepRefreshThreshold = 150.0,
     this.normalRefreshText = 'Refresh',
     this.deepRefreshText = 'Deep refresh',
     this.pullDownToRefreshText = 'Pull down to refresh',
-    this.refreshingText = 'Refreshing',
-    this.pullingText = 'Pulling',
     this.releaseToText = 'Release to',
-    this.pullFurtherForText = 'Pull further for',
     this.pullHarderForText = 'Pull harder for',
+    this.pullDownToRefreshIcon = const Icon(Icons.arrow_downward, size: 20),
     this.normalRefreshIcon = const Icon(Icons.refresh, size: 20),
     this.deepRefreshIcon = const Icon(Icons.autorenew, size: 20),
   }) : super(key: key);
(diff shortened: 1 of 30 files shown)


core/lib/presentation/views/pull_to_refresh/pull_to_refresh_widget.dart
```diff
@@ -0,0 +1,153 @@
...
+  String? _lastAction;
+
+  @override
+  Widget build(BuildContext context) {
+    return NotificationListener<ScrollNotification>(
+      onNotification: (notification) {
+        if (notification is ScrollUpdateNotification) {
+          setState(() {
+            _scrollOffset = notification.metrics.pixels;
+          });
```
tddang-linagora, line 58: Should we only call `setState` when pass certain threshold? To avoid rebuild the widget too many time unnecessarily.

dab246: No, we must always update this value to check the current pull to refresh display state

tddang-linagora: For example, `deepRefreshThreshold` is 200. If `_scrollOffset` is changing any where between 0-199, the deep search won't be activated. However, even 1 pixel change for `_scrollOffset` triggers `setState`. We can simply optimize this callback with:
```suggestion
          if (notification.metrics.pixels > widget.deepRefreshThreshold && _scrollOffset.abs() <= widget.deepRefreshThreshold) {
            setState(() {
              _scrollOffset = notification.metrics.pixels;
            });
          } else if (notification.metrics.pixels <= widget.deepRefreshThreshold && _scrollOffset.abs() > widget.deepRefreshThreshold) {
            setState(() {
              _scrollOffset = notification.metrics.pixels;
            });
          } 
```

chibenwa:
MUCH NEEDED

That's a day to day pain to me and Alex expressed similar concerns today.

tddang-linagora (approved):
LGTM

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -101,11 +104,11 @@ class ThreadController extends BaseController with EmailActionController {
   bool canSearchMore = false;
   MailboxId? _currentMemoryMailboxId;
   final ScrollController listEmailController = ScrollController();
-  final FocusNode focusNodeKeyBoard = FocusNode();
   final latestEmailSelectedOrUnselected = Rxn<PresentationEmail>();
   @visibleForTesting
   bool isListEmailScrollViewJumping = false;
   WebSocketQueueHandler? _webSocketQueueHandler;
+  FocusNode? focusNodeKeyBoard;
```
hoangdat, line 111: why we need this?

hoangdat:
- [ ] Try to avoid screen blink when refreshing email list

lib/features/thread/presentation/thread_controller.dart
dab246: This variable was already initialized (used to listen for the `Shift` key event to select email). I just switched it to `optional` and changed its position.

dab246:
> * [x]  Try to avoid screen blink when refreshing email list

[demo-refresh.webm](https://github.com/user-attachments/assets/cac49a2c-dfe2-4e0a-8789-c28cd57be2fa)

hoangdat (approved):

tddang-linagora (approved):

hoangdat:
@chibenwa

PR #3616, opened 2025-04-03
Fix do not call refresh token or show session expires dialog when token expires in composer

dab246:
## Issue

```
When typing recipient in composer
Then receive token expires 401
And refresh token do not call, session expires dialog doesn't display
```

- Reproduced:

https://github.com/user-attachments/assets/cc481856-a26b-4c0d-bbc6-9a61cf32568d

## Root cause

Because we are handling `autocomplete` and getting the result directly from `await Future`, don't use `comsumeState` in `BaseController`, so the missing handler error token expired

## Resolved

https://github.com/user-attachments/assets/565e0310-adcf-42f5-90a1-e8224a6efac4


tddang-linagora (approved):

hoangdat:
Should also handle it for
- upload inline images
- upload attachments

tddang-linagora:
No logout dialog on 401 with cases

- [ ] Input recipients
- [ ] Upload inline image
- [ ] Upload attachment

https://github.com/user-attachments/assets/2f63b634-dd07-443b-917e-f34ed7ee3ec4

Config: dev env, refresh token disabled, Safari 18

dab246:
> No logout dialog on 401 with cases
> 
> * [x]  Input recipients
> * [x]  Upload inline image
> * [x]  Upload attachment
> 

Fixed. Please see demo:

- Autocomplete:



https://github.com/user-attachments/assets/46e70de9-d522-4bee-a0ce-8e32a712fc64


- Inline image


https://github.com/user-attachments/assets/5116a941-ef57-4a71-a8f9-d399c1b44f9e



-  Attachment:




https://github.com/user-attachments/assets/fcf1325c-1064-4f22-bbfa-c9df283b4e8c

lib/features/upload/domain/model/upload_attachment.dart
```diff
@@ -57,12 +60,16 @@ class UploadAttachment with EquatableMixin {
...
-      logError('UploadAttachment::upload():ERROR: $e');
-      if (e is DioError && e.type == DioErrorType.cancel) {
+    } catch (error, stackTrace) {
+      logError('UploadAttachment::upload():ERROR: $error');
+      if (error is DioError && error.type == DioErrorType.cancel) {
         _updateEvent(Left(CancelAttachmentUploadState(uploadTaskId)));
       } else {
-        _updateEvent(Left(ErrorAttachmentUploadState(uploadId: uploadTaskId)));
+        try {
+          exceptionThrower.throwException(error, stackTrace);
```
hoangdat, line 69: I found that it was caught in `lib/features/upload/data/datasource_impl/attachment_upload_datasource_impl.dart`. So we only need to throw exception here

dab246: No, we execute multiple upload functions at the same time, so the try/catch in the datasource only catches the exception at the time the datasource finishes calling. Moreover, the `upload` function inside has been try/catch to update the result for StreamController, so we need to handle the exception in this function.

lib/features/upload/presentation/controller/upload_controller.dart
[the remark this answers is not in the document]
dab246: It is handled together with the commitState of the BaseController. And it only happens when one of the logics in the interactor, repository throws an error right at the moment it completes.

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -88,6 +88,7 @@ class UploadController extends BaseController {
                 (currentState) => currentState?.copyWith(uploadStatus: UploadFileStatus.uploadFailed));
             deleteFileUploaded(failure.uploadId);
             _showToastMessageWhenUploadAttachmentsFailure(failure);
+            consumeState(Stream.value((Left(failure))));
```
hoangdat, line 91: how about to consume other `failure` state too?

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -143,6 +144,7 @@ class UploadController extends BaseController {
               leadingSVGIconColor: Colors.white,
               leadingSVGIcon: imagePaths.icInsertImage);
           }
+          consumeState(Stream.value((Left(failure))));
```
hoangdat, line 147: idem

dab246:
- [ ] Check add inline image in identity

lib/features/upload/presentation/controller/upload_controller.dart
dab246: With upload only single errors from returned events are placed in ErrorAttachmentUploadState

lib/features/upload/presentation/controller/upload_controller.dart
dab246: With upload only single errors from returned events are placed in ErrorAttachmentUploadState

lib/features/upload/presentation/controller/upload_controller.dart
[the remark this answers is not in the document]
hoangdat: but with `UploadAttachmentFailure` I only see that we showing toast, not do anthing else

lib/features/upload/presentation/controller/upload_controller.dart
hoangdat: so put it out of if is also fair

lib/features/upload/presentation/controller/upload_controller.dart
hoangdat: idem

lib/features/upload/presentation/controller/upload_controller.dart
dab246: No, we have case Cancel upload

tddang-linagora (dismissed):

lib/features/upload/presentation/controller/upload_controller.dart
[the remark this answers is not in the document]
dab246: No `UploadAttachmentFailure` was handled with the BaseController's commitState, it handled the exception before returning onFailureView

dab246:
> * [x]  Check add inline image in identity ❌
> 


Fixed


https://github.com/user-attachments/assets/38837b22-27d8-415c-8293-d63a181226e3

tddang-linagora (approved):

dab246:
> please rebase

Done

dab246:
- [x] Work well on multiple composer

dab246:
- [x] Reconnect when sending/save as draft email with error 401




https://github.com/user-attachments/assets/63880599-3aa7-4937-8cf3-2f60e3452d07

tddang-linagora (approved):

hoangdat (approved):

PR #3619, opened 2025-04-04
[HOT-FIX] Email content in `text/plain` format does not automatically wrap when displayed on mobile

dab246:
## Issue

Email content in `text/plain` format does not automatically wrap when displayed on mobile

## Root cause

From fixing the display of `markdown` table content, we always use the `pre` tag to wrap the email content, resulting in the email content not automatically wrapping.

## Solution

We will check if the email content is a `markdown` table or `ASCII art` table. If so, we will use `style` to display as markdown, otherwise we will leave the text as is.

## Resolved

- Mobile:

- Web:

Shared

### DIFF
--- a/core/lib/presentation/utils/html_transformer/sanitize_autolink_filter.dart
+++ b/core/lib/presentation/utils/html_transformer/sanitize_autolink_filter.dart
@@ -32,7 +32,6 @@ class SanitizeAutolinkFilter {
         options: _linkifyOption,
         linkifiers: _linkifier,
       );
-      log('AutolinkFilter::process:elements: $elements');
       final htmlTextBuffer = StringBuffer();
 
       for (var element in elements) {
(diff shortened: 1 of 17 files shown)


core/lib/presentation/utils/html_transformer/sanitize_autolink_filter.dart
```diff
@@ -53,19 +52,20 @@ class SanitizeAutolinkFilter {
...
+      return textSanitized;
     } catch (e) {
       logError('$runtimeType::process:Exception = $e');
       return inputText;
     }
   }
 
   String _buildUrlLinkTag({required String urlLink, required String value}) {
-    return '<a href="$urlLink" target="_blank" rel="noreferrer">$value</a>';
+    return '<a href="$urlLink" target="_blank" rel="noreferrer" style="white-space: nowrap; word-break: keep-all;">$value</a>';
```
tddang-linagora, line 65: Would we not want to break the line of url?

dab246: It should wrap to the width of the screen.

tddang-linagora: So it should be `word-break: break-all;` not `word-break: keep-all;`?

dab246: Never break between words, only break lines at spaces

tddang-linagora (approved):

hoangdat (approved):

Issue #3621, opened 2025-04-04
[Bug][Search] return wrong result if searching with option "Doesn' hava"


thuyenh1081:
### Environment

https://mail.stg.lin-saas.com/  version 0.15.0
https://tmail.linagora.com/
I tested on Chrome

#### Reproduction Steps
1. Open advance search
2. Fill into textbox "Doesn't have" by the text that available as a part/whole of  email'subject
3. Click on Search button
4. See that email that contains the text in textbox "Doesn't have" is still shown in the result

### Expected result
The mail that has text in "Doesn't have" text box will not be shown in the result in this case

### Attached video

XXX

PR #3626, opened 2025-04-08
TF-3621 Fix search return wrong result with option does not have

dab246:
## Issue

#3621 

## Root cause 

Incorrect use of the `notKeyword` filter when searching with `Doesn't have`

## Solution

Use `text` filter with `Operator.NOT` replace for `notKeyword`

## Resolved

https://github.com/user-attachments/assets/3b322854-5690-4e1a-9717-2060cffaee3c


chibenwa (approved):

tddang-linagora (dismissed):

lib/features/thread/presentation/thread_view.dart
```diff
@@ -675,7 +675,18 @@ class ThreadView extends GetWidget<ThreadController>
...
+          return EmptyEmailsWidget(
+            key: const Key('empty_thread_view'),
+            isNetworkConnectionAvailable: controller.networkConnectionController.isNetworkConnectionAvailable(),
+            isSearchActive: controller.isSearchActive,
+            isFilterMessageActive: controller.mailboxDashBoardController.filterMessageOption.value != FilterMessageOption.all,
+            onCreateFiltersActionCallback: controller.goToCreateEmailRuleView,
+          );
+        }
+        return const SizedBox.shrink();
+      },
```
hoangdat, line 689: remove it

hoangdat: Handle it in #3646 #3624

dab246:
- [x] Add unit test for case multiple `notKeyword`

Issue #3640, opened 2025-04-10
Increase composer size in case of single composer openning


hoangdat:
### Desc
- user want to increase composer size in case of single composer
- but still want to leverage multiple composers

### DoD
- [x] increase the composer size in case of single composer opening as before
- [x] rearrange/resize multiple composer if user open more than one composer

PR #3641, opened 2025-04-10
TF-3640 Increase composer size when there is only one composer

dab246:
## Issue

#3640 

## Resolved

https://github.com/user-attachments/assets/88c7ba9f-46f7-41fd-942b-470551407d44


tddang-linagora (commented):
- [x] Should also have the composer normal size when 1 is `ScreenDisplayMode.normal` and the rest is `ScreenDisplayMode.hidden`

dab246:
> * [ ]  Should also have the composer normal size when 1 is `ScreenDisplayMode.normal` and the rest is `ScreenDisplayMode.hidden`

No we only need to increase composer size when there is only 1 composer

tddang-linagora (approved):

hoangdat (approved):

Issue #3659, opened 2025-04-15
Cannot remove forwards


chibenwa:
https://github.com/user-attachments/assets/e1e4deb9-2c66-4941-8adf-5f788df63067

https://tmail.linagora.com/ v 0.15.1

hoangdat:
@dab246  this issue relate to `positive` - `negative` buttons in confirm dialog. 
Please confirm. If yes, re-check all the confirm dialog.

dab246:
> [@dab246](https://github.com/dab246) this issue relate to `positive` - `negative` buttons in confirm dialog. Please confirm. If yes, re-check all the confirm dialog.

Oh, maybe it was affected by PR update new theme. I fixed it.

PR #3414, opened 2025-01-09
TF-3410 Reduce color noise and user load

dab246:
## Issue

#3410 

## Resolved

https://github.com/user-attachments/assets/3c83c4d4-71a8-43ee-b831-3f086119b882

PR #3662, opened 2025-04-16
TF-3659 Fix cannot remove forwards

dab246:
## Issue

#3659

## Reproduced

https://github.com/user-attachments/assets/df692935-a9e5-4549-8deb-4ebd99cb62a3

## Root cause

- Side effects from commit: https://github.com/linagora/tmail-flutter/pull/3414/commits/fec267dcbf8982c4ea98b2e24a524da7a14ed3d9#diff-962af98b988bd676daa119520af50b0646b0db6457032ba108275c00fad16838L222-R204

Image

- It only happens for the `showConfirmDialogAction` function of `MessageDialogActionMixin` when the conditions are satisfied:
```
+ showAsBottomSheet = true
+ alignCenter = false
+ hasCancelButton = true
+ autoPerformPopBack = true
+ responsiveUtils.isMobile(context) = false
+ action is negative button
```

So only the `remove forwards` action is affected.

## Resolved

https://github.com/user-attachments/assets/1855fb19-a081-4668-88be-879d8ee4a889

### DIFF
--- a/lib/features/base/mixin/message_dialog_action_mixin.dart
+++ b/lib/features/base/mixin/message_dialog_action_mixin.dart
@@ -206,9 +206,8 @@ mixin MessageDialogActionMixin {
             onCancelButtonAction: () {
               if (autoPerformPopBack) {
                 popBack();
-              } else {
-                onCancelAction?.call();
               }
+              onCancelAction?.call();
             },
             onCloseButtonAction: onCloseButtonAction ?? popBack,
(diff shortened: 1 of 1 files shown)


tddang-linagora (approved):

hoangdat:
- [x] tested with all `confirm dialog`

hoangdat (approved):

Issue #3627, opened 2025-04-08
[MULTI-COMPOSER] 2 scroll bar with differnet look


chibenwa:
![Image](https://github.com/user-attachments/assets/6d8acd2e-853a-40ca-92eb-db9d9221b156)

Note that I am fine with having 2 scroll bar, but they should have the same look IMO

Cc @hoangdat

PR #3664, opened 2025-04-16
TF-3627 Fix two scrollbar with different look

dab246:
## Issue

#3627

## Dependency

- Need merged: https://github.com/linagora/html-editor-enhanced/pull/51

## Solution

We use `::-webkit-scrollbar-*` to change the style of the html scrollbar. But on `Firefox` (on some operating systems like `Windows`) does not support `::-webkit-scrollbar-* `. Instead we use `scrollbar-width` and `scrollbar-color` but it cannot change the `border-radius` of the scrollbar. So by using `::-webkit-scrollbar-*` combined with `scrollbar-width` and `scrollbar-color` we have done:
- Change the style of the scrollbar on all browsers on `MacOS, Windows, Linux` operating systems (except` Firefox`) but always shown scrollbar when it exceeds size container
- Only the `Firefox` browser on `Windows and Linux` can only change the color and size.

**Note**
- There might be a way to change the style of Firefox on Windows, but we need a third party JavaScript library and it's quite a hassle, so we choose the simplest way, which supports most browsers, which is to use CSS style for html.

## Resolved

- All browser on MacOS:

https://github.com/user-attachments/assets/caa2f644-ce1d-467e-ba36-ca4c77617ffb

- Chrome on Windows

https://github.com/user-attachments/assets/befe3af9-b445-4780-b322-10cdc503687a

- Firefox on Windows

https://github.com/user-attachments/assets/54f41860-12cf-4d4e-b888-4301626d3619

### DIFF
--- a/core/lib/presentation/utils/theme_utils.dart
+++ b/core/lib/presentation/utils/theme_utils.dart
@@ -16,8 +16,8 @@ class ThemeUtils {
       dividerTheme: _dividerTheme,
       visualDensity: VisualDensity.adaptivePlatformDensity,
       scrollbarTheme: ScrollbarThemeData(
-        thickness: WidgetStateProperty.all(8.0),
-        radius: const Radius.circular(8.0),
+        thickness: WidgetStateProperty.all(6.0),
+        radius: const Radius.circular(10.0),
         thumbColor: WidgetStateProperty.all(AppColor.thumbScrollbarColor)),
     );
   }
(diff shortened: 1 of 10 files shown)


hoangdat:
- [ ] scrollbar of html editor seems never hide - Chrome 
image

hoangdat (dismissed):

tddang-linagora (dismissed):

tddang-linagora (approved):

hoangdat (approved):

hoangdat:
I also created a ticket to improve it also in some other place in app 

https://github.com/linagora/tmail-flutter/issues/3666

Issue #3674, opened 2025-04-20
REPLY TO LIST


chibenwa:
https://github.com/user-attachments/assets/4e344fbd-55c3-48a8-a75b-af7181bc3c60

DOD repair reply to list like TB

And thanks by the way because the team **do not write enough test** I answered to the list instead of to @quantranhong1999

 :-( 👎 😠

dab246:
> DOD repair reply to list like TB
> 
> And thanks by the way because the team **do not write enough test** I answered to the list instead of to [@quantranhong1999](https://github.com/quantranhong1999)
> 
> :-( 👎 😠

Hi @chibenwa, first of all we apologize for any inconvenience this feature may cause you. But actually we have compared with TB before implementing it, and we have written **unit test** for it. 

<img width="703" alt="Image" src="https://github.com/user-attachments/assets/ffd1180f-5cab-4b10-829a-83b36565d7bd" />

<img width="703" alt="Image" src="https://github.com/user-attachments/assets/7d86b5f7-19b7-463b-9d0c-9c1a7ced9df0" />


So to avoid further errors, mistakes. We REALLY NEED you to verify again about the LOGIC of `Reply / Reply To List / Reply all` email as follows, right?

**1. Reply:**

- If the user is the sender

```console
To = email.to
Cc, Bcc, ReplyTo = Empty
```

- Otherwise

    _If email.replyTo is not empty then:_
    
    ```console
    To = email.replyTo.withoutUser
    Cc, Bcc, ReplyTo = Empty
    ```
    
    _Otherwise:_
    
    ```console
    To = email.from.withoutUser
    Cc, Bcc, ReplyTo = Empty
    ```

**2. Reply To List:**

- If the email has a `List-Post` header then:

```console
To = email.List-Post.withoutUser
Cc, Bcc, ReplyTo = Empty
```

- Otherwise: There is no `Reply To List` feature

3. Reply all

- If the user is the sender

```console
To = email.to.withoutUser
Cc = email.cc.withoutUser
Bcc = email.bcc.withoutUser
ReplyTo = email.replyTo.withoutUser
```

- Otherwise


    _If email.replyTo is not empty then:_
    
    ```console
    To = (email.replyTo + email.to).withoutUser
    Cc = email.cc.withoutUser
    Bcc = email.bcc.withoutUser
    ReplyTo = Empty
    ```
    
    _Otherwise:_
    
    ```console
    To = (email.from + email.to).withoutUser
[...]

Issue #3670, opened 2025-04-17
Forwarding an email removes the attachments from the current email detail view


chibenwa:
https://github.com/user-attachments/assets/63d994a1-bc05-4844-ab04-4510c1fe4467

!

Issue #3661, opened 2025-04-15
Release new version v0.15.2


hoangdat:
### DoD:
- [ ] Request translating in weblate and merge to branch
- [x] Test in all platforms base on the check-list case: [tnr-tmail-template_20Mar.ods](https://github.com/user-attachments/files/19358057/tnr-tmail-template_20Mar.ods)
  - Chrome
  - Firefox
  - Edge
  - Safari
  - Android
  - iOS
  - Opera
- [x] Memory leak verifycation
  - View emails with heavy inline
  - Send multiple emails with heavy inline image and attachment
  - Create identity with heavy image
- [x] Tag new version
- [x] Push new docker image
- [x] iOS - Release app in Test Flight
- [x] Android - Release app in Internal Test

hoangdat:
- @dab246 Chrome, Android
- @hoangdat Edge, iOS
- @tddang-linagora Firefox, Safari

tddang-linagora:
Memory leak:
- View email OK
- Create identity OK
- Compose email will be fixed https://github.com/linagora/tmail-flutter/pull/3669

tddang-linagora:
Multiple composer - Inline images lost when dropping an image when there are another image loading in another composer

https://github.com/user-attachments/assets/d0dc04df-1d10-4cd2-af76-86ab313994d4

dab246:
> Mark mailbox as read doesn't redirect to log in when 401
> 


Fixed at https://github.com/linagora/tmail-flutter/pull/3677

hoangdat:
New version was deployed, please test on stg @dab246 @tddang-linagora

hoangdat:
Released with 0.15.3

PR #3680, opened 2025-04-22
Bump version to v0.15.3

hoangdat:
### Fixed
- #3670 Forwarding email make attachment bar disappear
- #3661 Fix memory leak for file picker
- #3674 Refactor reply email base on thunderbird logic
- Handle session in case of using isolate for Mark as Read, Empty Trash, Empty Spam


tddang-linagora (approved):

dab246 (approved):
lgtm

Issue #3671, opened 2025-04-17
Cannot cancel sending when no network WEB


chibenwa:
It looks like all composer interaction are blocked without network on WEB

https://github.com/user-attachments/assets/b093385e-407a-4eba-a928-12b4a481ab9d

v0.15.1 chrome

dab246:
Same #3150

PR #3682, opened 2025-04-22
TF-3671 Fix all composer interaction are blocked without network

dab246:
## Issue 

#3671

## Root cause 

Screenshot 2025-04-21 at 18 43 32

- Files that are broken offline are because the `WebView` (or the internal `iframe` of `html_editor_enhanced`) doesn't understand the `assets/...` path:

```html
<script src="assets/packages/html_editor_enhanced/assets/jquery.min.js"></script>
```
In a `WebView` or `iframe` environment, the `assets/...` path won't work like in a `Flutter` app (unless you have a `local server` to serve). It will try to fetch from the main domain like:

```
https://domain/assets/packages/html_editor_enhanced/assets/jquery.min.js
```
And when offline → `ERR_INTERNET_DISCONNECTED`

- First time opening the app without internet → Flutter never loaded `summernote-no-plugins.html` → the `assets/summernote-no-plugins.html` file was not `materialized` (attached to the file system) → `HtmlEditorWidget` could not find the HTML file when offline.

## Solution

- Inject the content of `script, css,..` files directly into html

```
<!-- INSTEAD OF -->
<link href="summernote-lite.min.css" rel="stylesheet">

<!-- USE -->
<style>
/* summernote-lite.min.css content here */
</style>
```

- Pre-cache HTML

## Dependency

Need merged: 

 - https://github.com/linagora/html-editor-enhanced/pull/52 

## Resolved

https://github.com/user-attachments/assets/f259eb58-0341-4e53-b794-235642cae6ba


dab246:
- [ ] Should we support `preloading assets` so that all `icons` that have never been displayed online are displayed offline? @hoangdat

lib/main/utils/app_utils.dart
tddang-linagora, line 1: This class holds no variable. Why's there a need for singleton?

dab246: Reverted

tddang-linagora (approved):

hoangdat:
> Should we support preloading assets so that all icons that have never been displayed online are displayed offline?

Any impaction? 
First loading maybe longer?

dab246:
> 
> Any impaction? 
>

It will take some time for us to load all the assets, helping the browser cache it.

>
> First loading maybe longer?
> 

It may take some time, depending on the size of the asset to load.

dab246:
> * merged [Support cache asset offline html-editor-enhanced#52](https://github.com/linagora/html-editor-enhanced/pull/52)
> * please update and rebase @dab246

Updated

Issue #3684, opened 2025-04-22
[Bug] Signature is located in a wrong place if you move an inline image right-left


tprudentova:
0.15.3, mail.stg.lin-saas.com, Windows, tested on Chrome, Opera 

1. Open composer 
2. Enter an inline image 
3. Add some text + have a signature 
4. Use the button Float right for inline image, then float back left 
5. Signature and the text below the image are located in the right part of the composer 

Expectation: they are below the image as before

https://github.com/user-attachments/assets/50ac115d-0535-41e4-a4fe-b820be3c3abf

PR #3688, opened 2025-04-23
TF-3684 Fix signature is located in a wrong place if you move an inline image right-left

dab246:
## Issue

#3684 

## Root cause

- According to [MDN's document](https://developer.mozilla.org/en-US/docs/Web/CSS/float), the `float` attribute changes the order of elements in html. So images set to `float: right` will float to the right and content behind it can `float` up to the edge of the image.
- Only the `Web` is affected because the new web has the feature to change `float` for images

## Solution

Set `clear: both` CSS style for Signature to remove the effect of `float`. See more at [MDN-Web/CSS/clear](https://developer.mozilla.org/en-US/docs/Web/CSS/clear)

## Dependency

Need merged: 

 - https://github.com/linagora/html-editor-enhanced/pull/53 

## Resolved

https://github.com/user-attachments/assets/58e94683-fafd-4264-995b-504988783160

### DIFF
--- a/pubspec.lock
+++ b/pubspec.lock
@@ -1245,10 +1245,10 @@ packages:
     description:
       path: "."
       ref: main
-      resolved-ref: "3279b9d164723ca349ca3d030abc27efbadd0d34"
+      resolved-ref: d23d675f8e3f7d8909abb9707aa2700437673d01
       url: "https://github.com/linagora/html-editor-enhanced.git"
     source: git
-    version: "3.1.7"
+    version: "3.1.8"
   html_unescape:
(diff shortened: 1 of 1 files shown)


tddang-linagora (dismissed):

hoangdat:
merged https://github.com/linagora/html-editor-enhanced/pull/53. please update

dab246:
> merged [linagora/html-editor-enhanced#53](https://github.com/linagora/html-editor-enhanced/pull/53). please update

Updated

hoangdat (approved):

Issue #3644, opened 2025-04-10
[Thread] [POC] Can scroll EmailView (`HtmlElementView`) within a scroll view (`ListView, SingleScrollView,...`)


dab246:
## Description

Currently on the web platform, we use [HtmlElementView](https://api.flutter.dev/flutter/widgets/HtmlElementView-class.html) to display email content as an `iframe`. So when wrapping HtmlElementView in a scroll view, the scroll event of the outer scroll view can be swallowed by the iframe of the inner HtmlElementView, resulting in only being able to scroll the content in the iframe but the outer scroll view cannot be scrolled.

## Approach

Try one of the following:

- [ ] Try with different scroll views (ListView, SingleScrollView, CustomScrollView,..) on flutter versions from `3.24.5` and above.
- [ ] Handle scroll event with `javascript` in iframe and update scroll view outside.

PR #3716, opened 2025-05-09
TF-3644 Handle scroll thread detail when cursor is in email content area

dab246:
## Issue

#3644 

## Root cause

`HtmlElementView (iframe)` does not allow Flutter to intervene at the mouse scroll level due to platform view class limitation.

## Solution

Use `JavaScript` in `iframe` to forward scroll event to `Flutter`

- Send scroll event from iframe to Flutter
- Catch message event from iframe in Flutter Web by create a `JS interop` to listen for `postMessage` from iframe

## Resolved

https://github.com/user-attachments/assets/329ccd5a-a3ca-454b-ab4a-c65d7cee9e72

### DIFF
--- a/core/lib/presentation/views/html_viewer/html_content_viewer_on_web_widget.dart
+++ b/core/lib/presentation/views/html_viewer/html_content_viewer_on_web_widget.dart
@@ -32,6 +32,7 @@ class HtmlContentViewerOnWeb extends StatefulWidget {
   final bool allowResizeToDocumentSize;
 
   final bool keepWidthWhileLoading;
+  final ScrollController? scrollController;
 
   const HtmlContentViewerOnWeb({
     Key? key,
@@ -45,6 +46,7 @@ class HtmlContentViewerOnWeb extends StatefulWidget {
     this.onClickHyperLinkAction,
     this.keepWidthWhileLoading = false,
     this.contentPadding,
+    this.scrollController,
   }) : super(key: key);
 
   @override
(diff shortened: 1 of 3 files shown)


core/lib/presentation/views/html_viewer/html_content_viewer_on_web_widget.dart
```diff
@@ -137,6 +144,24 @@ class _HtmlContentViewerOnWebState extends State<HtmlContentViewerOnWeb> {
     });
   }
 
+  void _handleIframeOnScrollChangedListener(
```
hoangdat, line 147: to prevent handle multiple times on one even, we should return true or false, base on the `data`. If we handle scroll -> return true, else return false. Base on the result, we can break the invoke function to do other logic.

dab246: Updated

core/lib/presentation/views/html_viewer/html_content_viewer_on_web_widget.dart
```diff
@@ -246,6 +297,17 @@ class _HtmlContentViewerOnWebState extends State<HtmlContentViewerOnWeb> {
           
           resizeObserver.observe(document.body);
         }
+        
+        ${widget.scrollController != null ? '''
+          window.addEventListener('wheel', function (event) {
+            const deltaY = event.deltaY;
```
hoangdat, line 303: What happens if user try to scroll horizontal? We will not handle this?

dab246: If the user scrolls horizontally of course, `window.addEventListener('wheel')` will be triggered, depending on the deltaY value the scroll handler will handle, but it will always ensure consistent scrolling between the iframe and flutter's scroll view

tddang-linagora (approved):

tddang-linagora (approved):

hoangdat (approved):

Issue #3739, opened 2025-05-22
[Bug] [ANDROID] Error when uploading several attachments


tprudentova:
0.15.4, stg.lin-saas.com, Android 

1. Open composer 
2. Click Attach 
3. Select several files 
4. There's an error, not all files are attached 

https://github.com/user-attachments/assets/c42905b7-dcd5-4fcb-afb6-9c19f7f39bda

tprudentova:
@hoangdat @chibenwa

dab246:
Can't reproduce it as you described.
- Do you have any other preconditions before adding attachment?
- Have you tried uploading other files besides `PDF` to see if you get the error? If not, can you share your PDF file with us for further investigation?
- Maybe there was some problem with the server yesterday. If possible, try again today to see if the problem persists?

https://github.com/user-attachments/assets/c64c3f13-8f49-451c-a7a8-31844bee757b

tprudentova:
@dab246 yes, I think it was a server error, cause it seems to be fixed now. Thank you!

PR #3751, opened 2025-05-26
TF-3739 Fix error when uploading several attachments on android

dab246:
## Issue

#3739

Cannot upload multiple attachment when token expired.

```console
PlatformException(token_failed, Concurrent operations detected: token, token, null, null)
```

## Root cause

The application uses `Dio` with `AuthorizationInterceptor` to automatically attach access tokens to `HTTP requests`. When the token expires, the interceptor will refresh the token to get a new token. However, when multiple uploads (multi-file) run in parallel, or uploads run in separate isolates, requests with `401` errors will simultaneously trigger token refresh, leading to multiple calls to the `refreshingTokensOIDC(...)` function at the same time. This causes internal errors from the platform channel because it does not support simultaneous operations

## Solution

- Use `Async, await` to replace for `isolate` when upload attachment

## Resolved

https://github.com/user-attachments/assets/68be51d5-b660-448d-b219-b5b4368417ea


tddang-linagora (approved):

chibenwa:
To be rebased.
This was out of radar. 
Fix is small.
No reason to delay this.

hoangdat:
I cannot reproduce the error, and the issue was raised and closed by Taisya. 
Moreover, I still not agree on uploading in main isolate.

So I propose to close this PR.

Issue #3767, opened 2025-05-27
Recent search item not working in twake mail web


hoangdat:
### Desc

- It is cleaned whenever re-login 
- should not clean it in log out
- add `TuppleKey` to distinguish account in recent search

PR #3797, opened 2025-06-11
TF-3767 Recent search item not working in Twake mail web

dab246:
## Issue

#3767 

## Resolved

https://github.com/user-attachments/assets/f8bcbd10-4e6b-437a-b291-314b4ae64e43

### DIFF
--- a/lib/features/cleanup/data/local/recent_search_cache_manager.dart
+++ b/lib/features/cleanup/data/local/recent_search_cache_manager.dart
@@ -1,4 +1,5 @@
 
+import 'package:core/utils/app_logger.dart';
 import 'package:jmap_dart_client/jmap/account_id.dart';
 import 'package:jmap_dart_client/jmap/core/user_name.dart';
 import 'package:model/extensions/account_id_extensions.dart';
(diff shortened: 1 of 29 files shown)


lib/features/cleanup/data/local/recent_search_cache_manager.dart
```diff
@@ -76,6 +88,10 @@ class RecentSearchCacheManager {
         ? listRecentSearch.sublist(0, newLimit)
         : listRecentSearch;
 
+    for (var recentSearch in newListRecentSearch) {
+      log('RecentSearchCacheManager::getAllLatest: Recent search: ${recentSearch.toString()}');
+    }
```
tddang-linagora:
```suggestion
    if (kDebugMode) {
      for (var recentSearch in newListRecentSearch) {
        log('RecentSearchCacheManager::getAllLatest: Recent search: ${recentSearch.toString()}');
      }
    }
```

IMO we should not make the app wait to log on release build.

dab246: We will delete it.

dab246: Done

tddang-linagora (approved):

hoangdat (approved):

Issue #3792, opened 2025-06-06
Can't reply to mail


X-Ryl669:
### Description
I'm using Stalwart. It works well for sending mail and fetching mails.
However, whenever I try to reply to an HTML email, forward it or save as draft and reopen it for sending it fails.

I've tried sending mail with only links (I don't know if it's HTML underneath, there's no option in Twake mail to see the mail's source) and it works.

### Expected result
Save mail to draft should work in all cases.
Sending mail some HTML should work in all cases.

### Current behavior
Twake mail fails with a dialog: "Saving the message in draft folder failed. An error happened while saving the message". It has 2 buttons "Message" "Modify". Clicking message close the current editor and you're loosing your email's reply (WHAT'S THE USE OF THIS?) and clicking modify returns to the editor

### Preconditions (optional)
A mail with HTML in it is required. Can't forward the mail to check the source, since forwarding also fails.
I can send you the HTML mail in a private context (found from /data/data/com.linagora.android.teammail/app_flutter/opened_email), if it's of any use, but it's plain old standard garbage made by outlook.

#### Reproduction Steps
Reply or forward a mail with HTML. It should work.
Save as draft while replying. It should work.

### Acceptance criteria
1. No mail should be lost (if the draft can't be saved, save the mail to a local file on the device, so we can, at least send it with another client). Or don't have a button that'll loose the current work, it's so frustrating.
2. Replying / Forwarding / Saving to draft are all basic features, it should work.

### Context
Android 9.
Twake mail 0.15.5

### Additional information

I don't know how to get an advanced information about the error, there's no option in the application to display the error log. I haven't found any log or anything like that in /data/data/com.linagora.android.teammail/

chibenwa:
Hello

This may be an inter-operability issue.

Can you please:

 - Share stallwart logs ?

 - Share the JMAP client exchanges that fails?

Thanks in advance.

X-Ryl669:
Here are the Stalwart's log (FWIW). Maybe it would be a good idea to save the communication log of tmail upon error in the application too? 

I manage the JMAP server so I can increase the log level to "TRACE" temporarily, but I think it's not the case for every user here, and the default is just "INFO" where it just says it received a request (with no details about it).


```
2025-06-08T05:27:46Z INFO HTTP connection started (http.connection-start) listenerId = "http", localPort = 60800, remoteIp = 127.0.0.1, remotePort = 37630
[...]

PR #3810, opened 2025-06-19
TF-3792 Fix cannot reply to email on Stalwart server

dab246:
## Issue

#3792

## Resolved

https://github.com/user-attachments/assets/28f49716-98c0-48b3-8fc9-06413bca1e3b

### DIFF
--- a/model/lib/extensions/email_body_part_extension.dart
+++ b/model/lib/extensions/email_body_part_extension.dart
@@ -11,4 +11,19 @@ extension EmailBodyPartExtension on EmailBodyPart {
     cid: cid,
     charset: charset,
     disposition: disposition.toContentDisposition());
+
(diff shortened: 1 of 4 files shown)


model/lib/extensions/email_body_part_extension.dart
```diff
@@ -11,4 +11,19 @@ extension EmailBodyPartExtension on EmailBodyPart {
     cid: cid,
     charset: charset,
     disposition: disposition.toContentDisposition());
+
+  EmailBodyPart onlyUseBlobIdOrPartId() => EmailBodyPart(
+    partId: blobId == null ? partId : null,
```
hoangdat, line 16: can we have the case which include partId only?

dab246: No we always have blobId, partId is only available when we get available attachments from the server

hoangdat (approved):

Issue #3177, opened 2024-09-24
[Bug] Drag and dropping images sometimes breaks UI


tprudentova:
0.13.2, mail.stg.lin-saas.com, tested on Safari, MacOS

1. Open the composer (identity should have a signature), but the cursor right above the signature
2. Drag and drop an image to the middle of the screen (I think it should be visually below the signature)
3. UI breaks, there's an additional signature arrow, if you try to move the image, the app crashes, need to reload 

https://github.com/user-attachments/assets/932973a6-4791-4472-859c-e09d6fdb78e4

tddang-linagora:
@tprudentova ~Please help me verify if this issue is still happening. I cannot reproduce it on tmail.linagora.com.~

Edit: Nevermind, reproducible on Google Chrome.

PR #3831, opened 2025-06-24
TF-3177 Fix drag and dropping images sometimes breaks UI in composer on web

dab246:
## Issue

#3177 

## Reproduced

https://github.com/user-attachments/assets/0c60983e-c131-45b5-b37a-418495ac4b1b

## Root cause

-  `insertHtml` is called while the caret is inside `.tmail-signature-button`. Summernote may clone the parent `.tmail-signature`, but skips` .tmail-signature-content` because it's hidden (display: none), leading to duplication.
- The button has `contenteditable="false"`. This prevents precise caret placement, so Summernote may insert HTML at the wrong place or clone the parent incorrectly.

## Dependency

Need merged:

https://github.com/linagora/html-editor-enhanced/pull/55 

## Resolved

https://github.com/user-attachments/assets/ff52c9bd-74a8-4a05-9509-7cac9d83d5e1


tddang-linagora (dismissed):

pubspec.yaml
```diff
@@ -63,7 +63,7 @@ dependencies:
   html_editor_enhanced:
     git:
       url: https://github.com/linagora/html-editor-enhanced.git
-      ref: main
+      ref: bugfix/duplicated-signature-button-when-insert-image
```
hoangdat: please update. merged https://github.com/linagora/html-editor-enhanced/commit/27272e278bcd94d48522ce1944f8650c9366e51a

dab246: Updated

hoangdat (approved):

PR #3855, opened 2025-07-02
Replace blinking screens when login in with one splash screen

dab246:
## Issue

[Replace blinking screens when login in with one splash screen](https://www.notion.so/linagora/Replace-blinking-screens-when-login-in-with-one-splash-screen-20262718bad180eb8080fba1a622ee1f?source=copy_link)

## Resolved

https://github.com/user-attachments/assets/7c255a76-c2f6-4149-a0d9-117a697f18e0

### DIFF
--- a/lib/features/caching/utils/session_storage_manager.dart
+++ b/lib/features/caching/utils/session_storage_manager.dart
@@ -7,11 +7,11 @@ class SessionStorageManager {
 
   final html.Storage sessionStorage = html.window.sessionStorage;
 
-  Future<void> save(String key, String value) async {
-    return sessionStorage.addAll({key: value});
+  void save(String key, String value) {
+    sessionStorage.addAll({key: value});
   }
 
-  Future<String> get(String key) async {
+  String get(String key) {
     final entry = sessionStorage
       .entries
       .firstWhereOrNull((entry) => entry.key == key);
@@ -23,7 +23,7 @@ class SessionStorageManager {
     }
   }
 
-  Future<void> remove(String key) async {
+  void remove(String key) {
     if (sessionStorage.containsKey(key)) {
       sessionStorage.remove(key);
     }
--- a/lib/features/caching/utils/session_storage_manager.dart
+++ b/lib/features/caching/utils/session_storage_manager.dart
@@ -5,12 +5,14 @@ import 'package:universal_html/html.dart' as html;
 
 class SessionStorageManager {
 
-  void save(String key, String value) {
-    html.window.sessionStorage.addAll({key: value});
+  final html.Storage sessionStorage = html.window.sessionStorage;
+
+  Future<void> save(String key, String value) async {
+    return sessionStorage.addAll({key: value});
   }
 
-  String get(String key) {
-    final entry = html.window.sessionStorage
+  Future<String> get(String key) async {
+    final entry = sessionStorage
       .entries
       .firstWhereOrNull((entry) => entry.key == key);
 
@@ -21,7 +23,9 @@ class SessionStorageManager {
     }
   }
 
-  void remove(String key) {
-    html.window.sessionStorage.remove(key);
+  Future<void> remove(String key) async {
+    if (sessionStorage.containsKey(key)) {
+      sessionStorage.remove(key);
+    }
   }
 }
--- a/lib/features/home/presentation/extensions/handle_web_finger_to_get_token_extension.dart
+++ b/lib/features/home/presentation/extensions/handle_web_finger_to_get_token_extension.dart
@@ -1,0 +1,69 @@
+import 'package:core/presentation/state/failure.dart';
+import 'package:core/utils/platform_info.dart';
+import 'package:dartz/dartz.dart';
+import 'package:model/oidc/oidc_configuration.dart';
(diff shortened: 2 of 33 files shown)


lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -709,6 +737,9 @@ class MailboxDashBoardController extends ReloadableController
       _handleSessionFromArguments(arguments);
     } else if (arguments is MailtoArguments) {
       _handleMailtoURL(arguments);
+    } else if (PlatformInfo.isWeb) {
+      dispatchRoute(DashboardRoutes.thread);
```
tddang-linagora, line 743: Should we wait for new session to be assigned before changing this view?

web/login-callback.html
```diff
@@ -1,31 +1,38 @@
...
 
-<body>
+        // Build the redirect URL
+        let redirectUrl;
+        if (destination.includes("/login")) {
+          redirectUrl = destination.replace("/login", "/dashboard?type=normal");
+        } else {
+          const separator = destination.endsWith("/") ? "" : "/";
+          redirectUrl = destination + separator + "dashboard?type=normal";
+        }
```
tddang-linagora, line 24: When possible, this should be checked in case app is inside Cozy dashboard

dab246: Please check it on Cozy help me

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
dab246: Don't need, we should to user see loading bar in thread

tddang-linagora (dismissed):

lib/features/caching/utils/session_storage_manager.dart
```diff
@@ -5,12 +5,14 @@ import 'package:universal_html/html.dart' as html;
...
-    html.window.sessionStorage.addAll({key: value});
+  final html.Storage sessionStorage = html.window.sessionStorage;
+
+  Future<void> save(String key, String value) async {
+    return sessionStorage.addAll({key: value});
   }
 
-  String get(String key) {
-    final entry = html.window.sessionStorage
+  Future<String> get(String key) async {
```
hoangdat: why we need to change this? still not see any async function inside

web/login-callback.html
hoangdat: @tddang-linagora please help us verify this case

lib/features/home/presentation/extensions/handle_web_finger_to_get_token_extension.dart
```diff
@@ -0,0 +1,69 @@
...
+      );
+    }
+  }
+
+  void getOIDCConfiguration(OIDCResponse oidcResponse) {
+    consumeState(getOIDCConfigurationInteractor.execute(oidcResponse));
+  }
+
+  void handleOIDCConfigurationSuccess(OIDCConfiguration oidcConfig) {
+    performUserAuthenticationOnWeb(oidcConfig);
```
hoangdat, line 35: how about to have a condition to check platform here?

lib/features/login/data/datasource_impl/authentication_oidc_datasource_impl.dart
```diff
@@ -131,14 +141,25 @@ class AuthenticationOIDCDataSourceImpl extends AuthenticationOIDCDataSource {
   @override
   Future<String> getAuthenticationInfo() {
     return Future.sync(() async {
-      return await _authenticationClient.getAuthenticationInfo();
-    }).catchError(_exceptionThrower.throwException);
+      return await _sessionStorageManager.get(
```
hoangdat: why not keep the abstract layer via `_authenticationClient`?

lib/features/home/presentation/extensions/handle_web_finger_to_get_token_extension.dart
dab246: We don't need it, because the condition calling it is only handled on the web.

lib/features/login/data/datasource_impl/authentication_oidc_datasource_impl.dart
dab246: Getting authentication info is only available on the web, and the datasource that handles it is the local SessionStorageManager, so it wouldn't make sense to put it in the AuthenticationClient that handles the network.

lib/features/caching/utils/session_storage_manager.dart
dab246: Reverted

web/login-callback.html
tddang-linagora: https://github.com/user-attachments/assets/09d4989e-66cf-489a-a401-9e65e6769bdf

Looks fine by me

tddang-linagora (dismissed):

hoangdat (dismissed):
please rebase

dab246:
> please rebase

Done

tddang-linagora (approved):

hoangdat (approved):

PR #3857, opened 2025-07-03
Apply new style preferences part in Settings

dab246:
## Issue

https://www.notion.so/linagora/Preferences-part-in-settings-22262718bad180e78aaaede1a3b4b29f?source=copy_link

## Resolved

https://github.com/user-attachments/assets/d83f6cd9-fdd9-457a-ad9d-8f618a9a84eb

### DIFF
--- a/core/lib/presentation/utils/theme_utils.dart
+++ b/core/lib/presentation/utils/theme_utils.dart
@@ -6,12 +6,13 @@ import 'package:flutter/services.dart';
 class ThemeUtils {
   ThemeUtils._();
 
-  static ThemeData get appTheme {
+  static ThemeData buildAppTheme(BuildContext context) {
     return ThemeData(
       scaffoldBackgroundColor: Colors.white,
       fontFamily: ConstantsUI.fontApp,
       appBarTheme: _appBarTheme,
       textTheme: _textTheme,
+      hoverColor: Theme.of(context).colorScheme.outline.withOpacity(0.08),
       textSelectionTheme: _textSelectionTheme,
       dividerTheme: _dividerTheme,
       visualDensity: VisualDensity.adaptivePlatformDensity,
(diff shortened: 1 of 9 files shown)


core/lib/presentation/utils/theme_utils.dart
```diff
@@ -6,12 +6,13 @@ import 'package:flutter/services.dart';
 class ThemeUtils {
   ThemeUtils._();
 
-  static ThemeData get appTheme {
+  static ThemeData buildAppTheme(BuildContext context) {
```
tddang-linagora, line 9:
```suggestion
  static ThemeData buildAppTheme(ThemeData themeData) {
```

IMO we should avoid passing `BuildContext` when not necessary.

lib/features/manage_account/presentation/menu/manage_account_menu_view.dart
```diff
@@ -20,149 +19,136 @@ class ManageAccountMenuView extends GetWidget<ManageAccountMenuController> {
...
+           bottom: 16,
+         ),
+         child: Column(
+           crossAxisAlignment: CrossAxisAlignment.start,
+           children: [
+             Padding(
+                padding: const EdgeInsets.all(8),
+                child: TMailButtonWidget(
+                  key: const Key('back_to_dashboard_button'),
+                  text: AppLocalizations.of(context).back,
```
tddang-linagora, line 37: Should we cache `AppLocalizations.of(context)` before `return` statement to reuse?

core/lib/presentation/utils/theme_utils.dart
dab246: We need use it.

lib/features/manage_account/presentation/menu/manage_account_menu_view.dart
dab246: Why do we need cache it? What is affect? IMO, don't need?

tddang-linagora (dismissed):

hoangdat (dismissed):

PR #3862, opened 2025-07-08
[HOT] Fix font error and lost text

dab246:
## Issue 

- The text `Select all messages of this page` is missing

Screenshot 2025-07-08 at 11 21 16

- Font error when using Arabic language

Screenshot 2025-07-08 at 12 01 53

- Font error when text contain emoji

Screenshot 2025-07-08 at 13 08 35

## Root cause

We set `fontFallbackBaseUrl` to avoid loading fonts from `https://fonts.gstatic.com` which leads to fallback fonts not being downloaded, causing font errors in some languages ​​(like Arabic), emoji font errors, and even losing text because `fontFamily` has not been set for the style.

## Solution

- Use fontFamily for all style in Text widget
- Restore `fontFallbackBaseUrl` google by default set by Flutter

## Side effect

https://github.com/linagora/tmail-flutter/pull/3854

## Resolved

- Web:

https://github.com/user-attachments/assets/e51c9688-611e-44d8-9e16-7989091f8cb5

- Mobile:

[demo-mobile.webm](https://github.com/user-attachments/assets/ed83c759-8794-4df0-b491-837ac8c26beb)

### DIFF
--- a/core/lib/presentation/utils/app_toast.dart
+++ b/core/lib/presentation/utils/app_toast.dart
@@ -1,6 +1,7 @@
 import 'package:core/presentation/extensions/color_extension.dart';
 import 'package:core/presentation/resources/image_paths.dart';
 import 'package:core/presentation/utils/responsive_utils.dart';
+import 'package:core/presentation/utils/theme_utils.dart';
 import 'package:core/presentation/views/toast/tmail_toast.dart';
 import 'package:core/presentation/views/toast/toast_position.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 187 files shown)


tddang-linagora (dismissed):

assets/fonts/noto/NotoColorEmoji.ttf
hoangdat, line 1: it increase size of the app much, and also increase loading time in web app. 
How we can only apply it in web, still keep google static for mobile?

dab246: We don't use font fallback from assets, we will restore `fontFallbackBaseUrl` google by default set by Flutter and add `fontFamily` for all style in Text widget

hoangdat (approved):

Issue #3885, opened 2025-07-17
[CUSTOMER] Language defaults


chibenwa:
On a customer platform, v0.16.3

Tmail dans Settings/get that returns no language (no prior value)

<img width="1841" height="746" alt="Image" src="https://github.com/user-attachments/assets/e45c2bc8-c69b-4dca-b20c-86ff3eba7c1d" />

The language used is then the default english

Expected behaviour:

 - Use language in Settings/get
 - If not then use browser defaults
 - If not use english

Please fix asap and tag a 0.16.3.1 with this fix

PR #3887, opened 2025-07-18
TF-3885 Fix language defaults

dab246:
## Issue

#3885

## Reproduce

https://github.com/user-attachments/assets/6520c2e4-4fc3-4121-800d-0c06b18dfc4c

## Root cause

Since device local returns a Locale without `countryCode` so the test condition `supportedLocales.contains(deviceLocale)` returns `false` resulting in the default locale `en` being used

https://github.com/linagora/tmail-flutter/blob/78cb93ef1cbbd47959f1207e35d7b145efc05f6c/lib/main/localizations/localization_service.dart#L97-L102

## Resolved

- When server & cached locale is NULL

https://github.com/user-attachments/assets/5ea48f15-c0d3-4329-bd40-6a2b126b2f92

- When server & cached locale is not NULL

https://github.com/user-attachments/assets/667fa9ed-e421-4be2-9c6e-beb4e697a93a

### DIFF
--- a/lib/features/composer/presentation/extensions/create_email_request_extension.dart
+++ b/lib/features/composer/presentation/extensions/create_email_request_extension.dart
@@ -158,7 +158,7 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
             IndividualHeaderIdentifier.acceptLanguageHeader: LocalizationService.supportedLocalesToLanguageTags()
           },
           contentLanguageHeader: {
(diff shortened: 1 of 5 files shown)


chibenwa (approved):

chibenwa:
Please duplicate also the PR into master ?

tddang-linagora (approved):

hoangdat (approved):

Issue #3894, opened 2025-07-21
Blank email content when opening email


Gictorbit:
### Description
When I click on an email from the email list, the email opens but the content area is completely blank. The email list itself loads correctly and shows senders, subjects, and previews without issue.

### Expected result
After clicking on an email, the email content should be fully displayed, including text, attachments, and formatting.

### Current behavior
The email content area shows up blank after clicking on an email, even though the email list displays correctly.

### Preconditions (optional)
+ Logged in as a valid user
- Emails are properly synced and visible in the list

#### Reproduction Steps
1. Open the app and go to the inbox
2. Wait until the list of emails loads
3. Click on any email from the list
4. Observe that the email content is not displayed

### Acceptance criteria
 + Email content is shown correctly after clicking on an email
 - No blank content area is observed
 + Email body, formatting, and attachments are properly rendered

### Context
docker image: `linagora/tmail-web:v0.17.0`

### Additional information

<img width="1220" height="756" alt="Image" src="https://github.com/user-attachments/assets/54b68224-71ba-4dbc-8476-d743e775410f" />

hoangdat:
thank for reporting, we are investigating on it

hoangdat:
you had blank email all the time or sometimes? we can not reproduce it in web app? can you more detail on this topic?

Gictorbit:
Hi, just to clarify — before the new release everything was working perfectly and email content displayed correctly. After the latest update, opening emails shows a blank content area.

I'm using Twake Mail with Stalwart as the backend. Let me know if you need any logs or more details to help troubleshoot this.

Gictorbit:
When I open an email, this JMAP request is triggered:

```sh
curl 'https://mail.example.com/jmap/' \
  -H 'accept: application/json;jmapVersion=rfc-8621' \
  -H 'authorization: Bearer [redacted]' \
  --data-raw '{"using":["urn:ietf:params:jmap:core","urn:ietf:params:jmap:mail"],"methodCalls":[["Email/get",{"accountId":"c","ids":["daaaaaa3"],"properties":["bodyValues","htmlBody","attachments","headers","keywords","mailboxIds","messageId","references","header:X-SMIME-Status:asText"],"fetchHTMLBodyValues":true},"c0"]]}'
```

The API response actually contains the correct email content, but it doesn’t show up in the application interface.

chibenwa:
Can you please share the EML of the underlying email?

Please also share your browser version.

[...]

Issue #3717, opened 2025-05-11
Unable to access settings in Tmail web


dougmeredith:
### Description

I can see a round circle in the top right corner, which I'm guessing is supposed to be a user avatar. When I hover over it, the mouse pointer changes, as if I should be able to click on it. Clicking has no effect.

### Expected result

I'm guessing this is intended to get me into the settings.

### Context

Version 0.15.3 using Docker.

Browser it the current Chrome, on Linux.

hoangdat:
Hi, 

Thanks for your reporting.
But it still works in our side

<img width="364" alt="Image" src="https://github.com/user-attachments/assets/d1bd0072-ff63-4f7d-a794-ec90ec93a8bd" />

Can you open devTools/console of browser, then send us anymore relate information?

dougmeredith:
Thanks for the quick reply!

Firstly I should note that the problem also occurs for me using Firefox. Secondly, the top right corner of my browser tab doesn't look like yours:

![Image](https://github.com/user-attachments/assets/b0dd0478-4a74-4256-b561-2d511f0d378c)

This is from the Chrome console when I click:

`main.dart.js:58775 Instance of 'minified:aQE'
main.dart.js:58775 at Object.k (https://tmail.test.dstrc.org/main.dart.js:3619:19)
main.dart.js:58775     at Object.ad (https://tmail.test.dstrc.org/main.dart.js:3628:15)
main.dart.js:58775     at Object.lp (https://tmail.test.dstrc.org/main.dart.js:42423:16)
main.dart.js:58775     at rA.ab0 (https://tmail.test.dstrc.org/main.dart.js:202982:18)
main.dart.js:58775     at bSM.$1 (https://tmail.test.dstrc.org/main.dart.js_5.part.js:7960:38)
main.dart.js:58775     at bgF.$1 (https://tmail.test.dstrc.org/main.dart.js:80063:14)
main.dart.js:58775     at akf.bE0 (https://tmail.test.dstrc.org/main.dart.js:97724:14)
main.dart.js:58775     at tear_off.<anonymous> (https://tmail.test.dstrc.org/main.dart.js:3755:65)
main.dart.js:58775     at cjs.$0 (https://tmail.test.dstrc.org/main.dart.js:91859:22)
main.dart.js:58775     at mD.aC6 (https://tmail.test.dstrc.org/main.dart.js:91416:9)`

dougmeredith:
[...]

PR #3896, opened 2025-07-22
TF-3894 Fix blank email content when opening email

dab246:
## Issue

#3894 

## Root cause 

The `Stalwart` server does not return any email address in the session (https://github.com/linagora/tmail-flutter/issues/3717#issuecomment-3036701594) lead to `session.getOwnEmailAddress()` throwing exception.

https://github.com/linagora/tmail-flutter/blob/be0ef818d16141afb23c6f070e22a514b3b85070/lib/features/thread_detail/presentation/thread_detail_view.dart#L44

## Reproduce

https://github.com/user-attachments/assets/3834d6be-3a07-4114-a282-5ad3e2b45419

## Resolved

- With both `ownEmailAddress` & identity

https://github.com/user-attachments/assets/1390b3f2-ed3a-45e2-b83f-f67702d10538

- Without `ownEmailAddress` &  with identity

https://github.com/user-attachments/assets/1f322d88-5122-44ac-8c3c-71ff24f25716

- Without both `ownEmailAddress` & identity

https://github.com/user-attachments/assets/e42fefa0-c449-46e5-8acb-5c489e10e829


tddang-linagora (dismissed):

model/lib/extensions/session_extension.dart
```diff
@@ -129,6 +129,26 @@ extension SessionExtension on Session {
     return null;
   }
 
+  String generateOwnEmailAddressFromDomain(String domain) {
```
hoangdat, line 132: why we need this function?

lib/features/base/mixin/own_email_address_mixin.dart
```diff
@@ -0,0 +1,42 @@
...
+
+    if (listIdentities.isEmpty) {
+      synchronizeOwnEmailAddress(emptyOwnEmailAddress);
+      ownEmailAddress.refresh();
+      return;
+    }
+
+    final identityEmailAddress = listIdentities.firstOrNull?.email ?? '';
+    final domain = EmailUtils.getDomainByEmailAddress(identityEmailAddress);
+    final userEmailAddress =
```
hoangdat, line 31: still not understand this point

model/lib/extensions/session_extension.dart
dab246: We create the exact email the user is using. Because identity has many different emails, only the domain is the same.

lib/features/base/mixin/own_email_address_mixin.dart
dab246: We create the exact email the user is using. Because identity has many different emails, only the domain is the same.

Gictorbit:
thank you

Gictorbit:
gentle ping on this PR.
would appreciate it if we could get a review or merge @hoangdat. thanks

tddang-linagora (dismissed):

Gictorbit:
I’ve tested this branch, and everything is working as expected. Is it ready to be merged?

Antti:
Can someone please move this forward? This is a major issue for those connecting to stalwart

tddang-linagora (approved):

hoangdat (approved):

Issue #3920, opened 2025-07-30
Too many docker overlay in tmail-web


guimard:
### Description

The tmail-web container isn't well built. Then when it starts, it creates a lot of files which generated docker massive overlays.

See [container-diff](https://github.com/user-attachments/files/21505454/container-diff.txt)

PR #3921, opened 2025-07-30
Build gz files during build, not during startup (Closes: #3920)

guimard:


guimard:
Of course then, customization might embed their ".gz" to increase _(a little)_ performances

hoangdat (approved):

PR #3958, opened 2025-08-13
fix the bugs

Monther-debug:


Monther-bug:
done

hoangdat:
hi @Monther-debug , can you describe your PR?
What bugs you are trying to fix?

Monther-bug:
I just fix the packages problem

core/pubspec.yaml
```diff
@@ -39,7 +39,7 @@ dependencies:
   ### Dependencies from pub.dev ###
   cupertino_icons: 1.0.6
 
-  flutter_svg: 2.0.9
+  flutter_svg: ^2.1.0
```
dab246, line 42: We always use `specific versions` to ensure project stability and avoid the impact of version upgrades of dependencies.

Monther-bug:
ok i will do it

Monther-bug:
done

Monther-bug:
done

chibenwa:
Thanks a lot for this contribution.

This PR has a title `fix the bugs`.

Which bugs? Can we have an exhaustive list of bugs fixed? Can we have 1 commit per bugs?

Finally this pull request is not reviewable. The changeset is too huge. I believe this work needs serious rework and especially be split in explicit subparts (pull requests) so that we have a decent chance to review the quality of it and ensure it is safe.

image

=> this is not ok

Thanks for your understanding.

android/app/.cxx/Debug/m5h3x372/arm64-v8a/configure_fingerprint.bin
```diff
@@ -0,0 +1,28 @@
+C/C++ Structured Logr
```
chibenwa, line 1: I am reluctant to commit binaries from external sources.

core/pubspec.yaml
```diff
@@ -39,7 +39,7 @@ dependencies:
   ### Dependencies from pub.dev ###
   cupertino_icons: 1.0.6
 
-  flutter_svg: 2.1.0
+  flutter_svg: ^2.1.0
```
chibenwa, line 42: Please use strict version. Having tolerance here open a vulnerability to potential suply chain attacks.

chibenwa:
Thanks again for the contribution!

This pull request is inactive for over 6 month. It is very complex with numerous commits and huge diff change, the team will not be able to integrate it as is. Thus, I will close it for now.

If this is still relevant we can welcome those changes in new pull requests, but they would need to be focussed on a signle topic, with a reasonable amount of code change and commits that allow for a review, testing and improvments.

Best regards,

Benoit TELLIER

PR #4021, opened 2025-09-17
Fix cannot open email on iPad/Tablet device

dab246:
## Issue

Cannot open detail email on iPad/Tablet device

## Root cause

ThreadDetailView on device's MailboxDashboardView not updated (PlatformInfo.isMobile == true)

## Resolved

- iPad:

https://github.com/user-attachments/assets/d09db9e0-6b18-409e-affa-5ffaa1611d3d

- Mobile:

https://github.com/user-attachments/assets/d71c3c99-e2f4-4340-8068-0d0eb73ef79d

- Web:

https://github.com/user-attachments/assets/14c4de75-b39a-40a3-94ec-dcad9c2f75e3

### DIFF
--- a/lib/features/composer/presentation/composer_view.dart
+++ b/lib/features/composer/presentation/composer_view.dart
@@ -43,7 +43,7 @@ class ComposerView extends GetWidget<ComposerController> {
         onCloseViewAction: () => controller.handleClickCloseComposer(context),
         onClearFocusAction: () => controller.clearFocus(context),
         backgroundColor: MobileAppBarComposerWidgetStyle.backgroundColor,
-        childBuilder: (context, constraints) => SafeArea(
+        childBuilder: (_, constraints) => SafeArea(
           left: !controller.responsiveUtils.isLandscapeMobile(context),
           right: !controller.responsiveUtils.isLandscapeMobile(context),
           child: Container(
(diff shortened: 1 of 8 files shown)


lib/features/composer/presentation/view/mobile/mobile_container_view.dart
```diff
@@ -32,7 +32,7 @@ class MobileContainerView extends StatelessWidget {
         child: Scaffold(
           backgroundColor: backgroundColor ?? MobileContainerViewStyle.outSideBackgroundColor,
           resizeToAvoidBottomInset: false,
-          body: LayoutBuilder(builder: (context, constraints) {
+          body: LayoutBuilder(builder: (_, constraints) {
```
tddang-linagora, line 35: Why do we skip this `context`

lib/features/composer/presentation/composer_view.dart
```diff
@@ -43,7 +43,7 @@ class ComposerView extends GetWidget<ComposerController> {
         onCloseViewAction: () => controller.handleClickCloseComposer(context),
         onClearFocusAction: () => controller.clearFocus(context),
         backgroundColor: MobileAppBarComposerWidgetStyle.backgroundColor,
-        childBuilder: (context, constraints) => SafeArea(
+        childBuilder: (_, constraints) => SafeArea(
```
tddang-linagora, line 46: idem

lib/features/composer/presentation/view/mobile/mobile_container_view.dart
dab246: We should use the context of the root view to avoid the case of invalid context when the widget tree is rebuilt.

lib/features/composer/presentation/composer_view.dart
dab246: idem

tddang-linagora (approved):

hoangdat (approved):

PR #3598, opened 2025-03-26
TF-3358 Save draft locally periodically

dab246:
## Issue

#3358 

## Resolved

- Reload page & Token Expired:

https://github.com/user-attachments/assets/571f4354-8c99-4731-98f2-0cf81700177a

- Save after 15s:

https://github.com/user-attachments/assets/e86f6edf-ba79-4565-801f-1b1d0115e079

## Summary by CodeRabbit

* **New Features**
  * Local Drafts: save drafts locally and manage them in a "Local Draft List" dialog with edit, restore, discard, and restore-all actions.
* **Improvements**
  * Autosave: periodic automatic saving of open composer windows.
  * Dialog handling: dialogs now block background interactions reliably.
  * UI: simplified avatar rendering and added localized strings for draft flows.
* **Bug Fixes**
  * Fixed a download cancel behavior to use correct cancel reason.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #4100, opened 2025-10-13
TF-4099 Mobile adding an attachment causes composer UI issues

dab246:
## Issue

#4099

## Root cause

- Log analysis:

```console
[ERROR:android_webview/browser/aw_browser_terminator.cc:165]
Renderer process (14064) crash detected (code -1).
[ERROR:android_webview/browser/aw_browser_terminator.cc:113]
Render process (14064) kill (OOM or update) wasn't handled by all associated webviews, killing application.

```

This is a typical crash of Android WebView renderer being `OOM (Out Of Memory)` or being force-killed due to `low-memory` policy.

Special:

```console
Process com.linagora.android.teammail (pid 12911) has died: svc +1 LAST
skip restart com.linagora.android.teammail because this device is a lowmemory device!

```

Confirmed MIUI (system_server) killed the process to reclaim memory, as the device recognized it as a “lowmemory device”.

- Conclude:

`flutter_inappwebview` creates its own sandbox renderer (Chromium engine) -> consumes a lot of RAM.

When file_picker turns on `Intent.ACTION_OPEN_DOCUMENT`, Android temporarily switches foreground to another app → causing MIUI or Android low-memory killer to think your app is in the background → kill WebView renderer.

Then WebView cannot recover because Chromium sandbox is dead → Flutter is terminated.

## Workaround

- Reduce WebView memory footprint
- Enable `android:largeHeap="true"` in AndroidManifest => Allow process to keep more RAM, limit OOM kill when DocumentsUI is open.

## Summary by CodeRabbit

* **New Features**
  * Enhanced editor state preservation when accessing file and image pickers on mobile devices.

* **Bug Fixes**
  * Fixed editor content loss during file/image picker interactions on mobile.

* **Chores**
  * Optimized memory allocation to support larger operations.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -45,6 +45,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
+            android:largeHeap="true"
             android:launchMode="singleTask"
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -36,7 +36,8 @@
    <application
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
-        android:icon="@mipmap/ic_launcher">
+        android:icon="@mipmap/ic_launcher"
+        android:largeHeap="true">
 
        <meta-data
            android:name="com.google.firebase.messaging.default_notification_icon"
@@ -45,7 +46,6 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
-            android:largeHeap="true"
             android:launchMode="singleTask"
             android:theme="@style/LaunchTheme"
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -10,7 +10,6 @@ import 'package:dio/dio.dart';
 import 'package:dropdown_button2/dropdown_button2.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:filesize/filesize.dart';
-import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:get/get.dart';
@@ -612,6 +611,9 @@ class ComposerController extends BaseController
 
   void onCreatedMobileEditorAction(BuildContext context, HtmlEditorApi editorApi, String? content) {
     richTextMobileTabletController?.htmlEditorApi = editorApi;
+    richTextMobileTabletController?.onInitWebViewLifecycleManager(
+      editorApi.webViewController,
+    );
     richTextMobileTabletController?.richTextController.onCreateHTMLEditor(
       editorApi,
       onEnterKeyDown: _onEnterKeyDown,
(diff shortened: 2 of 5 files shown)


hoangdat:
what is the progress for this ticket?

dab246:
> what is the progress for this ticket?

In the process of being completely resolved, the error is related to the device's RAM usage rights. Only occurs on low-configuration devices, the error does not appear frequently.

chibenwa:
This fix do not fully solve the issue

However it delivers an improvments.

I propose to rebase it, merge it, but keep the issue open.

tddang-linagora (dismissed):

android/app/src/main/AndroidManifest.xml
```diff
@@ -45,6 +45,7 @@
         <activity
             android:name=".MainActivity"
             android:exported="true"
+            android:largeHeap="true"
```
hoangdat: should place it in <application>

lib/features/upload/domain/usecases/local_file_picker_interactor.dart
```diff
@@ -14,6 +17,7 @@ class LocalFilePickerInteractor {
 
   Stream<Either<Failure, Success>> execute({FileType fileType = FileType.any}) async* {
     try {
+      log('$runtimeType::execute:Memory info: ${(ProcessInfo.currentRss / (1024 * 1024)).toStringAsFixed(2)} MB');
```
hoangdat: how about to add `breadscrumb` here?

android/app/src/main/AndroidManifest.xml
```diff
@@ -45,6 +45,7 @@
...
             android:launchMode="singleTask"
```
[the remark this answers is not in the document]
dab246: fixed

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1157,14 +1159,18 @@ class ComposerController extends BaseController
...
   }
 
-  void openFilePickerByType(BuildContext context, FileType fileType) async {
-    if (!kIsWeb) {
+  void openFilePickerByType(BuildContext context, FileType fileType) {
+    if (PlatformInfo.isMobile) {
       popBack();
+      richTextMobileTabletController?.webViewLifecycleManager?.pause();
     }
     consumeState(_localFilePickerInteractor.execute(fileType: fileType));
```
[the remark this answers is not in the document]
dab246: fixed

lib/features/upload/domain/usecases/local_file_picker_interactor.dart
```diff
@@ -1,6 +1,9 @@
 
+import 'dart:io';
```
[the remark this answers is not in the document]
dab246: fixed

android/app/src/main/AndroidManifest.xml
dab246: fixed

lib/features/upload/domain/usecases/local_file_picker_interactor.dart
dab246: removed

dab246:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > lib/features/composer/presentation/composer_controller.dart (1)> `1191-1235`: _⚠️ Potential issue_ | _🟠 Major_
> > > **Async handlers called from void context run unawaited.**
> > > These handlers (`_handlePickFileSuccess`, `_handlePickFileFailure`, `_handlePickImageSuccess`, `_handlePickImageFailure`) are now `Future<void>` but are invoked from `handleSuccessViewState`/`handleFailureViewState`, which are `void` methods called via `consumeState`'s `Stream.listen(onData)`. Since `onData` is typed as `void Function(...)`, the returned Futures are discarded, meaning:
> > > 
> > > 1. `await resume()` and `await _restoreEditorContentIfBlank()` execute fire-and-forget
> > > 2. Any exceptions not caught within the handlers won't propagate
> > > 
> > > The existing try-catch in `_restoreEditorContentIfBlank` provides some protection, but errors in `resume()` could go unobserved.
> > > Consider wrapping the async block in the handlers with an additional safety layer:
> > > 
> 

It's not necessary, as both functions already have try/catch handling built in.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -235,6 +235,7 @@ class ComposerController extends BaseController
 
   List<Attachment> initialAttachments = <Attachment>[];
   String? _textEditorWeb;
+  String? _savedEditorContentBeforePicker;
```
hoangdat, line 238: IMO, please remove this mechanism, because the problem is system acquire memory from app, so it restart the app. we should find the better way to cache the content.

hoangdat:
it still not fix the problem, will verify it along side with #3598

Issue #4103, opened 2025-10-14
Display storage with GB instead of GiB


dab246:
# Context

In pricing page, we display storage on GB but in TMail, storage seems displayed by **GiB**. This creates confusion for the user

# Specs

This issue consists of display storage in TMail in GB (instead of GiB)

<img width="1680" height="939" alt="Image" src="https://github.com/user-attachments/assets/d7f33908-3d70-4fd1-a4a2-6b92f31ce6ff" />

poupotte:
Product review: I want

PR #4104, opened 2025-10-14
TF-4103 Display storage with GB instead of GiB

dab246:
## Issue

#4103 

## Resolved

- Before:

Screenshot 2025-10-14 at 7 05 48 PM

- After:

Screenshot 2025-10-14 at 7 11 49 PM


dab246:
Quota values ​​on 2 platforms `linagora.com` and `stg.lin-saas.com` are different.

- On `linagora.com`: Using binary system – IEC, GiB (gibibyte) so use division unit 1024 i.e. 1 GiB = 1024 bytes


Screenshot 2025-10-15 at 12 10 03 PM

 
- On `stg.lin-saas.com`: Using decimal system – SI, GB (gigabyte) so use division unit 1000 i.e. 1 GB = 1000 bytes


Screenshot 2025-10-15 at 12 11 55 PM

PR #4110, opened 2025-10-17
[HOT-FIX] Duplicated email subject after enabling thread mode

dab246:
## Issue

Duplicated email subject after enabling thread mode

## Reproduce

Steps to reproduce

```
Open any email while thread mode is disabled.
Go to Settings → Enable thread mode.
Go back to the previously opened email view.
```

https://github.com/user-attachments/assets/ba733eb3-d546-4619-9128-0e7754f71ef2

## Root cause

The issue occurs because the EmailView uses a GlobalKey, which prevents the widget from being disposed of when switching views. As a result, when the EmailView is rebuilt, it retains the previous object state within the class instead of creating a fresh instance, causing outdated or incorrect data (such as the old email subject) to be displayed.

## Resolved

https://github.com/user-attachments/assets/3998e534-f284-4f17-8419-2abebed7e355

### DIFF
--- a/lib/features/email/presentation/styles/email_subject_styles.dart
+++ b/lib/features/email/presentation/styles/email_subject_styles.dart
@@ -12,7 +12,6 @@ class EmailSubjectStyles {
     overflow: PlatformInfo.isWeb ? TextOverflow.ellipsis : null,
     color: Colors.black,
     fontSize: 24,
-    height: 1,
     letterSpacing: -0.24,
   );
 }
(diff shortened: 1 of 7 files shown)


tddang-linagora (dismissed):

hoangdat (dismissed):

hoangdat:
- [x] please rebase and update

dab246:
> * [x]  please rebase and update

Done

Issue #4128, opened 2025-10-27
Attachment download fails if I come back to mailbox view


chibenwa:
GIVEN I start downloading an attachment
WHEN I navigate back to emailList view
THEN the download eventually fails

Expectations: it INDEED succeed

https://github.com/user-attachments/assets/ac415a78-f98f-44ce-a594-7857566c7dd3

PR #4131, opened 2025-10-28
TF-4128 Fix attachment download fails if come back to mailbox view 

dab246:
## Issue

#4128 

## Root cause

The interactor responsible for handling attachment downloads is created and managed inside the `SingleEmailController`. When the user navigates back to the mailbox, the email view is closed, which causes the `SingleEmailController` to be disposed. As a result, the interactor tied to it is also destroyed and can no longer continue processing. However, the download process still attempts to emit new events into a stream that has already been closed, leading to the error:

```console
Bad state: Cannot add new events after calling close
```

## Resolved

https://github.com/user-attachments/assets/f4c4cd59-bd11-4b7c-a349-5c35873fc817


dab246:
- [x] Rebase on master @hoangdat

hoangdat (approved):

Issue #1715, opened 2023-04-04
Quickly access favourite emails


chibenwa:
# Summary

Alongside system mailboxes there is a `Favourites` entry.

When I click on it, it displays all `starred` emails. (Same as "Search all Mails + star filter)

PR #4134, opened 2025-10-30
TF-1715 Add Favorites Tab in Sidebar

dab246:
## Issue

#1715 

## Resolved

- Web:

https://github.com/user-attachments/assets/a8fcebce-703f-4e54-85aa-02a09eb2e411

- Mobile:

https://github.com/user-attachments/assets/1f7bd08a-60b8-4d20-862b-6281cfaab0ba


hoangdat:
- [x] should mark as star when drag n drop or move email to `Favorite` folder

lib/l10n/intl_messages.arb
```diff
@@ -2688,13 +2688,13 @@
     "placeholders_order": [],
     "placeholders": {}
   },
-  "noEmailInYourCurrentFolder": "There are no emails in your current folder",
+  "noEmailInYourCurrentFolder": "We're sorry, there are no emails in your current folder",
```
hoangdat: should not change it

lib/l10n/intl_messages.arb
```diff
@@ -2688,13 +2688,13 @@
...
   "@noEmailInYourCurrentFolder": {
     "type": "text",
     "placeholders_order": [],
     "placeholders": {}
   },
-  "noEmailMatchYourCurrentFilter": "There are no emails that match your current filter.",
+  "noEmailMatchYourCurrentFilter": "We're sorry, there are no emails that match your current filter.",
```
hoangdat: idem

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -188,12 +188,13 @@ class ThreadController extends BaseController with EmailActionController {
...
         }
       }
+    } else {
+      super.handleSuccessViewState(success);
     }
   }
 
   @override
   void handleFailureViewState(Failure failure) {
-    super.handleFailureViewState(failure);
```
hoangdat, line 200: why we need to change order of this method. I look risky for the whole thread controller

dab246: To keep the semantic logic of `if/else` consistent, avoid misunderstandings about the parent function of this function always being called first. We only execute the parent function when the child function's conditions are not satisfied. This change does not affect the logic of the current threads.

lib/l10n/intl_messages.arb
dab246: Updated

lib/l10n/intl_messages.arb
dab246: Updated

dab246:
> * [x]  should mark as star when drag n drop or move email to `Favorite` folder

Added


https://github.com/user-attachments/assets/5333d183-b15e-4c7e-bb10-eb3d413d5398

lib/features/thread/presentation/thread_controller.dart
dab246: reverted

hoangdat (approved):

Issue #4145, opened 2025-11-12
Changing the way to get the "URL”


dab246:
# Context

Today we implemented something that will not work in 1 or 2% of the use case. And tomorrow it'll not work for b2b with custom domain. It is used for instance for mail to create a link to the "premium page".

# Specs

This issue consists to change the way the frond-end get the "FQDN” to be more consistent

dab246:
Currently, the `workplaceFqdn` attribute in `user_info` is only available in the `stg.lin-saas.com` environment.
In the `prod` and `lin-saas.dev` environments, this attribute has not been implemented yet and therefore remains unavailable.

PR #4147, opened 2025-11-12
TF-4145 Changing the way to get the `URL`

dab246:
## Issue

#4145 

## Summary

Refactor the logic to generate the `Paywall `URL using the user's `workplaceFqdn` instead of relying solely on static domain resolution.

## Context

Previously, the Paywall URL was built using a default fallback pattern (`pattern.getQualifiedUrl`), which combined the user's email and a root domain from `RouteUtils.getRootDomain()`. This approach worked, but did not respect the user’s actual workspace domain (`FQDN`).

To fix this, the system now leverages the `workplaceFqdn` obtained from the authenticated OIDC user info (`oidcUserInfo.workplaceFqdn`) to dynamically generate the `Paywall` URL for the correct tenant.

## Implementation Details

1. Using `WebLinkGenerator.safeGenerateWebLink`

- The `workplaceFqdn` value (`e.g. alice.example.app`) is passed to` WebLinkGenerator` to construct the Paywall URL.
- The method automatically handles malformed or missing values and guarantees a valid `https://` link.
- Example

```
WebLinkGenerator.safeGenerateWebLink(workplaceFqdn: 'alice.example.app');
// → https://alice.example.app/
```

2. Improved fallback behavior

If `workplaceFqdn` is unavailable or invalid, the code now gracefully falls back to:

```dart
pattern.getQualifiedUrl(
  ownerEmail: ownEmailAddress,
  domainName: RouteUtils.getRootDomain(),
);
```
This ensures that even in the worst case, the Paywall link remains functional.

### DIFF
--- a/core/lib/utils/web_link_generator.dart
+++ b/core/lib/utils/web_link_generator.dart
@@ -1,0 +1,86 @@
+import 'package:core/core.dart';
+
+/// Utility class for generating web links in flat subdomain format.
+class WebLinkGenerator {
+  const WebLinkGenerator();
+
+  /// Ensures that a path starts with `/`.
+  String ensureFirstSlash(String? path) {
+    if (path == null || path.isEmpty) return '/';
+    return path.startsWith('/') ? path : '/$path';
+  }
+
+  /// Validates if a given FQDN looks valid (has at least 2 parts and non-empty).
+  bool isValidFqdn(String fqdn) {
+    if (fqdn.trim().isEmpty || !fqdn.contains('.')) return false;
+    final parts = fqdn.split('.').where((p) => p.trim().isNotEmpty).toList();
+    return parts.length >= 2;
+  }
+
+  /// Generates a web link based on a workplace FQDN and app slug.
+  ///
+  /// Throws [ArgumentError] when input is invalid.
+  String generateWebLink({
+    required String workplaceFqdn,
+    required String slug,
+    String? pathname,
+    String? hash,
+    List<List<String>>? searchParams,
+  }) {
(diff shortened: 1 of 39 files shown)


core/lib/utils/web_link_generator.dart
```diff
@@ -0,0 +1,88 @@
...
+  static bool isValidFqdn(String fqdn) {
+    if (fqdn.trim().isEmpty || !fqdn.contains('.')) return false;
+    final parts = fqdn.split('.').where((p) => p.trim().isNotEmpty).toList();
+    return parts.length >= 2;
+  }
+
+  /// Generates a web link based on a workplace FQDN and app slug.
+  ///
+  /// Throws [ArgumentError] when input is invalid.
+  static String generateWebLink({
```
hoangdat, line 22: for paywall, it should not the same pattern: `https://tdvudev.stg.lin-saas.com/settings/premium`, so IMO, we should have a different way to generate paywall link

dab246: Not according to the ADR description of this issue we need to generate `paywall` from `workplaceFqdn`. The `generateWebLink` function is the function used to handle `workplaceFqdn`, so we still need it. And if we handle it according to this function our current paywall will be `https://tdvudev.stg.lin-saas.com/premium` instead of `https://tdvudev.stg.lin-saas.com/settings/premium`. And we also have a fallback for if this `generateWebLink` process fails. We will get the paywall link generated from eco-system

dab246: Updated paywall url correctly from `workplaceFqdn`

hoangdat (approved):

Issue #4169, opened 2025-11-24
[Label] Add Labels module to tmail-flutter


dab246:
## Desc

- [ ] Add Label capability, Label object
- [ ] Implement `Label/get`, `Label/set`, `Label/changes`
- [ ] Write unit test

## Docs

https://github.com/linagora/tmail-backend/blob/master/docs/modules/ROOT/pages/tmail-backend/jmap-extensions/jmapLabels.adoc

PR #4170, opened 2025-11-24
TF-4169 Add `Labels` module in tmail-flutter

dab246:
## Issue

#4169 

## Resolved

- [x] Add Label capability, Label object
- [x] Implement `Label/get`, `Label/set`, `Label/changes`
- [x] Write unit test


labels/.metadata
hoangdat, line 1: do we need it?

labels/lib/model/hex_color.dart
```diff
@@ -0,0 +1,10 @@
+import 'package:equatable/equatable.dart';
+
+class HexColor with EquatableMixin {
+  final String value;
+
+  HexColor(this.value);
```
hoangdat: should we precheck the color must be start with `#`?

labels/.metadata
dab246: Need, automatically generated when creating flutter package and it is not recommended to edit manually

labels/lib/model/hex_color.dart
dab246: Added

hoangdat (approved):

PR #4183, opened 2025-11-27
Cache bug on web

dab246:
## Issue

Users cannot see a specific email in its corresponding folder (e.g., `Inbox`). However, when performing a search with related keywords, the email still appears in the search results. Multiple attempts to resolve the issue were unsuccessful, including logging out and logging back in, refreshing the browser (F5), and reloading the entire page.

## Workaround

- Add a new env configuration: `forceEmailQuery = true`
- When switching between mailboxes, trigger a full reload of the email list using `Email/query` followed by `Email/get`.
- Clear all caches before logout
- Clear email and state cache on browser refresh (F5) or reload button
- Clear email and state cache before login


chibenwa:
1. I would need a docker image to functionally triage this, please
2. We still need to address the root cause. Please let's plan a meeting with me to share user elements

lib/features/thread/data/repository/thread_repository_impl.dart
```diff
@@ -138,6 +138,83 @@ class ThreadRepositoryImpl extends ThreadRepository {
...
+      sort: sort,
+      limit: limit,
+      filterOption: emailFilter?.filterOption,
+    );
+
+    final cachedState = await stateDataSource.getState(
+      accountId,
+      session.username,
+      StateType.email,
+    );
```
hoangdat: if it failed, we can ignore it?

lib/features/thread/domain/usecases/get_emails_in_mailbox_interactor.dart
```diff
@@ -19,48 +22,66 @@ class GetEmailsInMailboxInteractor {
...
+        'GetEmailsInMailboxInteractor::execute() - Unexpected exception: $error, stack: $stack',
+      );
+      yield Left(GetAllEmailFailure(error));
     }
   }
 
   Either<Failure, Success> _toGetEmailState({
     required EmailsResponse emailResponse,
-    MailboxId? currentMailboxId,
+    required MailboxId? currentMailboxId,
```
hoangdat, line 73: why need to require it?

lib/features/thread/domain/usecases/get_emails_in_mailbox_interactor.dart
```diff
@@ -19,48 +22,66 @@ class GetEmailsInMailboxInteractor {
...
+    Properties? propertiesCreated,
+    Properties? propertiesUpdated,
+    bool getLatestChanges = true,
+    bool forceEmailQuery = false,
+  }) async* {
     try {
-      yield Right<Failure, Success>(GetAllEmailLoading());
+      yield Right(GetAllEmailLoading());
+
+      final Stream<EmailsResponse> sourceStream = forceEmailQuery
```
hoangdat, line 37: can we have the tests for this?

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: We will add try/catch for it

lib/features/thread/domain/usecases/get_emails_in_mailbox_interactor.dart
dab246: We need to compare with the selectedMailboxId variable in memory, to avoid displaying wrong emails for mailbox

lib/features/thread/domain/usecases/get_emails_in_mailbox_interactor.dart
dab246: Added

hoangdat:
- [x] client in mobile read/star email, but app run with `forceEmailQuery==true` not update

dab246:
> * [x]  client in mobile read/star email, but app run with `forceEmailQuery==true` not update

I checked again and it still works. Describe more clearly how you reproduced it.

hoangdat:
Hi @chibenwa ,

I built image `v0.19.8-rc1`, please deploy it in canary to test more detail. And don't forget to enable env

```
FORCE_EMAIL_QUERY=true
```

to use Email/query in switching mailbox.

hoangdat:
- [x] checking `state` clearing

dab246:
> * [x]  checking `state` clearing

Fixed 

https://github.com/user-attachments/assets/5c5d5b48-f561-4e3e-8899-9e5c29d4aba0

lib/features/thread/data/repository/thread_repository_impl.dart
```diff
@@ -138,6 +138,101 @@ class ThreadRepositoryImpl extends ThreadRepository {
...
+    UnsignedInt? limit,
+    int? position,
+    Set<Comparator>? sort,
+    EmailFilter? emailFilter,
+    Properties? propertiesCreated,
+  }) async* {
+    final localDataSource = mapDataSource[DataSourceType.local];
+    final networkDataSource = mapDataSource[DataSourceType.network];
+
+    if (localDataSource == null || networkDataSource == null) {
```
hoangdat: I dont think we really need `localDataSource`

lib/features/thread/data/repository/thread_repository_impl.dart
```diff
@@ -138,6 +138,101 @@ class ThreadRepositoryImpl extends ThreadRepository {
...
+    );
+
+    log(
+      'ThreadRepositoryImpl::forceQueryAllEmailsForWeb(): '
+      'Local cache count = ${cachedList.length}; '
+      'State = ${cachedState?.value}',
+    );
+
+    if (localResponse.hasEmails()) {
+      yield localResponse;
```
hoangdat: IMO, no need to yield email cache to view, because it sometime make a big blink when open an old mailbox. (state is still ok)

lib/features/caching/caching_manager.dart
```diff
@@ -171,21 +161,71 @@ class CachingManager {
...
 
+  List<Future<void>> _clearEmailAndStateWithUser(
+    AccountId accountId,
+    UserName userName,
+  ) {
+    final emailKey = TupleKey(accountId.asString, userName.value).encodeKey;
+    final stateKey = StateType.email.getTupleKeyStored(accountId, userName);
+
+    return [
+      _emailCacheManager.deleteContainKey(emailKey),
```
hoangdat: why not `_emailCacheManager.clear(),`?

lib/features/caching/caching_manager.dart
```diff
@@ -73,66 +73,62 @@ class CachingManager {
   );
 
   Future<void> clearAll() async {
+    await Future.wait([
+      clearMailDataCached(),
+      _fcmCacheManager.clear(),
+      _accountCacheManager.clear(),
+      if (PlatformInfo.isIOS)
+        _keychainSharingManager.delete()
+    ], eagerError: true);
```
hoangdat: Do we really need `eagerError: true` here? Why not to make it run other clear?
Moreover, how about to handle error if one of these tasked throw?

lib/features/caching/caching_manager.dart
```diff
@@ -73,66 +73,62 @@ class CachingManager {
...
+
+    return Future.wait([
+      if (hasUserContext)
+        ..._clearDetailedEmailWithUser(accountId, userName)
+      else
+        ..._clearDetailedEmailWithoutUser(),
+    ]);
   }
 
   Future<void> clearAllEmailAndStateCache({AccountId? accountId, UserName? userName}) {
```
hoangdat: I found some of these method was call without try/catch? Should we add it? 
Fx: in `upgrade_hive_db` or in `onBeforeUnloadBrowserListener`

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: IMO, we need to make sure that the email appears immediately the next time we open the mailbox, when there is an email cache.

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: We have a loading bar to let users know that data is being updated, it's quite normal to blink the list when there is new data.

lib/features/caching/caching_manager.dart
dab246: We only delete emails for current account, as we will support delegates, multi accounts in future

lib/features/caching/caching_manager.dart
dab246: We should use it to avoid having to wait for a future when it fails.

lib/features/caching/caching_manager.dart
dab246: We should consider whether we prioritize data accuracy or application usability. If data accuracy is the priority, we don't need try/catch, to make it easier to track errors. If we prioritize application usability in case of unexpected errors, we should try/catch it, so that users can still use it in case of errors, even though user data might be incorrect.

lib/features/thread/data/repository/thread_repository_impl.dart
hoangdat: But no `localDataSource` don't need `return`

lib/features/caching/caching_manager.dart
hoangdat: We still should control the error, at least to log, than crashing the app.

hoangdat: Moreover, some place IMO, it is very important to catch the error and make user can continue use the app
- ThreadController::refreshAllEmail
- onBeforeUnloadBrowserListener (it relates to user's input)
- login

chibenwa:
Deployed (and configured!) on canary.

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: Why is there no localDataSource?

lib/features/caching/caching_manager.dart
dab246: Ok, we will use try/catch at here

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: Ok We will remove it

lib/features/caching/caching_manager.dart
hoangdat: should print this kind of errors to console web

lib/features/caching/caching_manager.dart
dab246: Oke, we will delete all email & state

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: Removed

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: Removed

lib/features/caching/caching_manager.dart
dab246: Removed

lib/features/thread/data/repository/thread_repository_impl.dart
dab246: Removed

lib/features/caching/caching_manager.dart
dab246: Updated

core/lib/utils/app_logger.dart
```diff
@@ -1,51 +1,70 @@
...
+      case Level.debug:
+      case Level.verbose:
+        html.window.console.debug('[TwakeMail] $value');
+        break;
+    }
+    return;
+  }
 
-  String logsStr = value ?? '';
-  logHistory.value = '$logsStr\n${logHistory.value}';
```
hoangdat, line 13: why we dont need it anymore?

dab246: We don't use it from the beginning, so delete it to avoid wasting more time displaying the log.

lib/features/base/base_controller.dart
```diff
@@ -548,42 +549,20 @@ abstract class BaseController extends GetxController
...
-      if (isAuthenticatedWithOidc) {
-        await _clearOidcAuthData();
-      } else {
-        await _clearBasicAuthData();
-      }
+      await Future.wait([
+        if (isAuthenticatedWithOidc)
+          deleteCredentialInteractor.execute()
+        else
+          deleteAuthorityOidcInteractor.execute(),
```
hoangdat: is it ok `if/else` inside Future.wait?

dab246: Dart allows using `if/else` inside a list, so placing it inside `Future.wait([...]) `is completely valid.

hoangdat (approved):

PR #4196, opened 2025-12-09
AI scribe data layer and requirements

zatteo:
The logical part of this PR that started to get too big https://github.com/linagora/tmail-flutter/pull/4187

Included : 
- creation of the scribe module
- data layer
- text selection mixin

## Release Notes

* **New Features**
  * Added AI-powered text enhancement capabilities including grammar correction, brevity adjustment, tone modification, translation, and formatting options.
  * Multi-language support for AI features (English, French, Russian, Vietnamese).
  * Text selection tracking with coordinate detection in composer.
  * Custom prompt support for AI text generation.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #4187, opened 2025-12-02
AI Scribe in the composer

zatteo:
Follow the work of https://github.com/linagora/tmail-flutter/pull/4172. Mockups are finalized and the goal of this PR is to go to production contrary to the previous one which was more a draft to play and learn.

Still some work to do, but already creating the PR to get first feedbacks especially about usage of components, controllers, mixins, code sharing, respect of convention, etc.

## Mockups

Screenshot 2025-12-02 at 16 05 38
Screenshot 2025-12-02 at 16 05 23

## Current status

https://github.com/user-attachments/assets/ef395662-65fb-4f86-a094-28d6a6e58e35

## Todo

To start, I will not implement every features from the mockup. I consider enough for a first version if I add:
- [x] ability to write custom prompt
- [x] copy result in clipboard
- [ ] mobile version (full screen instead of popup menu and popup modal)

## To play with it

Add in .env
```
AI_ENABLED=true
AI_API_KEY=<key from ai.linagora.com>
AI_API_URL=https://ai.linagora.com/api/v1
```

## Summary by CodeRabbit

* **New Features**
  * AI Scribe assistant added to the composer: generate, correct, rephrase, translate, change tone, and format text.
  * Selection-aware workflow across web, mobile, and iframe/webview: detect text selection and show an in-editor AIScribe trigger to act on selected text or full message.
  * In-editor UI: selection-triggered AIScribe button, bottom-bar AI button, dialog and suggestion modal for inserting AI results.
  * New localized label "AI Assistant" with translations (en/fr/ru/vi).

* **Chores**
  * New AI package integrated and build/localization steps updated.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #4197, opened 2025-12-10
AI scribe UI

zatteo:
Big Scribe PR https://github.com/linagora/tmail-flutter/pull/4187 has been split in two:
- logical part https://github.com/linagora/tmail-flutter/pull/4196 (merged in the features/ai branch)
- UI part (here)

## About prompts

A small PR will come soon that slighty update them to make them work as good as we can quickly. But later, we will handle differently prompts: 
- they will be fetched from a remote location
- they will be managed by a dedicated library or SDK 

## Video

https://github.com/user-attachments/assets/43b869a2-b9bc-47b8-82ea-08cb120f9960

## Summary by CodeRabbit

* **New Features**
  * Added AI Scribe integration enabling users to generate text suggestions directly in the composer via AI-powered assistance.
  * Introduced AI assistant button in composer with menu for predefined actions and custom prompts.
  * Added multilingual support for AI assistant across French, Russian, and Vietnamese.

### DIFF
--- a/core/lib/utils/string_convert.dart
+++ b/core/lib/utils/string_convert.dart
@@ -196,4 +196,18 @@ class StringConvert {
         .map((e) => NamedAddress(name: '', address: e))
         .toList();
   }
+
+  static String convertHtmlContentToTextContent(String htmlContent) {
+    String textContent = htmlContent.replaceAll(RegExp(r'<[^>]*>'), '');
+
+    textContent = textContent
+      .replaceAll('&nbsp;', ' ')
+      .replaceAll('&amp;', '&')
+      .replaceAll('&lt;', '<')
+      .replaceAll('&gt;', '>')
+      .replaceAll('&quot;', '"')
+      .replaceAll('&#39;', "'");
+
+    return textContent.trim();
+  }
 }
--- a/core/lib/utils/string_convert.dart
+++ b/core/lib/utils/string_convert.dart
@@ -4,6 +4,7 @@ import 'dart:typed_data';
 import 'package:core/utils/app_logger.dart';
 import 'package:core/domain/exceptions/string_exception.dart';
 import 'package:core/utils/mail/named_address.dart';
+import 'package:html/dom.dart';
 import 'package:html/parser.dart';
 import 'package:http_parser/http_parser.dart';
 
@@ -198,15 +199,29 @@ class StringConvert {
   }
 
   static String convertHtmlContentToTextContent(String htmlContent) {
-    String textContent = htmlContent.replaceAll(RegExp(r'<[^>]*>'), '');
+    final document = parse(htmlContent);
 
-    textContent = textContent
-      .replaceAll('&nbsp;', ' ')
-      .replaceAll('&amp;', '&')
-      .replaceAll('&lt;', '<')
-      .replaceAll('&gt;', '>')
-      .replaceAll('&quot;', '"')
-      .replaceAll('&#39;', "'");
+    // Each paragraph is surrounded by block tags so we add a /n for each block tag
+    // Even <br> are surrounded by block tags so we can ignore <br> and treat them
+    // as paragraph
+    const blockTags = [
+      'p',
+      'div',
+      'li',
+      'section',
+      'article',
+      'header',
+      'footer',
+      'h1','h2','h3','h4','h5','h6'
+    ];
+
+    for (final tag in blockTags) {
+      document.querySelectorAll(tag).forEach((element) {
+        element.append(Text('\n'));
+      });
+    }
+
+    final String textContent = document.body?.text ?? '';
 
     return textContent.trim();
   }
--- a/core/lib/utils/string_convert.dart
+++ b/core/lib/utils/string_convert.dart
@@ -204,22 +204,11 @@ class StringConvert {
     // Each paragraph is surrounded by block tags so we add a /n for each block tag
     // Even <br> are surrounded by block tags so we can ignore <br> and treat them
     // as paragraph
-    const blockTags = [
-      'p',
-      'div',
-      'li',
-      'section',
-      'article',
-      'header',
-      'footer',
-      'h1','h2','h3','h4','h5','h6'
-    ];
+    const blockTags = 'p, div, li, section, blockquote, article, header, footer, h1, h2, h3, h4, h5, h6';
 
-    for (final tag in blockTags) {
-      document.querySelectorAll(tag).forEach((element) {
-        element.append(Text('\n'));
-      });
-    }
+    document.querySelectorAll(blockTags).forEach((element) {
+      element.append(Text('\n'));
+    });
 
     final String textContent = document.body?.text ?? '';
 
--- a/core/test/utils/string_convert_test.dart
+++ b/core/test/utils/string_convert_test.dart
@@ -831,5 +831,68 @@ void main() {
 
       expect(result, 'First paragraph\n\nSecond paragraph');
     });
+
+    test('should preserve list items for ul li', () {
+      const htmlContent = '<ul><li>First item</li><li>Second item</li><li>Third item</li></ul>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First item\nSecond item\nThird item');
+    });
+
+    test('should preserve list items for ol li', () {
+      const htmlContent = '<ol><li>First item</li><li>Second item</li><li>Third item</li></ol>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First item\nSecond item\nThird item');
+    });
+
+    test('should strip inline style about text alignment', () {
+      const htmlContent = '<p>First paragraph</p><p style="text-align: right; ">Aligned right text</p><p>Last paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First paragraph\nAligned right text\nLast paragraph');
+    });
+
+    test('should strip inline style about indentation', () {
+      const htmlContent = '<p>First paragraph</p><p style="margin-left: 25px; ">Indented text</p><p>Last paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First paragraph\nIndented text\nLast paragraph');
+    });
+
+    test('should decode HTML entities', () {
+      const htmlContent = '<p>Text with &nbsp; and &amp; and &lt; and &gt; and &quot;</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'Text with \u00A0 and & and < and > and "');
+    });
+
+    test('should preserve text with nested inline formatting', () {
+      const htmlContent = '<p>Text with <strong>bold</strong> and <em>italic</em> and <u>underline</u></p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'Text with bold and italic and underline');
+    });
+
+    test('should handle empty content', () {
+      const htmlContent = '<p></p><div></div><p></p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, '');
+    });
+
+    test('should preserve header text with line breaks', () {
+      const htmlContent = '<h1>Main Title</h1><p>Content paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'Main Title\nContent paragraph');
+    });
+
+    test('should preserve link text only', () {
+      const htmlContent = '<p>Click <a href="https://example.com">here</a> for more info</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'Click here for more info');
+    });
   });
 }
--- a/core/test/utils/string_convert_test.dart
+++ b/core/test/utils/string_convert_test.dart
@@ -816,4 +816,20 @@ void main() {
       );
     });
   });
+
+  group('StringConvert.convertHtmlContentToTextContent', () {
+    test('should preserve line breaks between paragraphs', () {
+      const htmlContent = '<p>First paragraph</p><p>Second paragraph</p><p>Third paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First paragraph\nSecond paragraph\nThird paragraph');
+    });
+
+    test('should preserve line breaks for br', () {
+      const htmlContent = '<p>First paragraph</p><div><br></div><p>Second paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First paragraph\n\nSecond paragraph');
+    });
+  });
 }
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -27,6 +27,7 @@ import 'package:model/model.dart';
 import 'package:permission_handler/permission_handler.dart';
 import 'package:pointer_interceptor/pointer_interceptor.dart';
 import 'package:rich_text_composer/rich_text_composer.dart';
+import 'package:scribe/scribe/ai/presentation/widgets/ai_scribe.dart';
 import 'package:super_tag_editor/tag_editor.dart';
 import 'package:tmail_ui_user/features/base/base_controller.dart';
 import 'package:tmail_ui_user/features/base/before_reconnect_handler.dart';
@@ -89,6 +90,7 @@ import 'package:tmail_ui_user/features/composer/presentation/widgets/mobile/from
 import 'package:tmail_ui_user/features/composer/presentation/widgets/saving_message_dialog_view.dart';
 import 'package:tmail_ui_user/features/composer/presentation/widgets/saving_template_dialog_view.dart';
 import 'package:tmail_ui_user/features/composer/presentation/widgets/sending_message_dialog_view.dart';
+import 'package:scribe/scribe.dart';
 import 'package:tmail_ui_user/features/email/domain/state/get_email_content_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/save_template_email_state.dart';
 import 'package:tmail_ui_user/features/email/domain/state/update_template_email_state.dart';
@@ -198,6 +200,7 @@ class ComposerController extends BaseController
   final GlobalKey<TagsEditorState> keyReplyToEmailTagEditor = GlobalKey<TagsEditorState>();
   final GlobalKey headerEditorMobileWidgetKey = GlobalKey();
   final GlobalKey<DropdownButton2State> identityDropdownKey = GlobalKey<DropdownButton2State>();
+  final GlobalKey aiScribeButtonKey = GlobalKey();
   final double defaultPaddingCoordinateYCursorEditor = 8;
 
   FocusNode? subjectEmailInputFocusNode;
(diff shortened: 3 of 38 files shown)


zatteo:
@coderabbitai review

dab246:
Screenshot 2025-12-11 at 13 13 41

- [ ] IMO, We should show submenus when hovering over an item, instead of clicking, so users know which item is selected and which submenu it belongs to.

dab246:
- [ ] What's about mobile?

lib/features/composer/presentation/widgets/mobile/tablet_bottom_bar_composer_widget.dart
```diff
@@ -71,6 +75,18 @@ class TabletBottomBarComposerWidget extends StatelessWidget {
...
+            const SizedBox(width: TabletBottomBarComposerWidgetStyle.space),
+            TMailButtonWidget.fromIcon(
+              key: aiScribeButtonKey,
+              icon: imagePaths.icSparkle,
+              borderRadius: TabletBottomBarComposerWidgetStyle.iconRadius,
+              padding: TabletBottomBarComposerWidgetStyle.iconPadding,
+              iconSize: TabletBottomBarComposerWidgetStyle.iconSize,
+              tooltipMessage: AppLocalizations.of(context).aiAssistant,
+              onTapActionCallback: onOpenAIScribe!,
+            ),
```
dab246, line 88: We should create a separate widget for it so we can reuse it in multiple places.

lib/features/composer/presentation/widgets/web/bottom_bar_composer_widget.dart
```diff
@@ -31,6 +31,8 @@ class BottomBarComposerWidget extends StatelessWidget {
   final VoidCallback toggleMarkAsImportantAction;
   final VoidCallback saveAsTemplateAction;
   final VoidCallback onOpenInsertLink;
+  final VoidCallback? onOpenAIScribe;
+  final GlobalKey? aiScribeButtonKey;
```
dab246, line 35: IMO, the aiScribe widget should hold the GlobalKey so that it can manage its own key, preventing widgets holding keys from causing memory leaks.

lib/features/composer/presentation/widgets/web/web_editor_widget.dart
```diff
@@ -101,6 +111,19 @@ class _WebEditorState extends State<WebEditorWidget> {
...
     }
+
+    _selectionChangeListener = (event) {
+      if (event is MessageEvent) {
+        final data = jsonDecode(event.data);
+
+        if (data['name'] == HtmlUtils.registerSelectionChangeListener.name) {
+          handleSelectionChange(data);
+        }
+      }
```
dab246: Should use try/catch for this

lib/features/composer/presentation/widgets/web/web_editor_widget.dart
```diff
@@ -101,6 +111,19 @@ class _WebEditorState extends State<WebEditorWidget> {
...
+    };
+    if (_selectionChangeListener != null) {
+      window.addEventListener("message", _selectionChangeListener!);
```
dab246: We should listen for the `message` event in one place to avoid calling it multiple times. We could reuse `window.addEventListener("message", _dropListener!);`, but it should be modified appropriately for general use, as we might have multiple features that need to be handled in this function.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -156,6 +159,10 @@ class ComposerController extends BaseController
   final isMarkAsImportant = Rx<bool>(false);
   final isContentHeightExceeded = Rx<bool>(false);
 
+  final selectedText = Rxn<String>();
+  final hasTextSelection = false.obs;
+  final textSelectionCoordinates = Rxn<Offset>();
```
dab246: IMO, these values ​​are always set together, so we don't need too many values ​​to observe here. If you absolutely need these three values, you should create an object to contain them and just observe this object to update the UI.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -859,6 +869,100 @@ class ComposerController extends BaseController
...
+      textContent = textContent
+        .replaceAll('&nbsp;', ' ')
+        .replaceAll('&amp;', '&')
+        .replaceAll('&lt;', '<')
+        .replaceAll('&gt;', '>')
+        .replaceAll('&quot;', '"')
+        .replaceAll('&#39;', "'");
+
+      return textContent.trim();
+  }
```
dab246: It should be placed in the util class within the core module for reuse.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -859,6 +869,100 @@ class ComposerController extends BaseController
...
+      context: context,
+      imagePaths: imagePaths,
+      content: fullText,
+      onInsertText: insertTextInEditor,
+      interactor: _generateAITextInteractor,
+      buttonPosition: buttonPosition,
+    );
+  }
+
+  void showAIScribeMenuForSelectedText(BuildContext context, {Offset? buttonPosition}) {
```
dab246: We should move all of AIScribe's handling to a separate mixin, extension, or class to avoid bloating the Composer controller code.

lib/features/composer/presentation/composer_view.dart
```diff
@@ -413,28 +425,38 @@ class ComposerView extends GetWidget<ComposerController> {
...
+                                  );
+                                } else {
+                                  return const SizedBox.shrink();
+                                }
+                              }),
+                              SizedBox(height: MediaQuery.viewInsetsOf(context).bottom + 64),
+                            ],
+                          ),
+                          _buildAIScribeSelectionButton(context),
+                        ],
```
dab246, line 458: Should only add the Stack widget and _buildAIScribeSelectionButton when AIScribe is enabled and AI capability is supported.

scribe/lib/scribe/ai/presentation/bindings/ai_scribe_bindings.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:dio/dio.dart';
+import 'package:get/get.dart';
+import 'package:scribe/scribe/ai/data/datasource/ai_datasource.dart';
+import 'package:scribe/scribe/ai/data/datasource_impl/ai_datasource_impl.dart';
+import 'package:scribe/scribe/ai/data/repository/ai_repository_impl.dart';
+import 'package:scribe/scribe/ai/domain/repository/ai_scribe_repository.dart';
+import 'package:scribe/scribe/ai/domain/usecases/generate_ai_text_interactor.dart';
+
+class AIScribeBindings extends Bindings {
```
dab246, line 9: Should extends from `InteractorsBindings`

scribe/lib/scribe/ai/presentation/widgets/ai_scribe.dart
```diff
@@ -0,0 +1,213 @@
...
+          ),
+        );
+      }
+
+      return Center(child: modalContent);
+    },
+  );
+}
+
+({double left, double bottom}) _calculateModalPosition({
```
dab246: Can we write unit tests for it?

scribe/lib/scribe/ai/presentation/widgets/ai_scribe_bar.dart
```diff
@@ -0,0 +1,89 @@
...
+  void dispose() {
+    _controller.removeListener(_onTextChanged);
+    _controller.dispose();
+    super.dispose();
+  }
+
+  void _onTextChanged() {
+    setState(() {
+      _isButtonEnabled = _controller.text.trim().isNotEmpty;
+    });
```
dab246: If only a small widget is being changed, use `ValueNotifier` instead of `setState`.

scribe/lib/scribe/ai/presentation/widgets/ai_scribe_menu.dart
```diff
@@ -0,0 +1,250 @@
...
+  @override
+  Widget build(BuildContext context) {
+    final categories = widget.availableCategories ?? AIScribeMenuCategory.values;
+    return Column(
+      key: _menuKey,
+      mainAxisSize: MainAxisSize.min,
+      crossAxisAlignment: CrossAxisAlignment.stretch,
+      children: categories.map((category) {
+        return _buildCategoryItem(category);
+      }).toList(),
```
dab246:
```suggestion
children: categories.map(_buildCategoryItem).toList(),
```

scribe/lib/scribe/ai/presentation/widgets/ai_scribe_menu.dart
```diff
@@ -0,0 +1,250 @@
...
+    );
+  }
+
+  Widget _buildCategoryItem(AIScribeMenuCategory category) {
+    if (category.hasSubmenu) {
+      return Container(
+        key: _categoryKeys[category],
+        child: _buildMenuItem(
```
dab246, line 119: Why not just pass the key directly to _buildMenuItem?

dab246:
- [x] Fix failed unit tests

Screenshot 2025-12-11 at 13 44 43

lib/features/composer/presentation/widgets/web/bottom_bar_composer_widget.dart
zatteo: @dab246 Who is "the aiScribe widget"? In my case I need it at the top in the controller to position the scribe menu according the position of the bottom bar button.

lib/features/composer/presentation/composer_view.dart
zatteo: First step done here https://github.com/linagora/tmail-flutter/pull/4197/changes/75758eb93487f840d5cff4fcd1975c0b36cfbf60 and I plan to add the AI capability in a next PR

scribe/lib/scribe/ai/presentation/bindings/ai_scribe_bindings.dart
zatteo: I can not import directly `import 'package:tmail_ui_user/features/base/interactors_bindings.dart';` from the scribe package. And maybe it is not relevant to have double way dependency? What do you suggest?

scribe/lib/scribe/ai/presentation/widgets/ai_scribe.dart
zatteo: Good idea. It will also help me to fix some edge cases I saw this morning.

scribe/lib/scribe/ai/presentation/widgets/ai_scribe_menu.dart
zatteo: Actually it has no purpose anymore and I forgot to remove it during refactoring and commit merging. I removed it.

zatteo:
> Screenshot 2025-12-11 at 13 13 41
> 
>     * [ ]  IMO, We should show submenus when hovering over an item, instead of clicking, so users know which item is selected and which submenu it belongs to.

Done in a241bdb5e77d4fcc07ece3e52b161d19cc6312a0. Idea confirmed by the designer 🧑‍🎨

zatteo:
> * [ ]  What's about mobile?

If you don't mind, I propose to do another PR for it. See mockups below. Very different.

image

lib/features/composer/presentation/widgets/mobile/tablet_bottom_bar_composer_widget.dart
zatteo: Sure? I don't know, except mutualizing the icon, it does not mutualize a lot because it must have the style of bottom bar buttons.

scribe/lib/scribe/ai/presentation/bindings/ai_scribe_bindings.dart
hoangdat: I propose to move binding to `lib/features`. WDYT?

chibenwa:
I see in last comment we rely on `.env` to selectively enable AI support.

I understand this is simple but:

 - It adds configuration that can easily be overlooked. I'd prefer have only conf on the back and have the front self adapt.
 - It won't work on mobile

The way we generally move forward is by leveraging a JMAP capability to advertize back support.

CF https://github.com/linagora/tmail-backend/blob/master/tmail-backend/tmail-third-party/ai-bot/AIChatCompletionRoutes.adoc#additions-to-the-capability-object

Given that we already have such mechanism we can very likely just drop the `.env` stuff.

lib/l10n/intl_messages.arb
```diff
@@ -5129,5 +5129,11 @@
     "type": "text",
     "placeholders_order": [],
     "placeholders": {}
+  },
```
chibenwa, line 5132: I do really love the work done on modularization and only seeing onXXX events in main files, and a simple binding for the feature.

We did not manage however to get a modular system for the I10n ?

(notpick I know.... feel free to overlook it for now...)

Edit: I did spot some I10n for scribe below. I bet it's for the main menu entry?

chibenwa:
<3 <3 <3 <3 <3 <3 <3

Crash--:
> The way we generally move forward is by leveraging a JMAP capability to advertize back support.


@chibenwa next PR https://github.com/linagora/tmail-flutter/pull/4203

chibenwa:
> @chibenwa next PR #4203

Too much PR exception, the recovery from OSXP is a bit harsh :-p

(and I did do homeworks in the train!)

core/lib/utils/string_convert.dart
```diff
@@ -196,4 +196,18 @@ class StringConvert {
...
+    String textContent = htmlContent.replaceAll(RegExp(r'<[^>]*>'), '');
+
+    textContent = textContent
+      .replaceAll('&nbsp;', ' ')
+      .replaceAll('&amp;', '&')
+      .replaceAll('&lt;', '<')
+      .replaceAll('&gt;', '>')
+      .replaceAll('&quot;', '"')
+      .replaceAll('&#39;', "'");
+
```
chibenwa, line 214: > What do you think of this piece of Flutter code? Should I roast it?

# 🔍 **What’s good**

* The function is simple and readable.
* For very basic HTML, it might work.

But…

---

# ⚠️ **The issues (yes, you can roast this code 😄)**

## 1. **Using a regex to strip HTML tags…**

This is *always* fragile.

Problems you will eventually hit:

* Tags with attributes containing `>`
* Comments, scripts, styles
* Broken or unusual HTML
* Nested content where regex fails

👉 **There’s a reason HTML parsers exist.**

---

## 2. **Only a few HTML entities are handled**

There are **thousands** of HTML entities.
Your code handles maybe 5.

Anything like `&eacute;`, `&copy;`, `&hellip;`, etc. will remain unchanged.

---

## 3. **No handling of `<br>`, `<p>`, etc.**

All line breaks will disappear.
Your output becomes a single messy block:

> `"Line1Line2Line3"`

---

## 4. **Whitespace is not normalized**

HTML often contains multiple spaces, newlines, indentation.
Your output might be weirdly spaced.

---

# ✔️ **A proper solution in Flutter**

### Use the `html` package:

```dart
import 'package:html/parser.dart' as html;

String convertHtmlContentToTextContent(String htmlContent) {
  final document = html.parse(htmlContent);
  final String parsedString = document.body?.text ?? '';
  return parsedString.trim();
}
```

### Why this is better:

* Correctly strips tags
* Correctly decodes *all* HTML entities (`&eacute;` → `é`)
* More robust
* Easier to maintain


---

# 🧠 **Conclusion**

Your current code works **only for very simple HTML** and is **not reliable** for real-world HTML content.

👉 Strong recommendation: use an HTML parser instead.

chibenwa, line 214: > Redos vulnerable?

Short answer: **Yes — this regex *can* be ReDoS-vulnerable depending on the input.**

Let’s break it down clearly.

---

# ✅ **Your regex**

```dart
RegExp(r'<[^>]*>')
```

# ⚠️ **Is it vulnerable to ReDoS?**

**Potentially, yes — in specific edge cases.**

Reason:

* `<[^>]*>` uses `[^>]` with `*`, which is a **greedy, unbounded quantifier**.
* On malformed or adversarial input designed to confuse the engine, it can trigger **catastrophic backtracking**.

### Example of problematic input

A long string like:

```
<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<!
```

Or nested angled brackets like:

```
< < < < < < < < < < < < < < x
```

The regex tries to match:

* The opening `<`
* Then tries to find the next `>`, but finds none
* It backtracks repeatedly through a huge number of combinations

This can cause:

* CPU spike
* App freeze
* DoS in worst cases

---

# 🔍 **Why this happens**

The `[^>]` is fine by itself, but when used like this:

```
< [^>]* >
```

…it forces the engine to try **every possible way** to match the sequence when no matching `>` exists.

This is a textbook case of a **super-linear** regex pattern.

---

# ✔️ **How to avoid ReDoS**

## 1. Use a *non-greedy* version

```dart
RegExp(r'<[^>]*?>');
```

Still fragile, but **less prone** to catastrophic backtracking.

---

## 2. Use a *possessive* quantifier (not available in Dart RegExp 😞)

In languages like Java:

```
<[^>]*+>
```

Dart does **not** support this — so you can’t fully fix it with regex alone.

---

## 3. **Best solution: don’t use regex for HTML parsing**

Dart has an HTML parser that **avoids regex pitfalls** entirely:

```dart
import 'package:html/parser.dart' as html;

String convertHtmlContentToTextContent(String htmlContent) {
  final document = html.parse(htmlContent);
  return document.body?.text.trim() ?? '';
}
```

This eliminates:

* ReDoS risk
* Invalid HTML issues
* Missed entities
* Missing line breaks

---

# 🧠 **Conclusion**

**Yes**, your regex can be ReDoS-vulnerable with adversarial input.

If this code runs on:

* user-generated content
* emails
* scraped HTML
* chat messages

👉 **Do NOT use regex. Use an HTML parser.**

If you want, I can also analyze your *specific* use-case and tell you the actual risk level.

chibenwa, line 214: And yes, I fucking hate REGEXPs...

chibenwa: (Or use a re2 engine without backtracking if really forced with a regexp)

lib/l10n/intl_messages.arb
zatteo: Every l10n from the scribe is in the scribe module folder. What you can see in the app folder is the translation of the tooltip of the button in the composer bar. I considered that this button is part of the tmail app.

Screenshot 2025-12-15 at 07 28 59

lib/features/composer/presentation/widgets/mobile/tablet_bottom_bar_composer_widget.dart
dab246: What if another feature needs to use this button? When we need to integrate it into other apps, do we call the widget again with so many parameters like this?

lib/features/composer/presentation/widgets/web/bottom_bar_composer_widget.dart
dab246: If you create a GlobalKey here, when you want to use the `aiScribeButtonKey` button elsewhere, you'll have to create a GlobalKey for the `aiScribeButtonKey` again. And if the `aiScribeButtonKey` isn't rendered, the GlobalKey will still exist in the controller and won't be released, potentially leading to a memory leak.

core/lib/utils/string_convert.dart
zatteo: Very good point. I proposed a new version here https://github.com/linagora/tmail-flutter/pull/4197/changes/5f3b46dedacb59b7cad87dc40e907a053c8e25c6. FYI in the future (but I don't know if we talk about days or weeks or more), we may need to keep HTML here to be able to keep the styling. But for now it is simpler like this as it avoid to render HTML in the scribe.

chibenwa: MUCH better indeed. Thanks.

chibenwa (approved):
Still many changesout of lib/scribe but I'm unsure committing to more refactorings here on this specific feature is worth it.

dab246:
@coderabbitai please review again

core/lib/utils/string_convert.dart
```diff
@@ -196,4 +197,32 @@ class StringConvert {
...
+    const blockTags = [
+      'p',
+      'div',
+      'li',
+      'section',
+      'article',
+      'header',
+      'footer',
+      'h1','h2','h3','h4','h5','h6'
+    ];
```
dab246: Is this enough?

core/lib/utils/string_convert.dart
```diff
@@ -196,4 +197,32 @@ class StringConvert {
...
+
+    for (final tag in blockTags) {
+      document.querySelectorAll(tag).forEach((element) {
+        element.append(Text('\n'));
+      });
+    }
```
dab246: IMO, this process takes a considerable amount of time.

core/test/utils/string_convert_test.dart
```diff
@@ -816,4 +816,20 @@ void main() {
...
+      expect(result, 'First paragraph\nSecond paragraph\nThird paragraph');
+    });
+
+    test('should preserve line breaks for br', () {
+      const htmlContent = '<p>First paragraph</p><div><br></div><p>Second paragraph</p>';
+      final result = StringConvert.convertHtmlContentToTextContent(htmlContent);
+
+      expect(result, 'First paragraph\n\nSecond paragraph');
+    });
+  });
```
dab246, line 897: Write more test cases for it.

core/lib/utils/string_convert.dart
zatteo: I did a more complete test by doing a lot of thing in the composer. As a result, I added `blockquote` in the composer but otherwise it is enough. I also added more tests to demonstrate it.

See 3393df36580166a3ccf512fa2e5374f66d50d48b

core/lib/utils/string_convert.dart
zatteo: Good point. I updated the behavior to use a single querySelecterAll with a comma nested selector in 3393df36580166a3ccf512fa2e5374f66d50d48b which must be better.

lib/features/composer/presentation/widgets/web/bottom_bar_composer_widget.dart
zatteo: Understood. WDYT about this totally different approach that removes the aiScribeButtonKey ? https://github.com/linagora/tmail-flutter/pull/4207

hoangdat:
All things will be addressed in #4203

hoangdat (approved):

PR #4207, opened 2025-12-15
Tmail button widget action callback

zatteo:


core/lib/presentation/views/button/tmail_button_widget.dart
```diff
@@ -302,6 +308,7 @@ class TMailButtonWidget extends StatelessWidget {
     return TMailContainerWidget(
       onTapActionCallback: onTapActionCallback,
       onTapActionAtPositionCallback: onTapActionAtPositionCallback,
+      onTapActionAtOriginCallback: onTapActionAtOriginCallback,
```
chibenwa, line 311: I see this as a common pattern in this changeset that the calling context forward tons of of argument thought class fields downsteam.

While i understand this is an established pattern that we might not want to challenge here ( :-( ) I would like us to think how we Improve the overall structure of the code over here.

Ideas (uncontextualized) out of the box:

 - Grouping arguments together within case class

EG

```
case class WidgetStyle(Width with, Height height, MaxWidth maxWidth, MaxHeight maxHeight, BorderRadius borderRadius)
case class WidgetInteractions(onTapActionCallback: onTapActionCallback,  onTapActionAtPositionCallback, onTapActionAtOriginCallback, onTapActionAtOriginCallback);

And return TMailContainerWidget(widgetInteractions, widgetStyle, child: childWidget);
```

Bonus points: style convertions would be in WidgetStyle, interactions transformation in the case class as well

Extra bonus point - breaking evil combo: reuse.In over class

 - Other idea 2 staged builders: we only build a minimal object to which the caller would add it's state so that this class do not need to be aware.

 -   Very likely zillion of other good ideas.

Cc @hoangdat

dab246:
We don't need this anymore, please close it. @zatteo

PR #4251, opened 2026-01-14
[Proposal] Add AGENTS.md

zatteo:
I encountered multiple times the case where my AI agent did some work and then wanted to update Flutter packages by running `flutter pub get`. However because we use module we must run the prebuild script instead. It is a good example of what make sense in the [AGENTS.md](https://agents.md/) file that is more and more used.

What do you think
- about adding this file ?
- with this first content ?

It solved the issue in my case.
 
For the record, my current AI agent, but I think it will be the same for every other AI agent:
- tool: Kilo Code extension for VS Code
- model: Mistral Devstral 2

## Summary by CodeRabbit

* **Documentation**
  * Added AGENTS.md documentation file with build setup guidance, including instructions for updating Flutter packages via the prebuild script.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,5 @@
+# Twake Mail
+
+ ## Build
+
+ To update Flutter packages, run `./scripts/prebuild.sh` instead of `flutter pub get`.


chibenwa:
> about adding this file ?

Mitigated. On one side it disclose 'tooling' in the code base (I'm against IDE in project .gitignore for isntance), on the other it's valuable for fellow human too.

dab246:
IMO, It's unnecessary and redundant because `prebuild.sh` not only updates dependencies but also updates `l10n`. If you want to solve your problem, you should only use it locally.

zatteo:
> > about adding this file ?
> 
> Mitigated. On one side it disclose 'tooling' in the code base (I'm against IDE in project .gitignore for isntance), on the other it's valuable for fellow human too.

What is the issue in "disclosing" tooling in the code base ? Especially since this file is agent agnostic ?

zatteo:
> IMO, It's unnecessary and redundant because `prebuild.sh` not only updates dependencies but also updates `l10n`. If you want to solve your problem, you should only use it locally.

Hmmm agreeing that regenerating `l10n` will waste time. 

More generally, I have the feeling that this file _could_ be a lot useful (in this repo but in general as well) but unsure that this is the good usecase here.

chibenwa:
> What is the issue in "disclosing" tooling in the code base ?

More of an historical posture regarding IDEs, indeed regarding "agents" this might not be relevant.

Crash--:
I think adding AGENTS.md is the way to go today to share "agent" knowledge between us. I'm not sure this exemple is the right one, maybe we can just "/init" and push there. And we should update it every time. 

I see the AGENTS.md as a documentation, so, to me, it makes sense to add it here.

hoangdat:
IMO, we should keep `AGENTS.md` as the standard for agents; people with various coding agents can configure based on it in local.

hoangdat (approved):

Issue #15, opened 2021-07-28
Mock API to test jmap-dart-client


hoangdat:

hoangdat:
Mock with `DioAdapter`.
Merged https://github.com/linagora/jmap-dart-client/pull/8

PR #4281, opened 2026-01-27
fix: Relevance sort order now explicitly sorts by newest first

Tsopic:
## Summary

- Fix search results showing old emails (Jan 20) in full search view while quick search preview shows newest emails correctly
- Make `relevance` sort order return `receivedAt DESC` comparator instead of `None()`

## Problem

The previous fix (PR #15) removed `relevance` from `isScrollByPosition()` to prevent sending `position=0`. However, there was still a mismatch in the **sort comparator**:

| Component | Sort Sent to JMAP |
|-----------|-------------------|
| Quick search dropdown | `receivedAt DESC` (hardcoded) |
| Full search view | `null` (from `None().toNullable()`) |

When no sort comparator is sent, the JMAP server decides ordering, which doesn't consistently return newest emails first.

## Solution

Changed `EmailSortOrderType.relevance.getSortOrder()` to return `receivedAt DESC` instead of `None()`. This ensures:
- Full search matches dropdown preview behavior
- Consistent user experience across search paths
- Newest emails shown first as users expect

## Note on Semantics

While "relevance" suggests server-side relevance scoring, the JMAP server doesn't provide visible relevance-based ordering when no comparator is specified. This change prioritizes consistent UX over semantic naming. A future improvement could rename this to `mostRecentMatch` or similar.

## Test plan

- [x] All 1274 tests pass
- [ ] Manual test: Search for a term, verify full view shows same newest emails as dropdown preview

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/lib/features/mailbox_dashboard/presentation/model/search/email_sort_order_type.dart
+++ b/lib/features/mailbox_dashboard/presentation/model/search/email_sort_order_type.dart
@@ -59,7 +59,11 @@ enum EmailSortOrderType {
               ..setIsAscending(true))
         );
       case EmailSortOrderType.relevance:
-        return const None();
+        return Some(
+          <Comparator>{}
+            ..add(EmailComparator(EmailComparatorProperty.receivedAt)
+              ..setIsAscending(false))
+        );
       case EmailSortOrderType.senderAscending:
         return Some(
           <Comparator>{}


chibenwa:
Hello

Thanks for the report, that is interesting!

The issue is more the inconsistency between the preview and the actual sort used for results: we should apply the sort on both.

I opened https://github.com/linagora/tmail-flutter/issues/4282 to that end.

Regarding the default sort order it is intendeed to rely on server defaults when none. TMail server uses a relevance sort based on OpenSearch scoring and thus delivers best response first. I would suggest to see with Stallwart community if a related behaviour could be implemented too on top of Stallwart.

Issue #4274, opened 2026-01-26
Sometimes, Email disappear in email list


hoangdat:
#### Behavior 
```
She received emailA yesterday in her inbox, read it
This morning the mail disapeared from mail list
She switched mailbox + clicked refresh -> no effect
The mail could be seen with search
I asked her to logout login again and the mail magically reappeared
```

### DoD
- [ ] find the way to reproduce
- [ ] write the test and ADR to document it

PR #4288, opened 2026-01-28
TF-4274 Mitigating sometimes email disappear in email list

dab246:
## Issue

#4274

## Workaround

- `F5, Logout, Refresh button`-> clear cache 
=> No need flag `FORCE_EMAIL_QUERY`
- `Switch mailbox` -> query 
=> Need flag `FORCE_EMAIL_QUERY`

## Config Helm

- https://ci.linagora.com/linagora/lrs/saas/tools/helm-charts/tmail-frontend/-/merge_requests/13 

## Summary by CodeRabbit

* **New Features**
  * Configurable "force email query" flag to control heavier network fetches for web users

* **Bug Fixes**
  * Reduced missing/disappearing emails by standardizing cache invalidation on refresh, logout, and UI refresh

* **Improvements**
  * Consolidated and more reliable cache-clearing flows across sessions and mailbox switches
  * Safer reload/unload behavior on web to improve email visibility

* **Tests**
  * Added tests covering the force-query behavior and fallback scenarios


chibenwa:
I'd like that we clarify the PR title: as far as I understand this, we are **mitigating** not **fixing**. Correct?

Can we update the PR title accordingly?

dab246:
> I'd like that we clarify the PR title: as far as I understand this, we are **mitigating** not **fixing**. Correct?
> 
> Can we update the PR title accordingly?

Yes, because the solution we're using for this issue is similar to the previous CNB cache fix, we're using the previous commits to resolve this issue on master. So I'll update the title to be more accurate.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
```diff
@@ -611,10 +611,17 @@ class MailboxDashBoardController extends ReloadableController
...
         twakeAppManager.hasComposer &&
-        !twakeAppManager.isExecutingBeforeReconnect) {
+        !twakeAppManager.isExecutingBeforeReconnect;
+
+    if (shouldPrevent) {
       event.preventDefault();
+      return;
+    }
+
+    if (PlatformInfo.isWeb) {
```
tddang-linagora: Would this check neccessary if `onBeforeUnloadBrowserListener` only triggers on browser?

lib/features/manage_account/presentation/manage_account_dashboard_controller.dart
```diff
@@ -136,6 +137,13 @@ class ManageAccountDashBoardController extends ReloadableController
     _getParametersRouter();
   }
 
+  @override
+  Future<void> onBeforeUnloadBrowserListener(html.Event event) async {
+    if (PlatformInfo.isWeb) {
```
tddang-linagora: idem

hoangdat:
- [x] cannot jump back to app if token expired

dab246:
> * [x]  cannot jump back to app if token expired

Fixed

Crash--:
I asked Sonnet about this issue and it tells me that https://github.com/linagora/tmail-flutter/compare/master...fix/Test4274#diff-76bcb6d2435718cd09b0cb2ad8686022302023338e4394efd188fcc71e650bda 

Is it something that can explain the root cause of the issue?

hoangdat:
Who is Sonet?

Because it is sometimes issues, cache unsynchronize well.  still try to put Sentry to understand what is it. In the meanwhile time, we make user can easier to revalidate cache when refresh/logout/f5

dab246:
> I asked Sonnet about this issue and it tells me that [master...fix/Test4274#diff-76bcb6d2435718cd09b0cb2ad8686022302023338e4394efd188fcc71e650bda](https://github.com/linagora/tmail-flutter/compare/master...fix/Test4274#diff-76bcb6d2435718cd09b0cb2ad8686022302023338e4394efd188fcc71e650bda)
> 
> Is it something that can explain the root cause of the issue?

We can't be certain about it, as we haven't been able to replicate the problem. Based on the fact that the app works normally when users log out and log back in, we've come up with the best solution at this time to avoid similar errors. Regarding Sonet's analysis, the scenarios it outlines are also inaccurate because it hasn't been successfully replicated.

lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart
dab246: Fixed

lib/features/manage_account/presentation/manage_account_dashboard_controller.dart
dab246: Fixed

Crash--:
@dab246 Do we have a list of all the tests / scenarios we've tested? Since we're struggling with this issue for a few months now, we need to be methodic to not lose time for nothing. Also maybe with all the scenarios / tests we've done, if we read them, maybe we'll have ideas on what else to test.

tddang-linagora (approved):

lib/features/manage_account/presentation/manage_account_dashboard_controller.dart
```diff
@@ -136,6 +137,13 @@ class ManageAccountDashBoardController extends ReloadableController
     _getParametersRouter();
   }
 
+  @override
+  Future<void> onBeforeUnloadBrowserListener(html.Event event) async {
+    if (event is html.BeforeUnloadEvent) {
+      await cachingManager.clearAllEmailAndStateCache();
+    }
+  }
```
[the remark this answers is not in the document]
dab246: In this case, it still works correctly because the cache clearing happens synchronously.

dab246:
> @dab246 Do we have a list of all the tests / scenarios we've tested? Since we're struggling with this issue for a few months now, we need to be methodic to not lose time for nothing. Also maybe with all the scenarios / tests we've done, if we read them, maybe we'll have ideas on what else to test.

Currently, no, because this problem is very difficult to reproduce. Errors happen sometimes, and they can have many preconditions that we can't detect. Therefore, we need the help of tracking tools like Sentry to better understand user activity before and after this error occurs. Anyway, @Crash--  if you have any ideas about this issue, please feel free to share them so we can discuss it. Thanks.

hoangdat (approved):

hoangdat:
> Do we have a list of all the tests / scenarios we've tested?

@Crash-- 

We tried to focus to test in the synchronization in client: web socket in web/mobile, push notificaiton in mobile. More and more case to attach to this, but it is similar like these 

### in Web/Mobile :
- Device A: try to open a folder, read emails
- Device B: make change in other devices (read/star/unstar/delete/send email to myself/receive email) with the same account to trigger sync in Device A
- Observe in Device A

### in Mobile:
#### Case 1
- Device A: try to open a folder, read emails
- Turn off network in Device A, try to do sth in Device B
- Turn on network in Device B, open app to observe

#### Case 2
- Device A: try to open a folder, read emails, kill app (app in terminated)
- Send message to Device A
- Try to click on notification to open email
- Open some folder
- Repeat multiple times

hoangdat:
@Crash-- 
cc @chibenwa 
This solution is designed as a safety net to mitigate cache-related risks.
In case the client cache becomes corrupted or out of sync, a refresh or logout will trigger a full data re-synchronization, restoring a clean and reliable state.

Issue #4284, opened 2026-01-27
Refreshing causes "virtual mailboxes" to be hidden


chibenwa:
WHEN I click refresh
THEN the Favourites + Needs action mailbox disappear from left side bar
AND reappears only once loading completes

Expectation: left side bar stays the same upon refresh

chibenwa:
https://github.com/user-attachments/assets/4c05da74-c212-4a8c-a77b-613b2130524a

PR #4290, opened 2026-01-29
TF-4284 Fix refreshing causes "virtual mailboxes" to be hidden

dab246:
## Issue

#4284 

## Root cause

Because each refresh involves rebuilding the tree before adding the virtual folders (Favorite/Action Required), the mailbox list view is re-rendered.

## Solution

Add virtual folders when creating the mailbox tree. Only then trigger the rendering to the UI.

## Resolved

https://github.com/user-attachments/assets/f82fc423-a1c3-4932-9e51-b5faca18a475

## Summary by CodeRabbit

* **New Features**
  * Added a consolidated MailboxCollection model and a single update entry point with optional transformation hooks; AI-driven actions are supported but off by default.

* **Refactor**
  * Mailbox tree/list updates moved to an immutable, functional flow with a centralized routine to keep default, personal, team and “all” views synchronized.
  * Favorite and action-required flows now return updated collections instead of mutating state.

* **Tests**
  * Tests updated to use the new MailboxCollection shape.

### DIFF
--- a/lib/features/base/base_mailbox_controller.dart
+++ b/lib/features/base/base_mailbox_controller.dart
@@ -37,6 +37,7 @@ import 'package:tmail_ui_user/features/mailbox/presentation/extensions/presentat
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_actions.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_categories.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_categories_expand_mode.dart';
+import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_collection.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_node.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_tree.dart';
 import 'package:tmail_ui_user/features/mailbox/presentation/model/mailbox_tree_builder.dart';
@@ -62,6 +63,7 @@ typedef OnMoveFolderContentActionCallback = void Function(
 );
 typedef DeleteMailboxActionCallback = void Function(PresentationMailbox mailbox);
 typedef AllowSubaddressingActionCallback = void Function(MailboxId, Map<String, List<String>?>?, MailboxActions);
+typedef OnUpdateMailboxCollectionCallback = MailboxCollection Function(MailboxCollection);
 
 abstract class BaseMailboxController extends BaseController
     with ExpandFolderTriggerScrollableMixin {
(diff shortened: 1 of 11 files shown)


lib/features/base/base_mailbox_controller.dart
```diff
@@ -675,10 +696,26 @@ abstract class BaseMailboxController extends BaseController
     }
   }
 
-  void autoCreateVirtualFolder(bool isAINeedsActionEnabled) {
-    addFavoriteFolderToMailboxList();
+  bool get isAINeedsActionEnabled => false;
```
tddang-linagora, line 702: Was this line for test?

lib/features/base/base_mailbox_controller.dart
```diff
@@ -675,10 +696,26 @@ abstract class BaseMailboxController extends BaseController
...
+
+  MailboxCollection autoCreateVirtualFolder({
```
tddang-linagora: Should it be private method if it's not used outside of this class?

tddang-linagora: Should the virtual folder be a part of `MailboxCollection`, so it's always there, and we can pull it out to use in the UI depends on the config? Just so that we don't have to remember to call `autoCreateVirtualFolder` when new `MailboxCollection` is created somewhere else. Just an opinion. Close the comment if you find it unreasonable.

dab246: Describe your algorithm here and we will review it.

dab246: It can be used with child classes, so we should make it public instead of private.

lib/features/base/base_mailbox_controller.dart
dab246: No. Why do you think so? Do you have any suggestions?

tddang-linagora: I see a hard-coded value, just curious.

lib/features/base/base_mailbox_controller.dart
tddang-linagora: Nevermind. It has external dependencies.

tddang-linagora (dismissed):

lib/features/mailbox/presentation/extensions/handle_action_required_tab_extension.dart
hoangdat: how we apply the same pattern with other, return a MailboxCollection instead of mutate directly?

dab246: fixed

tddang-linagora (dismissed):

lib/features/mailbox/presentation/extensions/handle_favorite_tab_extension.dart
```diff
@@ -2,47 +2,65 @@ import 'package:model/mailbox/presentation_mailbox.dart';
...
-    PresentationMailbox favoriteFolder,
-  ) {
-    final defaultMailboxNode = defaultMailboxTree.value.root;
-    List<MailboxNode> currentDefaultFolders =
-        defaultMailboxNode.childrenItems ?? [];
+  MailboxTree _addFavoriteFolderToDefaultMailboxTree({
+    required PresentationMailbox favoriteFolder,
+    required MailboxTree defaultTree,
+  }) {
+    final defaultMailboxNode = defaultTree.root;
```
hoangdat, line 40: maybe we should verify the `alreadyExists`?

dab246: The duplicate check is already handled inside `insertAfterByPriority` via  `_containsMailbox` (list_mailbox_node_extension.dart:34), which compares by mailbox ID and returns early if the node already exists. Adding another  `alreadyExists` guard here would be redundant.

hoangdat:
how about to add these tests to `mailbox_tree_builder_test`?

```dart
  group('MailboxCollection allMailboxes field', () {
    final inbox = PresentationMailbox(
      MailboxId(Id('inbox')),
      name: MailboxName('Inbox'),
      role: Role('inbox'),
    );
    final sent = PresentationMailbox(
      MailboxId(Id('sent')),
      name: MailboxName('Sent'),
      role: Role('sent'),
    );

    test(
        'generateMailboxTreeInUI: returned allMailboxes matches input list',
        () async {
      final input = [inbox, sent];

      final result = await TreeBuilder().generateMailboxTreeInUI(
        allMailboxes: input,
        currentCollection: MailboxCollection.empty(),
      );

      expect(result.allMailboxes.length, equals(input.length));
      expect(
        result.allMailboxes.map((m) => m.id).toList(),
        containsAll(input.map((m) => m.id)),
      );
    });

    test(
        'generateMailboxTreeInUIAfterRefreshChanges: returned allMailboxes matches input list',
        () async {
      final input = [inbox, sent];

      final result =
          await TreeBuilder().generateMailboxTreeInUIAfterRefreshChanges(
        allMailboxes: input,
        currentCollection: MailboxCollection.empty(),
      );

      expect(result.allMailboxes.length, equals(input.length));
      expect(
        result.allMailboxes.map((m) => m.id).toList(),
        containsAll(input.map((m) => m.id)),
      );
    });

    test(
        'generateMailboxTreeInUIAfterRefreshChanges: allMailboxes in result excludes virtual folders from current collection',
        () async {
      final serverMailboxes = [inbox, sent];
      final virtualFavorite = PresentationMailbox.favoriteFolder;

      final previousCollection = MailboxCollection(
        allMailboxes: [inbox, virtualFavorite],
        defaultTree: MailboxTree(MailboxNode.root()),
        personalTree: MailboxTree(MailboxNode.root()),
        teamMailboxTree: MailboxTree(MailboxNode.root()),
      );

      final result =
          await TreeBuilder().generateMailboxTreeInUIAfterRefreshChanges(
        allMailboxes: serverMailboxes,
        currentCollection: previousCollection,
      );

      expect(
        result.allMailboxes.every((m) => !m.isVirtualFolder),
        isTrue,
        reason:
            'Tree builder must not inject virtual folders — they are added by the callback after buildTree/refreshTree',
      );
      expect(result.allMailboxes.length, equals(serverMailboxes.length));
    });
  });
  ```

dab246:
> how about to add these tests to `mailbox_tree_builder_test`?

Test case 3 has been applied, along with some additional test cases. Test cases 1 and 2 have been skipped because:

Test 1 is technically correct but weak — it only checks no items are lost/duplicated, while the meaningful behavior (selected mailbox → deactivated) is already covered by Test 4. It adds overlap without value.

Test 2 is a tautology — it asserts `list == list` since the method returns input as-is. It can never fail, so it provides no real safety.

More valuable tests are those ensuring no stale data (virtual folders, deleted mailboxes) leaks after refresh, which are already covered elsewhere.

hoangdat (approved):

Issue #4265, opened 2026-01-21
Attachment reminder should be scoped to user issue


chibenwa:
We have recently received several user complaints regarding the prompt “Forgot to attach a file?” being displayed even when users have not typed the words attach or attachment in their email content.

Upon verification, we observed that this occurs because emails sent to external domains automatically include a disclaimer, which contains the word “attachment”. As a result, the attachment reminder is being triggered based on the disclaimer text rather than the actual email body written by the user.

Could we adjust the detection logic so that the verification for the keywords attach / attachment applies only to the user-composed email body, explicitly excluding the disclaimer section?

This would help avoid false positives and improve the user experience.

CF Email/get bacting

Crash--:
Also, let's write an email with "filetage", (which is a french word absolutely not related to file), i've the popup saying that I've forgot to attach my file... So it means that we're not looking for "file" but for "file*". This is something we need to improve

chibenwa:
>  we're not looking for "file" 

Actually we need language specific normalizer (case, plural, gender, etc) coupled with a decent tokenizer

CF https://docs.opensearch.org/latest/analyzers/

Is there some good dart package for doing NLP ?

hoangdat:
- [ ] exclude our signature by CSS class, but if user copy/paste or chane anything in signature, we cannot

hoangdat:
### Current logic
- we depend on scanning texts with keywords which provide by Zoe, we still not care to the languagle 

### Compare with thunderbird 
- now thunderbird also support a preference to make user control himself

### Next actions
- find the package can handle NLP logic
- work with Zoe on the direction for this ticket 

@poupotte

Crash--:
I think you can split the work in 3 phases:

- First only analyze the text written by the user (to answer benoit use case about the disclamer). I don't think you need @poupotte on that. 
- Second: We should check words like "file " "file.", but "filetage" should not trigger the alert. I don't think you need @poupotte on that. 
- Third: Add NPL (because even with npl, it the second point is this there, we'll still have issue). Maybe yes, you'll need @poupotte to see if we prioritize that or not ^^

chibenwa:
Agreed that in the absence of nice Flutter library addressing `hird` is likely expensive and not worth the complexity and return over investment.

hoangdat:
### DoD
- [x] exclude our signature by CSS class
- [x] check for `file`, `files`, `file.`, `file?`, ....
- [x] exclude list?
[...]

PR #4291, opened 2026-01-30
TF-4265 Fix attachment reminder 

dab246:
## Issue

#4265 

## Resolved

https://github.com/user-attachments/assets/2d100b6e-a3b7-4553-b19f-2c2db00c410e

## Additional

- Configuration for Attachment Keyword Detection: https://github.com/linagora/tmail-flutter/compare/bugfix/tf-4265-attachment-reminder-should-be-scoped-to-user-issue?expand=1#diff-60220cfca73c8577d3c8c00572ccb6c157fb530534a54131f4a203f41f566c29

## Integration test

[Screen_recording_20260130_160727.webm](https://github.com/user-attachments/assets/78d0cb69-ea97-4315-ab40-c964045b3a1f)

```console
🧪 Should not see attachment reminder when send email with attachment keyword in signature
        ✅   1. waitUntilVisible widgets with type "TwakeWelcomeView".
        ✅   2. tap widgets with text "Use company server".
        ✅   3. waitUntilVisible widgets with type "LoginView".
        ✅   4. tap widgets with type "TextField" descending from widgets with key [<'dns_lookup_input_form'>].
        ✅   5. enterText widgets with type "TextField" descending from widgets with key [<'dns_lookup_input_form'>].
        ✅   6. waitUntilVisible widgets with text "bob".
        ✅   7. tap widgets with text "Next".
        ✅   8. waitUntilVisible widgets with key [<'base_url_form'>] descending from widgets with type "LoginView".
        ✅   9. enterText widgets with type "TextField" descending from widgets with key [<'base_url_form'>].
        ✅  10. waitUntilVisible widgets with text containing a368b69d925b.ngrok-free.app.
        ✅  11. tap widgets with text "Next".
        ✅  12. waitUntilVisible widgets with key [<'credential_input_form'>] descending from widgets with type "LoginView".
        ✅  13. enterText widgets with type "TextField" descending from widgets with key [<'login_username_input'>].
        ✅  14. waitUntilVisible widgets with text "bob@example.com".
        ✅  15. tap widgets with type "TextField" descending from widgets with key [<'login_password_input'>].
        ✅  16. enterText widgets with type "TextField" descending from widgets with key [<'login_password_input'>].
        ⏳  17. tap widgets with key [<'loginSubmitForm'>].
        : 
        ✅  17. tap widgets with key [<'loginSubmitForm'>].
        ✅  18. isPermissionDialogVisible (native)
        ✅  19. waitUntilVisible widgets with type "ThreadView".
        ✅  20. tap widgets with type "InkWell" descending from widgets with type "ComposeFloatingButton".
        ✅  21. waitUntilVisible widgets with type "ComposerView".
        ✅  22. isPermissionDialogVisible (native)
        ✅  23. grantPermissionWhenInUse (native)
        ✅  24. tap widgets with type "SubjectComposerWidget".
        ✅  25. enterText widgets with type "SubjectComposerWidget".
        ⏳  26. tap widgets with type "InAppWebView" descending from widgets with type "HtmlEditor" descending from widgets with type "MobileEditorView" descending from widg
        ✅  26. tap widgets with type "InAppWebView" descending from widgets with type "HtmlEditor" descending from widgets with type "MobileEditorView" descending from widgets with widget matching predicate descending from widgets with type "ComposerView" inclusive.
        ✅  27. tap widgets with key [<'prefix_to_recipient_expand_button'>].
        ✅  28. enterText widgets with widget matching predicate descending from widgets with type "RecipientComposerWidget" inclusive.
        ✅  29. tap widgets with widget matching predicate descending from widgets with type "RecipientSuggestionItemWidget" inclusive.
        ⏳  30. tap widgets with widget matching predicate descending from widgets with type "TMailButtonWidget" descending from widgets with type "AppBarComposerWidget" inc
        ✅  30. tap widgets with widget matching predicate descending from widgets with type "TMailButtonWidget" descending from widgets with type "AppBarComposerWidget" inclusive.
✅ Should not see attachment reminder when send email with attachment keyword in signature (integration_test/tests/composer/attachment_reminder_test.dart) (33s)

Test summary:
📝 Total: 1
✅ Successful: 1
❌ Failed: 0
⏩ Skipped: 0
📊 Report: file:///Users/datvu/WorkingSpace/tmail-flutter/build/app/reports/androidTests/connected/index.html
⏱️  Duration: 113s
```

## Release Notes

* **New Features**
  * Added attachment reminder that alerts users when composing emails mentioning attachments but without actual files attached.
  * Introduced configurable keyword detection for attachment reminders with support for multiple languages and custom keyword lists.
  * Enhanced email signature detection and removal for more accurate keyword matching.

* **Documentation**
  * Added comprehensive guide for configuring attachment keyword detection settings.


docs/configuration/attachment_keywords_configuration.md
```diff
@@ -0,0 +1,77 @@
...
+
+```
+
+* `includeList`: List of keywords to **add** to the detection dictionary.
+  * Used when the default dictionary (hardcoded in app) is missing specific business terms.
+  * Matches are case-insensitive.
+  * Example: Adding `"cv"` will trigger the alert for "Attached is my CV".
+
+
+* `excludeList`: List of tokens to **ignore/block** from detection.
```
chibenwa: I do not think we can provide an exhaustive list. In itself this seems inherently brittle.

I bet the idea is to address the example raised by @Crash-- but bear in mind what this is - examples! There's likely 100s of others...

chibenwa (commented):
Thanks for the comprehensive tests. While I do not believe in the blacklist approach however the improvement of the tokenization is obvious.

lib/features/composer/presentation/model/keyword_config.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:equatable/equatable.dart';
+import 'package:json_annotation/json_annotation.dart';
+
+part 'keyword_config.g.dart';
+
+@JsonSerializable(includeIfNull: false, explicitToJson: true)
+class KeywordConfig with EquatableMixin {
```
chibenwa:
```suggestion
class AttachmentKeywordConfig with EquatableMixin {
```

chibenwa (commented):
Let's streamline this!

=> Fix my naming comment
=> run claude locally to get a feedback ?
=> Push and get this audited by code scene

Merged by 09/04

lib/features/composer/presentation/model/keyword_config.dart
dab246: fixed

docs/configuration/attachment_keywords_configuration.md
```diff
@@ -0,0 +1,95 @@
...
+  * For emails **< 20,000 characters**: runs synchronously on the UI thread (fast, < 10ms).
+  * For emails **≥ 20,000 characters**: runs in a Dart Isolate via `compute()` to avoid UI jank.
+
+* **Execution Flow:**
+  1. User clicks "Send".
+  2. Email HTML is converted to plain text (signatures and quotes stripped).
+  3. App loads `includeList` and `excludeList` from config (cached after first load).
+  4. `includeList` is merged with the built-in multi-language keyword dictionary.
+  5. A Regex pattern is built (or retrieved from cache) and scans the plain text.
+  6. Each match is validated against the `excludeList` using full surrounding token context.
```
tddang-linagora: Can the `excludeList` be used in the same regex with the `includeList`? So we don't have to evaluate the pattern twice. Maybe this regex `(?<![\p{L}])(?!(?:exclude1|exclude2)(?![\p{L}]))(keyword1|keyword2)(?![\p{L}])` ?

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -38,51 +46,81 @@ class AttachmentTextDetector {
...
+  /// Technical Rationale:
+  /// 1. Frame Budget (16ms): To maintain 60fps, operations on the UI thread must stay under 16ms.
+  /// 2. Isolate Overhead: `compute` takes about 2-4ms just to copy data and start the thread.
+  /// 3. Real-world Benchmark:
+  ///    - For < 20k chars, Sync execution is usually faster (< 10ms) than the Async overhead.
+  ///    - For > 20k chars, Regex processing might exceed 16ms on low-end devices, causing jank.
+  ///
+  /// -> Below 20k: Run Sync (Instant response).
+  /// -> Above 20k: Run Async (Safety for UI).
+  static const int _kAsyncExecutionThreshold = 20000;
```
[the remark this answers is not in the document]
dab246: fixed

test/features/composer/presentation/manager/exclude_list_filter_test.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:tmail_ui_user/features/composer/presentation/manager/exclude_list_filter.dart';
+
+/// Helper: simulate a regex [Match] at [start]..[end] within [text].
+Match _fakeMatch(String text, int start, int end) {
+  final regex = RegExp(RegExp.escape(text.substring(start, end)));
+  return regex.firstMatch(text)!;
+}
```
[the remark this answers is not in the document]
dab246: fixed

lib/features/composer/presentation/manager/attachment_text_detector.dart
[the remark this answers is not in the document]
dab246: comment is updated. please check again

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -882,14 +882,18 @@ class ComposerController extends BaseController
     }
 
     final emailContent = await getContentInEditor();
+
+    final attachmentKeywords = await validateAttachmentReminder(
+      emailSubject: subjectEmail.value ?? '',
+      emailContent: emailContent,
+    );
```
tddang-linagora, line 905: Is there any reason it was move before the `mounted` check?

dab246: Because `validateAttachmentReminder` is `async` (runs in an Isolate for large emails). The widget can be disposed during that await. Checking `context.mounted` only makes sense after an `async` gap, placing it before would be a no-op since the context is obviously still mounted the moment the user taps Send.                Standard Flutter pattern: `await → check mounted → use context`.

docs/configuration/attachment_keywords_configuration.md
dab246: No. The negative lookahead (?!(?:exclude1|exclude2)) is anchored at the keyword's start position. It fails when the keyword appears in the middle or end of an excluded  token.                                                                                                                                                                 Example: exclude = "icon-app-file", keyword = "file", the lookahead at "file" sees "file here", not "icon-app-file", so the exclusion is missed. The current ExcludeListFilter uses getSurroundingToken() to extract the full token around each match before comparing, context a single regex cannot replicate.

tddang-linagora (dismissed):

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -38,51 +46,86 @@ class AttachmentTextDetector {
...
+  /// Long drafts (> 20k chars) may still block the event loop on web. A proper web
+  /// strategy (e.g., Service Worker or chunked async processing) is not yet implemented.
+  static const int _kAsyncExecutionThreshold = 20000;
+
   /// Detect if the text contains keywords suggesting there is an attachment.
   /// [lang] is the language code (`en`, `fr`, `ru`, `vi`, `ar`, ...).
+  ///
+  /// **Note:** Uses plain `String.contains()` — no word-boundary check.
+  /// Prefer [matchedKeywordsUnique] for production send-path detection.
   static bool containsAttachmentKeyword(String text, {required String lang}) {
```
hoangdat: should we remove some method in this file which not in use

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -38,51 +46,86 @@ class AttachmentTextDetector {
...
-    return keywords.any((k) => lowerText.contains(k.toLowerCase()));
+    return keywords.any((k) => lowerText.contains(k));
   }
 
-  /// Returns a list of matched keywords (if any) for the specified language
+  /// Returns a list of matched keywords (if any) for the specified language.
+  ///
+  /// **Note:** Uses plain `String.contains()` — no word-boundary check.
+  /// Prefer [matchedKeywordsUnique] for production send-path detection.
   static List<String> matchedKeywords(String text, {required String lang}) {
```
hoangdat: idem

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -38,51 +46,86 @@ class AttachmentTextDetector {
...
+        .where((k) => lowerText.contains(k))
         .toList();
   }
 
-  /// Detect if text contains keyword in any language
+  /// Detect if text contains keyword in any language.
+  ///
+  /// **Note:** Uses plain `String.contains()` — no word-boundary check.
+  /// Prefer [matchedKeywordsUnique] for production send-path detection.
   static bool containsAnyAttachmentKeyword(String text) {
```
hoangdat: idem

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -38,51 +46,86 @@ class AttachmentTextDetector {
...
     }
     return false;
   }
 
-  /// Returns a map of languages with a list of matched keywords
+  /// Returns a map of languages with a list of matched keywords.
+  ///
+  /// **Note:** Uses plain `String.contains()` — no word-boundary check.
+  /// Prefer [matchedKeywordsUnique] for production send-path detection.
   static Map<String, List<String>> matchedKeywordsAll(String text) {
```
hoangdat: idem

lib/features/base/base_controller.dart
```diff
@@ -574,6 +575,8 @@ abstract class BaseController extends GetxController
       await cachingManager.closeHive();
     } catch (e) {
       logWarning('BaseController::clearAllData: Cannot clear all data: $e');
+    } finally {
+      AttachmentKeywordConfigManager().clearCache();
```
hoangdat, line 603: should get it via binding?

lib/features/composer/presentation/extensions/attachment_detection_extension.dart
```diff
@@ -4,24 +4,26 @@ import 'package:flutter/material.dart';
...
     try {
       final fullContent = '$emailSubject $emailContent';
       final plainText = HtmlUtils.extractPlainText(fullContent);
-      final keywords = AttachmentTextDetector.matchedKeywordsUnique(plainText);
-      if (keywords.isEmpty) {
-        return [];
-      } else {
-        return keywords;
-      }
+      final attachmentKeywordConfig = await AttachmentKeywordConfigManager().getConfig();
```
hoangdat, line 20: binding

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -882,14 +882,18 @@ class ComposerController extends BaseController
     }
 
     final emailContent = await getContentInEditor();
+
+    final attachmentKeywords = await validateAttachmentReminder(
```
hoangdat, line 902: how about skip check reminder if user uploaded attachment?

test/features/composer/presentation/manager/attachment_text_detector_test.dart
hoangdat, line 1: how about to add these tests to combine the test with config, htmlUtils, and text detector?
[attachment_detection_pipeline_test.dart.txt](https://github.com/user-attachments/files/26505494/attachment_detection_pipeline_test.dart.txt)

docs/configuration/attachment_keywords_configuration.md
hoangdat, line 1: please shorter the configuration guide

lib/features/composer/presentation/manager/attachment_text_detector.dart
```diff
@@ -91,19 +134,113 @@ class AttachmentTextDetector {
...
+  /// The core logic function.
+  /// 1. Builds/Retrieves the Regex.
+  /// 2. Scans the text.
+  /// 3. Filters results using Exclude List.
+  /// Can be executed on the main thread (sync path) or inside an Isolate (async path).
+  static List<String> _detectKeywords(DetectionParams params) {
+    final text = params.text;
+    if (text.isEmpty) return [];
+
+    final regex = RegExp(
```
hoangdat, line 132: this was compile everytime `_detectKeywords` was call. So maybe caching also the regex. WDYT?

lib/features/base/base_controller.dart
dab246: AttachmentKeywordConfigManager is an app-scoped singleton (pattern factory with static _instance), not a user-scoped service, so it doesn't go through GetX binding. Reasons:
- It contains configuration from assets/ (independent of user session)
- It's used in both BaseController.clearAllData() (without binding context) and ComposerController extensions
- If it were included in binding, it would have to use Get.find<AttachmentKeywordConfigManager>() in a place without a suitable scope (e.g., BaseController)
=> The singleton approach is appropriate here. clearCache() is called in the correct place within the finally block of clearAllData().

lib/features/composer/presentation/extensions/attachment_detection_extension.dart
dab246: idem

test/features/composer/presentation/manager/attachment_text_detector_test.dart
dab246: added

lib/features/composer/presentation/composer_controller.dart
dab246: fixed

lib/features/composer/presentation/manager/attachment_text_detector.dart
dab246: _detectKeywords runs inside the compute() isolate, each isolate has its own heap, and the static cache is not shared between isolates. RegExp also cannot be sent over the isolate boundary, so the pattern string must be passed and compiled in the isolate each time. With a sync path (< 20k characters), caching can save ~0.1ms, but it's not worth the added complexity. The pattern string is cached in _getPattern(), which is the most expensive part (sort + join + escape). Compiling RegExp from the cached string is faster. => Keeping as-is is the right approach.

hoangdat (approved):
pleae rebase and squash

dab246:
> pleae rebase and squash


Done

PR #4298, opened 2026-02-03
HOTFIX Integration test 0.24.8

tddang-linagora:
Test summary:
📝 Total: 2
✅ Successful: 2
❌ Failed: 0
⏩ Skipped: 0
📊 Report: file:///Users/dangdat/Desktop/dev/linagora-master/tmail-flutter/build/app/reports/androidTests/connected/index.html
⏱️  Duration: 190s

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rich text editor cleanup to avoid issues when closing the composer.

* **Tests**
  * Added a UI action to discard composer changes in integration tests.
  * Updated forwarding-email tests to reliably detect attachments.
  * Consolidated and improved template-creation tests to assert templates by subject and streamline cleanup.

### DIFF
--- a/lib/features/composer/presentation/controller/rich_text_mobile_tablet_controller.dart
+++ b/lib/features/composer/presentation/controller/rich_text_mobile_tablet_controller.dart
@@ -5,6 +5,7 @@ import 'package:core/utils/html/html_utils.dart';
 import 'package:core/utils/platform_info.dart';
 import 'package:file_picker/file_picker.dart';
 import 'package:flutter/cupertino.dart';
+import 'package:flutter/foundation.dart';
 import 'package:get/get.dart';
 import 'package:rich_text_composer/rich_text_composer.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/header_style_type.dart';
@@ -75,10 +76,8 @@ class RichTextMobileTabletController extends GetxController {
   @override
   void onClose() {
     richTextController.dispose();
-    try {
+    if (!kDebugMode) {
       htmlEditorApi?.webViewController.dispose();
-    } catch (e) {
-      logWarning('RichTextMobileTabletController::onClose:Exception: $e');
     }
     htmlEditorApi = null;
     super.onClose();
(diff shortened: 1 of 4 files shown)


hoangdat (dismissed):

lib/features/composer/presentation/controller/rich_text_mobile_tablet_controller.dart
```diff
@@ -75,7 +75,11 @@ class RichTextMobileTabletController extends GetxController {
   @override
   void onClose() {
     richTextController.dispose();
-    htmlEditorApi?.webViewController.dispose();
+    try {
+      htmlEditorApi?.webViewController.dispose();
+    } catch (e) {
+      logWarning('RichTextMobileTabletController::onClose:Exception: $e');
+    }
```
dab246: why do we need try/catch at here?

tddang-linagora: When run test in Android, it throws with message the controller is already disposed

dab246: Does running app release cause this issue, or only in debugging? Should we check if it has already been disposed before calling dispose?

tddang-linagora: Only in debug, and there's no way of checking whether it has been disposed or not. Should we replace it with `if (!kDebugMode)`?

dab246: IMO, we should add it.

tddang-linagora: Done

dab246 (dismissed):
```suggestion
if (kDebugMode) {
  try {
    htmlEditorApi?.webViewController.dispose();
  } catch (e) {
    logWarning('Dispose error: $e');
  }
} else {
  htmlEditorApi?.webViewController.dispose();
}
```

dab246 (approved):

hoangdat (approved):

PR #4313, opened 2026-02-05
[Fix] Handle `400 Bad Request` during Token Refresh flow

dab246:
## Issue

This PR addresses a issue where the application fails to handle `400 Bad Request` errors during the OIDC token refresh process. Previously, when the refresh token expired or was revoked, the app would either loop or return a generic error, 
leaving the user stuck in an invalid state without being logged out.

## Root cause

Missing Error Handling: The `AuthorizationInterceptors` logic previously treated all errors during the refresh step as generic network errors. It did not specifically check for `400 Bad Request`

## Solution

- Catch 400: Specifically catch 400 error within the `refreshingTokensOIDC` function.
- Force Logout Flow: Upon refresh failure 400:
  - Invoke clear() to reset the current authentication state.
  - Propagate a `RefreshTokenFailedException` (wrapped inside a `DioException`) to allow the UI layer to detect the failure and redirect to the Login screen.
- Testing: Add Unit Test coverage for scenarios where the refresh token request returns a 400 error.

## Resolved

Screenshot 2026-02-05 at 15 42 16

## Summary by CodeRabbit

* **New Features**
  * Stronger handling for refresh-token failures that triggers immediate recovery flows
  * iOS keychain session updates for improved session persistence

* **Bug Fixes**
  * Prevented token duplication during refresh and clarified session-expired handling
  * Improved detection and propagation of invalid/revoked sessions

* **Tests**
  * Added coverage for refresh-token failure (400 Bad Request) scenarios

### DIFF
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -142,6 +142,15 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         log('AuthorizationInterceptors::onError: Request using old token');
         isRetryRequest = true;
       } else {
+        logTrace(
+          'AuthorizationInterceptors::onError: '
+          '401 received but refresh skipped. '
+          'statusCode = ${err.response?.statusCode} | '
+          'authType = $_authenticationType | '
+          'hasConfig = ${_configOIDC != null} | '
+          'url = ${err.requestOptions.uri}',
+          webConsoleEnabled: true,
+        );
         return super.onError(err, handler);
       }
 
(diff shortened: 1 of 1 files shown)


chibenwa:
Thanks for having a look into it 

However in the network inspector I did not see 400 error codes, mostly just 400

Also in sentry top 2 errors are:

image

(For canary alone!)

This let me but think that the issue also lies down in the way we handle 401 !

Could you have a look please?

hoangdat:
@chibenwa 
- for 400, we want to handle the case when refreshToken also expire (not happen in canary because lifetime of refreshToken in canary is 30 days) - we found during the time we debugging
- for 401, we still try to find the way to handle it

Issue #4314, opened 2026-02-05
Unread emails disapear (blink!)


chibenwa:
https://github.com/user-attachments/assets/b57efd85-71e7-4c3b-8880-6d18a3a479b8

v 0.24.8 canary

PR #4316, opened 2026-02-06
TF-4314 Fix unread emails disappear blink

dab246:
## Issue

#4314

## Reproduce

https://github.com/user-attachments/assets/82b077fa-db6d-4a9c-bbbd-4a30599ce060

## Root cause

Inconsistent caching strategy when `FORCE_EMAIL_QUERY=true`. While the `Open Folder` action bypasses caching, the `Load More` action persists data to the cache. This leads to state inconsistency upon app restart. The app initializes with the latest server state, prompting the WebSocket to push only incremental updates. However, the UI hydrates with stale data from the `Load More` cache, resulting in a discrepancy between the displayed list and the actual server state.

## Solution

- Update email to cache when load mailbox with `FORCE_EMAIL_QUERY=true`
- Add `limit` property for refresh change emails method

## Resolved

https://github.com/user-attachments/assets/12af14c2-0f31-40b4-bf5a-89d5395a3c94

## Summary by CodeRabbit

* **New Features**
  * Optional per-request limit when fetching emails, allowing refreshes and forced queries to return a capped number of items.

* **Bug Fixes**
  * Prevented local email data from being overwritten by empty server responses during refresh.
  * More consistent paging during email loads and refreshes to honor per-request limits.

* **Tests**
  * Updated tests to cover the new per-request limit behavior.

### DIFF
--- a/lib/features/thread/data/repository/thread_repository_impl.dart
+++ b/lib/features/thread/data/repository/thread_repository_impl.dart
@@ -188,7 +188,8 @@ class ThreadRepositoryImpl extends ThreadRepository {
       'Server email count = $serverCount',
     );
 
-    if (serverCount > 0) {
+    if (serverCount > 0 ||
+        (serverResponse.notFoundEmailIds?.isNotEmpty ?? false)) {
       await _updateEmailCache(
         accountId,
         session.username,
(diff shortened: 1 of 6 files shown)


hoangdat:
IMO, we should have dedicated method for `refreshChange` in case of `forceEmailQuery = true`.

Please check the case loadmore many email then do an action, refresh change only get in first page, so the email list very short

dab246:
> IMO, we should have dedicated method for `refreshChange` in case of `forceEmailQuery = true`.
> 
> Please check the case loadmore many email then do an action, refresh change only get in first page, so the email list very short

Ok, we've revised the refresh change logic to be more accurate for this case.

hoangdat (approved):

Issue #4308, opened 2026-02-04
[Label] Add labels for several messages


dab246:
# Context

When user selects several mail, today he can’t add a label to all this email

# Specs

This issue consists to add the possibility to categorize a list of messages.

- Add label icon in the list of action with tooltip ‘Label as’ between ‘Move’ and ‘Mark as spam’ action
- When user click on it, we display the list of all label
    - User selects the label(s)
- We add label(s) in all messages selected

<img width="591" height="368" alt="Image" src="https://github.com/user-attachments/assets/c018e37f-b07f-44fa-bb70-01b4edfdacee" />

poupotte:
Product review: I want

PR #4319, opened 2026-02-09
TF-4308 Add labels for several messages

dab246:
## Issue

#4308 

## Resolved

https://github.com/user-attachments/assets/074ae176-15aa-432b-9e75-c82a53cd465b

## Summary by CodeRabbit

* **New Features**
  * Bulk label assignment: apply multiple labels to many emails at once.
  * Label selection modal for choosing labels during batch actions.
  * Top-bar label action to trigger labeling for selected emails.

* **UI/UX Improvements**
  * Visual feedback for selected labels in lists.
  * Configurable close-button alignment.
  * User-facing success/failure messages and localizations for batch labeling.


tddang-linagora (approved):

hoangdat:
- [ ] how to display the selected tag for a bunch of emails?

image

hoangdat:
- [ ] opened email was not well updated if we tag multiple emails in email list in tablet mode 

image

lib/features/mailbox_dashboard/presentation/extensions/update_current_emails_flags_extension.dart
hoangdat, line 81: should check in case no `keywords` exist?

lib/features/mailbox_dashboard/presentation/bindings/email_action_interactor_bindings.dart
```diff
@@ -13,5 +14,6 @@ class EmailActionInteractorBindings extends Bindings {
   void dependencies() {
     Get.lazyPut(() => AddALabelToAnEmailInteractor(_emailRepository));
     Get.lazyPut(() => RemoveALabelFromAnEmailInteractor(_emailRepository));
+    Get.put(AddListLabelToListEmailsInteractor(_emailRepository));
```
hoangdat, line 17: why not lazyPut?

hoangdat:
- [x] close button in the top-left not follow other dialog

image

lib/main/localizations/app_localizations.dart
```diff
@@ -5776,4 +5776,39 @@ class AppLocalizations {
...
+  String get selectLabel {
+    return Intl.message(
+      'Select label',
+      name: 'selectLabel',
+    );
+  }
+
+  String get addListLabelToListEmailSuccessfullyMessage {
+    return Intl.message(
+      'Labels have been successfully added to these emails.',
```
hoangdat, line 5796: message here and message at `addListLabelToListEmailHasSomeFailureMessage` are the same, maybe make user confused

chibenwa:
image

...

That's too much

chibenwa (changes requested):
I disagree with this change which denotes clear misdesign of the overall application

chibenwa:
```
● Yes, there is a real design problem here, but the comment needs to be precise to be actionable. Here's the breakdown:
                                                                                                                                                                                                            
  ---                                                                                                                                                                                                       
  What the change does
                                                                                                                                                                                                            
  The old AddLabelToListEmailsActionMixin (26 lines) just opened a modal. The new AddListLabelsToListEmailsMixin (177 lines) absorbs the entire interactor chain: session validation, account ID checks,  
  state consumption, toast display, and UI sync callback.                                                                                                                                                   
   
  ---                                                                                                                                                                                                       
  The concrete misdesign     
                        
  1. The self-reference anti-pattern — the strongest argument
                                                                                                                                                                                                            
  The mixin declares:
  BaseController get currentController;                                                                                                                                                                     
  And MailboxDashBoardController implements it as:
  BaseController get currentController => this;                                                                                                                                                             
                                                                                                                                                                                                            
  The mixin is mixed into a BaseController, so it could call consumeState(...) directly. Instead it routes through a getter that returns this. This is circular indirection with no benefit — it just       
  obscures the fact that the mixin could use this directly if it constrained its on clause to BaseController.                                                                                               
                             
  2. The mixin requires too many abstract contracts                                                                                                                                                         
                             
  mixin AddListLabelsToListEmailsMixin on EmitStateMixin {                                                                                                                                                  
    AccountId? get currentAccountId;                                                                                                                                                                        
    Session? get currentSession;                                                                                                                                                                            
    ToastManager get currentToastManager;                                                                                                                                                                   
    BaseController get currentController;         // self-reference                                                                                                                                         
    AddListLabelToListEmailsInteractor? get addListLabelToListEmailInteractor;                                                                                                                              
    OnSyncListLabelForListEmail? get onSyncListLabelForListEmail;                                                                                                                                           
  }                          
                                                                                                                                                                                                            
  6 abstract getters to mix in one behavior. This is the mixin pattern being used as a substitute for proper dependency injection. The consuming class is forced to wire all these manually, making the     
  mixin hard to reuse and easy to misuse.
                                                                                                                                                                                                            
  3. Cross-controller delegation for UI                                                                                                                                                                     
   
  Both ThreadController and SearchEmailController call:                                                                                                                                                     
  mailboxDashBoardController.openChooseLabelModal(...)
                                                                                                                                                                                                            
  The dashboard controller should not own "open a label modal" as a method. This creates a hidden dependency where thread/search features cannot function without the dashboard controller being alive and
  injected — a coupling that will bite in tests and in future refactors.                                                                                                                                    
   
  ---                                                                                                                                                                                                       
  Is "misdesign of the overall application" too strong?
                                                                                                                                                                                                            
  Partially. The self-reference (currentController => this) and the 6-getter contract are clear local misdesigns. The cross-controller delegation is a broader symptom of the app's pattern of routing UI
  actions through MailboxDashBoardController as a god object.                                                                                                                                               
                             
```

For the record...

tddang-linagora:
>That's too much

Split into 2:
- #4430
- #4431 

>misdesign

A new approach is in #4431

chibenwa:
Please @tddang-linagora explain this new approach and why it shall be prefered.

Which difference with this work?

Thanks by advance!

tddang-linagora:
> Please @tddang-linagora explain this new approach and why it shall be prefered.
> 
> Which difference with this work?
> 
> Thanks by advance!

Instead of creating a mixin but inject a lot of information into it making it tightly coupled, my approach is like extending the capability of an existing feature without modify the code inside it, typical open-closed. It is fully testable, minimal dependency, and single responsibility. You can add endless feature to a core functionality this way.

chibenwa:
> Instead of creating a mixin but inject a lot of information into it making it tightly coupled, my approach is like extending the capability of an existing feature without modify the code inside it, typical open-closed. It is fully testable, minimal dependency, and single responsibility. You can add endless feature to a core functionality this way.

Ok

And part one is the refactoring? Part 2 the implem?

tddang-linagora:
> > Instead of creating a mixin but inject a lot of information into it making it tightly coupled, my approach is like extending the capability of an existing feature without modify the code inside it, typical open-closed. It is fully testable, minimal dependency, and single responsibility. You can add endless feature to a core functionality this way.
> 
> Ok
> 
> And part one is the refactoring? Part 2 the implem?

yes exactly.

chibenwa:
@dab246 opinion on approach 2?

If this is concensual then I have no problem to move forward with approach 2.

dab246:
> @dab246 opinion on approach 2?
> 
> If this is concensual then I have no problem to move forward with approach 2.

IMO, approach 2 doesn't significantly reduce the amount of code and file changes for this issue. Approach 2 simply uses a controller instead of a mixin, but we still have to manage it, initialize it before use, and release it when not in use. Nevertheless, approach 2 is still worth considering, and we can continue with that approach.

Issue #4329, opened 2026-02-11
Rogue Mailbox/set request for favorite


chibenwa:
```
[...]

PR #4331, opened 2026-02-12
TF-4329 Fix rogue mailbox set request for favorite

dab246:
## Issue

#4329 

## Reproduce

https://github.com/user-attachments/assets/fb58e6b5-3012-4604-a62d-f97d77d44e92

## Root Cause

The `Starred` folder is currently included in the parent mailbox list within the folder creation modal. Since `Starred` is a virtual folder with the ID `"favorite"` (instead of a standard UUID), selecting it as a parent triggers an `Invalid UUID string: favorite` error. Virtual folders do not possess a valid UUID and cannot support child folders.

## Solution

Filter out the `Starred` folder from the parent folder selection list. As a virtual folder, it does not support nesting, so users should be prevented from selecting it as a parent directory.

## Resolved

https://github.com/user-attachments/assets/e13b7cf6-e292-4c6b-93ba-993100b2de0e

## Summary by CodeRabbit

* **Bug Fixes**
  * Virtual mailboxes are now excluded from the mailbox creation interface.

* **Tests**
  * Added comprehensive tests to verify virtual mailbox filtering, order preservation, non-mutation, and performance on large lists.

### DIFF
--- a/model/lib/extensions/list_presentation_mailbox_extension.dart
+++ b/model/lib/extensions/list_presentation_mailbox_extension.dart
@@ -1,4 +1,4 @@
-
+import 'package:collection/collection.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
 import 'package:model/extensions/presentation_mailbox_extension.dart';
 import 'package:model/mailbox/presentation_mailbox.dart';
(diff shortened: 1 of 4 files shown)


model/lib/extensions/list_presentation_mailbox_extension.dart
```diff
@@ -18,4 +18,7 @@ extension ListPresentationMailboxExtension on List<PresentationMailbox> {
   bool get isAllUnreadMailboxes => every((mailbox) => mailbox.countUnReadEmailsAsString.isNotEmpty);
 
   List<MailboxId> get mailboxIds => map((mailbox) => mailbox.id).toList();
+
+  List<PresentationMailbox> get withoutVirtualMailbox =>
```
hoangdat, line 22: please add unit-test

dab246: Added

hoangdat (approved):

Issue #4343, opened 2026-02-23
Load Sentry from Sentry CDN


Crash--:
### Description
We're currently loading sentry from sentry CDN 
### Expected result
Not loading from external source 
### Current behavior

js_util_patch.dart:173  GET https://browser.sentry-cdn.com/10.6.0/bundle.tracing.min.js net::ERR_BLOCKED_BY_CLIENT

On Canary Tmail

hoangdat:
- what is the context of the issue? customer block this? 
- and the expectation is what?

chibenwa:
Loading executable from other sources is not a good practice 

We can consider it as a potential for code injection

chibenwa:
So this ticket is a MUST

Crash--:
@hoangdat we already have this discussion about loading fonts and wasm from external sources. I think you still have a ticket for some JS. 

For privacy issue (since you’re leaking your users informations) we should never rely on third party loading. (Fonts / JS / whatever).

poupotte:
@hoangdat : I let you validate this ticket

hoangdat:
<img width="687" height="208" alt="Image" src="https://github.com/user-attachments/assets/543c20a9-c449-4faa-a16c-9c77c4e76f97" />

PR #4348, opened 2026-02-25
TF-4343 Fix load Sentry from Sentry CDN

dab246:
## Issue

#4343 

## Reproduce

https://github.com/user-attachments/assets/711714b3-5269-46f0-af3f-e44f41f52af0

## Root Cause

By default, the `sentry_flutter` SDK on Flutter Web automatically injects a `<script>` tag to load the `bundle.tracing.min.js` file from an external CDN (`browser.sentry-cdn.com`). Browsers and ad-blocking extensions often flag this as a third-party tracker and block the network request, resulting in the `ERR_BLOCKED_BY_CLIENT` error.

Attempting to fix this by disabling `autoInitializeNativeSdk` and manually calling `Sentry.init` via JavaScript Interop is **not recommended** for the Web platform. It violates Sentry's shared environment guidelines and leads to **Duplicate Events** because both the Dart SDK and JS SDK attach global error handlers simultaneously.

## Solution

Implement a **DOM Hijacking (Interceptor)** approach to serve the script locally without altering the native behavior of the `sentry_flutter` Dart SDK:

* **Self-hosting & Cache Busting**: Download the `bundle.tracing.min.js` file, place it in the `web/js/` directory, and load it statically in the `<head>` of `web/index.html` using `<script src="js/sentry-tracing.min.js?v=1.0.0"></script>`. The `?v=...` query string ensures cache busting on future updates.
* **DOM Interceptor (`sentry-interceptor.js`)**: Create a lightweight JavaScript file and load it *before* the Flutter bootstrap script. This script overrides the browser's native `document.createElement` method to act as a middleware.
* **Block CDN at Birth**: When the Flutter Dart SDK attempts to dynamically create a script tag for the Sentry CDN, the interceptor catches the `src` assignment. It blocks the external URL, prevents the network request, and artificially dispatches a `load` event. This tricks the Dart SDK into believing the CDN script loaded successfully, forcing it to seamlessly use the local Sentry instance we already provided.
* **Trusted Types Compatibility**: The interceptor safely stringifies values (e.g., handling `TrustedScriptURL` objects used by Flutter's strict CSP during bootstrap) before checking for the CDN URL. This ensures the interceptor does not crash the Flutter engine's initialization process.

## Resolved

https://github.com/user-attachments/assets/442d589c-d0da-4cda-a39d-ef44ac11256a

## Summary by CodeRabbit

* **Chores**
  * Added self-hosted telemetry/tracing scripts to the page head.
  * Added a client-side interceptor that prevents specific external tracking scripts from loading, blocks their network requests, and simulates successful script loads so the app continues to run without disruption.


core/lib/utils/sentry/sentry_initializer.dart
```diff
@@ -25,6 +27,16 @@ class SentryInitializer {
...
 
+    if (PlatformInfo.isWeb) {
+      initSentryWeb(
+        config.dsn,
+        config.environment,
+        config.release,
+        config.tracesSampleRate,
+        config.isDebug,
+      );
+    }
```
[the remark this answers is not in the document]
dab246: Fixed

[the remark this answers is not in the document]
dab246: @coderabbitai

web/js/sentry-interceptor.js
```diff
@@ -0,0 +1,63 @@
...
+          if (shouldBlockSentryUrl(urlString)) {
+            console.log('[Sentry Interceptor] 🛑 Blocked CDN request (Property):', urlString);
+            setTimeout(() => this.dispatchEvent(new Event('load')), 10);
+          } else {
+            originalSetAttribute.call(this, 'src', val);
+          }
+        },
+        get: function() {
+          return this.getAttribute('src') || '';
+        }
```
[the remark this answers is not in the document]
dab246: Fixed

tddang-linagora (dismissed):

web/js/sentry-interceptor.js
```diff
@@ -0,0 +1,69 @@
...
+       'src'
+      );
+
+      Object.defineProperty(element, 'src', {
+        configurable: true,
+        set: function(val) {
+          const urlString = val ? String(val) : '';
+
+          if (shouldBlockSentryUrl(urlString)) {
+            console.log('[Sentry Interceptor] 🛑 Blocked CDN request (Property):', urlString);
```
hoangdat: console error

web/js/sentry-interceptor.js
```diff
@@ -0,0 +1,69 @@
...
+          return nativeSrcDescriptor?.get?.call(this) ?? '';
+        }
+      });
+
+      element.setAttribute = function(name, value) {
+        if (name === 'src' || name === 'SRC') {
+          const urlString = value ? String(value) : '';
+
+          if (shouldBlockSentryUrl(urlString)) {
+            console.log('[Sentry Interceptor] 🛑 Blocked CDN request (Attribute):', urlString);
```
hoangdat: console error

web/index.html
```diff
@@ -41,6 +41,8 @@
   <script src="worker_service/worker_service.js"></script>
   <script src="i18n/translater.js"></script>
   <script src="https://unpkg.com/@dotlottie/player-component@2.7.12/dist/dotlottie-player.mjs" type="module"></script>
+  <script src="js/sentry-tracing.min.js?v=1.0.0"></script>
```
hoangdat: If this file is ever swapped (compromised CI, malicious PR), the browser will execute it silently. SRI makes the browser verify the hash before running — tampered file gets blocked automatically.

hoangdat: Fx:  
```
<script src="js/sentry-tracing.min.js?v=1.0.0" integrity="sha384-<hash>"               
  crossorigin="anonymous"></script>  
```

web/index.html
```diff
@@ -41,6 +41,8 @@
...
+  <script src="js/sentry-interceptor.js?v=1.0.0"></script>
```
hoangdat: idem for this

web/js/sentry-interceptor.js
```diff
@@ -0,0 +1,69 @@
...
+            setTimeout(() => this.dispatchEvent(new Event('load')), 10);
```
hoangdat: imo, no need this. WDYT?

web/index.html
hoangdat: should have a comment to ensure the order of script be followed well
```
  <!-- Load Sentry SDK first, then interceptor to block subsequent CDN injection -->     
```

dab246: Why use this method? Where did you get this information from? Is there an official document?

web/js/sentry-interceptor.js
dab246: This shouldn't be an error, it's just a warning, so we don't need console.error.

web/js/sentry-interceptor.js
dab246: This shouldn't be an error, it's just a warning, so we don't need console.error.

web/index.html
hoangdat: to ensure the check to prevent the load of malicious JS. And moreover, can fetching new script when we update the new one.

web/js/sentry-interceptor.js
dab246: The `setTimeout + dispatchEvent('load')` is intentional. When the Sentry SDK dynamically injects a CDN script, it waits for the load event before resolving its internal Promise. If we silently drop the src without firing load, the SDK hangs indefinitely. This synthetic event acts as a "polite block" — the CDN request never happens, but the SDK's initialization flow continues normally.

web/index.html
dab246: added

web/index.html
dab246: added

web/index.html
dab246: added

web/js/sentry-interceptor.js
hoangdat: but it show to every people in console. keep it as error will hide it for user, make it less scare

web/js/sentry-interceptor.js
hoangdat: but it show to every people in console. keep it as error will hide it for user, make it less scare

hoangdat (approved):

hoangdat:
Tested in STG: no load SDK from CDN, and log still good

Issue #4363, opened 2026-03-05
Thread - Display by collapsed thread in search


dab246:
# Context

Some month ago, we added thread mode in our application. However, we have some views (search, inbox) which continue to display list of messages by unique messages instead of by thread.

# Specs

This issue consists in results view to display message by thread instead of by messages.

<img width="738" height="385" alt="Image" src="https://github.com/user-attachments/assets/8fe5ff2a-39c7-4481-a057-ec048b92de9f" />

poupotte:
Product review: I want

PR #4364, opened 2026-03-05
TF-4363 [Thread] Display by collapsed thread in search

dab246:
## Issue

#4363 

## Resolved

- With `Thread is enabled`:

https://github.com/user-attachments/assets/85ca430f-d04e-4db3-a3a0-b3cfacaaba6a

- With `Thread is disabled`:

https://github.com/user-attachments/assets/78c4fc59-d366-4574-a3f6-ce6469f50126

## Summary by CodeRabbit

* **New Features**
  * Thread-collapse option for searches/lists tied to local settings; UI reacts and refreshes when the setting changes.
  * Local settings service added to persist and reload thread-collapse preferences.

* **Refactor**
  * Search requests and controllers now propagate the collapse-threads flag end-to-end; query construction uses availability-safe helpers.

* **Bug Fixes**
  * Thread-detail selection now no-ops when no email is opened.

* **Tests**
  * Tests updated to cover collapse-setting behavior and integrations.

### DIFF
--- a/docs/adr/0085-riverpod-state-management-for-local-settings.md
+++ b/docs/adr/0085-riverpod-state-management-for-local-settings.md
@@ -17,7 +17,7 @@ As part of implementing the collapsed thread feature in search email (TF-4363), 
 - `ThreadController` — must react when the setting changes to re-trigger search
 - `SearchEmailController` — must read the current value at query time
 
-The setting is written by `PreferencesController` via `UpdateLocalSettingsInteractor` when the user toggles it in `PreferencesView`, and must also be reset on logout to prevent stale cross-account data.
+The setting is written by `PreferencesController` via `UpdateLocalSettingsInteractor` when the user toggles it in `PreferencesView`.
 
 Rather than routing this through a GetX-based `LocalSettingsService` with a reactive `Rx` field and `ever(...)` listeners — which would require manual equality tracking and expose mutable state broadly — we chose to use Riverpod's `StateNotifier` as the single source of truth from the start.
 
@@ -30,17 +30,17 @@ Introduce `localSettingsNotifierProvider` (`StateNotifier<PreferencesSetting>`) 
 ### Architecture
 
 ```text
-┌─────────────────────────────────────────────────────┐
-│  appProviderContainer  (global ProviderContainer)   │
-│                                                     │
-│   localSettingsNotifierProvider                     │
-│   StateNotifier<PreferencesSetting>                 │
-│       ▲ write          ▲ write         read ▼       │
-│       │                │                   │        │
-│  LocalSettings    Preferences         Thread /      │
-│  Service          Controller          Search        │
-│  (initial load)   (on toggle)         Controllers   │
-└─────────────────────────────────────────────────────┘
+┌──────────────────────────────────────────────────────────┐
+│  appProviderContainer  (global ProviderContainer)        │
+│                                                          │
+│   localSettingsNotifierProvider                          │
+│   StateNotifier<PreferencesSetting>                      │
+│       ▲ write              ▲ write        read ▼         │
+│       │                    │                  │          │
+│  LocalSettingsService  PreferencesController  Thread /   │
+│  (app start /          (on toggle)            Search     │
+│   route reload)                               Controllers│
+└──────────────────────────────────────────────────────────┘
 ```
 
 ### Write paths
@@ -50,9 +50,6 @@ Introduce `localSettingsNotifierProvider` (`StateNotifier<PreferencesSetting>`) 
 
 **On user toggle in PreferencesView:**
 `PreferencesController.handleSuccessViewState(UpdateLocalSettingsSuccess)` → `localSettingsNotifierProvider.notifier.update()`
-
-**On logout:**
-`BaseController.clearAllData()` → `localSettingsNotifierProvider.notifier.update(PreferencesSetting.initial())`
 
 ### Read paths
 
--- a/docs/adr/0085-riverpod-state-management-for-local-settings.md
+++ b/docs/adr/0085-riverpod-state-management-for-local-settings.md
@@ -1,0 +1,90 @@
+# 85. Riverpod State Management Pilot — Local Settings & Collapsed Thread in Search
+
+Date: 2026-04-23
+
+## Status
+
+Accepted
+
+## Related ADRs
+
+- [ADR-0071](./0071-collapse-threads-in-email-query.md) — collapseThreads in Email/query
+
+## Context
+
+As part of implementing the collapsed thread feature in search email (TF-4363), two controllers need to read the `collapseThreads` local setting:
+
+- `ThreadController` — must react when the setting changes to re-trigger search
+- `SearchEmailController` — must read the current value at query time
+
+The setting is written by `PreferencesController` via `UpdateLocalSettingsInteractor` when the user toggles it in `PreferencesView`, and must also be reset on logout to prevent stale cross-account data.
+
+Rather than routing this through a GetX-based `LocalSettingsService` with a reactive `Rx` field and `ever(...)` listeners — which would require manual equality tracking and expose mutable state broadly — we chose to use Riverpod's `StateNotifier` as the single source of truth from the start.
+
+This is also a **pilot** to validate the GetX + Riverpod coexistence pattern before applying it more broadly across the codebase.
+
+## Decision
+
+Introduce `localSettingsNotifierProvider` (`StateNotifier<PreferencesSetting>`) backed by a global `ProviderContainer` (`appProviderContainer`). All reads and writes go through this provider directly.
+
+### Architecture
+
+```text
+┌─────────────────────────────────────────────────────┐
+│  appProviderContainer  (global ProviderContainer)   │
+│                                                     │
+│   localSettingsNotifierProvider                     │
+│   StateNotifier<PreferencesSetting>                 │
+│       ▲ write          ▲ write         read ▼       │
+│       │                │                   │        │
+│  LocalSettings    Preferences         Thread /      │
+│  Service          Controller          Search        │
+│  (initial load)   (on toggle)         Controllers   │
+└─────────────────────────────────────────────────────┘
+```
+
+### Write paths
+
+**On app start / settings route reload:**
+`LocalSettingsService.onInit()` → `_loadInitialSettings()` → cache read → `localSettingsNotifierProvider.notifier.update()`
+
+**On user toggle in PreferencesView:**
+`PreferencesController.handleSuccessViewState(UpdateLocalSettingsSuccess)` → `localSettingsNotifierProvider.notifier.update()`
+
+**On logout:**
+`BaseController.clearAllData()` → `localSettingsNotifierProvider.notifier.update(PreferencesSetting.initial())`
+
+### Read paths
+
+**`ThreadController`** — subscribes via `appProviderContainer.listen(localSettingsNotifierProvider, ...)`. Riverpod skips notification when `PreferencesSetting` equality is unchanged (`EquatableMixin`), so no manual dedup is needed.
+
+**`SearchEmailController`** — reads on demand via `appProviderContainer.read(localSettingsNotifierProvider)`.
+
+### `LocalSettingsService` role
+
+`LocalSettingsService` is a `GetxService` acting as a **temporary bridge** between the GetX DI graph and Riverpod. It remains a `GetxService` because all DI wiring in the project uses GetX bindings — introducing a plain class would be inconsistent with the existing layer. Its sole responsibility is loading the initial value from cache into the provider on startup via `onInit`.
+
+It will be removed entirely once `ManageAccountRepository` and its datasources are migrated to Riverpod providers, at which point `LocalSettingsNotifier` can inject and call `GetLocalSettingsInteractor` directly.
+
+### Settings route reload on web
+
+When the user reloads the browser directly on the settings route, `MailboxDashBoardBindings` does not run. `ManageAccountDashBoardBindings` guards against this:
+
+```dart
+if (!Get.isRegistered<LocalSettingsService>()) {
+  Get.put(LocalSettingsService(Get.find<GetLocalSettingsInteractor>()));
+}
+```
+
+## Consequences
+
+**Positive**
+- No redundant cache round-trip — provider is updated directly from `UpdateLocalSettingsSuccess` result
+- No manual equality tracking needed — Riverpod + `EquatableMixin` handles it
+- Single write surface for `PreferencesSetting`; no mutable `Rx` exposed across the codebase
+- Validates the GetX + Riverpod coexistence pattern for future migrations
+
+**Negative**
+- `appProviderContainer` is a global singleton — controllers depend directly on a concrete implementation rather than an abstraction, which is a known DIP violation. The correct solution is to wrap it behind an interface (e.g. `LocalSettingsRepository`), but doing so now would add complexity before the pattern is proven at scale. This should be addressed when the migration reaches the repository layer.
+- Two DI systems (GetX + Riverpod) coexist during the migration period; developers must know which layer owns which state
+- `LocalSettingsService` is a temporary class that must be removed when the migration reaches the repository layer
--- a/lib/features/manage_account/presentation/services/local_settings_service.dart
+++ b/lib/features/manage_account/presentation/services/local_settings_service.dart
@@ -17,10 +17,10 @@ class LocalSettingsService extends GetxService {
     _loadLocalSettings();
   }
 
-  void refresh() => _loadLocalSettings();
+  Future<void> refresh() => _loadLocalSettings();
 
-  void _loadLocalSettings() {
-    _getLocalSettingsInteractor.execute().last.then(
+  Future<void> _loadLocalSettings() {
+    return _getLocalSettingsInteractor.execute().last.then(
       (result) => result.fold(
         (failure) => logWarning(
           'LocalSettingsService::_loadLocalSettings:failure: $failure',
@@ -31,6 +31,10 @@ class LocalSettingsService extends GetxService {
           }
         },
       ),
-    );
+    ).catchError((error, stackTrace) {
+      logWarning(
+        'LocalSettingsService::_loadLocalSettings:error: $error | stackTrace: $stackTrace',
+      );
+    });
   }
 }
--- a/lib/features/manage_account/presentation/services/local_settings_service.dart
+++ b/lib/features/manage_account/presentation/services/local_settings_service.dart
@@ -1,0 +1,59 @@
+import 'package:core/utils/app_logger.dart';
+import 'package:get/get.dart';
+import 'package:tmail_ui_user/features/manage_account/domain/model/preferences/preferences_setting.dart';
+import 'package:tmail_ui_user/features/manage_account/domain/state/get_local_settings_state.dart';
+import 'package:tmail_ui_user/features/manage_account/domain/usecases/get_local_settings_interactor.dart';
+import 'package:tmail_ui_user/features/thread_detail/presentation/action/thread_detail_ui_action.dart';
+
+class LocalSettingsService extends GetxService {
+  final GetLocalSettingsInteractor _getLocalSettingsInteractor;
+  final Rxn<ThreadDetailUIAction> _threadDetailUIAction;
+
+  LocalSettingsService(
+    this._getLocalSettingsInteractor,
+    this._threadDetailUIAction,
+  );
+
+  final localSettings = PreferencesSetting.initial().obs;
+
+  Worker? _settingsWorker;
+
+  @override
+  void onInit() {
+    super.onInit();
+    _listenToSettingsChanges();
+    _loadLocalSettings();
+  }
+
+  @override
+  void onClose() {
+    _settingsWorker?.dispose();
+    super.onClose();
+  }
+
+  void _listenToSettingsChanges() {
+    _settingsWorker = ever(
+      _threadDetailUIAction,
+      (action) {
+        if (action is UpdatedThreadDetailSettingAction) {
+          _loadLocalSettings();
+        }
+      },
+    );
+  }
+
+  void _loadLocalSettings() {
+    _getLocalSettingsInteractor.execute().last.then(
+      (result) => result.fold(
+        (failure) => logWarning(
+          'LocalSettingsService::_loadLocalSettings:failure: $failure',
+        ),
+        (success) {
+          if (success is GetLocalSettingsSuccess) {
+            localSettings.value = success.preferencesSetting;
+          }
+        },
+      ),
+    );
+  }
+}
--- a/lib/features/manage_account/presentation/services/local_settings_service.dart
+++ b/lib/features/manage_account/presentation/services/local_settings_service.dart
@@ -1,42 +1,41 @@
 import 'package:core/utils/app_logger.dart';
 import 'package:get/get.dart';
-import 'package:tmail_ui_user/features/manage_account/domain/model/preferences/preferences_setting.dart';
 import 'package:tmail_ui_user/features/manage_account/domain/state/get_local_settings_state.dart';
 import 'package:tmail_ui_user/features/manage_account/domain/usecases/get_local_settings_interactor.dart';
+import 'package:tmail_ui_user/features/manage_account/presentation/providers/local_settings_notifier.dart';
+import 'package:tmail_ui_user/main/providers/app_provider_container.dart';
 
+/// Responsible for loading local settings from cache on startup and pushing
+/// the result into [localSettingsNotifierProvider] so all Riverpod consumers
+/// receive the initial value automatically.
 class LocalSettingsService extends GetxService {
   final GetLocalSettingsInteractor _getLocalSettingsInteractor;
 
   LocalSettingsService(this._getLocalSettingsInteractor);
 
-  final localSettings = PreferencesSetting.initial().obs;
-
-  Future<void>? _pendingLoad;
-
   @override
   void onInit() {
     super.onInit();
-    reload();
+    _loadInitialSettings();
   }
 
-  Future<void> reload() => _pendingLoad ??= _loadLocalSettings()
-      .whenComplete(() => _pendingLoad = null);
-
-  Future<void> _loadLocalSettings() {
+  Future<void> _loadInitialSettings() {
     return _getLocalSettingsInteractor.execute().last.then(
       (result) => result.fold(
         (failure) => logWarning(
-          'LocalSettingsService::_loadLocalSettings:failure: $failure',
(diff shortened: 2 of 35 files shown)


hoangdat:
please rebase with `master` branch

dab246:
- [x] Remove `Count` in email item of email list view

chibenwa:
Rebase needed

This work is opened for over a month.

Also diff is massive. I'd like we get opinion of @9clg6 before moving forward and merging.

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -425,6 +433,17 @@ class ThreadController extends BaseController with EmailActionController {
         _peakEmailCount = countEmails;
       }
     });
+
+    _registerLocalSettingsListener();
+  }
+
+  void _registerLocalSettingsListener() {
+    if (!Get.isRegistered<LocalSettingsService>()) return;
+    ever(_localSettingsService.localSettings, (_) {
```
hoangdat: should we dispose it?

lib/features/manage_account/presentation/services/local_settings_service.dart
```diff
@@ -0,0 +1,59 @@
...
+import 'package:tmail_ui_user/features/manage_account/domain/usecases/get_local_settings_interactor.dart';
+import 'package:tmail_ui_user/features/thread_detail/presentation/action/thread_detail_ui_action.dart';
+
+class LocalSettingsService extends GetxService {
+  final GetLocalSettingsInteractor _getLocalSettingsInteractor;
+  final Rxn<ThreadDetailUIAction> _threadDetailUIAction;
+
+  LocalSettingsService(
+    this._getLocalSettingsInteractor,
+    this._threadDetailUIAction,
```
hoangdat: i still not get this idea? why should to put UIAction here?

lib/features/thread/data/extensions/query_email_method_extension.dart
```diff
@@ -0,0 +1,28 @@
...
+
+  void addSortsIfNotNull(Set<Comparator>? sort) {
+    if (sort != null) addSorts(sort);
+  }
+
+  void addFiltersIfNotNull(Filter? filter) {
+    if (filter != null) addFilters(filter);
+  }
+
+  void addCollapseThreadsIfValid(bool? collapseThreads) {
```
hoangdat:
```suggestion
  void addCollapseThreadsIfAvailable(bool? collapseThreads) {
```

lib/features/thread/data/extensions/query_email_method_extension.dart
```diff
@@ -0,0 +1,28 @@
...
+import 'package:jmap_dart_client/jmap/core/sort/comparator.dart';
+import 'package:jmap_dart_client/jmap/core/unsigned_int.dart';
+import 'package:jmap_dart_client/jmap/mail/email/query/query_email_method.dart';
+
+extension QueryEmailMethodExtension on QueryEmailMethod {
+  void addLimitIfNotNull(UnsignedInt? limit) {
+    if (limit != null) addLimit(limit);
+  }
+
+  void addPositionIfValid(int? position) {
```
hoangdat:
```suggestion
  void addPositionIfAvailable(int? position) {
```

hoangdat:
we should find another approach for setting changes sync. should be groom it before implement it

lib/features/manage_account/presentation/services/local_settings_service.dart
hoangdat, line 1: how about to keep it as a singleton. Keep the `localSettings` for the whole app read from.
it should have a `refresh` method to reload `localSettings` if any setting change

lib/features/thread/data/extensions/query_email_method_extension.dart
dab246: fixed

lib/features/thread/data/extensions/query_email_method_extension.dart
dab246: fixed

lib/features/manage_account/presentation/services/local_settings_service.dart
dab246: Removed `_threadDetailUIAction `

lib/features/thread/presentation/thread_controller.dart
dab246: disposed

lib/features/manage_account/presentation/services/local_settings_service.dart
dab246: Updated

tddang-linagora (dismissed):

lib/features/manage_account/presentation/services/local_settings_service.dart
chibenwa: > keep it as a singleton

By no mean I am a flutter expert but singleton have bad press in the IT ecosystem:
 - uses of singleton are hard to audit
 - Testability
 - Asynchronous loading is indeed an issue

IMO that's the kind of easy right-now choices that-backfire-like-crazy in the future that we shall refrain to do.

Can't we rather inject this?

lib/features/search/email/presentation/search_email_controller.dart
```diff
@@ -161,6 +162,13 @@ class SearchEmailController extends BaseController
 
   Set<String> get listHasKeywordFiltered => searchEmailFilter.value.hasKeyword;
 
+  late final LocalSettingsService _localSettingsService = Get.find<LocalSettingsService>();
+
+  bool get _isCollapseThreadsEnabled {
+    if (!Get.isRegistered<LocalSettingsService>()) return false;
+    return _localSettingsService.localSettings.value.threadConfig.isEnabled;
+  }
```
chibenwa: Mmmmmmmmm 

That's the kind of tricky bad code that gets unpredicable in prod... 

Does GetX provides a way to do this in a cleaner fashion?

--------------------------

Claude opinion

<html><head></head><body><p><strong>What's wrong here</strong></p>
<pre><code class="language-dart">// 1. lazy field — crashes if accessed before registration
late final LocalSettingsService _localSettingsService = Get.find&lt;LocalSettingsService&gt;();

// 2. defensive guard — symptom of the real problem
if (!Get.isRegistered&lt;LocalSettingsService&gt;()) return false;
</code></pre>
<p>The <code>isRegistered</code> guard is a <strong>smell</strong> — it means callers can't trust the service is ready, so every caller must defensively check. This spreads the loading concern everywhere.</p>
<hr>
<p><strong>Why it's desirable to fix</strong></p>
<ul>
<li>Callers should <strong>not care</strong> about initialization order</li>
<li>The guard <code>return false</code> silently swallows an uninitialized state — hard to debug</li>
<li><code>late final</code> + <code>Get.find</code> at field init time is a time-bomb if the widget builds before <code>put()</code></li>
</ul>
<hr>
<p><strong>The GetX way — initialize first, then build</strong></p>
<p>Use <code>GetX</code> bindings to guarantee initialization order before any dependent widget builds:</p>
<pre><code class="language-dart">class AppBinding extends Bindings {
  @override
  void dependencies() {
    // Eager, synchronous — registered before any route builds
    Get.put(LocalSettingsService(), permanent: true);
  }
}
</code></pre>
<pre><code class="language-dart">// In your app entrypoint
GetMaterialApp(
  initialBinding: AppBinding(),
  home: HomeScreen(),
);
</code></pre>
<p>Now <code>LocalSettingsService</code> is <strong>guaranteed registered</strong> before any widget tree builds.</p>
<hr>
<p><strong>Clean consumer — no guard needed</strong></p>
<pre><code class="language-dart">class MyWidget extends StatelessWidget {
  // Safe — AppBinding guarantees it's registered
  final LocalSettingsService _localSettingsService = Get.find&lt;LocalSettingsService&gt;();

  bool get _isCollapseThreadsEnabled =&gt;
      _localSettingsService.localSettings.value.threadConfig.isEnabled;
}
</code></pre>
<p>No <code>isRegistered</code>, no <code>late</code>, no <code>false</code> fallback.</p>
<hr>
<p><strong>If initialization is async</strong> (e.g. reading from disk), use <code>asyncPut</code>:</p>
<pre><code class="language-dart">class AppBinding extends Bindings {
  @override
  void dependencies() {
    Get.putAsync&lt;LocalSettingsService&gt;(() async {
      final service = LocalSettingsService();
      await service.init(); // load from SharedPreferences etc.
      return service;
    }, permanent: true);
  }
}
</code></pre>
<p>And show a loading screen until ready:</p>
<pre><code class="language-dart">GetMaterialApp(
  initialBinding: AppBinding(),
  home: FutureBuilder(
    future: Get.putAsync&lt;LocalSettingsService&gt;(...),
    builder: (context, snapshot) =&gt;
        snapshot.hasData ? HomeScreen() : SplashScreen(),
  ),
);
</code></pre>
<hr>
<p><strong>Summary</strong></p>

  | Current code | With Bindings
-- | -- | --
Init order guaranteed | ❌ | ✅
Guards needed | ✅ everywhere | ❌ never
Callers care about loading | ✅ | ❌
Testable | ⚠️ | ⚠️ (still GetX)


<p>The binding centralizes the "when is it ready" concern in <strong>one place</strong>, so every consumer can trust the service exists.</p></body></html>

dab246: This was already inject binding  to before use,  checking `isRegistered ` is only to pass the unit test.

lib/features/manage_account/presentation/services/local_settings_service.dart
dab246: We don't use singletons; we use them through DI.

hoangdat: I mean: it should have app life cycle (app_binding should manage instance of it).

lib/features/search/email/presentation/search_email_controller.dart
chibenwa: You don't get the point of injecting not yet initialized component

Handling that mutability/state in all the code base is a poison.

Do we have alternatives?

hoangdat: IMO, we can inject the service into the controller constructor.

dab246: > You don't get the point of injecting not yet initialized component
> 

It seems you don't understand either. When we inject a binding into a class, we initialize its constructor. Here, GetX uses the Get.put function to push it onto the GetX binding stack. And when using it, you just need to call Get.find.

dab246: Before switching to the new State Management (Riverpod), we should still retain GetX's binding management mechanism.

hoangdat:
Previously, LocalSettingsService relied on manually calling reload() after settings were saved, also need dispatch action, and worker in controller to listen.
 
I propose a `LocalSettingsNotifier` which have a broadcast stream, will stay in domain layer.
```dart
class LocalSettingsNotifier {
  final _controller = StreamController<PreferencesSetting>.broadcast();

  Stream<PreferencesSetting> get stream => _controller.stream;

  void notify(PreferencesSetting settings) => _controller.add(settings);

  void dispose() => _controller.close();
}

 ```
 
 Whenever `GetLocalSettingsInteractor` or `UpdateLocalSettingsInteractor` have the new local setting,  LocalSettingsNotifier will notify, LocalSettingsService will watch the stream in LocalSettingsNotifier and update value for localSettings.

**Comment by dab246** [comment]
> Previously, LocalSettingsService relied on manually calling reload() after settings were saved, also need dispatch action, and worker in controller to listen.
> 
> I propose a `LocalSettingsNotifier` which have a broadcast stream, will stay in domain layer.
> 
> ```dart
> class LocalSettingsNotifier {
>   final _controller = StreamController<PreferencesSetting>.broadcast();
> 
>   Stream<PreferencesSetting> get stream => _controller.stream;
> 
>   void notify(PreferencesSetting settings) => _controller.add(settings);
> 
>   void dispose() => _controller.close();
> }
> ```
> 
> Whenever `GetLocalSettingsInteractor` or `UpdateLocalSettingsInteractor` have the new local setting, LocalSettingsNotifier will notify, LocalSettingsService will watch the stream in LocalSettingsNotifier and update value for localSettings.

Why do we need to create an additional `StreamController` when we can simply use the `Obx` variable in `LocalSettingsService` to listen for changes? Creating an additional `StreamController` means we have to manage it.

**Comment by chibenwa** [comment]
> Whenever GetLocalSettingsInteractor or UpdateLocalSettingsInteractor have the new local setting, LocalSettingsNotifier will notify, LocalSettingsService will watch the stream in LocalSettingsNotifier and update value for localSettings.

Sounds reasonable to me.

Does this need to be a preliminary refactoring?

**Comment by dab246** [comment]
> Previously, LocalSettingsService relied on manually calling reload() after settings were saved, also need dispatch action, and worker in controller to listen.
> 
> I propose a `LocalSettingsNotifier` which have a broadcast stream, will stay in domain layer.
> 
> ```dart
> class LocalSettingsNotifier {
>   final _controller = StreamController<PreferencesSetting>.broadcast();
> 
>   Stream<PreferencesSetting> get stream => _controller.stream;
> 
>   void notify(PreferencesSetting settings) => _controller.add(settings);
> 
>   void dispose() => _controller.close();
> }
> ```
> 
> Whenever `GetLocalSettingsInteractor` or `UpdateLocalSettingsInteractor` have the new local setting, LocalSettingsNotifier will notify, LocalSettingsService will watch the stream in LocalSettingsNotifier and update value for localSettings.
>

The core idea of decoupling the write path from the read path via a notifier is exactly right.

We went with `Riverpod`’s `StateNotifier` over a broadcast stream for a few reasons:

- **No current value**: a broadcast `StreamController` does not hold state — new subscribers miss the last emission.  
  `StateNotifier` always exposes the current value synchronously, which `SearchEmailController` needs to read at query time without subscribing.

- **Domain purity**: a stateful singleton with a `StreamController` to close does not belong in the domain layer — it has no lifecycle owner there.

- **Fewer moving parts**: the proposed flow still needs `LocalSettingsService` to bridge the stream into its own state, then controllers subscribe separately.  
  With `StateNotifier`, controllers read/listen directly from the provider.

This is also a pilot for our `GetX → Riverpod` migration (ADR-0085), so establishing the pattern now avoids replacing a custom stream solution later.

File: docs/adr/0085-riverpod-state-management-for-local-settings.md [hunk 6]
```diff
@@ -0,0 +1,90 @@
...
+- [ADR-0071](./0071-collapse-threads-in-email-query.md) — collapseThreads in Email/query
+
+## Context
+
+As part of implementing the collapsed thread feature in search email (TF-4363), two controllers need to read the `collapseThreads` local setting:
+
+- `ThreadController` — must react when the setting changes to re-trigger search
+- `SearchEmailController` — must read the current value at query time
+
+The setting is written by `PreferencesController` via `UpdateLocalSettingsInteractor` when the user toggles it in `PreferencesView`, and must also be reset on logout to prevent stale cross-account data.
```
hoangdat: we can remove this: `and must also be reset on logout to prevent stale cross-account data.`

tddang-linagora (dismissed):

docs/adr/0085-riverpod-state-management-for-local-settings.md
dab246: removed

chibenwa:
Thanks for carring @dab246, efforts appreciated

lib/main/providers/app_provider_container.dart
```diff
@@ -0,0 +1,5 @@
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+/// Singleton [ProviderContainer] shared between the [ProviderScope] and
+/// non-widget code (e.g. GetX controllers / services).
+final appProviderContainer = ProviderContainer();
```
hoangdat, line 5: should we hand the provider to the widgetTree?

```dart
runApp(                                                                           
    UncontrolledProviderScope(
      container: appProviderContainer,                     
      child: TMailApp(),
    ),                                                                              
  );       
```

dab246: We intentionally skipped `UncontrolledProviderScope` here because no  Flutter widget in the current implementation uses `ref.watch` or `ConsumerWidget`.  All consumers are GetX controllers that access the provider directly via `appProviderContainer`. Adding `UncontrolledProviderScope` would introduce overhead with no benefit at this stage.  We can revisit this when a widget needs to consume the provider directly via `ref`.

hoangdat (approved):

hoangdat:
please rebase

dab246:
> please rebase 

It has no conflict. Please squash and merge.

docs/adr/0085-riverpod-state-management-for-local-settings.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Decision
+
+Introduce `localSettingsNotifierProvider` (`StateNotifier<PreferencesSetting>`) backed by a global `ProviderContainer` (`appProviderContainer`). All reads and writes go through this provider directly.
+
+### Architecture
+
+```text
+┌──────────────────────────────────────────────────────────┐
+│  appProviderContainer  (global ProviderContainer)        │
```
9clg6, line 34: You don't need a appProviderContainer and it's a anti-pattern, it's doesn't work like get it.
In you providers, made with @riverpod has in parameters a ref used to inject instance. Same for viewmodel.
In the view you use the ref.watch and ref.read with the ref provided by ConsumerWidget.

Can't be merge as it is

docs/adr/0085-riverpod-state-management-for-local-settings.md
```diff
@@ -0,0 +1,87 @@
...
+│  LocalSettingsService  PreferencesController  Thread /   │
+│  (app start /          (on toggle)            Search     │
+│   route reload)                               Controllers│
+└──────────────────────────────────────────────────────────┘
+```
+
+### Write paths
+
+**On app start / settings route reload:**
+`LocalSettingsService.onInit()` → `_loadInitialSettings()` → cache read → `localSettingsNotifierProvider.notifier.update()`
```
9clg6, line 49: Notifier.update is not needed if you use ref.watch inside your providers (riverpod pattern)

lib/features/search/email/presentation/search_email_controller.dart
```diff
@@ -161,6 +163,9 @@ class SearchEmailController extends BaseController
 
   Set<String> get listHasKeywordFiltered => searchEmailFilter.value.hasKeyword;
 
+  bool get _isCollapseThreadsEnabled =>
+      appProviderContainer.read(localSettingsNotifierProvider).threadConfig.isEnabled;
```
9clg6, line 167: Don't use global provider + here .watch is needed

chibenwa:
Please @dab246 we need to take the time to open a PR addressing @9clg6 review comments in a follow up PR.
Thanks.

docs/adr/0085-riverpod-state-management-for-local-settings.md
dab246: Thank you for the feedback. The architecture you are describing is correct for a **greenfield pure Riverpod app**. However, this codebase is a **GetX + Riverpod hybrid** — `ThreadController` and `SearchEmailController` are `GetxController` subclasses managed by GetX's DI graph. GetX controllers have no `ref`, no `BuildContext`, and no connection to a `ProviderScope` — the framework simply does not provide it.

**Why `appProviderContainer` is necessary here:**

Riverpod's own documentation acknowledges `ProviderContainer` as the correct tool for accessing providers outside of the Flutter widget tree — in plain Dart classes, services, and any code where `WidgetRef` is unavailable. This is not equivalent to a GetIt service locator: `appProviderContainer` is a real `ProviderContainer` that participates in Riverpod's dependency graph, respects provider lifecycle, and enables `listen`/`read`/`invalidate` semantics.

`ConsumerWidget` + `ref.watch` requires the **entire controller and widget hierarchy** to be migrated to Riverpod first. The codebase has hundreds of GetX controllers — that is the full migration described in ADR-0075 (Riverpod + go_router + get_it, ~8–12 months, 4 phases). Doing it for one feature in isolation would create a split architecture with no clear boundary.

**The trade-off is documented:** ADR-0085 (Consequences section) explicitly acknowledges `appProviderContainer` as a known DIP violation and explains that it will be removed when the migration reaches the repository layer. This PR is a **pilot** to validate the coexistence pattern — not the final architecture.

**On `@riverpod` annotation:** As noted in the reply for the previous PR, `riverpod_annotation` and `riverpod_generator` are not present in the project's dependency tree. The project is on `flutter_riverpod: 2.6.1`. Adopting `@riverpod` code generation is part of the planned Riverpod 3.0 upgrade PR.

docs/adr/0085-riverpod-state-management-for-local-settings.md
dab246: This is a valid pattern in Riverpod — but it applies when state is **derived reactively from another Riverpod provider**. That is not the case here.

`PreferencesSetting` is produced by two **external, non-Riverpod sources**:

1. `LocalSettingsService.onInit()` — a GetX service that reads from a Hive cache at app startup
2. `PreferencesController.handleSuccessViewState(UpdateLocalSettingsSuccess)` — a GetX controller reacting to a use case result

Neither source is a Riverpod provider, so there is no upstream `ref` to watch inside `LocalSettingsNotifier`. Using `ref.watch(someProvider)` inside the notifier would require the Hive datasource, repository, and interactor to all be Riverpod providers first.

That migration is described in ADR-0085 (Consequences → `LocalSettingsService` removal condition): once `ManageAccountRepository` and its datasources are migrated to Riverpod, `LocalSettingsNotifier` can inject and call `GetLocalSettingsInteractor` directly, eliminating `LocalSettingsService` and the `.update()` write path entirely. Until then, `.update()` is the correct bridge.

lib/features/search/email/presentation/search_email_controller.dart
dab246: **On `appProviderContainer`:** Same as Comment 1 — `GetxController` subclasses have no `ref`. `appProviderContainer` is the established bridge.

**On `.read` vs `.watch`:**

```dart
bool get _isCollapseThreadsEnabled =>
    appProviderContainer.read(localSettingsNotifierProvider).threadConfig.isEnabled;
```

This getter is called **at query construction time** — it reads the current snapshot of `PreferencesSetting` at the moment a search query is built (called at lines 380, 455, 525, 586 inside query methods). It is a **point-in-time read**, not a reactive subscription.

`ref.watch` / `container.watch` (in the `WidgetRef` sense) only makes sense inside `ConsumerWidget.build` where it registers a rebuild listener. In a `ProviderContainer` context outside widgets, watching is done via `container.listen(...)` — which already exists in this codebase:

```dart
// ThreadController._registerLocalSettingsListener()
_localSettingsSubscription = appProviderContainer.listen(
  localSettingsNotifierProvider,
  (previous, next) {
    if (previous?.threadConfig == next.threadConfig) return;
    // re-trigger search
  },
  fireImmediately: false,
);
```

The reactive behavior (re-triggering search when the setting changes) is handled by `listen`. The `_isCollapseThreadsEnabled` getter correctly uses `.read` to pick up the latest value when a query runs. These two roles are complementary, not conflicting.

dab246:
> Please @dab246 we need to take the time to open a PR addressing @9clg6 review comments in a follow up PR. Thanks.

I took the time to address @9clg6's comments. And the final review is all good. Thanks

docs/adr/0085-riverpod-state-management-for-local-settings.md
9clg6: The fact that this is a hybrid codebase is precisely why the pilot should demonstrate the target pattern, not reproduce the current one with different tooling. appProviderContainer as a global singleton is functionally identical to a service locator — Riverpod was designed to eliminate that pattern, not to be accessed through it.

"ConsumerWidget + ref.watch requires the entire hierarchy to be migrated first" — this is circular. The whole point of an incremental migration is to convert widgets one at a time. Each ConsumerWidget introduced is a step forward; refusing to introduce any until everything is ready means nothing ever moves. Coexistence is the boundary — GetX widgets keep their controllers, Riverpod widgets get ref. That's not a "split architecture", that's a migration in progress.

On @riverpod codegen: starting the migration on flutter_riverpod: 2.6.1 with StateNotifier (deprecated since Riverpod 2.4) to then plan a separate "Riverpod 3.0 upgrade PR" creates migration debt on top of migration debt. The 3.0 upgrade should land before the pilot so we build on the target API from day one, not retrofit it later.

docs/adr/0085-riverpod-state-management-for-local-settings.md
9clg6: This is exactly the problem. If the pilot's conclusion is "we can't use ref.watch because the sources aren't Riverpod providers yet", then the pilot isn't validating the Riverpod pattern at all — it's validating a StateNotifier fed by imperative .update() calls from GetX, which is just a reactive wrapper around the old architecture.

A meaningful pilot should migrate the source too: wrap the Hive datasource in a Riverpod provider, inject GetLocalSettingsInteractor via ref.watch, and let the LocalSettingsNotifier derive its state reactively. That's the pattern we need to prove works in this codebase. If we defer that to "when ManageAccountRepository is migrated", we've deferred the only part of the pilot that actually tests Riverpod's value proposition.

The .update() bridge is a pragmatic shortcut, but a pilot that only validates shortcuts doesn't tell us anything about the real migration.

lib/features/search/email/presentation/search_email_controller.dart
9clg6: The distinction between .read for snapshot and .listen for reactivity is technically correct in a ProviderContainer context — but it normalizes imperative state access as the standard pattern. In idiomatic Riverpod, you watch and the framework handles rebuild granularity. Reasoning about "when to read the right value" is GetX thinking applied to Riverpod APIs.

"ref.watch / container.watch only makes sense inside ConsumerWidget.build" — this is the argument for keeping the status quo, not for migrating. The goal is to move SearchEmailController logic into widgets or Riverpod notifiers where ref.watch is natural. If we accept that controllers will never have ref, we're not migrating — we're just adding a Riverpod layer on top of GetX without changing the consumption pattern.

The .read + .listen combo works, but it's two manual pieces doing what a single ref.watch does declaratively. The pilot should aim for the simpler target pattern, not optimize the bridge.

9clg6:
@chibenwa @dab246

chibenwa:
Please @dab246 we should invest 2 man-day trying to set up the patterns described by @9clg6 

From an old experience setting up new patterns takes time and is very frustrating at the begining but pays off on the long run. 

Just ask @Arsnael how painful it was to establish scala coding practice at the beginning and how much it delayed JMAP implementation, but how beneficial it was on the long run - we rocketted through the rest of the implem!

hoangdat:
@9clg6 

Thank for your spending time in reviewing the riverpod implementation in Twake Mail. Can you help us summarize what is the thing we should change to have good adaptation. It is perfect if we have a set of bullet points for the whole team can understand clearly.

PR #4380, opened 2026-03-13
ADR-71 CollapseThread

hoangdat:
ADR to implement `collapseThreads` in `Email/Query`

cc @chibenwa @Crash-- @poupotte 

## Summary by CodeRabbit

* **Documentation**
  * Added ADR describing collapseThreads behavior: explains the two query paths (force vs cache-first), the cache inconsistency when toggling grouped/ungrouped views, and the decision to clear the email cache on toggle while leaving force-query unchanged.
  * Added ADR for thread-aware bulk actions: defines a thread-expansion approach, thread-aware bulk read/move flows, per-thread error isolation, deduplication, caching/invalidation rules, and expected results.

### DIFF
--- a/docs/adr/0071-collapse-threads-in-email-query.md
+++ b/docs/adr/0071-collapse-threads-in-email-query.md
@@ -1,0 +1,64 @@
+# 0071 - Enable collapseThreads in Email/query
+
+Date: 2026-03-12
+
+## Status
+
+Proposed
+
+## Context
+
+JMAP `Email/query` supports `collapseThreads` (RFC 8621 §4.4). When `true`, server returns only the **latest email per thread**, avoiding duplicate rows in the mail list.
+
+Currently, our app does **not** set `collapseThreads`. When thread is enabled, users see multiple emails from the same thread as separate rows.
+
+### Two Query Paths (see ADR-0070)
+
+| Path | `FORCE_EMAIL_QUERY` | Method | Cache |
+|------|---------------------|--------|-------|
+| Force query | `true` | `forceQueryAllEmailsForWeb()` | No local cache — server only |
+| Cache-first | `false` | `getAllEmail()` | Local DB first, then sync |
+
+### Problem: Cache Inconsistency on Toggle
+
+On the cache-first path, the local DB reflects the `collapseThreads` mode it was built under. When thread mode changes:
+
+- **OFF → ON**: Cache has individual emails (5 rows) but server now returns collapsed (3 rows) → stale first yield
+- **ON → OFF**: Cache has collapsed emails (3 rows) but server now returns all individuals (5 rows) → missing emails in first yield
+
+Force-query path has **no problem** — always queries server directly.
+
+## Decision
+
+### 1. Set `collapseThreads: true` in Email/query When Thread Is Enabled
+
+In `MailAPIMixin.fetchAllEmail()` and `ThreadAPI.searchEmails()`, pass `collapseThreads: true` when thread is enabled.
+
+### 2. Clear Email Cache on Thread Setting Toggle
+
+Reuse the existing thread-enabled setting. **No new flag needed.** When thread is toggled (enable ↔ disable), clear email cache immediately in the toggle handler. Next `getAllEmail()` will rebuild cache under correct mode.
+
+### 3. No Change for Force-Query Path
+
+Already server-only. Just pass `collapseThreads: true` in the query.
+
+## Consequences
+
+**Positive**: Correct thread display, reduced data per page, spec-compliant.
+
+**Negative**: One-time full reload on each toggle; toggle handler coupled with cache management.
+
+## Implementation Steps
+
+1. Add `collapseThreads` to `QueryEmailMethod` in `MailAPIMixin.fetchAllEmail()` and `ThreadAPI.searchEmails()`
+2. Pass thread-enabled status from `Session` to query builder
+3. Clear email cache in thread setting toggle handler
+4. Update `GetEmailsInMailboxInteractor` and related interactors to propagate thread status
+
+## References
+
+- [RFC 8621 §4.4](https://www.rfc-editor.org/rfc/rfc8621#section-4.4)
+- [ADR-0070](./0070-sync-strategy-disappearing-emails.md)
+- `lib/features/base/mixin/mail_api_mixin.dart` — `fetchAllEmail()`
+- `lib/features/thread/data/repository/thread_repository_impl.dart` — `getAllEmail()`
+- `lib/features/thread/domain/usecases/get_emails_in_mailbox_interactor.dart`
--- a/docs/adr/0071-collapse-threads-in-email-query.md
+++ b/docs/adr/0071-collapse-threads-in-email-query.md
@@ -51,7 +51,7 @@ Already server-only. Just pass `collapseThreads: true` in the query.
 ## Implementation Steps
 
 1. Add `collapseThreads` to `QueryEmailMethod` in `MailAPIMixin.fetchAllEmail()` and `ThreadAPI.searchEmails()`
-2. Pass thread-enabled status from `Session` to query builder
+2. Pass thread-enabled status from Settings to query builder
 3. Clear email cache in thread setting toggle handler
 4. Update `GetEmailsInMailboxInteractor` and related interactors to propagate thread status
 
--- a/docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
+++ b/docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
@@ -56,14 +56,14 @@ When users perform actions on:
 the system must resolve:
 
 ```text
-ThreadId → List<EmailId>
+ThreadId → List<EmailInThreadDetailInfo>
 ```
 
 Key challenges:
 
 * N+1 API calls (`getThreadById`)
 * Partial failures (one thread fails while others succeed)
-* Duplicate `EmailId`s when merging results
+* Duplicate `EmailInThreadDetailInfo`s when merging results
 * High latency with large selections
 * Existing interactors are not designed for thread-level inputs
 
@@ -82,7 +82,7 @@ Thread-aware Interactor (Stream<UIState>)
     ↓
 ThreadExpansionService
     ↓
-List<EmailId> (deduplicated & pagination-safe)
+List<EmailInThreadDetailInfo> (deduplicated & pagination-safe)
     ↓
 Existing Email Interactor (Stream)
     ↓
@@ -93,7 +93,7 @@ UI State Update (Optimistic + Server Sync)
 
 ### Responsibility
 
-* Convert `List<ThreadId>` → `List<EmailId>`
+* Convert `List<ThreadId>` → `List<EmailInThreadDetailInfo>`
 * Reuse `ThreadDetailRepository.getThreadById`
 * Handle:
 
@@ -112,11 +112,11 @@ so we just need to call and use them.
 
 ```dart
 class ThreadExpansionResult {
-  final List<EmailId> emailIds;
+  final List<EmailInThreadDetailInfo> emailThreadInfos;
   final Map<ThreadId, Object> errors;
 
   ThreadExpansionResult({
-    required this.emailIds,
+    required this.emailThreadInfos,
     required this.errors,
   });
 }
@@ -128,7 +128,7 @@ class ThreadExpansionResult {
 class ThreadExpansionService {
   final ThreadDetailRepository threadDetailRepository;
 
-  final Map<ThreadId, List<EmailId>> _cache = {};
+  final Map<ThreadId, List<EmailInThreadDetailInfo>> _cache = {};
 
   ThreadExpansionService(this.threadDetailRepository);
 
@@ -139,7 +139,7 @@ class ThreadExpansionService {
     required MailboxId sentMailboxId,
     required String ownEmailAddress,
   }) async {
-    final emailIds = <EmailId>{};
+    final emailThreadInfos = <EmailInThreadDetailInfo>{};
     final errors = <ThreadId, Object>{};
 
     final futures = threadIds.map((threadId) async {
@@ -158,13 +158,8 @@ class ThreadExpansionService {
           ownEmailAddress,
         );
 
-        final ids = threadDetails
-            .map((e) => e.emailId)
-            .whereType<EmailId>()
-            .toList();
-
-        _cache[threadId] = ids;
-        emailIds.addAll(ids);
+        _cache[threadId] = threadDetails;
+        emailIds.addAll(threadDetails);
       } catch (e) {
         // Error isolation
         errors[threadId] = e;
@@ -174,7 +169,7 @@ class ThreadExpansionService {
     await Future.wait(futures);
 
     return ThreadExpansionResult(
-      emailIds: emailIds.toList(),
+      emailThreadInfos: emailThreadInfos.toList(),
       errors: errors,
     );
   }
@@ -201,7 +196,7 @@ import 'package:jmap_dart_client/jmap/core/id.dart';
 import 'package:jmap_dart_client/jmap/core/error/set_error.dart';
 
 class ThreadActionResult {
-  final List<EmailId> success;
+  final List<EmailInThreadDetailInfo> success;
   final Map<Id, SetError> actionErrors;
   final Map<ThreadId, Object> expansionErrors;
 
@@ -234,7 +229,7 @@ class MarkAsThreadReadInteractor {
     required String ownEmailAddress,
     required ReadActions readAction,
     List<ThreadId> threadIds = const [],
-    List<EmailId> emailIds = const [],
+    List<EmailInThreadDetailInfo> emailThreadInfos = const [],
   }) async* {
     yield Right(LoadingMarkAsThreadRead());
     
@@ -246,12 +241,12 @@ class MarkAsThreadReadInteractor {
       ownEmailAddress: ownEmailAddress,
     );
 
-    final allEmailIds = {
-      ...emailIds,
-      ...expansion.emailIds,
+    final allEmailThreadInfos = {
+      ...emailThreadInfos,
+      ...expansion.emailThreadInfos,
     }.toList();
 
-    if (allEmailIds.isEmpty) {
+    if (allEmailThreadInfos.isEmpty) {
       yield Left(MarkAsThreadReadFailure(ThreadActionResult(
         success: [],
         actionErrors: {},
@@ -263,7 +258,7 @@ class MarkAsThreadReadInteractor {
     final result = await emailRepository.markAsRead(
       session,
       accountId,
-      emailIds,
+      allEmailThreadInfos.emailIds,
       readAction,
     );
 
@@ -293,9 +288,9 @@ class MoveThreadInteractor {
     required AccountId accountId,
     required MailboxId sentMailboxId,
     required String ownEmailAddress,
-    required MoveToMailboxRequest moveRequest,
+    required MailboxId destinationMailboxId,
     List<ThreadId> threadIds = const [],
-    List<EmailId> emailIds = const [],
+    List<EmailInThreadDetailInfo> emailThreadInfos = const [],
   }) async* {
     yield Right(LoadingMoveThread());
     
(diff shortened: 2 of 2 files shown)


chibenwa (commented):
Overall ok to me

docs/adr/0071-collapse-threads-in-email-query.md
```diff
@@ -0,0 +1,64 @@
...
+## Decision
+
+### 1. Set `collapseThreads: true` in Email/query When Thread Is Enabled
+
+In `MailAPIMixin.fetchAllEmail()` and `ThreadAPI.searchEmails()`, pass `collapseThreads: true` when thread is enabled.
+
+### 2. Clear Email Cache on Thread Setting Toggle
+
+Reuse the existing thread-enabled setting. **No new flag needed.** When thread is toggled (enable ↔ disable), clear email cache immediately in the toggle handler. Next `getAllEmail()` will rebuild cache under correct mode.
+
```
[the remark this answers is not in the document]
hoangdat: Thread mode has 2 level of settings: first in Session and if it in Session, it also in Settings to make user can control it himself

hoangdat:
- [x] how to do actions (read/unread, star/unstar, move...) for Thread in EmailList? 
  - fetch all email ids then trigger all?

chibenwa:
> fetch all email ids then trigger all?

Ok with me.

dab246:
- [x] how to do actions (read/unread, star/unstar, move, drag & drop...) for multiple thread selected in EmailList?

dab246:
> * [x]  how to do actions (read/unread, star/unstar, move...) for Thread in EmailList?
> * [x]  how to do actions (read/unread, star/unstar, move, drag & drop...) for multiple thread selected in EmailList?
>

Added `ADR-0072` is the solution to resolve this issue.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,372 @@
...
+Date: 2026-03-17
+
+## Status
+
+Proposed
+
+## Context
+
+After enabling `collapseThreads = true` (ADR-0071), each item in the `EmailList` represents a **ThreadId** instead of an individual `EmailId`.
+
```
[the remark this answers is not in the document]
dab246: Fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,386 @@
...
+      accountId: accountId,
+      emailIds: allEmailIds,
+      destinationMailboxId: destinationMailboxId,
+    );
+
+    return ThreadActionResult(
+      success: result.emailIdsSuccess,
+      actionErrors: result.mapErrors,
+      expansionErrors: expansion.errors,
+    );
```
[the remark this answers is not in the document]
dab246: Fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,386 @@
...
+      session: session,
+      accountId: accountId,
+      emailIds: allEmailIds,
+    );
+
+    return ThreadActionResult(
+      success: result.emailIdsSuccess,
+      actionErrors: result.mapErrors,
+      expansionErrors: expansion.errors,
+    );
```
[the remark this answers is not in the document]
dab246: Fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,386 @@
...
+  final Map<Id, SetError> actionErrors;
+  final Map<ThreadId, Object> expansionErrors;
+
+  ThreadActionResult({
+    required this.success,
+    required this.actionErrors,
+    required this.expansionErrors,
+  });
+}
+```
```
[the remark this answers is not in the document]
dab246: Fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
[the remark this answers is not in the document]
dab246: fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+    );
+
+    return ThreadActionResult(
+      success: result.emailIdsSuccess,
+      actionErrors: result.mapErrors,
+      expansionErrors: expansion.errors,
+    );
+  }
+}
+```
```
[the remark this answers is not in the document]
dab246: No, we don't want to separate the interactors and reuse the repositories to avoid affecting the old logic.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+    final result = await emailRepository.moveToMailbox(session, accountId, moveRequest);
+
+    return ThreadActionResult(
+      success: result.emailIdsSuccess,
+      actionErrors: result.mapErrors,
+      expansionErrors: expansion.errors,
+    );
+  }
+}
+```
```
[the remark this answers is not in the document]
dab246: No, we don't want to separate the interactors and reuse the repositories to avoid affecting the old logic.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+
+### Implementation Points
+
+Hook `expansionService.clearCache()` in:
+
+- `MailboxRepository` after successful sync completion
+- Email interactor wrappers after successful state mutations (read/unread, star, move, delete)
+- Settings controller when `collapseThreads` toggle changes
+- Thread detail update callbacks when thread membership changes
+
```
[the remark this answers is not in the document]
dab246: Updated

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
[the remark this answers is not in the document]
dab246: fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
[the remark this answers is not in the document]
dab246: fixed

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+}
+```
+
+### Implementation
+
+```dart
+class ThreadExpansionService {
+  final ThreadDetailRepository threadDetailRepository;
+
+  final Map<ThreadId, List<EmailId>> _cache = {};
```
hoangdat: why we should to use cache here?

dab246: Reduce the number of times we need to call getThread from the server when users repeatedly perform actions on the thread.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+class MarkAsThreadReadInteractor {
+  final ThreadExpansionService expansionService;
+  final EmailRepository emailRepository;
+
+  MarkAsThreadReadInteractor(
+    this.expansionService,
+    this.emailRepository,
+  );
+
+  Future<ThreadActionResult> execute({
```
hoangdat: why not `Either<Succses, Failure>`?

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+```
+
+* Minimizes latency for multi-thread operations
+
+## 5. Cache Invalidation
+
+Cache must be cleared when:
+
+* Mailbox sync occurs
+* Email state changes (read, star, move, delete)
```
hoangdat: how about web socket?

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+## Architecture Overview
+
+```text
+Thread Interactor
+    ↓
+ThreadExpansionService
+    ↓
+List<EmailId>
+    ↓
+Existing Email Interactor
```
hoangdat: should address and ensure the pagination for `List<EmailId>` ations are well addressed before.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,401 @@
...
+### Implementation Points
+
+Hook `expansionService.clearCache()` in:
+
+- `ThreadController.refreshAllEmail()` after mailbox sync completes (see lib/features/thread/presentation/thread_controller.dart)
+- `MarkAsThreadReadInteractor`, `MoveThreadInteractor`, etc. after calling `clearCache()` post-action
+- Settings controller's `onCollapseThreadsToggled()` callback (or equivalent setter)
+- `ThreadDetailRepository` observers when `getThreadById` detects membership changes
+
+## Consequences
```
hoangdat, line 604: how about the time from user click in mark as star in UI, until he see it in updated? 
how about the UI base on the update of this single email?

dab246: We will directly update the email list in memory after the user triggers the action, and then update it again based on the results returned by the websocket.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: We've already implemented the limitation of the Set/Get method within the markAsRead/Star/Move functions, so we just need to retrieve the list of email IDs and pass it down to those functions to handle the rest.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: Since websockets always call refreshChange, we only need to clearCache in that function.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: Updated

hoangdat:
it is ok to me

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,637 @@
...
+}
+```
+
+### Implementation
+
+```dart
+class ThreadExpansionService {
+  final ThreadDetailRepository threadDetailRepository;
+
+  final Map<ThreadId, List<EmailId>> _cache = {};
```
Crash--: no "max size" for this cache? No TTL ? No eviction strategy?

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,637 @@
...
+
+    final futures = threadIds.map((threadId) async {
+      try {
+        // Cache hit
+        if (_cache.containsKey(threadId)) {
+          emailIds.addAll(_cache[threadId]!);
+          return;
+        }
+
+        final threadDetails = await threadDetailRepository.getThreadById(
```
Crash--, line 153: Since we only need to have `ids` at the end, do we really need to use `getThreadById`? It seems overkill (lot of mapping and maybe other stuff). Can't we have a simple jmap thread/get id request (https://datatracker.ietf.org/doc/html/rfc8621#section-3.1) ? Or is it the same? (naive question)

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,637 @@
...
+```
+
+* Minimizes latency for multi-thread operations
+
+## 6. Cache Invalidation
+
+Cache must be cleared when:
+
+* Mailbox sync occurs
+* Email state changes (read, star, move, delete)
```
Crash--: If the cache is `Map<ThreadId, List<EmailId>>`, why do we need to invalide the cache when we have a read or star change? For move & delete I understand.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,637 @@
...
+* Thread content changes
```
Crash--, line 587: what is "thread content"?`<List<MailId>>`?

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
```diff
@@ -0,0 +1,637 @@
...
+```text
+Thread Action Flow:
+
+ThreadIds
+  → expandThreads (parallel + cache + error-isolated)
+  → extract EmailIds
+  → deduplicate
+  → call existing Email Interactor
+  → return (success + actionErrors + expansionErrors)
+```
```
Crash--, line 633: `ThreadId → List<EmailId>`, In the list, can we have a mailId that is not in the same mailbox than the other ones? 

If yes, then for mark as read it can be ok to apply the action to all the thread, but for the "move" action for exemple, I don't know. Should ThreadExpansionService.expandThreads() accept an optional mailbox filter ?

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
hoangdat: No. it will be update whenever `email changes` happen, user switchs `folder`.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: `getThreadById` is the function used to call the `Thread/get` method. We need to retrieve the list of email IDs in each thread to handle email actions.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: This descriptive name might not be accurate (I'll delete it), but the point is that we should clear the cache every time we open the mailbox and refresh for changes.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: These are actions related to threads, such as when a thread changes, or when an email is added or removed from the thread.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: You're right, there might be emails that aren't in the same mailbox, so we need to handle retrieving the mailbox ID of each email via `Email/get`. We'll add this processing step as well.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: `getThreadById` is a function used to retrieve a list of email IDs (from Thread/get) along with some metadata (such as keywords, mailboxId) (from Email/get). Since emails in a thread can be in different mailboxes, we need these properties to support processing those actions.

docs/adr/0072-thread-aware-bulk-actions-email-list-collapse-threads.md
dab246: What you said is correct; the emails in the thread can be in different mailboxes, but we've already handled retrieving some email metadata (like mailbox ID) when calling `getThreadById`. Therefore, performing the previous actions becomes easier.

Crash--: Great to now have access to the mailboxid. 

So if I’ve in my thread id 1: 
- mail 1 -> mailbox 1
- Mail 2 -> mailbox 2 

do we give to the user the ability to move the thread? If yes what will be the behavior in this case? We’ll move mail 1 and 2 to the new selected mailbox id?

dab246: Yes, we allow it to be done because we already have enough email data (Id, mailboxId) after performing `getThreadById`. We will convert `List<EmailInThreadDetailsInfo>` to `Map<MailboxId, List<EmailId>>` , then simply pass it to the `moveToMailbox` function in EmailAPI to handle it automatically. Basically, the `moveToMailbox` function creates PatchObjects:

```
{
  emailId1: {
    mailboxIds/mailboxId1: null,
    mailboxIds/newMailboxId: true,
  },
  emailId2: {
    mailboxIds/mailboxId2: null,
    mailboxIds/newMailboxId: true,
  }
}
```

Crash--: So for a user who chooses the thread view, we display all the emails in the thread, even if they’re not in the same mailbox. Then, on that thread page, we allow the user to move the entire thread to another mailbox.

So that would potentially change the mailbox of an email that the user had specifically placed in a particular mailbox?

A more concrete example:

I’m in a conversation with a client, so my thread is growing. For one of the emails, I move it into a specific mailbox called “client files,” because it contains an important attachment, then the thread continues.

A few days later, I come back, select the thread, and move it to “finalized client.”

Does that mean I won’t find the email with the attachment anymore in my “client files” mailbox? Or do we keep both mailboxes, but then it’s not really a move anymore?

Not sure if my question is clear.

dab246: If we follow the current logic, when you move your current thread to the `finalized client` folder, all emails from that thread (regardless of which folder they belong to) will also be moved to the `finalized client` folder.

dab246: I think my code caused you to misunderstand, so I've updated it :))

```
{
  emailId1: {
    mailboxIds/mailboxId1: null,
    mailboxIds/newMailboxId: true,
  },
  emailId2: {
    mailboxIds/mailboxId2: null,
    mailboxIds/newMailboxId: true,
  }
}
```

Crash--: OK thanks for the clarification @dab246 . I'm still not convince from a user point of view, but in the meantime I don't know how, as a user, I'll be able to have mails from different mailboxs in the same thread anyway. 

As long as we do not display emails from the Trash in threads, that works for me. (and for Draft, I don't know ^^')

dab246: What you said makes sense. Perhaps we need to adjust some of the old logic and compare it with other mail client approaches to create a reasonable UX here.

chibenwa: >  (and for Draft, I don't know ^^')

That's interesting!

Do we have detailed spec for product team on this (handling of system mailbox in threads?)

Cc @poupotte

dab246:
This ADR will be implemented through the following tickets:

- Ticket 1 — Enable collapseThreads on Force Query path: https://github.com/linagora/tmail-flutter/issues/4400

- Ticket 2 — Enable collapseThreads on Cache-first path + Cache Invalidation: #4401 

- Ticket 3 — Thread-aware Actions + Optimistic UI: #4402 

- Ticket 4 — Filter out emails in Trash/Drafts/Templates from Thread view: #4403

Issue #4388, opened 2026-03-18
Support configurable OIDC authority separate from JMAP server URL


nka11:
## Context

When deploying Twake Mail against a **third-party JMAP server** (e.g. Stalwart) that delegates authentication to an **external OIDC identity provider** (e.g. Kanidm, Keycloak, Authentik), the OIDC issuer lives on a completely different origin than the JMAP server.

Example topology:

| Component | URL |
|---|---|
| JMAP server (Stalwart) | `https://mail.example.org` |
| OIDC Identity Provider (Kanidm) | `https://id.example.org` |
| Twake Mail web | `https://webmail.example.org` |

The JMAP server validates Bearer tokens issued by the IdP (via its userinfo endpoint), but does **not** serve `/.well-known/openid-configuration` itself — that endpoint lives on the IdP.

## Current behavior

Twake Mail assumes the OIDC authority can be discovered from the JMAP server URL:

1. `LoginController._checkOIDCIsAvailable()` sends a WebFinger / OIDC discovery request to the `SERVER_URL` origin.
2. Since the JMAP server does not expose `/.well-known/openid-configuration`, the request returns 404 or times out.
3. The app falls back to basic auth, which may not be available (the server only accepts Bearer tokens via OIDC).

This means there is currently **no way to use Twake Mail with a JMAP server whose OIDC provider is on a different domain**, unless the JMAP server proxies the OIDC discovery endpoint (which is unusual and creates coupling).

## Expected behavior

Allow operators to configure the OIDC authority independently from the JMAP server URL via `env.file`:

```env
SERVER_URL=https://mail.example.org
AUTHORITY_OIDC=https://id.example.org/oauth2/openid/tmail-web
WEB_OIDC_CLIENT_ID=tmail-web
OIDC_SCOPES=openid,profile,email
DOMAIN_REDIRECT_URL=https://webmail.example.org
```

When `AUTHORITY_OIDC` is set:

1. **Skip WebFinger/OIDC discovery** against `SERVER_URL` entirely.
2. Build the `OIDCConfiguration` directly from the configured authority, client ID, and scopes.
3. Proceed with the standard Authorization Code + PKCE flow against the configured IdP.
4. Present the resulting access token as a Bearer token to the JMAP server.

When `AUTHORITY_OIDC` is **not** set, the current WebFinger discovery behavior is preserved (no breaking change).

## Authenticated flow (with `AUTHORITY_OIDC`)

```
[...]

PR #4389, opened 2026-03-18
Federate to choosen OIDC issuer

nka11:
proposal for feature #4388 

## Release Notes

* **New Features**
  * Added OIDC identity provider URL configuration support, allowing users to specify custom Single Sign-On (SSO) endpoints.
  * Introduced manual OIDC authority input form on mobile with option to skip and use password authentication instead.

* **Improvements**
  * Fixed environment file caching to ensure fresh fetches from the server.
  * Enhanced device orientation handling during login for improved user experience.

### DIFF
--- a/lib/features/login/presentation/login_controller.dart
+++ b/lib/features/login/presentation/login_controller.dart
@@ -374,15 +374,6 @@ class LoginController extends ReloadableController {
     _checkOIDCIsAvailable();
   }
 
-  bool _canAutoStartOidc() {
-    return AppConfig.authorityOidc.isNotEmpty && AppConfig.baseUrl.isNotEmpty;
-  }
-
-  void _autoStartOidc() {
-    _baseUri = Uri.tryParse(AppConfig.baseUrl);
-    _checkOIDCIsAvailable();
-  }
-
   void _checkOIDCIsAvailable() {
     if (_currentBaseUrl == null) {
       dispatchState(Left(CheckOIDCIsAvailableFailure(CanNotFoundBaseUrl())));


core/lib/presentation/extensions/uri_extension.dart
```diff
@@ -6,6 +6,19 @@ extension URIExtension on Uri {
   Uri toQualifiedUrl({required Uri baseUrl}) {
     log('SessionUtils::toQualifiedUrl():baseUrl: $baseUrl | sourceUrl: $this');
     if (hasOrigin) {
+      // When behind a reverse proxy, the JMAP server may return URLs with
+      // the internal scheme (e.g. http://host:8080) while the client
+      // connects via https://host. Rewrite the origin to match baseUrl when
+      // the hostname is the same but the scheme differs (http vs https).
```
chibenwa, line 12: We had explicit code changes in ASF james to allow solve this. 

I would encourage you to bring this to the stalwart community, the jmap client is not a good place to fix jmap server issue.

lib/features/home/presentation/extensions/handle_web_finger_to_get_token_extension.dart
```diff
@@ -20,11 +21,23 @@ extension HandleWebFingerToGetTokenExtension on HomeController {
...
     if (baseUri == null) {
       goToLogin();
-    } else {
-      consumeState(
-        checkOIDCIsAvailableInteractor.execute(OIDCRequest.fromUri(baseUri)),
-      );
+      return;
     }
+
+    // When AUTHORITY_OIDC is explicitly configured, bypass WebFinger
```
chibenwa, line 27: Configured at build time through the build environment?

Personally I by it but as far as I am aware of we shall support . well-known/openid-configuration which shall be enough for stalwart.

What's the real need here?

lib/features/login/presentation/login_view.dart
```diff
@@ -217,6 +219,73 @@ class LoginView extends BaseLoginView {
     return !(controller.responsiveUtils.isMobile(context) && controller.responsiveUtils.isPortrait(context));
   }
 
+  Widget _buildOidcAuthorityInput(BuildContext context) {
```
chibenwa, line 225: We are wrapping the oidc form in flutter, right? But does this handle advanced verification like biometric or card?

Standard practice as far as I am aware of is to use a web page...

chibenwa:
Wonderful work!

Please do split this uin smaller chucks for it to be digestible by the team: 1 PR per topic would be ideal...

Issue #4383, opened 2026-03-16
Handle read only labels


chibenwa:
CF https://github.com/linagora/tmail-backend/issues/2160

When `readOnly` then we shall prevent editing / deleting a label

chibenwa:
If, and onlu if session

```
"com:linagora:params:jmap:labels": {
  "version": 2
}
```

If version field is absent or < 1 we need to not ask for readOnly field

PR #4412, opened 2026-03-25
TF-4383 Handle read only labels

dab246:
## Issue

#4383 

## Resolved

- With `readOnly=false`:

https://github.com/user-attachments/assets/76afd112-d0fa-4a20-a2e5-1cc73d1851e2

- With `readOnly=true`:

https://github.com/user-attachments/assets/c930e00f-4805-4396-b833-954ae86b59ba

## Summary by CodeRabbit

* **New Features**
  * Added label read-only state to the model and exposed a read-only indicator.
  * UI respects read-only labels: context menu and long-press can be disabled and prompting shown based on capability.
  * Session layer can read a labels capability to determine whether to ask about read-only behavior.

* **Chores**
  * Re-exported labels capability types for consumer access.

### DIFF
--- a/labels/lib/extensions/label_extension.dart
+++ b/labels/lib/extensions/label_extension.dart
@@ -29,12 +29,15 @@ extension LabelExtension on Label {
     }
   }
 
+  bool get isReadOnly => readOnly == true;
+
   Label copyWith({
     Id? id,
     KeyWordIdentifier? keyword,
     String? displayName,
     HexColor? color,
     String? description,
+    bool? readOnly,
   }) {
     return Label(
       id: id ?? this.id,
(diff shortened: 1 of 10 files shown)


hoangdat:
- [ ] how about adding toast when user try to long press

dab246:
> * [ ]  how about adding toast when user try to long press

What is the purpose? No long press action is performed when readyOnly = true.

hoangdat:
> What is the purpose? No long press action is performed when readyOnly = true.

To notice user that label cannot change

dab246:
> > What is the purpose? No long press action is performed when readyOnly = true.
> 
> To notice user that label cannot change

We haven't performed any action that would require a user notification.

hoangdat (approved):

PR #4431, opened 2026-04-01
TW-4308 Add labels for several messages part 2

tddang-linagora:
## Issue
- #4308 

## Demo
https://github.com/user-attachments/assets/074ae176-15aa-432b-9e75-c82a53cd465b

## Summary by CodeRabbit

* **New Features**
  * Label selection modal to apply labels to multiple emails.
  * "Label as" action added to selection toolbars in mailbox, search, and thread views.
  * Option to show a selectable checkbox icon on label list items and customizable item padding.
  * Close button alignment configurable (start or end).

* **Improvements**
  * Support applying multiple label keywords to many emails at once.
  * Improved multi-email label sync with callback propagation and UI feedback.

### DIFF
--- a/lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
+++ b/lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
@@ -25,7 +25,7 @@ class AddListLabelsToListEmailsSuccess extends UIState {
   });
 
   @override
-  List<Object?> get props => [emailIds, labelKeywords, labelDisplays, onSync];
+  List<Object?> get props => [emailIds, labelKeywords, labelDisplays];
 }
 
 class AddListLabelsToListEmailsHasSomeFailure
--- a/lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
+++ b/lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
@@ -3,21 +3,29 @@ import 'package:core/presentation/state/success.dart';
 import 'package:jmap_dart_client/jmap/mail/email/email.dart';
 import 'package:jmap_dart_client/jmap/mail/email/keyword_identifier.dart';
 
+typedef OnSyncListLabelForListEmail = void Function(
+  List<EmailId> emailIds,
+  List<KeyWordIdentifier> labelKeywords,
+  {bool shouldRemove}
+);
+
 class AddingListLabelsToListEmails extends LoadingState {}
 
 class AddListLabelsToListEmailsSuccess extends UIState {
   final List<EmailId> emailIds;
   final List<KeyWordIdentifier> labelKeywords;
   final List<String> labelDisplays;
+  final OnSyncListLabelForListEmail? onSync;
 
   AddListLabelsToListEmailsSuccess(
     this.emailIds,
     this.labelKeywords,
-    this.labelDisplays,
-  );
+    this.labelDisplays, {
+    this.onSync,
+  });
 
   @override
-  List<Object> get props => [emailIds, labelKeywords, labelDisplays];
+  List<Object?> get props => [emailIds, labelKeywords, labelDisplays, onSync];
 }
 
 class AddListLabelsToListEmailsHasSomeFailure
@@ -25,11 +33,12 @@ class AddListLabelsToListEmailsHasSomeFailure
   AddListLabelsToListEmailsHasSomeFailure(
     super.emailIds,
     super.labelKeywords,
-    super.labelDisplays,
-  );
+    super.labelDisplays, {
+    super.onSync,
+  });
 
   @override
-  List<Object> get props => [...super.props, 'hasSomeFailure'];
+  List<Object?> get props => [...super.props, 'hasSomeFailure'];
 }
 
 class AddListLabelsToListEmailsFailure extends FeatureFailure {
--- a/lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
+++ b/lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
@@ -1,0 +1,92 @@
+import 'package:core/presentation/resources/image_paths.dart';
+import 'package:core/presentation/state/failure.dart';
+import 'package:core/presentation/state/success.dart';
+import 'package:flutter/material.dart';
+import 'package:model/email/presentation_email.dart';
+import 'package:model/extensions/list_presentation_email_extension.dart';
+import 'package:tmail_ui_user/features/base/base_controller.dart';
+import 'package:tmail_ui_user/features/email/domain/state/labels/add_list_label_to_list_email_state.dart';
+import 'package:tmail_ui_user/features/email/domain/usecases/labels/add_list_label_to_list_emails_interactor.dart';
+import 'package:tmail_ui_user/features/home/data/exceptions/session_exceptions.dart';
+import 'package:tmail_ui_user/features/labels/domain/exceptions/label_exceptions.dart';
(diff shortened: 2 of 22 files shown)


lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
```diff
@@ -0,0 +1,126 @@
...
+    if (labelKeywords.isEmpty) {
+      emitFailure(
+        controller: this,
+        failure: AddListLabelsToListEmailsFailure(
+          exception: const LabelKeywordIsNull(),
+          labelDisplays: labelDisplays,
+        ),
+      );
+      return;
+    }
```
dab246: Avoid boilerplate by using a validate method:

```dart
Exception? _validateParams(List<KeyWordIdentifier> keywords) {
  if (_labelController.session == null) return NotFoundSessionException();
  if (_labelController.accountId == null) return NotFoundAccountIdException();
  if (keywords.isEmpty) return const LabelKeywordIsNull();
  return null;
}
```

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  @override
+  void handleSuccessViewState(Success success) {
+    if (success is AddListLabelsToListEmailsSuccess) {
+      toastManager.showMessageSuccess(success);
+      _pendingOnSync?.call(success.emailIds, success.labelKeywords, shouldRemove: false);
+    } else if (success is AddListLabelsToListEmailsHasSomeFailure) {
+      toastManager.showMessageSuccess(success);
+      _pendingOnSync?.call(success.emailIds, success.labelKeywords, shouldRemove: false);
+    }
```
dab246, line 84: Avoid boilerplate:

```dart
  final (emailIds, labelKeywords) = switch (success) {
    AddListLabelsToListEmailsSuccess s => (s.emailIds, s.labelKeywords),
    AddListLabelsToListEmailsHasSomeFailure s => (s.emailIds, s.labelKeywords),
    _ => return,
  };
  toastManager.showMessageSuccess(success);
  _pendingOnSync?.call(emailIds, labelKeywords, shouldRemove: false);
```

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
```diff
@@ -0,0 +1,126 @@
...
+        failure: AddListLabelsToListEmailsFailure(
+          exception: const LabelKeywordIsNull(),
+          labelDisplays: labelDisplays,
+        ),
+      );
+      return;
+    }
+
+    // Store callback before consumeState (stream-based async)
+    _pendingOnSync = onSync;
```
dab246: Be careful with race condition: 

`_pendingOnSync` is a single field. If the user opens the modal and calls `_addLabels` twice in a row (double-tap, or two different email lists), the second call will overwrite the `_pendingOnSync` of the first call. When the stream from the first call resolves, it will call the callback from the second call, resulting in incorrect data.

dab246: Beware of memory leaks.

`_pendingOnSync` stores a closure that can capture `BuildContext`, widget state, or large objects from the caller. If the `consumeState` stream doesn't emit (timeout, uncaught error, dispose early), this closure will be held indefinitely until `onClose()` is called. If the controller is reused (GetX singleton scope), `onClose()` may never be called at the right time.

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
tddang-linagora: Done

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
tddang-linagora: `else if` clause deleted.

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
tddang-linagora: `_pendingOnSync` is deleted.

tddang-linagora: `_pendingOnSync` is deleted.

lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
```diff
@@ -3,33 +3,42 @@ import 'package:core/presentation/state/success.dart';
...
     this.labelKeywords,
-    this.labelDisplays,
-  );
+    this.labelDisplays, {
+    this.onSync,
+  });
 
   @override
-  List<Object> get props => [emailIds, labelKeywords, labelDisplays];
+  List<Object?> get props => [emailIds, labelKeywords, labelDisplays, onSync];
```
hoangdat: no need to put it on `props`.

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -1272,63 +1274,77 @@ class ThreadController extends BaseController with EmailActionController {
...
-        break;
-      default:
-        break;
+        moveEmailsToArchive(emails);
+      },
+      EmailActionType.compose: (_) =>
+          mailboxDashBoardController.openComposer(ComposerArguments()),
+      EmailActionType.labelAs: (emails) {
+        cancelSelectEmail();
+        openChooseLabelModal(emails);
```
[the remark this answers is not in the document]
hoangdat: please fix this to prevent user need to select emails again

lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart
```diff
@@ -0,0 +1,92 @@
...
+  Exception? _validateParams(List<KeyWordIdentifier> keywords) {
+    if (_labelController.session == null) return NotFoundSessionException();
+    if (_labelController.accountId == null) return NotFoundAccountIdException();
+    if (keywords.isEmpty) return const LabelKeywordIsNull();
+    return null;
+  }
+
+  @override
+  void handleSuccessViewState(Success success) {
+    if (success is AddListLabelsToListEmailsSuccess) {
```
hoangdat, line 81: it also match `AddListLabelsToListEmailsHasSomeFailure`, is it ok?

dab246: `AddListLabelsToListEmailsHasSomeFailure` extends `AddListLabelsToListEmailsSuccess`, so is `AddListLabelsToListEmailsSuccess` already covers both  cases. The handling logic is identical for both, show a toast and call `onSync` with the successfully labeled email IDs. The toast message difference is handled internally by `toastManager.showMessageSuccess()`, which already checks the concrete type. Keeping a single check is intentional and avoids duplicating identical code.

lib/features/email/domain/state/labels/add_list_label_to_list_email_state.dart
dab246: removed

dab246:
> 🧹 Nitpick comments (1)
> > lib/features/labels/presentation/delegates/add_list_label_to_list_emails_delegate.dart (1)> `69-69`: **Remove/update stale inline comment.**
> > > The comment says callback is stored before `consumeState`, but no callback field is stored anymore in this implementation.

Removed

hoangdat (approved):

PR #4453, opened 2026-04-15
Refactoring plan

hoangdat:
2 Options:
- EventBus with ***EmailAction + EmailActionQueue
- Riverpod + Service

### DIFF
--- a/docs/adr/0076-refactor-via-eventbus-state-propagation.md
+++ b/docs/adr/0076-refactor-via-eventbus-state-propagation.md
@@ -1,0 +1,108 @@
+# 0076 - Refactor via AppEventBus + State Providers (OCP-Compliant State Propagation)
+
+## Status
+Proposed
+
+## Context
+
+Controllers in TMail historically use the GetX `consumeState(stream, onSuccess: ...)` pattern plus the `EmailActionController` mixin. `consumeState` did three things at once:
+
+1. Execute a stream and dispatch its result into `viewState`.
+2. Route the success type via `handleSuccessViewState` — a growing `if/else` chain.
+3. Run a controller-local reaction callback (toast, nav).
+
+Consequences:
+
+- `MailboxDashBoardController` and `ThreadController` ballooned past 2K LOC.
+- Every new feature (archive, star, move, label, ...) required editing `handleSuccessViewState` in multiple controllers — **OCP violation**.
+- Duplicated flag-update logic across controllers.
+- Cross-screen state (email list, mailbox map) had no single owner — updates were scattered.
+
+We want to add new email actions **without editing existing controllers** while keeping GetX as the DI / reactive container.
+
+## Decision
+
+Keep GetX as the DI container and introduce three decoupled building blocks on top of it:
+
+### Building Blocks
+
+| Component | Type | Lifecycle | Purpose |
+|---|---|---|---|
+| `EmailAction` (abstract) | Plain class | Per call | Wraps interactor call + pre/post processing, returns `Stream<Either<Failure, Success>>` |
+| `EmailActionQueue` | `GetxService` | App | Central submit point; fires results on bus; manages `CancellationToken` |
+| `AppEventBus` | `GetxService` | App | Single pub/sub bus for all domain success/failure events |
+| `*StateProvider` (Session / Mailbox / EmailList) | `GetxService` | App | Pure `Rx` state stores mutated only by handlers |
+| `*BusHandler` | `GetxService` | App | Subscribes to specific bus events, calls provider mutations |
+| `BusHandlerRegistry` | Const map | Compile-time | Central list of handler factories — registering is one entry, cannot be forgotten |
+
+### Flow
+
+```
+Controller._action()
+  → EmailActionQueue.submit(XEmailAction(...), token: _cancelToken, onSuccess: toastOnly)
+    → action.execute(session, accountId) → Stream
+      → eventBus.fire(XSuccess | XFailure)
+        ├── XBusHandler.on<XSuccess>() → EmailListStateProvider.updateFlag(...)
+        │     → Obx(...) rebuilds automatically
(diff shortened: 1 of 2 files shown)


tddang-linagora (approved):
Both options are solid. Personally, I would want to replace GetX with Riverpod.

docs/adr/0076-refactor-via-eventbus-state-propagation.md
```diff
@@ -0,0 +1,108 @@
...
+
+### OCP Boundary
+
+Adding a new email action (e.g. Archive):
+
+| Task | Edit | Create |
+|---|---|---|
+| `ArchiveEmailAction extends EmailAction` | — | new file |
+| `ArchiveBusHandler extends GetxService` | — | new file |
+| Register in `BusHandlerRegistry` | 1-line entry | — |
```
chibenwa, line 63: This can be avoided with dependency injection as well I presume

docs/adr/0076-refactor-via-eventbus-state-propagation.md
```diff
@@ -0,0 +1,108 @@
...
+- **Channel 2 — UI reactions (`onSuccess` callback):** controller-local toasts, snackbars, navigation. **No state mutations allowed** (enforced by `_show*Toast` naming convention).
+
+### OCP Boundary
+
+Adding a new email action (e.g. Archive):
+
+| Task | Edit | Create |
+|---|---|---|
+| `ArchiveEmailAction extends EmailAction` | — | new file |
+| `ArchiveBusHandler extends GetxService` | — | new file |
```
chibenwa, line 62: Have a more specialized interface that GetXService

That interface actually can extend it.

More specialized interfaces préserve intent, helps di, etc...

docs/adr/0076-refactor-via-eventbus-state-propagation.md
```diff
@@ -0,0 +1,108 @@
...
+- Phase 2a — `BusHandlerRegistry` prevents silent registration failure — **Done**
+- Phase 2b — Strip state mutations from `onSuccess` (double-update fix) — **Done**
+- Phase 3 — Remove email-action branches from `handleSuccessViewState` — Todo
+- Phase 4 — Optional `ActionLoadingState`/`ActionDoneState` bus events — Optional
+
+## Consequences
+
+### Positive
+
+- **OCP-compliant** for adding new email actions — zero edits to existing files beyond a 1-line registry entry.
```
chibenwa, line 83: With Dependency Injection  we should even be able to sneak this 1 line centralized registry entry and can turn it into a 1 line decentralized registry......

docs/adr/0076-refactor-via-eventbus-state-propagation.md
```diff
@@ -0,0 +1,108 @@
...
+- BusHandler subscriptions must be disposed properly (mitigated by `GetxService` lifecycle).
+- Still depends on GetX reactive (`.obs`/`Obx`) — does not solve concerns around GetX's long-term maintenance or global singletons.
+- `viewState` as a field lingers for controller-lifecycle loading; mental model is split across two channels.
+- Debugging pub/sub events is harder than direct calls — needs disciplined logging.
+- Test setup still requires `Get.put(...)` of providers + handlers.
+
+## Alternatives Considered
+
+- **Keep `consumeState` + mixin** — fails OCP, controllers keep growing.
+- **Migrate to Riverpod + service architecture** (see ADR 0077) — cleaner long-term boundaries but high effort and framework risk.
```
chibenwa, line 104: Stupid question: does this move makes moving in the future to riverpod easier? Or harder?

docs/adr/0077-refactor-via-riverpod-service-architecture.md
```diff
@@ -0,0 +1,154 @@
...
+- Shared state (email list, mailbox map, session)
+- Controller-lifecycle state (loading, paging, errors)
+- UI reactions (toast, navigation, dialogs)
+- Cross-controller coordination via `Get.find<OtherController>()`
+
+Pain points:
+
+- God-controllers are hard to test — setup requires spinning up DI graphs, bindings, and mocking `Get.context`.
+- Cross-controller `Get.find` creates implicit, untyped dependencies.
+- Integration tests are tied to GetX lifecycle (robots use `Get.put`).
```
chibenwa, line 20: That's a real problem. 
If test depends on internal framework then it is not it tests. 

Can we imagine writing such tests without getX?

Imo it is the most priority topic...

docs/adr/0077-refactor-via-riverpod-service-architecture.md
```diff
@@ -0,0 +1,154 @@
...
+- **Isolate-safe** container access pattern for FCM / WebSocket.
+- **Aligned with modern Flutter community** — long-term maintenance and hiring.
+- `go_router` declarative routing enables deep linking and web URL sync properly.
+
+### Negative
+
+- **High effort** — ~77 engineer-days across 16 phases.
+- **High risk** — two DI systems coexist during months of migration; partial states possible on main.
+- **Routing migration is invasive** — deferred loading, 15+ routes, web/mobile parity all at once.
+- **Test suite rewrite** — all widget and integration tests must move off `Get.put` / robots.
```
chibenwa, line 101: :-/

MUST happen in priority to *really* protect the refactoring stage.

docs/adr/0077-refactor-via-riverpod-service-architecture.md
```diff
@@ -0,0 +1,154 @@
...
+- **Team learning curve** — Riverpod generator, code-gen workflow, `ref` lifecycle rules.
+- **Feature-delivery slowdown** during routing and auth phases (2-3 week partial freeze likely).
```
chibenwa, line 103: Acceptable

hoangdat:
will address in mob session with team in next monday

PR #4460, opened 2026-04-17
TF-4444 Enable User Control for Needs Action Label Removal

### LINKED ISSUE #4444: Enable User Control for “Needs Action” Label Removal
### Why

To evaluate the relevance and accuracy of the *needs-action* classification, users need the ability to remove this label when it is incorrectly assigned.

This enables direct comparison between **system-predicted labels** and **user-expected labels**.

### What

Currently:
- The *needs-action* label **cannot be removed by the user**, unlike user-defined labels.

Expected behavior:
- Users should be able to **remove the *needs-action* label** when it is incorrectly applied.

### How

Implement a UX flow for system label deletion in order to allow *needs-action* to be removable via the same UI interaction as user labels .

AlaeMghirbi:
This PR enables users  to remove `action-required` label when it is incorrectly assigned.

**Notes:** Running Tmail front locally with a backend Docker image documentation needs to be updated.

closes #4444

[Screencast from 24-04-26 10:25:00.webm](https://github.com/user-attachments/assets/828393e6-397c-46e3-bbde-54e973778f12)

## Release Notes

* **New Features**
  * Added visual "Action Required" indicator on emails that need attention.
  * Users can now remove the "Action Required" status from individual emails with a dedicated action button.

### DIFF
--- a/lib/features/email/presentation/email_view.dart
+++ b/lib/features/email/presentation/email_view.dart
@@ -40,6 +40,7 @@ import 'package:tmail_ui_user/features/email/presentation/widgets/email_view_loa
 import 'package:tmail_ui_user/features/email/presentation/widgets/information_sender_and_receiver_builder.dart';
 import 'package:tmail_ui_user/features/email/presentation/widgets/mail_unsubscribed_banner.dart';
 import 'package:tmail_ui_user/features/email/presentation/widgets/view_entire_message_with_message_clipped_widget.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/handle_ai_needs_action_extension.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/handle_open_context_menu_extension.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/labels/handle_logic_label_extension.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/verify_display_overlay_view_on_iframe_extension.dart';
@@ -275,6 +276,20 @@ class EmailView extends GetWidget<SingleEmailController> {
               imagePaths: controller.imagePaths,
               isMobileResponsive: isMobileResponsive,
               labels: emailLabels,
+              showNeedsAction: controller
+                      .mailboxDashBoardController.isAINeedsActionEnabled &&
+                  presentationEmail.hasNeedAction,
+              onDeleteNeedsAction: () {
+                final context = Get.context;
+                final labelDisplay = context == null
+                    ? 'Action Required'
+                    : AppLocalizations.of(context).actionRequired;
+
+                controller.onRemoveNeedsActionKeyword(
+                  presentationEmail.id,
+                  labelDisplay: labelDisplay,
+                );
+              },
               onDeleteLabelAction: (label) => controller.onToggleLabelAction(
                 presentationEmail.id,
                 label,
--- a/lib/features/labels/presentation/widgets/ai_action_widget.dart
+++ b/lib/features/labels/presentation/widgets/ai_action_widget.dart
@@ -1,0 +1,44 @@
+import 'package:core/core.dart';
+import 'package:flutter/material.dart';
+import 'package:tmail_ui_user/features/base/widget/labels/tag_widget.dart';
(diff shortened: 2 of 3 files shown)


lib/features/email/presentation/email_view.dart
```diff
@@ -280,10 +276,24 @@ class EmailView extends GetWidget<SingleEmailController> {
...
               isMobileResponsive: isMobileResponsive,
               labels: emailLabels,
+              showNeedsAction: controller
+                      .mailboxDashBoardController.isAINeedsActionEnabled &&
+                  presentationEmail.hasNeedAction,
+              onDeleteNeedsAction: () {
+                final context = Get.context;
+                final labelDisplay = context == null
+                    ? 'Action Required'
+                    : AppLocalizations.of(context).actionRequired;
```
dab246, line 286: Why do not use `context` of _buildEmailMessage function?

lib/features/email/presentation/email_view.dart
```diff
@@ -280,10 +276,24 @@ class EmailView extends GetWidget<SingleEmailController> {
               imagePaths: controller.imagePaths,
               isMobileResponsive: isMobileResponsive,
               labels: emailLabels,
+              showNeedsAction: controller
+                      .mailboxDashBoardController.isAINeedsActionEnabled &&
+                  presentationEmail.hasNeedAction,
```
dab246, line 281: should create variable for this:

```
...
final showNeedsAction = controller .mailboxDashBoardController.isAINeedsActionEnabled && presentationEmail.hasNeedAction;

return EmailSubjectWidget(
```

then

```
...
showNeedsAction: showNeedsAction,
...
```

lib/features/labels/presentation/widgets/ai_action_widget.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:tmail_ui_user/features/base/widget/labels/tag_widget.dart';
+import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
+
+class AiActionWidget extends StatelessWidget {
+  final double horizontalPadding;
+  final double? maxWidth;
+  final Widget? actionWidget;
+  final EdgeInsetsGeometry? padding;
+
+  const AiActionWidget({
```
hoangdat, line 12: why not use directly `LabelWidget`?

AlaeMghirbi: Because `LabelWidget` requires a `label` field:
```
class LabelWidget extends StatelessWidget {
  final Label label;
  final double horizontalPadding;
  ...
}
```

which is not the case for `AiNeedsAction`. It is not a label yet, so I don’t think hardcoding it as a label is a clean solution.

dab246: We already have `AiActionTagWidget` available, you can use it.

AlaeMghirbi: Yes, I know I used it initially for testing, but it is only a tag. What I actually need is a real widget that supports the **delete action** for required items.

Reference: https://github.com/linagora/tmail-flutter/pull/4460#issue-4281168349

You can check this video for a clearer idea — it demonstrates exactly what I mean. It is the same distinction as the one between `LabelTagWidget` and `LabelWidget`.

dab246: Oh no, you can add the necessary properties to `AiActionTagWidget`. Ex: 

```dart
class AiActionTagWidget extends StatelessWidget {
  final EdgeInsetsGeometry? margin;
  final Widget? actionWidget;

  const AiActionTagWidget({super.key, this.margin, this.actionWidget});

  @override
  Widget build(BuildContext context) {
    return TagWidget(
      text: AppLocalizations.of(context).actionRequired,
      backgroundColor: AppColor.aiActionTag,
      textColor: Colors.white,
      isTruncateText: true,
      showTooltip: PlatformInfo.isWeb,
      actionWidget: actionWidget,
      margin: margin,
    );
  }
}

```

We should use a single widget for easier management. And maybe later the `ActionRequired` tag in the email list screen will also have a delete button. We can easily reuse it.

AlaeMghirbi: Done!

AlaeMghirbi:
#### PR moved here: https://github.com/linagora/tmail-flutter/pull/4471

PR #4452, opened 2026-04-15
docs: add ADRs for Patrol web integration test — setup, architecture, migration

dab246:
## Summary                                                                                                                                    
                                                                             
  Add 4 ADRs covering the full plan for extending Patrol integration testing to web:                                                            
                                                                             
  - **ADR-0080**: Setup & execution — prerequisites, environment variables, how to run locally and in CI (headless), how to view results. Notes key differences from mobile setup (no ngrok/jq needed).                    
  - **ADR-0081**: Cross-platform test architecture — Abstract Factory + Dart conditional exports to achieve full SOLID compliance. One test file per scenario runs on all platforms; no platform `if/else` in any logic class.                                                                
  - **ADR-0082**: Migration guide — impact analysis on existing mobile tests (~200 files), shared base robot pattern, and a complete Send Email implementation example.                                                                                                                       
  - **ADR-0083**: Migration plan — 5-PR breakdown. PRs 1–4 are additive and keep mobile CI green; PR-5 is the single atomic breaking change that activates the new architecture on both platforms.           

## Related

- https://github.com/linagora/tmail-flutter/pull/4450                                                                                 

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Documentation**
  * Added guidance for web (Chrome) integration test setup, prerequisites, local and CI run modes, and where results appear.
  * Defined a cross-platform Test → Scenario → Robot architecture with platform-specific robot implementations wired via a factory/provider.
  * Added a migration guide with examples and a step-by-step migration plan for moving existing tests to the new model.
  * Described CI orchestration and platform-aware test selection/skip behavior.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

PR #4450, opened 2026-04-14
Patrol web POC

tddang-linagora:
## Docs
- https://github.com/linagora/tmail-flutter/pull/4452

Introduces Abstract Factory + conditional export infrastructure from ADR-0081. Platform selection resolved at compile time via dart.library.html — no if/else in any logic class.

New structure:
- factories/: RobotFactory, Mobile/WebRobotFactory, conditional provider
- robots/abstract/: AbstractLogin/Thread/ComposerRobot interfaces
- robots/mobile/: thin wrappers over existing robots
- robots/web/: WebComposerRobot overrides addContent for web

Migrates all scenarios/tests to new BaseTestScenario(super.$, super.robots) API. SendEmailScenario is the reference implementation using abstract robots.

## Demo

https://github.com/user-attachments/assets/cfdbc518-a76c-4943-9e72-8a0d9a95e989

PR #4461, opened 2026-04-17
poc: implement cross-platform robot factory architecture

tddang-linagora:
## Docs
- https://github.com/linagora/tmail-flutter/pull/4452

Introduces Abstract Factory + conditional export infrastructure from ADR-0081. Platform selection resolved at compile time via dart.library.html — no if/else in any logic class.

New structure:
- factories/: RobotFactory, Mobile/WebRobotFactory, conditional provider
- robots/abstract/: AbstractLogin/Thread/ComposerRobot interfaces
- robots/mobile/: thin wrappers over existing robots
- robots/web/: WebComposerRobot overrides addContent for web

Migrates all scenarios/tests to new BaseTestScenario(super.$, super.robots) API. SendEmailScenario is the reference implementation using abstract robots.


integration_test/base/test_base.dart
```diff
@@ -16,8 +17,7 @@ class TestBase {
 
   void runPatrolTest({
     required String description,
-    required BaseScenario Function(PatrolIntegrationTester $) scenarioBuilder,
-    List<TestTags> tags = const [],
```
hoangdat, line 22: we should keep, but for other use not for mobile/web seperately

tddang-linagora: Done in https://github.com/linagora/tmail-flutter/pull/4450

tddang-linagora:
Cherry picked to https://github.com/linagora/tmail-flutter/pull/4450

Issue #4473, opened 2026-04-24
Android app crash when composing and app goes in backgrounds


chibenwa:
This HORRIBLE bug happened to me again

I reported it to @hoangdat 3 weeks ago and I am VERY disapointed to see that the ticket is not even opened!

@hoangdat staff one people on it to fix it!

chibenwa:
Loosing user input is not acceptable.

Please print it and stamp it on the wall !

chibenwa:
<img width="704" height="283" alt="Image" src="https://github.com/user-attachments/assets/415ea355-9e39-421e-88a4-071519ac29c0" />

chibenwa:
@hoangdat has video on his private signal chat with me, ask him when working on it.

hoangdat:
- [ ] observe app mobile life cycle and trigger save draft locally `onPause`
- [ ] store in Hive email in draft & state of composer
- [ ] how to distinguish app be kill by user or kill by system ????
- [ ] in case of app was force to close check in `onRestart` to open composer again

dab246:
Same with #4424

PR #4477, opened 2026-04-28
docs(adr): add ADR-0086 for Android composer draft loss on background

dab246:
## Issue

#4473 

## Summary by CodeRabbit

* **Documentation**
  * Added an ADR describing a comprehensive Android composer draft-preservation strategy to prevent content loss when the app or renderer is terminated in background.
  * Documents encrypted local caching of editor content with short-guard and periodic snapshot semantics to avoid transient or missed saves.
  * Describes silent, best-effort server saves and a two-phase recovery that auto-restores drafts on resume or next launch.


docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+
+---
+
+## Decision
+
+### Four-Layer Defence Strategy
+
+#### Layer 1 — Proactive Local Cache at `AppLifecycleState.inactive`
+
+Add an `AppLifecycleListener` to `ComposerController` (mobile only). On `onInactive`, extract the editor HTML and all composer fields (recipients, subject, identity, attachments metadata), then persist them to a new Hive box (`composerMobileCache`) keyed by `composerId`.
```
hoangdat: should link to the definition of `composerId`? maybe a ADR ?

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+
+This is the primary fix for both RC1 and RC2. It is fast, offline, and does not depend on network connectivity.
+
+**Fallback for `getText()` failure**: a periodic content snapshot (`Timer.periodic(30s)`) runs while the editor is loaded, storing the last known HTML in `_lastKnownContent`. If `getText()` times out (> 2 s) at `inactive`, the snapshot is used.
+
+#### Layer 2 — Silent Server-Side Draft Save at `AppLifecycleState.inactive`
+
+After Layer 1 completes (still within `onInactive`), trigger `CreateNewAndSaveEmailToDraftsInteractor` silently in the background (best-effort, no UI feedback, no dialog). This saves the draft to the JMAP server, protecting against RC2 even if the local cache is later lost (e.g. app data cleared).
```
hoangdat: should care to the duplicate call if user switch state too rapid

docs/adr/0086-android-composer-draft-loss-on-background.md
hoangdat: I also aware that, system dialog can also make the `inActive` trigger. we should carefully handle the case

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+  final bool hasReadReceipt;
+  final bool isMarkAsImportant;
+  final bool isCleanClose;
+  final int timestampMs;
+  final String? draftEmailId;     // null if new compose
+  final String emailActionType;   // EmailActionType.name
+}
+```
+
+Attachments store only metadata (name, size, blobId) — not the binary content. This is sufficient to reconstruct the `CreateEmailRequest` on restore since the blobs are already on the server.
```
hoangdat: how about the case if user switch app during the uploading attachment?

chibenwa (dismissed):
Overall OK to me.

Thanks for the deep analysis.

Question is what exactly is riverpod used for?

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+
+**Fallback for `getText()` failure**: a periodic content snapshot (`Timer.periodic(30s)`) runs while the editor is loaded, storing the last known HTML in `_lastKnownContent`. If `getText()` times out (> 2 s) at `inactive`, the snapshot is used.
+
+#### Layer 2 — Silent Server-Side Draft Save at `AppLifecycleState.inactive`
+
+After Layer 1 completes (still within `onInactive`), trigger `CreateNewAndSaveEmailToDraftsInteractor` silently in the background (best-effort, no UI feedback, no dialog). This saves the draft to the JMAP server, protecting against RC2 even if the local cache is later lost (e.g. app data cleared).
+
+**Why `inactive` and not `paused`**: at `AppLifecycleState.paused`, Android Doze mode and app-standby buckets may throttle or block outbound network traffic before the request completes. `inactive` is before the compositor is frozen, giving the network call the best chance of completing while the renderer is still alive.
+
+This is **fire-and-forget** — failure is logged but not surfaced to the user. The draft will appear in the Drafts folder on success.
```
hoangdat: maybe a sentry message log?

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+|---|---|
+| `lib/features/composer/presentation/composer_controller.dart` | Add `AppLifecycleListener`, periodic snapshot timer, `_lastKnownContent` |
+| `lib/features/composer/presentation/composer_bindings.dart` | Register `ComposerMobileCacheClient` |
+| `lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart` | Add RC2 recovery check in `onReady()` — reads cache and auto-opens composer |
+
+### Cache eviction
+
+- Entries older than 24 hours are ignored on restore (treated as expired).
+- Entries are deleted after successful restoration or after `markCleanClose`.
+- On successful server-side draft save (Layer 2), the local cache entry is also cleared.
```
hoangdat, line 316: still confused about this point? what happen if it is not the `clean close`?

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,280 @@
...
+- Migrating `ComposerController` from GetX to Riverpod — that is part of the broader ADR-0075 migration roadmap.
+- Restoring uploaded attachment binaries — blobs are already on the server; only metadata is needed.
+
+---
+
+## Consequences
+
+### Positive
+
+- Draft content is preserved across Android background kills (both renderer-only and full process), without any user action required. On full process kill (RC2), the composer is automatically re-opened on next app launch.
```
hoangdat, line 333: how about the case of iOS?

dab246:
> Question is what exactly is riverpod used for?
>

ComposerAutoSaveNotifier (Riverpod StateNotifier) owns all cache read/write logic: `saveLocally()`, `restore()`, `markCleanClose()`, and `clearCache()`. `ComposerController` (GetX) acts purely as a consumer via `appProviderContainer.read(...)`. This keeps persistence concerns isolated, independently testable, and aligned with the ADR-0075 migration roadmap. Added explanation to "Riverpod pattern" section

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: Ok. Failure is logged to the console and recorded as a Sentry breadcrumb (severity: `info`), allowing the team to monitor failure rates without surfacing anything to the user. Updated "failure is logged" to explicitly include Sentry breadcrumb reporting.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: The entry remains in cache until one of three conditions is met: 
(a) it is successfully restored by Layer 3/4, 
(b) it is replaced by a newer snapshot, or 
(c) it expires after 24 hours. 
This is intentional, every involuntary termination must trigger a restore attempt. Added an explicit bullet under "Cache eviction" to document this behavior.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: RC1 (renderer kill) is Android-specific. However, Layers 1–4 also run on iOS, providing the same protection when the app is terminated in the background. Although WKWebView is generally less aggressively terminated on iOS than on Android, the recovery flow remains fully applicable. Added a note under "Positive Consequences" to clarify this.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: `composerId` is a web-only concept managed by `ComposerManager` for multiple simultaneous composers. On mobile, `ComposerBindings` is always instantiated with `composerId = null`. The mobile cache uses a separate `_mobileSessionId`, a UUID generated in `ComposerController.onInit()` on mobile only. Since mobile supports only one active composer at a time, a single UUID per session is sufficient as the cache key. Updated the ADR accordingly: replaced all `composerId` references with `_mobileSessionId` in Layer 1, Layer 3, the architecture diagrams, and the `SavedComposerSnapshot` model.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: A boolean guard, `_isSavingToDraftInProgress`, ensures that only one server-side save runs at a time. If `onInactive()` fires again while a save is already in flight, the duplicate request is silently skipped. The flag is cleared once the operation completes. Added a "Deduplication Guard" section to Layer 2.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: A short guard delay (~300 ms) is applied before proceeding with the save. If the app returns to the active state within that window, the save is cancelled. This prevents unnecessary Hive writes and network calls caused by transient lifecycle changes, such as system dialogs or permission prompts. Added a "System Dialog False Positive" note to Layer 1.

docs/adr/0086-android-composer-draft-loss-on-background.md
dab246: Attachments without a resolved `blobId` (that is, uploads still in progress) are excluded from `SavedComposerSnapshot`. Only fully uploaded attachments are persisted and restored. After recovery, any in-progress attachments must be re-uploaded manually. This is a known limitation. Added this clarification to Layer 1, Negative Consequences, and "What Is Not in Scope."

docs/adr/0086-android-composer-draft-loss-on-background.md
hoangdat: but maybe with breadcrumb, it will not be put to Sentry if no message or error happen?

dab246: Updated to use `SentryManager.instance.captureException(exception, level: SentryLevel.warning)` which sends the event immediately and independently, guaranteeing visibility in Sentry.

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,482 @@
...
+
+This is **fire-and-forget** — failure is logged to console and reported via `SentryManager.instance.captureException(exception, level: SentryLevel.warning)` for monitoring purposes, but not surfaced to the user. The draft will appear in the Drafts folder on success.
+
+> **Why `captureException` and not `addBreadcrumb`**: breadcrumbs are only uploaded to Sentry when a subsequent error event is captured. A silent fire-and-forget failure never triggers such an event, so the breadcrumb would be silently lost. `captureException` sends the event immediately and independently, guaranteeing visibility.
+
+**Deduplication guard**: a boolean `_isSavingToDraftInProgress` flag on `ComposerController` ensures only one server-side save runs at a time. If `onInactive` fires again (e.g., user rapidly switches apps back and forth) while a save is still in flight, the duplicate trigger is silently skipped. The flag is cleared on completion (success or failure).
```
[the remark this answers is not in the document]
hoangdat: yes, for sure, no plain content be log, only metadata to make us easier to debug

docs/adr/0086-android-composer-draft-loss-on-background.md
```diff
@@ -0,0 +1,482 @@
...
+#### Layer 4 — Re-open Composer on App Restart (RC2)
+
+When Android kills the entire app process, `ComposerController` is destroyed along with the route stack. `_onAppRestart()` inside `ComposerController` will never be called because the controller no longer exists.
+
+Recovery for RC2 must happen at a higher level — `MailboxDashBoardController.onReady()` — which always runs when the app restarts. On startup:
+
+1. `MailboxDashBoardController` reads all non-clean-close entries from `ComposerMobileCacheClient`.
+2. If one or more valid entries exist (age < 24 h, `isCleanClose = false`), the dashboard automatically opens the composer with the cached snapshot passed as `ComposerArguments`.
+3. `ComposerController.onReady()` then proceeds with Layer 3 to inject the content into the editor.
+4. The cache entry is cleared after successful restoration.
```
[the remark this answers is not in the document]
hoangdat: because multiple composer is not supported in mobile, so no way to have more valid entries

hoangdat (approved):

tddang-linagora (approved):

PR #4496, opened 2026-05-07
[Part 3] TF-4473 Add presentation layer for auto save composer on Android

dab246:
## Summary by CodeRabbit

* **New Features**
  * Android mobile draft auto-save with background snapshots, restore-when-blank flow, and explicit clean-close handling.
  * Composer restore from persistent cache across app restart; platform-specific composer wiring for web vs mobile.

* **Bug Fixes**
  * Treats old-format caches without timestamps as expired; tighter web unload/reconnect handling.
  * Reduced sensitive content in error logs.

* **Tests**
  * Expanded tests for auto-save, restore, cache expiry, and request-generation behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/tmail-flutter/pull/4496)


dab246:
@coderabbitai review

lib/features/thread_detail/presentation/extension/thread_detail_on_selected_email_updated.dart
```diff
@@ -18,7 +18,11 @@ extension ThreadDetailOnSelectedEmailUpdated on ThreadDetailController {
...
   ) {
-    if (selectedEmail?.id == null && mailboxDashBoardController.isEmailOpened) {
+    final selectedPresentationEmail = selectedEmail;
+    final selectedPresentationEmailId = selectedPresentationEmail?.id;
+    final selectedPresentationEmailThreadId = selectedPresentationEmail?.threadId;
+
+    if (selectedPresentationEmailId == null && mailboxDashBoardController.isEmailOpened) {
       closeThreadDetailAction();
       return;
     }
```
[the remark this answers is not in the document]
dab246: @coderabbitai review

lib/features/composer/presentation/extensions/setup_email_content_extension.dart
```diff
@@ -195,6 +197,23 @@ extension SetupEmailContentExtension on ComposerController {
...
+      return;
+    }
+
+    if (inlineImages.isEmpty) {
+      emailContentsViewState.value = Right(GetEmailContentSuccess(htmlEmailContent: content));
+      return;
+    }
+
+    await _restoreInlineImages(content, inlineImages);
+  }
```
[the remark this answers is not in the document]
dab246: @coderabbitai review

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
```diff
@@ -0,0 +1,265 @@
...
+  String _resolveAndSyncContent(
+    String? freshContent,
+    ComposerAutoSaveNotifier notifier,
+  ) {
+    if (freshContent != null && freshContent.isNotEmpty) {
+      if (notifier.mounted) notifier.updateLastKnownContent(freshContent);
+      return freshContent;
+    }
+    return notifier.lastKnownHtmlContent;
+  }
```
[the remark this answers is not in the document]
dab246: @coderabbitai review

lib/features/composer/presentation/mobile_composer_bindings.dart
```diff
@@ -0,0 +1,46 @@
...
+  MobileComposerBindings({String? composerId, ComposerArguments? composerArguments})
+      : super.base(composerId: composerId, composerArguments: composerArguments);
+
+  @override
+  void bindPlatformCacheDatasourceImpl() {
+    Get.lazyPut(() => ComposerHiveCacheClient(), tag: composerId);
+    Get.lazyPut(() => ComposerPersistentCacheDatasourceImpl(
+      Get.find<ComposerHiveCacheClient>(tag: composerId),
+      Get.find<CacheExceptionThrower>(),
+    ), tag: composerId);
```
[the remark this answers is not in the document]
dab246: with mobile we only use autoSaveComposerId for provider, composerId always is null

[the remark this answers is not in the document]
dab246: disposePlatformCacheImpl only call on web, mobile not called.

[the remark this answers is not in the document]
dab246: @coderabbitai review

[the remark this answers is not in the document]
dab246: @coderabbitai dispose in composerBinding, only call from composer_manager & it ONLY call on web. please check again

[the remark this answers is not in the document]
dab246: oke, so this comment is resolved?

hoangdat:
- [x] composer be openned but content be lost if we open then close the composer from background very quick


https://github.com/user-attachments/assets/e0bf63fe-9e84-4c72-9636-ee810464dfa7

hoangdat:
- [x] please very carefully check the case

   During save draft, token need to be refresh it succeed but write to hive new token is terminated by system

dab246:
> * [x]  composer be openned but content be lost if we open then close the composer from background very quick
> 
> * [x]  please very carefully check the case
>   During save draft, token need to be refresh it succeed but write to hive new token is terminated by system

 Fixed

hoangdat:
@coderabbit review

lib/features/thread_detail/presentation/extension/thread_detail_on_selected_email_updated.dart
```diff
@@ -31,23 +27,23 @@ extension ThreadDetailOnSelectedEmailUpdated on ThreadDetailController {
...
       loadThreadOnThreadChanged = isThreadDetailEnabled;
-      _preloadSelectedEmail(selectedPresentationEmail);
+      _preloadSelectedEmail(selectedEmail!);
       return;
     }
 
     // Thread setting updated, no need to dispose current single email controller
-    if (currentExpandedEmailId.value == selectedPresentationEmailId && selectedPresentationEmail != null) {
-      _preloadSelectedEmail(selectedPresentationEmail);
+    if (currentExpandedEmailId.value == selectedEmail!.id) {
```
hoangdat: it still be have this kind of error 

Image

lib/features/thread_detail/presentation/extension/thread_detail_on_selected_email_updated.dart
```diff
@@ -31,23 +27,23 @@ extension ThreadDetailOnSelectedEmailUpdated on ThreadDetailController {
     onKeyboardShortcutInit();
     scrollController ??= ScrollController();
 
-    if (currentExpandedEmailId.value == null && selectedPresentationEmail != null && selectedPresentationEmailId != null) {
+    if (currentExpandedEmailId.value == null) {
```
hoangdat: I still not get the idea of these changes in this file

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
```diff
@@ -85,13 +80,41 @@ extension HandleMobileAutoSaveExtension on ComposerController {
...
     if (content != null) _autoSaveNotifier()?.updateLastKnownContent(content);
+
+    if (WidgetsBinding.instance.lifecycleState != AppLifecycleState.resumed) return;
+
+    final notifier = _autoSaveNotifier();
+    if (notifier == null || notifier.isCleanClose) return;
+
+    final int currentHash;
+    try {
+      currentHash = await hashComposerStateForAutoSave().timeout(_getContentTimeout);
```
hoangdat: we still have the double get content from web view here

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
```diff
@@ -85,13 +80,41 @@ extension HandleMobileAutoSaveExtension on ComposerController {
...
+      unawaited(_periodicSaveTask());
     });
   }
 
-  Future<void> _updateLastKnownContent() async {
+  // Updates lastKnownHtmlContent and, when foregrounded, silently saves the
+  // draft to the server only if composer state changed since the last auto-save.
+  // Not called from inactive: background token refresh may be killed before
+  // writing the refreshed token to Hive, corrupting the account cache.
+  Future<void> _periodicSaveTask() async {
```
hoangdat, line 87: I found we still have some problem in the Sentry log when saving draft 

Image

hoangdat: we should improve it:
1. why it failed?
2. how to have more information to debug if it happen in the future?

lib/features/thread_detail/presentation/extension/thread_detail_on_selected_email_updated.dart
dab246: We will revert this file back to its original code.

lib/features/thread_detail/presentation/extension/thread_detail_on_selected_email_updated.dart
dab246: We will revert this file back to its original code.

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
dab246: fixed

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
dab246: fixed

hoangdat:
- [x] inline image be broken when comeback to composer

dab246:
> * [x]  inline image be broken when comeback to composer






https://github.com/user-attachments/assets/5abb5f7e-f278-4c9c-9e0a-ace6a4994177

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1260,8 +1261,8 @@ class ComposerController extends BaseController
     return _savedEmailDraftHash != newDraftHash;
   }
 
-  Future<int> _hashComposingEmail() async {
-    String emailContent = await getContentInEditor();
+  Future<int> _hashComposingEmail({String? htmlContent}) async {
+    String emailContent = htmlContent ?? await getContentInEditor();
```
hoangdat, line 1265:
```
String emailContent = htmlContent != null
      ? htmlContent.removeEditorStartTag()                                         
      : await getContentInEditor();
```

to have consistent with other place of compute hashing

dab246: No need because the content we saved has already been deleted using removeEditorStartTag.

dab246:
- [ ] Unable to perform `Save as Draft` a second time when reopening an existing draft email containing attachments or inline images.

  Note: This issue does not occur when composing a brand new email with attachments or inline images.

### Steps to Reproduce

1. Open an existing draft email containing attachments or inline images.
2. Perform `Save as Draft` successfully for the first time.
3. Keep the composer open.
4. Continue editing the email content.
5. Perform `Save as Draft` again.

### Actual Result

The second `Save as Draft` operation fails.

JMAP `Email/set` returns the following error:

```json
{
  "sessionState": "2c9f1b12-b35a-43e6-9af2-0106fb53a943",
  "methodResponses": [
    [
      "Email/set",
      {
        "accountId": "aaaaaaaaa",
        "oldState": "b3c78450-4ad6-11f1-a0bc-db8930321029",
        "newState": "b3c78450-4ad6-11f1-a0bc-db8930321029",
        "notCreated": {
          "b6638bf1-4ad6-11f1-9953-1f789dc35c3f": {
            "type": "invalidArguments",
            "description": "Attachment not found: 97ccb1e0-4ad5-11f1-a0bc-db8930321029_7",
            "properties": [
              "attachments"
            ]
          }
        }
      },
      "c0"
    ]
  ]
}
```

### Expected Result

`Save as Draft` should work successfully multiple consecutive times without requiring the composer to be closed, including draft emails containing attachments or inline images.

hoangdat:
please rebase @dab246

dab246:
> please rebase @dab246

Done

lib/features/composer/presentation/extensions/handle_mobile_auto_save_extension.dart
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  void _persistCleanCloseToCache() {
+    final accountId = mailboxDashBoardController.accountId.value;
+    final userName = mailboxDashBoardController.sessionCurrent?.username;
+    if (accountId == null || userName == null) return;
+    unawaited(appProviderContainer
+        .read(markComposerLocalCacheCleanCloseProvider)
+        .execute(accountId, userName));
+  }
```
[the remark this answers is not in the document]
dab246: What's your suggestion?

tddang-linagora (approved):
Worked on Oneplus 8T

hoangdat (approved):

lib/features/composer/presentation/extensions/setup_email_content_extension.dart
```diff
@@ -44,6 +44,8 @@ extension SetupEmailContentExtension on ComposerController {
       await _loadReplyForwardEmailContent(arguments);
     } else if (currentEmailActionType == EmailActionType.reopenComposerBrowser) {
       await _loadReopenBrowserEmailContent(arguments);
+    } else if (currentEmailActionType == EmailActionType.restoreComposerFromPersistentCache) {
```
9clg6, line 47: Why don't you use a switch case ?

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
```diff
@@ -0,0 +1,113 @@
...
+  final bool isCleanClose;
+  final String lastKnownHtmlContent;
+
+  const ComposerAutoSaveState({
+    this.hasRecoverableSnapshot = false,
+    this.isCleanClose = false,
+    this.lastKnownHtmlContent = '',
+  });
+
+  ComposerAutoSaveState copyWith({
```
9clg6, line 23: @freezed could help
or do a EquatableMixin to avoid copyWith declaration

lib/features/composer/presentation/providers/composer_cache_providers.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:tmail_ui_user/features/mailbox_dashboard/data/datasource/composer_cache_datasource.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/data/datasource_impl/composer_persistent_cache_datasource_impl.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/data/repository/composer_cache_repository_impl.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/repository/composer_cache_repository.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/mark_composer_cache_clean_close_interactor.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/resolve_composer_cache_for_restore_interactor.dart';
+import 'package:tmail_ui_user/main/exceptions/thrower/cache_exception_thrower.dart';
+
+final _composerHiveCacheClientProvider =
```
9clg6, line 12: Don't create 1 file for all providers, create 1 file per provider

lib/features/composer/presentation/providers/composer_cache_providers.dart
```diff
@@ -0,0 +1,39 @@
...
+    Provider<ComposerHiveCacheClient>((_) => ComposerHiveCacheClient());
+
+final _cacheExceptionThrowerProvider =
+    Provider<CacheExceptionThrower>((_) => CacheExceptionThrower());
```
9clg6, line 16: Don't use "Provider", but instead use @riverpod annotation available in Riverpod 3.0, 
like here : 
https://github.com/linagora/twake-on-matrix/pull/3008/changes#diff-5f840c481af4c2fba3fa9a2723299769aa23cfb511d924330a342ac4b94a1068

lib/features/composer/presentation/providers/composer_cache_providers.dart
```diff
@@ -0,0 +1,39 @@
...
+import 'package:tmail_ui_user/features/mailbox_dashboard/data/datasource_impl/composer_persistent_cache_datasource_impl.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/data/repository/composer_cache_repository_impl.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/repository/composer_cache_repository.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/mark_composer_cache_clean_close_interactor.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/resolve_composer_cache_for_restore_interactor.dart';
+import 'package:tmail_ui_user/main/exceptions/thrower/cache_exception_thrower.dart';
+
+final _composerHiveCacheClientProvider =
+    Provider<ComposerHiveCacheClient>((_) => ComposerHiveCacheClient());
```
9clg6, line 13: Same as above, don't use Provider, but @riverpod instead

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
```diff
@@ -0,0 +1,113 @@
...
+        isCleanClose,
+        lastKnownHtmlContent,
+      ];
+}
+
+/// Riverpod [StateNotifier] for Android composer auto-save (see ADR-0086).
+/// Keyed by composerId so each [ComposerController] gets an isolated instance.
+/// Must be invalidated from [ComposerController.onClose] to prevent unbounded
+/// family-instance growth in the global [ProviderContainer].
+class ComposerAutoSaveNotifier extends StateNotifier<ComposerAutoSaveState> {
```
9clg6, line 46: Don't use StateNotifier, it's deprecated in Riverpod 3.0
use instead : 

```
@riverpod
class ComposerAutoSave extends _$ComposerAutoSave {
  @override
  ComposerAutoSaveState build() => const ComposerAutoSaveState();
}
```

lib/features/mailbox_dashboard/presentation/extensions/handle_composer_restore_on_mobile_extension.dart
```diff
@@ -0,0 +1,52 @@
...
+    } catch (e, stackTrace) {
+      logWarning('HandleComposerRestoreOnMobileExtension::checkAndRestoreComposerOnMobile: error=$e\n$stackTrace');
+    }
+  }
+
+  Future<ComposerPersistentCache?> _resolveRestorableCache(
+    AccountId accountId,
+    UserName userName,
+  ) async {
+    final result = await appProviderContainer
```
9clg6, line 35: Using global variable to access to a provider is a anti-pattern in 3.0, you shouldn't do it
+
You should - in 95% of times - ref.watch and not ref.read. 
ref.read should be use only in on tap action on buttons for example

```Do not use [Ref.read](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/Ref/read.html) as a mean to "optimize" your code by avoiding [Ref.watch](https://pub.dev/documentation/hooks_riverpod/latest/hooks_riverpod/Ref/watch.html). This will make your code more brittle, as changes in your provider's behavior could cause your UI to be out of sync with the provider's state.```
https://arc.net/l/quote/icnguhxo

**Comment by chibenwa** [comment]
Please @dab246 we need to take the time to open a PR addressing @9clg6 review comments in a follow up PR.
Thanks.

File: lib/features/composer/presentation/extensions/setup_email_content_extension.dart (hunk 11 above)
Comment (dab246) [reply lib/features/composer/presentation/extensions/setup_email_content_extension.dart:47]: Thanks for the suggestion. We considered `switch` but kept `if-else` intentionally for two reasons:

**Semantic groupings with multi-value predicates.** Three of the branches use named boolean getters (`_isEditDraftContentAction`, `_isMailtoContentAction`, `_isReplyForwardContentAction`), each of which encapsulates a `Set.contains()` check over multiple `EmailActionType` values. Collapsing these into `switch` cases would require duplicating the enum values inline and losing the named, reusable predicates — those getters are also referenced elsewhere in the same file (e.g., line 287).

**CodeScene complexity budget.** A `switch` over all `EmailActionType` cases would introduce a high number of branches in a single method, which CodeScene flags as a cyclomatic complexity violation. The current structure distributes that complexity into small, named getters, keeping the dispatch method itself lean.

File: lib/features/composer/presentation/providers/composer_auto_save_notifier.dart (hunk 12 above)
Comment (dab246) [reply lib/features/composer/presentation/providers/composer_auto_save_notifier.dart:23]: Thanks for the pointer.

**On `EquatableMixin`:** `ComposerAutoSaveState` already `extends Equatable`, which is the standard inheritance-based approach recommended by the Equatable package. `EquatableMixin` is intended specifically for classes that cannot use inheritance (i.e., when the class already extends another type). So this is already handled correctly.

**On `@freezed`:** `freezed_annotation` appears only as a transitive dependency in this project — there is no `@freezed` usage anywhere in `lib/`. Adding `freezed` as a direct dev dependency and a generated `.freezed.dart` file for a 3-field state class would introduce code-gen overhead that isn't justified here. The manual `copyWith` is explicit, zero-magic, and easy to maintain.

File: lib/features/composer/presentation/providers/composer_cache_providers.dart (hunk 13 above)
Comment (dab246) [reply lib/features/composer/presentation/providers/composer_cache_providers.dart:12]: Appreciated, though we believe the current structure is the right choice for this specific file.

`composer_cache_providers.dart` contains a single **linear DI chain** for one feature:

```
ComposerHiveCacheClient → CacheExceptionThrower
  → ComposerCacheDatasource → ComposerCacheRepository
    → RemoveAll / MarkCleanClose / ResolveForRestore Interactors
```

The three infrastructure providers (`_composerHiveCacheClientProvider`, `_cacheExceptionThrowerProvider`, `_composerCacheDatasourceProvider`) are **private** (prefixed `_`) — they are implementation details that must not be exposed outside the file. Splitting them into separate files would force them to become public, leaking internal wiring. Grouping a tightly coupled dependency chain into one file is the same pattern used in GetX bindings across the rest of the codebase.

lib/features/composer/presentation/providers/composer_cache_providers.dart
dab246: Good point for future direction. However, the project currently uses **`flutter_riverpod: 2.6.1`** — this version constraint is intentional to maintain compatibility with existing dependencies and the current Flutter version pinned via FVM.

`@riverpod` code-generation requires `riverpod_annotation` and `riverpod_generator`, neither of which is present in the project's `pubspec.lock`. Introducing them for one new feature would also create inconsistency with `localSettingsNotifierProvider` and other existing providers that follow the Riverpod 2.x manual API.

We agree this should be improved. The plan is to migrate to Riverpod 3.0 (with full `@riverpod` annotation adoption) in a dedicated upgrade PR, where all providers are migrated consistently. Doing it piecemeal in a bugfix PR would leave the codebase in a mixed state that is harder to maintain.

lib/features/composer/presentation/providers/composer_cache_providers.dart
dab246: Good point for future direction. However, the project currently uses **`flutter_riverpod: 2.6.1`** — this version constraint is intentional to maintain compatibility with existing dependencies and the current Flutter version pinned via FVM.

`@riverpod` code-generation requires `riverpod_annotation` and `riverpod_generator`, neither of which is present in the project's `pubspec.lock`. Introducing them for one new feature would also create inconsistency with `localSettingsNotifierProvider` and other existing providers that follow the Riverpod 2.x manual API.

We agree this should be improved. The plan is to migrate to Riverpod 3.0 (with full `@riverpod` annotation adoption) in a dedicated upgrade PR, where all providers are migrated consistently. Doing it piecemeal in a bugfix PR would leave the codebase in a mixed state that is harder to maintain.

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
dab246: Same root cause as Comments 4 & 5 — `StateNotifier` is deprecated in **Riverpod 3.0**, but the project is on **2.6.1** where it remains a stable, first-class API.

Additionally, `StateNotifier` is already the established pattern in this codebase (`LocalSettingsNotifier` in `local_settings_notifier.dart` follows the exact same structure). Deviating from that in a single feature PR would introduce inconsistency without a broader migration strategy.

As noted above, the migration to the new Notifier API will be done project-wide in a dedicated Riverpod 3.0 upgrade PR.

lib/features/mailbox_dashboard/presentation/extensions/handle_composer_restore_on_mobile_extension.dart
dab246: Both points are valid in a **pure Riverpod** app — but this codebase is a **GetX + Riverpod hybrid**.

**On `appProviderContainer`:** GetX controllers do not have access to `BuildContext` or `ref`. `appProviderContainer` is a deliberate architectural bridge to allow GetX-managed code to read Riverpod providers. This is already the established pattern across the codebase (`local_settings_service.dart`, `search_email_controller.dart`, `thread_controller.dart`, `preferences_controller.dart`). This trade-off is intentional and will be revisited when the full migration to Riverpod (per the ongoing refactoring initiative) is completed.

**On `ref.read` vs `ref.watch`:** `_resolveRestorableCache()` is imperative business logic — a one-shot async call to execute a use case, not a reactive subscription. `ref.watch` is only meaningful inside a widget's `build` method where it can trigger rebuilds. The Riverpod documentation itself states that `ref.watch` should not be used outside of `build`, and explicitly lists "button tap actions" as the correct use case for `ref.read` — which is exactly the context here (a method called once on app resume). Using `ref.read` is correct and intentional.

dab246:
> Please @dab246 we need to take the time to open a PR addressing @9clg6 review comments in a follow up PR. Thanks.

HI @chibenwa , I took the time to address @9clg6's comments. And the final review is all good. I always respect you and @9clg6, so I think from now on we should add @9clg6 to our reviews whenever we create PRs to avoid having to go back and fix up PRs we've already merged. Thanks

PR #4509, opened 2026-05-12
Hotfix: Prevent `sentry_flutter` from loading Sentry bundle from CDN  

dab246:
## Issue

After deploying #4343 (self-hosted Sentry bundle), `bundle.tracing.min.js` was still being loaded from `browser.sentry-cdn.com` instead of the self-hosted server.

Screenshot 2026-05-12 at 10 36 42

### 1. Wrong interception hook

`sentry-interceptor.js` patched `document.createElement('script')`, but `sentry_flutter`'s `web_script_dom_api.dart` creates script elements via `new HTMLScriptElement()` (Dart `package:web` interop), which bypasses `document.createElement`.  As a result, the interceptor was never triggered.

### 2. SRI blocked script loading

The `integrity="sha384-..."` attributes in `index.html` caused SRI validation failures because the deployed file content differed from the local hash.

This prevented both self-hosted scripts from loading:

- `sentry-tracing.min.js`
- `sentry-interceptor.js`

Because the interceptor failed to load, requests to `browser.sentry-cdn.com` were not blocked.

### Update `sentry-interceptor.js`

Replace `document.createElement` interception with `HTMLScriptElement.prototype.src` descriptor patching.

This ensures all script `src` assignments are intercepted, regardless of whether the script element is created via:

- `document.createElement(...)`
- `new HTMLScriptElement()`

### Update `index.html`

Remove `integrity` and `crossorigin` attributes from the self-hosted Sentry scripts.

For same-origin resources, SRI provides limited security value but increases deployment risk because stale hashes can silently break script loading.

Also update interceptor version:

- `sentry-interceptor.js?v=1.0.1`

to force browser cache refresh.

## Summary by CodeRabbit

* **Chores**
  * Switched to a self‑hosted Sentry SDK, added a startup interceptor to block runtime CDN requests, preloaded the SDK before app boot, and updated script version query params while removing subresource integrity attributes.

* **Bug Fixes**
  * Prevents CDN loads at runtime and dispatches synthetic load events so initialization proceeds reliably even when external requests are blocked.

* **Documentation**
  * Added an ADR describing the approved approach and deployment considerations.

* **Tests**
  * Updated test setup to asynchronously initialize cache configuration before running tests.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/tmail-flutter/pull/4509)

### DIFF
--- a/web/js/sentry-interceptor.js
+++ b/web/js/sentry-interceptor.js
@@ -1,4 +1,10 @@
 (function() {
+  // Guard against double-execution (e.g. duplicate script tag, hot-reload).
+  // On a second load, nativeSrcDescriptor would resolve to our already-patched
+  // setter instead of the native one, causing infinite recursion.
+  if (window.__sentryInterceptorActive) return;
+  window.__sentryInterceptorActive = true;
+
   const BLOCKED_PATHNAME = '/bundle.tracing.min.js';
 
   function isSentryCdnHostname(hostname) {
@@ -20,7 +26,7 @@
     console.log('[Sentry Interceptor] 🛑 Blocked CDN request (' + channel + '):', urlString);
     // Dispatch a synthetic 'load' event so the Sentry SDK's internal Promise
     // resolves cleanly instead of hanging indefinitely.
-    setTimeout(() => element.dispatchEvent(new Event('load')), 10);
+    Promise.resolve().then(() => element.dispatchEvent(new Event('load')));
   }
 
   // sentry_flutter's web_script_dom_api.dart creates script elements via
@@ -59,14 +65,25 @@
   }
 
   // Defense-in-depth: also intercept setAttribute('src', ...) as a fallback.
+  // Patched on HTMLScriptElement.prototype only (not Element) to avoid
+  // intercepting setAttribute on every element in the DOM.
   const originalSetAttribute = Element.prototype.setAttribute;
-  Element.prototype.setAttribute = function(name, value) {
-    const urlString = value ? String(value) : '';
-    const isScriptSrc = this instanceof HTMLScriptElement && String(name).toLowerCase() === 'src';
-    if (isScriptSrc && shouldBlockSentryUrl(urlString)) {
-      blockAndNotify(this, 'setAttribute', urlString);
-      return;
-    }
-    originalSetAttribute.call(this, name, value);
-  };
+  try {
+    Object.defineProperty(HTMLScriptElement.prototype, 'setAttribute', {
+      configurable: true,
+      writable: true,
+      value: function(name, value) {
+        if (String(name).toLowerCase() === 'src') {
+          const urlString = String(value || '');
+          if (shouldBlockSentryUrl(urlString)) {
+            blockAndNotify(this, 'setAttribute', urlString);
+            return;
+          }
+        }
+        originalSetAttribute.call(this, name, value);
+      },
+    });
+  } catch (e) {
+    console.warn('[Sentry Interceptor] Failed to patch HTMLScriptElement.setAttribute:', e);
+  }
 })();
--- a/web/js/sentry-interceptor.js
+++ b/web/js/sentry-interceptor.js
@@ -1,8 +1,4 @@
 (function() {
-  const originalCreateElement = document.createElement;
-
-  // Only block the specific tracing bundle we self-host. Narrowing to this pathname
-  // prevents silently swallowing future CDN assets that have no local replacement.
   const BLOCKED_PATHNAME = '/bundle.tracing.min.js';
 
   function isSentryCdnHostname(hostname) {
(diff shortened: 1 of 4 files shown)


web/js/sentry-interceptor.js
```diff
@@ -13,71 +9,50 @@
...
-  // Load order: sentry-tracing.min.js (self-hosted SDK) must run before this interceptor.
-  // This interceptor patches document.createElement to block any subsequent CDN injection
-  // the SDK may attempt at runtime.
-  document.createElement = function(tagName, options) {
-    const element = originalCreateElement.call(document, tagName, options);
-    if (String(tagName).toLowerCase() === 'script') {
-      patchScriptElement(element);
+      nativeSrcDescriptor.set.call(this, val);
+    },
+  });
```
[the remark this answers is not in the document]
dab246: Good catch. The `canPatchSrc` guard + `try-catch` is a valid defensive measure — while `HTMLScriptElement.prototype.src` is configurable in all modern browsers we target, a third-party extension or prior script patching the descriptor could still cause `Object.defineProperty` to throw and silently disable the entire interceptor.

Applied the core of the suggestion: `canPatchSrc` pre-flight check, `try-catch` around the `defineProperty` call, and preserved `enumerable` from the native descriptor rather than hardcoding `true`.

Skipped the `makeSrcSetter` / `patchScriptElement` / `document.createElement` block from the committable suggestion — those functions don't exist in this implementation and appear to be noise from the PR diff context.

web/js/sentry-interceptor.js
```diff
@@ -13,71 +9,64 @@
...
+  Element.prototype.setAttribute = function(name, value) {
+    const urlString = value ? String(value) : '';
+    const isScriptSrc = this instanceof HTMLScriptElement && String(name).toLowerCase() === 'src';
+    if (isScriptSrc && shouldBlockSentryUrl(urlString)) {
+      blockAndNotify(this, 'setAttribute', urlString);
+      return;
     }
-    return element;
+    originalSetAttribute.call(this, name, value);
   };
```
hoangdat:
```suggestion
  const originalSetAttribute = Element.prototype.setAttribute;
  Object.defineProperty(HTMLScriptElement.prototype, 'setAttribute', {
    configurable: true,
    writable: true,
    value: function(name, value) {
      if (String(name).toLowerCase() === 'src' && shouldBlockSentryUrl(String(value || ''))) {
        blockAndNotify(this, 'setAttribute', String(value || ''));
        return;
      }
      originalSetAttribute.call(this, name, value);
    },
  });
```

dab246: @coderabbitai review this suggestion

web/js/sentry-interceptor.js
```diff
@@ -13,71 +9,64 @@
...
     }
   }
 
   function blockAndNotify(element, channel, urlString) {
     console.log('[Sentry Interceptor] 🛑 Blocked CDN request (' + channel + '):', urlString);
-    // Fire a synthetic 'load' event so the Sentry SDK's internal Promise resolves
-    // cleanly instead of hanging indefinitely waiting for a CDN script we blocked.
+    // Dispatch a synthetic 'load' event so the Sentry SDK's internal Promise
+    // resolves cleanly instead of hanging indefinitely.
     setTimeout(() => element.dispatchEvent(new Event('load')), 10);
```
hoangdat:
```suggestion
    Promise.resolve().then(() => element.dispatchEvent(new Event('load')));
```

web/js/sentry-interceptor.js
```diff
@@ -13,71 +9,64 @@
...
-    const nativeSrcDescriptor = Object.getOwnPropertyDescriptor(HTMLScriptElement.prototype, 'src');
-    Object.defineProperty(element, 'src', {
-      configurable: true,
-      set: makeSrcSetter(nativeSrcDescriptor),
-      get: makeSrcGetter(nativeSrcDescriptor),
-    });
-    element.setAttribute = makeSetAttribute(element.setAttribute);
+  if (canPatchSrc) {
+    try {
+      Object.defineProperty(HTMLScriptElement.prototype, 'src', {
```
hoangdat, line 45: Intercept appendChild instead of setAttribute; it reliably catches dynamically inserted scripts regardless of whether src uses properties or attributes.

hoangdat:
```
const originalAppendChild = Node.prototype.appendChild;

Node.prototype.appendChild = function(node) {
  if (
    node instanceof HTMLScriptElement &&
    shouldBlockSentryUrl(node.src)
  ) {
    console.warn("Blocked external Sentry SDK:", node.src);
    return node;
  }

  return originalAppendChild.call(this, node);
```

web/js/sentry-interceptor.js
```diff
@@ -1,8 +1,4 @@
 (function() {
```
hoangdat, line 1:
```js
    // Guard against double-execution (e.g. duplicate script tag, hot-reload).
    // On a second load, nativeSrcDescriptor would resolve to our already-patched
    // setter instead of the native one, causing infinite recursion.
    if (HTMLScriptElement.prototype.__sentryInterceptorActive) return;
    HTMLScriptElement.prototype.__sentryInterceptorActive = true;
```

web/js/sentry-interceptor.js
dab246: After evaluation, we're keeping the `src` setter + `setAttribute` approach and not switching to `appendChild` interception, for a few reasons:

1. **Incomplete coverage**  
   Intercepting `appendChild` alone isn't sufficient — scripts can also be inserted via `insertBefore`, `append`, `prepend`, `innerHTML`, `replaceChild`, etc. Covering all insertion points is far more invasive than a single prototype patch.

2. **Earlier interception**  
   The `src` setter fires at the moment of assignment, before the element is ever inserted into the DOM. This is strictly earlier and more surgical than waiting for DOM insertion.

3. **Already covered**  
   The current two-layer approach (`src` property setter + `setAttribute`) handles all known patterns used by `web_script_dom_api.dart` in `sentry_flutter`. There's no evidence that `appendChild`-only scripts bypass these layers.

web/js/sentry-interceptor.js
dab246: Fixed

web/js/sentry-interceptor.js
dab246: Fixed

web/js/sentry-interceptor.js
dab246: fixed

hoangdat:
@coderabbit review

hoangdat (approved):

Issue #4532, opened 2026-05-18
Some html email can not scroll to the end


hoangdat:
### Desc

Some html can not scroll to the bottom
- Alex reported a email from VNA
- I have an adverise email 

https://github.com/user-attachments/assets/44a0a751-ad4e-42b2-913e-ec9b41f19a8d

[What’s next in mobile app measurement_.eml](https://github.com/user-attachments/files/27953005/What.s.next.in.mobile.app.measurement_.eml)

poupotte:
@hoangdat : How can we validate this ticket ?

PR #4533, opened 2026-05-19
TF-4532 Fix some html email can not scroll to the end

dab246:
## Issue

#4532 

## Analyze

Some HTML emails (e.g. marketing newsletters) include `html, body { height: 100% !important; }`, which causes an infinite feedback loop:

1. iframe expands to fit content → body (`100%`) fills iframe  
2. `document.body.scrollHeight` equals iframe height  
3. `ResizeObserver` fires → Flutter adds `offsetHtmlContentHeight` → iframe grows  
4. Loop repeats → height grows indefinitely  

## Root cause

The email `<style>` block survives sanitization and is inserted inside `<div class="tmail-content">`, after document `<head>` styles. Its `!important` rule overrides ours and binds body height to iframe viewport.

### CSS (root cause)

Inject `html, body { height: auto !important; }` in a `<style>` block after the content div so it wins the cascade and decouples body height from iframe height.

### JS (defensive)

Add 50 ms debounce and last-height deduplication in `ResizeObserver` callback to prevent repeated `postMessage` bursts when multiple elements resize (e.g. lazy-loaded images).

### Dart (defensive)

`shouldUpdateHeight` returns `false` when `scrollHeightWithBuffer == _actualHeight`, avoiding redundant `setState` calls that would still trigger the second observer event.

## Tests

- `HtmlUtils.generateHtmlDocument`: verifies CSS override exists and is placed after email content.  
- `HtmlContentViewerOnWeb.shouldUpdateHeight`: 10 unit tests covering loop guard, first render, grow/shrink, minHeight boundaries, and strict `>` vs `>=` handling for `autoAdjust`.

## Demo

- Mobile:

https://github.com/user-attachments/assets/224a72f5-9a3c-4506-9ae5-e15bbd9393a7

- Web:

https://github.com/user-attachments/assets/4bb6327e-6c43-405d-b03d-551e004e8143

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced iframe height adjustment to reliably handle dynamic content changes and edge cases.
  * Improved performance with debounced height measurements and reduced redundant updates.
  * Better cleanup on page transitions to prevent resource leaks.
  * Injected CSS override in generated HTML to prevent infinite resize loops.

* **Tests**
  * Added comprehensive tests for height-stabilization behavior and HTML document generation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/tmail-flutter/pull/4533?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
+++ b/core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
@@ -65,5 +65,29 @@ void main() {
       // 100 + 30 = 130 < 200
       expect(check(reported: 100, current: 90, autoAdjust: true), isFalse);
     });
+
+    test('dedup alone CANNOT stop the loop — CSS override is load-bearing', () {
+      double iframe = 200;
+      const buffer = 30.0;
+      final history = <double>{iframe};
+
+      for (var i = 0; i < 10; i++) {
+        final scrollHeight = iframe; // body fills iframe — the bug condition
+        final reported = scrollHeight + buffer;
+        if (HtmlContentViewerOnWeb.shouldUpdateHeight(
+          scrollHeightWithBuffer: reported,
+          currentActualHeight: iframe,
+          minHeight: 200,
+          autoAdjust: false,
+        )) {
+          iframe = reported;
+          history.add(iframe);
+        }
+      }
+
+      expect(history.length, greaterThan(5),
+          reason: 'Dart dedup does not catch the buffer-delta loop; '
+              'the CSS override in generateHtmlDocument is the actual fix.');
+    });
   });
 }
--- a/core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
+++ b/core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
@@ -2,146 +2,68 @@ import 'package:core/presentation/views/html_viewer/html_content_viewer_on_web_w
 import 'package:flutter_test/flutter_test.dart';
 
 void main() {
-  // These tests guard against the infinite-resize loop triggered by emails that
-  // include `html, body { height: 100% !important }` in their CSS.  When that
-  // rule is active, document.body.scrollHeight equals the iframe's current
-  // height, so every postMessage adds offsetBuffer and grows the iframe again.
-  // shouldUpdateHeight must return false in the stabilized state to break the
-  // loop at the Dart layer.
+  // Guards against the infinite-resize loop: emails with `html, body { height: 100% }`
+  // make scrollHeight equal the iframe height, so every update grows it forever.
   group('HtmlContentViewerOnWeb.shouldUpdateHeight', () {
     const offset = 30.0;
     const minH = 200.0;
 
-    // ── loop-guard cases ──────────────────────────────────────────────────────
+    bool check({
+      required double reported,
+      required double current,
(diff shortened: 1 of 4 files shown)


hoangdat:
@coderabbit review

core/test/utils/html_utils_test.dart
```diff
@@ -1114,6 +1114,33 @@ void main() {
...
+
+    test('override rule is present regardless of optional parameters', () {
+      final doc = HtmlUtils.generateHtmlDocument(
+        content: '<p>Simple</p>',
+        styleCSS: '.custom { color: red; }',
+        minHeight: 400,
+      );
+
+      expect(doc, contains('html, body { height: auto !important; }'));
+    });
```
hoangdat, line 1141: add this tests 

```
test('override appears after styleCSS so user CSS cannot win cascade', () {
      // Guards against refactors that move the override into <head> before
      // styleCSS — if that happens, a competing rule in styleCSS would win.
      const competing = 'html, body { height: 100% !important; }';
      final doc = HtmlUtils.generateHtmlDocument(
        content: '<p>x</p>',
        styleCSS: competing,
      );

      final styleCssIndex = doc.indexOf(competing);
      final overrideIndex = doc.indexOf('html, body { height: auto !important; }');

      expect(styleCssIndex, isNot(-1));
      expect(overrideIndex, greaterThan(styleCssIndex),
          reason: 'override must come AFTER styleCSS in source order');
    });
```

core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
```diff
@@ -0,0 +1,69 @@
...
+
+    test('returns true when autoAdjust=true and withBuffer meets minHeight', () {
+      // 170 + 30 = 200 >= 200
+      expect(check(reported: 170, current: 150, autoAdjust: true), isTrue);
+    });
+
+    test('returns false when autoAdjust=true and withBuffer is below minHeight', () {
+      // 100 + 30 = 130 < 200
+      expect(check(reported: 100, current: 90, autoAdjust: true), isFalse);
+    });
```
hoangdat, line 67: add test

```
test('dedup alone CANNOT stop the loop — CSS override is load-bearing', () {
      double iframe = 200;
      const buffer = 30.0;
      final history = <double>{iframe};

      for (var i = 0; i < 10; i++) {
        final scrollHeight = iframe; // body fills iframe — the bug condition
        final reported = scrollHeight + buffer;
        if (HtmlContentViewerOnWeb.shouldUpdateHeight(
          scrollHeightWithBuffer: reported,
          currentActualHeight: iframe,
          minHeight: 200,
          autoAdjust: false,
        )) {
          iframe = reported;
          history.add(iframe);
        }
      }

      expect(history.length, greaterThan(5),
          reason: 'Dart dedup does not catch the buffer-delta loop; '
              'the CSS override in generateHtmlDocument is the actual fix.');
    });
```

core/test/utils/html_utils_test.dart
dab246: added

core/test/presentation/views/html_viewer/html_content_viewer_height_stabilization_test.dart
dab246: added

hoangdat (approved):

PR #4537, opened 2026-05-20
Fix prevent notification storm on burst FCM delivery on Android

dab246:
## Context

Android users experienced notification storms after overnight device reconnection: the server
delivered a large burst of FCM events in a short window, causing hundreds of local notifications,
service overload, and device freezing.

This PR implements the client-side countermeasures from
[ADR-0073](https://github.com/linagora/tmail-flutter/blob/master/docs/adr/0073-prevent-notification-storm-android.md)
(MUST and SHOULD items), complementary to the server-side `collapse_key` mechanism
([tmail-backend#2301](https://github.com/linagora/tmail-backend/issues/2301)).

> The needs-action + Inbox heuristic (SHOULD — pending product validation) is intentionally
> excluded from this PR and will be introduced separately after product sign-off.

### 1. Cap notifications at 20 per burst

Adds `selectEmailsForNotification` on `ListPresentationEmailExtension`.
On each FCM burst, at most 20 notifications are shown — always the **newest** emails sorted
by `receivedAt` descending. Emails with no date are treated as oldest and dropped first when
the burst exceeds the cap.

### 2. Keep FCM delivery state consistent during foreground sync

`EmailChangeListener` now stores the FCM delivery state after each foreground email sync on
Android. Previously this was skipped, causing the next background wake to re-process
already-seen emails and show stale notifications.

### 3. Unify mailbox cache lookup for notification exclusion

`getExcludedMailboxesForNotification` replaces the old `getMailboxesNotPutNotifications`.
It uses **inbox presence** as a cache quality signal: if the inbox is in the local cache the
full mailbox structure is considered reliable and no network call is made; otherwise it falls
back to the network. This avoids unnecessary network fetches while still filtering out
spam/trash/sent/drafts mailboxes.

### 4. Safe notification removal

`LocalNotificationManager.removeNotification()` now catches and logs `PlatformException`
(previously surfaced in Sentry as `PlatformException: Missing type parameter` from
`removeNotificationFromCache`) instead of propagating it and breaking the notification pipeline.

### 5. FCM token retry on SERVICE_NOT_AVAILABLE

`FcmReceiver._getInitialToken()` now retries up to `MAX_COUNT_RETRY_TO_GET_FCM_TOKEN` (3)
times with exponential backoff (`1 << (attempt - 1)` seconds: 1 s → 2 s) when token
retrieval fails. Transient Play Services failures are logged as `logWarning`; only a
definitive 3-attempt failure escalates to `logError`.

## Test coverage

- `selectEmailsForNotification`: cap at 20, newest-first ordering, `null receivedAt` treated
  as oldest, custom cap, empty input, zero cap.
- `getExcludedMailboxesForNotification`: cache hit (inbox present), cache miss (network
  fallback verified with `verify(...).called(1)`), empty cache, no-inbox cache.

## Summary by CodeRabbit

* **New Features**
  * Local notification selection now limits and prioritizes emails (default cap 20) to reduce noise.

* **Improvements**
  * Mailbox selection for push favors local cache when inbox is present, otherwise falls back to network.
  * Foreground sync persists delivery state only when account identity is present.
  * Notification removal is more robust with error handling.
  * FCM token retrieval retries with exponential backoff.

* **Tests**
  * Updated tests covering mailbox selection, cache/network scenarios, and email-cap behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/tmail-flutter/pull/4537?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/features/push_notification/presentation/listener/email_change_listener.dart
+++ b/lib/features/push_notification/presentation/listener/email_change_listener.dart
@@ -302,7 +302,10 @@ class EmailChangeListener extends ChangeListener {
       return;
     }
 
-    for (var presentationEmail in emailList) {
+    final selectedEmails = emailList.selectEmailsForNotification();
+    log('EmailChangeListener::_handleLocalPushNotification(): SELECTED = ${selectedEmails.length}');
+
+    for (var presentationEmail in selectedEmails) {
       await _showLocalNotification(
         userName: userName,
         presentationEmail: presentationEmail,
--- a/lib/features/push_notification/presentation/listener/email_change_listener.dart
+++ b/lib/features/push_notification/presentation/listener/email_change_listener.dart
@@ -110,6 +110,14 @@ class EmailChangeListener extends ChangeListener {
     if (PlatformInfo.isMobile) {
       _getNewReceiveEmailFromNotificationAction(action.session, action.accountId, action.newState);
     }
+    if (PlatformInfo.isAndroid) {
+      // On Android, foreground sync must persist the new delivery state so the
+      // next background FCM push does not replay already-seen email changes.
+      final userName = action.session?.username;
+      if (userName != null) {
+        _storeEmailDeliveryStateAction(action.accountId, userName, action.newState);
+      }
+    }
   }
 
   void _onPushNotification(PushNotificationAction action) {
--- a/lib/features/push_notification/presentation/services/fcm_receiver.dart
+++ b/lib/features/push_notification/presentation/services/fcm_receiver.dart
@@ -63,7 +63,7 @@ class FcmReceiver {
       } catch (e, st) {
         if (attempt < MAX_COUNT_RETRY_TO_GET_FCM_TOKEN) {
           logWarning('FcmReceiver::_getInitialToken: attempt $attempt failed: $e');
-          await Future.delayed(Duration(seconds: attempt));
+          await Future.delayed(Duration(seconds: 1 << (attempt - 1)));
         } else {
           logError(
             'FcmReceiver::_getInitialToken: all $MAX_COUNT_RETRY_TO_GET_FCM_TOKEN attempts failed',
--- a/lib/features/push_notification/presentation/services/fcm_receiver.dart
+++ b/lib/features/push_notification/presentation/services/fcm_receiver.dart
@@ -55,18 +55,25 @@ class FcmReceiver {
   }
 
   Future<String?> _getInitialToken() async {
-    try {
-      final token = await FirebaseMessaging.instance.getToken();
-      log('FcmReceiver::_getInitialToken:token: $token');
-      return token;
-    } catch (e, st) {
-      logError(
-        'FcmReceiver::_getInitialToken:',
-        exception: e,
-        stackTrace: st,
-      );
-      return null;
+    for (var attempt = 1; attempt <= MAX_COUNT_RETRY_TO_GET_FCM_TOKEN; attempt++) {
+      try {
+        final token = await FirebaseMessaging.instance.getToken();
+        log('FcmReceiver::_getInitialToken:token: $token');
+        return token;
+      } catch (e, st) {
+        if (attempt < MAX_COUNT_RETRY_TO_GET_FCM_TOKEN) {
+          logWarning('FcmReceiver::_getInitialToken: attempt $attempt failed: $e');
+          await Future.delayed(Duration(seconds: attempt));
+        } else {
+          logError(
+            'FcmReceiver::_getInitialToken: all $MAX_COUNT_RETRY_TO_GET_FCM_TOKEN attempts failed',
+            exception: e,
+            stackTrace: st,
+          );
+        }
+      }
     }
+    return null;
   }
 
   Future _onHandleFcmToken() async {
--- a/model/lib/extensions/list_presentation_email_extension.dart
+++ b/model/lib/extensions/list_presentation_email_extension.dart
@@ -121,4 +121,16 @@ extension ListPresentationEmailExtension on List<PresentationEmail> {
       return where((email) => email.pushNotificationActivated).toList();
     }
   }
+
+  /// Selects at most [maxNotifications] emails for push notification display,
+  /// returning the newest emails first (sorted by receivedAt descending).
+  List<PresentationEmail> selectEmailsForNotification({
+    int maxNotifications = 20,
+  }) {
+    if (maxNotifications <= 0) return [];
+    final sorted = toList()
+      ..sort((a, b) => (b.receivedAt?.value ?? DateTime(0))
+          .compareTo(a.receivedAt?.value ?? DateTime(0)));
+    return sorted.take(maxNotifications).toList();
+  }
 }
--- a/test/features/fcm/fcm_repository_test.dart
+++ b/test/features/fcm/fcm_repository_test.dart
@@ -4,7 +4,6 @@ import 'package:jmap_dart_client/jmap/core/id.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
 import 'package:mockito/annotations.dart';
 import 'package:mockito/mockito.dart';
-import 'package:model/extensions/mailbox_extension.dart';
 import 'package:model/extensions/presentation_mailbox_extension.dart';
 import 'package:model/extensions/session_extension.dart';
 import 'package:model/mailbox/presentation_mailbox.dart';
@@ -40,7 +39,16 @@ void main() {
   final accountIdFixture = SessionFixtures.aliceSession.accountId;
   final userNameFixture = SessionFixtures.aliceSession.username;
 
-  group('getMailboxesNotPutNotifications_method::test', () {
+  final inboxMailbox = Mailbox(
+    id: MailboxId(Id('inbox-id')),
+    role: PresentationMailbox.roleInbox,
+  );
+  final spamMailbox = Mailbox(
+    id: MailboxId(Id('spam-id')),
+    role: PresentationMailbox.roleSpam,
+  );
+
+  group('getMailboxesForNotification:', () {
     setUp(() {
       fcmDatasource = MockFCMDatasource();
       cacheFCMDatasourceImpl = MockCacheFCMDatasourceImpl();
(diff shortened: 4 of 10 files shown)


lib/features/push_notification/presentation/listener/email_change_listener.dart
```diff
@@ -110,6 +110,12 @@ class EmailChangeListener extends ChangeListener {
     if (PlatformInfo.isMobile) {
       _getNewReceiveEmailFromNotificationAction(action.session, action.accountId, action.newState);
     }
+    if (PlatformInfo.isAndroid) {
```
hoangdat, line 113: why only Android

dab246: Because we only fix issue storm on android

model/lib/extensions/list_presentation_email_extension.dart
```diff
@@ -121,4 +122,55 @@ extension ListPresentationEmailExtension on List<PresentationEmail> {
...
+  /// Emails outside the Inbox that are not actionable are silently dropped,
+  /// capping notification count to prevent notification storms.
+  List<PresentationEmail> selectEmailsForNotification(
+    MailboxId? inboxMailboxId, {
+    int maxNotifications = 20,
+  }) {
+    if (maxNotifications <= 0) return [];
+
+    if (inboxMailboxId == null) {
+      return take(maxNotifications).toList();
```
hoangdat: should have test to ensure it is the 20 latest emails

model/lib/extensions/list_presentation_email_extension.dart
```diff
@@ -121,4 +122,55 @@ extension ListPresentationEmailExtension on List<PresentationEmail> {
...
+    }
+
+    final actionEmails = where((e) => e.hasNeedAction)
```
hoangdat: it was not verified. Need wait product team validate it

hoangdat:
some time we have

```
FirebaseException: [firebase_messaging/unknown] java.io.IOException: java.util.concurrent.ExecutionException: java.io.IOException: SERVICE_NOT_AVAILABLE
```

https://errors.cozycloud.cc/organizations/cozycloud/issues/369216/?environment=tmail-linagora&project=83&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=22

chibenwa:
What do you want me to say if FCM return 500 ?

Our app needs to be resilient when faced with these 500...

dab246:
> some time we have
> 
> ```
> FirebaseException: [firebase_messaging/unknown] java.io.IOException: java.util.concurrent.ExecutionException: java.io.IOException: SERVICE_NOT_AVAILABLE
> ```
> 
> https://errors.cozycloud.cc/organizations/cozycloud/issues/369216/?environment=tmail-linagora&project=83&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=90d&stream_index=22

Added retry with linear backoff using the existing `MAX_COUNT_RETRY_TO_GET_FCM_TOKEN = 3` constant that was defined but previously unused. On failure (including `SERVICE_NOT_AVAILABLE`), the app retries up to 3 times — waiting 1 s then 2 s between attempts — before giving up. Transient Play Services unavailability is logged as `logWarning` to reduce Sentry noise; only a definitive 3-attempt failure escalates to `logError`.

model/lib/extensions/list_presentation_email_extension.dart
dab246: Removed. The needs-action + INBOX heuristic has been stripped out entirely from this PR — it is the SHOULD item in ADR-0073 that is pending product validation. `selectEmailsForNotification` now simply sorts newest-first and caps at 20 with no mailbox or keyword filtering. The heuristic will be re-introduced in a separate PR once the product team signs off on the selection criteria.

model/lib/extensions/list_presentation_email_extension.dart
dab246: Fixed. The old `take(maxNotifications)` without sorting was returning the first N elements in input order, not the newest. `selectEmailsForNotification` now always sorts by `receivedAt` descending before capping, so the result is guaranteed to be the newest emails regardless of input order. Test `caps at maxNotifications and selects the newest emails` explicitly verifies that the correct 20 emails are selected from a 50-email list.

test/features/fcm/fcm_repository_test.dart
```diff
@@ -40,7 +39,16 @@ void main() {
...
 
-  group('getMailboxesNotPutNotifications_method::test', () {
+  final inboxMailbox = Mailbox(
+    id: MailboxId(Id('inbox-id')),
+    role: PresentationMailbox.roleInbox,
+  );
+  final spamMailbox = Mailbox(
+    id: MailboxId(Id('spam-id')),
+    role: PresentationMailbox.roleSpam,
+  );
```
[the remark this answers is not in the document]
dab246: No need because pushNotificationDeactivated is a getter of PresentationMailbox, and it is set to true when mailboxes have the role of ` bool get pushNotificationDeactivated => isOutbox || isSent || isDrafts || isTrash || isSpam;`. And the Spam mailbox we created for testing has satisfied the condition and will have pushNotificationDeactivated set to true.

model/lib/extensions/list_presentation_email_extension.dart
```diff
@@ -121,4 +121,16 @@ extension ListPresentationEmailExtension on List<PresentationEmail> {
...
+  /// returning the newest emails first (sorted by receivedAt descending).
+  List<PresentationEmail> selectEmailsForNotification({
+    int maxNotifications = 20,
+  }) {
+    if (maxNotifications <= 0) return [];
+    final sorted = toList()
+      ..sort((a, b) => (b.receivedAt?.value ?? DateTime(0))
+          .compareTo(a.receivedAt?.value ?? DateTime(0)));
+    return sorted.take(maxNotifications).toList();
+  }
```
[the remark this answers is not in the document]
dab246: Because the limitation of this PR is only the limit on the number of the most recent emails received, and does not handle the classification of emails as (1) actionable emails first, then (2) newest Inbox emails.

test/features/fcm/fcm_repository_test.dart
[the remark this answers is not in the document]
dab246: So, Is this comment done?

model/lib/extensions/list_presentation_email_extension.dart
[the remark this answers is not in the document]
dab246: So, Is this comment done?

test/features/fcm/fcm_repository_test.dart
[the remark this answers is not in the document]
dab246: So, can we resolve this comment?

lib/features/push_notification/presentation/services/fcm_receiver.dart
```diff
@@ -55,18 +55,25 @@ class FcmReceiver {
...
-      logError(
-        'FcmReceiver::_getInitialToken:',
-        exception: e,
-        stackTrace: st,
-      );
-      return null;
+    for (var attempt = 1; attempt <= MAX_COUNT_RETRY_TO_GET_FCM_TOKEN; attempt++) {
+      try {
+        final token = await FirebaseMessaging.instance.getToken();
+        log('FcmReceiver::_getInitialToken:token: $token');
```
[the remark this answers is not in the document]
dab246: log() maps to Level.info, which is handled by the console handler only. SentryEventHandler exclusively accepts Level.error and Level.critical (see sentry_event_handler.dart:26), and SentryBreadcrumbHandler accepts Level.trace only. This token never reaches Sentry, the privacy rule applies to logError/logCritical, not log().

hoangdat (approved):

tddang-linagora (approved):

lib/features/push_notification/presentation/services/fcm_receiver.dart
```diff
@@ -55,18 +55,25 @@ class FcmReceiver {
...
+          logWarning('FcmReceiver::_getInitialToken: attempt $attempt failed: $e');
+          await Future.delayed(Duration(seconds: attempt));
+        } else {
+          logError(
+            'FcmReceiver::_getInitialToken: all $MAX_COUNT_RETRY_TO_GET_FCM_TOKEN attempts failed',
+            exception: e,
+            stackTrace: st,
+          );
+        }
+      }
```
Crash--, line 74: Can this error explain why sometimes we do not receive push notifications?

lib/features/push_notification/presentation/services/fcm_receiver.dart
```diff
@@ -55,18 +55,25 @@ class FcmReceiver {
...
-      return null;
+    for (var attempt = 1; attempt <= MAX_COUNT_RETRY_TO_GET_FCM_TOKEN; attempt++) {
+      try {
+        final token = await FirebaseMessaging.instance.getToken();
+        log('FcmReceiver::_getInitialToken:token: $token');
+        return token;
+      } catch (e, st) {
+        if (attempt < MAX_COUNT_RETRY_TO_GET_FCM_TOKEN) {
+          logWarning('FcmReceiver::_getInitialToken: attempt $attempt failed: $e');
+          await Future.delayed(Duration(seconds: attempt));
```
chibenwa, line 66: The delay is the count of seconds in ms?
Also this is bad, a general practice is to get exponential retries: `2^n` seconds
Otherwise even 10 retries completes uner 45s wich is not enough.
Adding an element of random also help avoiding retry hurds.

Cool read: https://brooker.co.za/blog/2015/03/21/backoff.html 
https://brooker.co.za/blog/2022/08/11/backoff.html

lib/features/push_notification/presentation/services/fcm_receiver.dart
hoangdat: yes, as the thing we found in Sentry indicate that

dab246: > Can this error explain why sometimes we do not receive push notifications?

**Yes, this is a direct cause of missing push notifications.**

When all retry attempts are exhausted and `_getInitialToken()` returns `null`, the failure propagates silently through the entire token registration chain:

```
_getInitialToken() → null
  └─ FcmService.handleToken(null)
       └─ fcmTokenStreamController.add(null)
            └─ FcmTokenController.onFcmTokenChanged(null)
                 └─ if (newToken != null) { ... }  ← guard skips registration entirely
```

The device never registers its FCM token with the JMAP server, so the server has no push endpoint to deliver notifications to. All push notifications are silently dropped until the next app cold start triggers a fresh token fetch.

This `logError` is the observable signal that we are in that failed state. The fix in this PR adds retry logic to reduce how often this path is reached

lib/features/push_notification/presentation/services/fcm_receiver.dart
dab246: > The delay is the count of seconds in ms?
> 

The original code you commented on used `Duration(seconds: attempt)` — the unit is **seconds**, not milliseconds. 

> Also this is bad, a general practice is to get exponential retries: 2^n seconds
> 

That has already been corrected in the follow-up commit. The current code is:

```dart
await Future.delayed(Duration(seconds: 1 << (attempt - 1)));
// attempt=1 → 2^0 = 1s
// attempt=2 → 2^1 = 2s
// attempt=3 → final attempt, no more delay
```

This is true exponential backoff (powers of 2), in **seconds**.

> Otherwise even 10 retries completes uner 45s wich is not enough.
> Adding an element of random also help avoiding retry hurds.
> Cool read: https://brooker.co.za/blog/2015/03/21/backoff.html
> https://brooker.co.za/blog/2022/08/11/backoff.html
> 

Your point about jitter is valid and not yet addressed. Without jitter, all devices that fail simultaneously (e.g., during a GMS service hiccup) will retry at the exact same intervals, causing a thundering herd against FCM. We should add a follow-up to apply full jitter:

```dart
final baseDelayMs = (1 << (attempt - 1)) * 1000; // exponential base in ms
final jitteredMs = Random().nextInt(baseDelayMs + 1); // full jitter: [0, base]
await Future.delayed(Duration(milliseconds: jitteredMs));
```

### On the total retry window being too short

Also valid. With `MAX_COUNT_RETRY_TO_GET_FCM_TOKEN = 3`, the current total window is at most **3 seconds** (1 s + 2 s). `SERVICE_NOT_AVAILABLE` from GMS can last **minutes**, so this window is insufficient for that scenario.

A more resilient configuration would be `MAX_COUNT_RETRY_TO_GET_FCM_TOKEN = 7` with a delay cap:

| Attempt | Delay (with jitter, capped at 64 s) |
|---------|--------------------------------------|
| 1       | [0, 1] s                             |
| 2       | [0, 2] s                             |
| 3       | [0, 4] s                             |
| 4       | [0, 8] s                             |
| 5       | [0, 16] s                            |
| 6       | [0, 32] s                            |
| 7       | final attempt                        |

This gives a total retry window of up to ~63 seconds — enough to survive most transient GMS outages — and jitter prevents coordinated retry storms. We can open a follow-up ticket to address this.

lib/features/push_notification/presentation/services/fcm_receiver.dart
Crash--: Great to hear thanks you! We’ll be able to explain why we had issues and how we fix that to some of our users.

PR #4558, opened 2026-05-25
Fix: Calendar event HTML description rendered as plain text

dab246:
## Problem

Calendar event (ICS) descriptions that contain HTML markup were displayed as raw, unrendered text in the email viewer. Instead of seeing formatted content, users saw the literal HTML tags — e.g. a description like:

```
<p>We are delighted to extend a warm invitation...</p>
```

was shown as the visible characters `<p>We are delighted to extend a warm invitation...</p>` instead of a rendered paragraph.

## Root Cause

The transformer pipeline in `_parseCalendarEventAction` was treating HTML descriptions as plain text:

```dart
// Before — plain-text pipeline
SanitizeAutolinkHtmlTransformers(),        // escapeHtml: true → HTML-escapes everything
SanitizePlainTextHtmlOutputTransformer(),  // strips all tags except <a>
NewLineTransformer(),
```

`SanitizeAutolinkHtmlTransformers` with `escapeHtml: true` calls `htmlEscape.convert()` on the raw text before any DOM parsing, converting `<` and `>` to `&lt;`/`&gt;`. This is correct for plain-text emails, but ICS descriptions are typically HTML — so the markup was rendered as visible escaped characters instead of formatted content.

## Fix

Introduce `AutolinkTextNodeTransformer` — a DOM transformer that walks the already-parsed DOM to linkify bare URLs and email addresses in text nodes — and replace the plain-text pipeline with an HTML-aware named factory:

```dart
// After — HTML-aware pipeline (TransformConfiguration.forCalendarEvent)
// Text transformers:
StandardizeHtmlSanitizingTransformers(),  // sanitizes HTML (strips dangerous tags/attrs)
NewLineTransformer(),
// DOM transformers:
AutolinkTextNodeTransformer(),            // linkifies bare URLs/emails in text nodes
SanitizeHyperLinkTagInHtmlTransformer(),  // sanitizes href, adds target="_blank" + rel="noreferrer"
```

`AutolinkTextNodeTransformer` walks the parsed DOM tree rather than using a regex over raw text, so tag boundaries come from the document structure. This eliminates any ReDoS risk and avoids double-linkifying content inside existing `<a>`, `<script>`, `<style>`, or `<textarea>` elements.

`SanitizeAutolinkFilter` was also refactored alongside this change: the `escapeHtml` flag was removed (it always HTML-escapes text elements now), and `_linkifyOption`/`_linkifier` were made `static` to avoid re-allocating per call.

## Security

XSS is blocked by `StandardizeHtmlSanitizingTransformers`, which strips `<script>` tags entirely, removes event-handler attributes (`onerror`, `onclick`, …), and strips `javascript:` URIs from `href` attributes. This is verified by the updated tests in `single_email_controller_test.dart` and `calendar_event_description_transform_test.dart`.

Note: the XSS neutralisation behaviour changed. The old pipeline rendered `<script>alert(1)</script>` as visible escaped text (`&lt;script&gt;…`). The new pipeline removes the element entirely. Both outcomes are safe; the new one produces cleaner output.

## Summary by CodeRabbit

* **New Features**
  * Calendar event descriptions now auto-linkify plain URLs and email addresses into safe, clickable links.

* **Improvements**
  * Safer, standardized HTML sanitization for calendar event content.
  * Preserves rich HTML formatting (links, lists, basic styling) while stripping unsafe attributes and scripts.

* **Tests**
  * Expanded test coverage for calendar description rendering, sanitization, and linkification behaviors.

[Review Change Stack](https://app.coderabbit.ai/change-stack/linagora/tmail-flutter/pull/4558?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


lib/features/email/presentation/controller/single_email_controller.dart
```diff
@@ -1257,8 +1256,8 @@ class SingleEmailController extends BaseController with AppLoaderMixin {
       blobIds,
       TransformConfiguration.create(
         customTextTransformers: const [
-          SanitizeAutolinkHtmlTransformers(),
-          SanitizePlainTextHtmlOutputTransformer(),
+          SanitizeAutolinkUnescapeHtmlTransformer(),
+          StandardizeHtmlSanitizingTransformers(),
           NewLineTransformer(),
```
hoangdat: Why not have a specific method to create transformer and use it in the test also?

chibenwa:
@coderabbitai review this please

lib/features/email/presentation/controller/single_email_controller.dart
dab246: fixed

test/features/email/data/local/calendar_event_description_transform_test.dart
```diff
@@ -83,25 +79,41 @@ void main() {
...
-        const description = 'Status: a < b && b > 0';
+      test('SHOULD preserve HTML entities in ICS description', () async {
+        const description = 'Status: a &lt; b &amp;&amp; b &gt; 0';
         final result = await transformDescription(description);
         expect(result, contains('&lt;'));
         expect(result, contains('&gt;'));
       });
     });
 
+    group('HTML passthrough — HTML descriptions rendered as formatted content', () {
```
hoangdat, line 88: how about the case 

```dart
    test('SHOULD preserve an existing HTML link destination and text', () async {
        // Real-world calendar invites embed join links as HTML <a href>.
        const description = '<a href="https://example.com/meeting">Join Zoom</a>';
        final result = await transformDescription(description);
        // The original destination URL must survive intact.
        expect(result, contains('href="https://example.com/meeting"'));
        // The visible link text must survive.
        expect(result, contains('Join Zoom'));
        // The markup must not be mangled into garbage by the linkifier.
        expect(result, isNot(contains('<ahref')));
      });
```

hoangdat, line 88: more coverage test

```dart
      test('SHOULD NOT corrupt an HTML link whose visible text is also a URL', () async {
        // href and anchor text both contain a URL — the linkifier must not
        // double-process the attribute and break the tag.
        const description =
            '<a href="https://example.com/a">https://example.com/a</a>';
        final result = await transformDescription(description);
        expect(result, contains('href="https://example.com/a"'));
        expect(result, isNot(contains('<ahref')));
      });

      test('SHOULD preserve a URL sitting in an <img> src attribute without mangling text', () async {
        // URL lives in the src attribute, not body text. <img> may be stripped
        // by the sanitizer, but the surrounding description must stay intact.
        const description =
            '<img src="https://cdn.example.com/a.png"> Team photo';
        final result = await transformDescription(description);
        expect(result, contains('Team photo'));
        expect(result, isNot(contains('<ahref')));
      });

      test('SHOULD autolink a bare URL that appears inside an HTML element', () async {
        // Positive guard: a URL in text content (not an attribute) must still
        // become a clickable link, even when wrapped in an element.
        const description =
            '<p>Join the call at https://meet.example.com/room/9</p>';
        final result = await transformDescription(description);
        expect(result, contains('href="https://meet.example.com/room/9"'));
        expect(result, contains('<p>'));
      });
```

dab246: Fixed

dab246: Added

core/lib/presentation/utils/html_transformer/sanitize_autolink_filter.dart
```diff
@@ -8,59 +8,81 @@ class SanitizeAutolinkFilter {
...
-  final _linkifier = <Linkifier>[
+  static final _linkifier = <Linkifier>[
     const EmailLinkifier(),
-    const UrlLinkifier()
+    const UrlLinkifier(),
   ];
 
+  // Matches a complete HTML tag, correctly handling quoted attribute values
+  // that may contain the > character (e.g. title="a > b").
+  static final _htmlTagPattern = RegExp(r'''<(?:[^>"']*|"[^"]*"|'[^']*')*>''');
```
hoangdat: it is ReDoS, should to find other approach to handle it. How about using DOM transformer?

dab246: Oke, we will use new transformer for auto link

hoangdat (approved):

PR #4566, opened 2026-05-28
ADR-0092 Upgrade flutter_riverpod to 3.x

dab246:
## Summary by CodeRabbit

* **Documentation**
  * Added an Architecture Decision Record for upgrading to Riverpod v3, including migration guidance, testing strategy updates, and runtime considerations.
  * Documents migration constraints, required verification steps (including Android), updated unit-test approaches, and follow-up items for linters/codegen and remaining call sites to address in future ADRs.

PR #4573, opened 2026-06-01
Upgrade `flutter_riverpod` to 3.x and adopt `@riverpod` codegen per ADR-0092

dab246:
## Description

Implements [ADR-0092](docs/adr/0092-upgrade-flutter-riverpod-3.md).

## Related

- ADR-0092: https://github.com/linagora/tmail-flutter/pull/4566

## Dependency changes

| Package | Before | After | Notes |
|---|---|---|---|
| `flutter_riverpod` | `^2.6.1` | `^3.3.1` | Resolves to `3.3.1` |
| `riverpod_annotation` | — | `^4.0.2` | Stable, runtime only |
| `riverpod_generator` | — | `4.0.4-dev.3` | Dev dep, build-time only |
| `freezed_annotation` | transitive `^2.4.4` | override `^3.0.0` | Annotation-only, safe to override |

## Code changes

| Commit | What changed |
|---|---|
| `main.dart` | Wrap `GetMaterialApp` with `UncontrolledProviderScope(container: appProviderContainer)` — shares container between widget tree and GetX layer |
| `LocalSettingsNotifier` | `StateNotifier` + manual `StateNotifierProvider` → `@Riverpod(keepAlive: true)` Notifier. Provider name `localSettingsProvider` unchanged; 4 call sites unaffected |
| Composer cache provider chain | Delete `composer_cache_providers.dart`. Move each provider inline to its class file with `@Riverpod(keepAlive: true)` and `ref.watch` |
| `ComposerAutoSaveNotifier` | `StateNotifier` + `StateNotifierProvider.family` → `@Riverpod(keepAlive: true)` Notifier family. `mounted` replaced by `_mounted` via `ref.onDispose`. Interactors resolved as `late final` in `build()` |
| Tests | `ComposerAutoSaveNotifier` tests switch from direct instantiation to `ProviderContainer` with overrides |

### Composer cache provider chain (after)

| File | Provider | Depends on |
|---|---|---|
| `composer_hive_cache_client.dart` | `composerHiveCacheClientProvider` | — |
| `cache_exception_thrower.dart` | `cacheExceptionThrowerProvider` | — |
| `composer_persistent_cache_datasource_impl.dart` | `composerCacheDatasourceProvider` | `composerHiveCacheClientProvider`, `cacheExceptionThrowerProvider` |
| `composer_cache_repository_impl.dart` | `composerCacheRepositoryProvider` | `composerCacheDatasourceProvider` |
| `resolve_composer_cache_for_restore_interactor.dart` | `resolveComposerCacheForRestoreProvider` | `composerCacheRepositoryProvider` |
| `remove_all_composer_cache_interactor.dart` | `removeAllComposerCacheProvider` | `composerCacheRepositoryProvider` |
| `mark_composer_cache_clean_close_interactor.dart` | `markComposerLocalCacheCleanCloseProvider` | `composerCacheRepositoryProvider` |

## What is NOT in this PR

`appProviderContainer` is frozen — no new call sites. The 6 existing call sites below remain and will be removed in future ADRs as each controller migrates to `ConsumerWidget`.

| File | Operation |
|---|---|
| `LocalSettingsService` | `.read(localSettingsProvider.notifier).update()` |
| `PreferencesController` | `.read(localSettingsProvider.notifier).update()` |
| `ThreadController` | `.read(localSettingsProvider)` + `.listen(...)` |
| `SearchEmailController` | `.read(localSettingsProvider)` |
| `HandleMobileAutoSaveExtension` | `.read(composerAutoSaveProvider(id).notifier)` + `.invalidate(...)` |
| `HandleComposerRestoreOnMobileExtension` | `.read(resolveComposerCacheForRestoreProvider)` |

## Release Notes

* **Refactor**
  * Modernized app state management architecture to use annotation-driven provider patterns for improved maintainability and scalability.
  * Reorganized composer cache and local settings management for better code organization.

* **Chores**
  * Updated Riverpod dependencies to latest stable versions.
  * Integrated Riverpod at the application level for enhanced provider management.


pubspec.yaml
```diff
@@ -270,14 +270,18 @@ dependencies:
 
   sentry_dio: 9.8.0
 
-  flutter_riverpod: ^2.6.1
+  flutter_riverpod: ^3.3.1
+
+  riverpod_annotation: ^4.0.2
```
tddang-linagora, line 275: Should this still be here if it already exist in `dependency_overrides`?

lib/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart
```diff
@@ -20,3 +24,7 @@ class RemoveAllComposerCacheInteractor {
     }
   }
 }
+
+@Riverpod(keepAlive: true)
```
hoangdat: I dont think a good idea to keep riverpod in domain layer should we move it to presentation layer. WDYT?

lib/features/mailbox_dashboard/domain/usecases/mark_composer_cache_clean_close_interactor.dart
```diff
@@ -46,3 +50,7 @@ class MarkComposerCacheCleanCloseInteractor {
     }
   }
 }
+
+@Riverpod(keepAlive: true)
```
hoangdat: idem

lib/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart
hoangdat: why we should have `keepAlive` here?

pubspec.yaml
dab246: `pubspec.yaml` do not remove `riverpod_annotation` from `dependency_overrides`. The reason is that `dependencies` needs it so `@riverpod` can be imported and used, while `dependency_overrides` is needed to pin a version that is compatible with `riverpod_generator 4.0.4-dev.3` across all workspace members. Please add a comment to explain this clearly so reviewers do not wonder why `riverpod_annotation` is declared in both places.

lib/main/exceptions/thrower/cache_exception_thrower.dart
```diff
@@ -20,4 +23,7 @@ class CacheExceptionThrower extends ExceptionThrower {
     }
     throw error;
   }
-}
\ No newline at end of file
+}
+
+@Riverpod(keepAlive: true)
+CacheExceptionThrower cacheExceptionThrower(Ref ref) => CacheExceptionThrower();
```
hoangdat: what is the problem if we keep 2 instances of CacheExceptionThrower in the app? one in GetX, one in Riverpod?

lib/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart
dab246: Agreed, and the reason is more specific than just "Riverpod in domain". The `@Riverpod` annotation itself is a lightweight factory and does not couple the interactor class to any framework. The real problem is the **import it forces on the domain file**:

```dart
import 'package:tmail_ui_user/.../data/repository/composer_cache_repository_impl.dart';
```

This import exists solely to reference `composerCacheRepositoryProvider`, which lives in the data layer. That creates a **domain → data dependency**, a direct violation of the Clean Architecture dependency rule.

Fixed: moved all three interactor provider functions (`removeAllComposerCache`, `markComposerLocalCacheCleanClose`, `resolveComposerCacheForRestore`) to a new file:

```
lib/features/mailbox_dashboard/presentation/providers/composer_cache_interactor_providers.dart
```

The interactor classes themselves remain unchanged in the domain layer. The data-layer import is removed from all three domain files.

lib/features/mailbox_dashboard/domain/usecases/mark_composer_cache_clean_close_interactor.dart
dab246: idem

lib/features/mailbox_dashboard/domain/usecases/remove_all_composer_cache_interactor.dart
dab246: Not needed. `keepAlive` is justified for the lower layers (`HiveCacheClient`, `Datasource`, `Repository`) because they manage I/O resources. Interactors are stateless wrappers with no resource to keep alive. Replaced with `@riverpod` (auto-dispose).

lib/main/exceptions/thrower/cache_exception_thrower.dart
dab246: No functional problem. `CacheExceptionThrower` is completely stateless. Bridging to `Get.find<>()` was considered but rejected: it would introduce a GetX dependency into the Riverpod chain, making it non-self-contained and harder to test. Keeping the separate Riverpod instance is the better trade-off.

lib/features/caching/clients/composer_hive_cache_client.dart
```diff
@@ -67,3 +70,7 @@ class ComposerHiveCacheClient extends HiveCacheClient<String> {
   ) =>
       deleteItem(_entryKey(accountId, userName, composerId));
 }
+
+@Riverpod(keepAlive: true)
+ComposerHiveCacheClient composerHiveCacheClient(Ref ref) =>
```
9clg6: WDYT to split the provider and the class ? 

Keeping the class here and move the provider to `lib/features/caching/providers/composer_hive_cache_client.provider.dart `for exemple ?

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
```diff
@@ -39,19 +42,23 @@ class ComposerAutoSaveState extends Equatable {
...
+  late final ResolveComposerCacheForRestoreInteractor _resolveInteractor;
+  late final RemoveAllComposerCacheInteractor _removeInteractor;
 
-  ComposerAutoSaveNotifier(
-    this._resolveInteractor,
-    this._removeInteractor,
-  ) : super(const ComposerAutoSaveState());
+  @override
+  ComposerAutoSaveState build(String composerId) {
+    _resolveInteractor = ref.read(resolveComposerCacheForRestoreProvider);
```
9clg6: 1) Could we use ref.watch instead of ref.read in build method ? 
["Do not use Ref.read as a mean to "optimize" your code by avoiding Ref.watch. This will make your code more brittle, as changes in your provider's behavior could cause your UI to be out of sync with the provider's state."](https://riverpod.dev/docs/concepts2/refs#interacting-with-the-providers-state-inside-user-interactions)

2) Do we have `riverpod_lint` ? Because normally because it should be flagged `avoid_manual_providers_as_generated_provider_dependency`.

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
```diff
@@ -39,19 +42,23 @@ class ComposerAutoSaveState extends Equatable {
...
+    _removeInteractor = ref.read(removeAllComposerCacheProvider);
```
9clg6: idem

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
```diff
@@ -97,17 +104,11 @@ class ComposerAutoSaveNotifier extends StateNotifier<ComposerAutoSaveState> {
     result.fold(
       (failure) => log('ComposerAutoSaveNotifier::clearCache: failure=${failure.runtimeType}'),
       (_) {
-        state = state.copyWith(hasRecoverableSnapshot: false);
+        if (mounted) {
```
9clg6: Personally I prefer early returns as 
`if(!mounted) return;`
`state = state.copyWith(hasRecoverableSnapshot: false);`

lib/features/mailbox_dashboard/data/repository/composer_cache_repository_impl.dart
```diff
@@ -46,5 +50,8 @@ class ComposerCacheRepositoryImpl extends ComposerCacheRepository {
         userName,
         composerId,
       );
-
 }
+
+@Riverpod(keepAlive: true)
+ComposerCacheRepository composerCacheRepository(Ref ref) =>
```
9clg6: idem

9clg6 (commented):
Well done ! 
Some comments but nothing huge

chibenwa:
Glad to see some consensual changes ;-)

I'm interested in @dab246 @tddang-linagora @hoangdat opinion:
 - does this change brings in your opinion a better code organisation / easier mechanics ? It's ok to say no or I do not know.
 - can it be generalized to other parts of the code base easily? Or will we have difficulties at some point? Or will some GetIT usage be harder to address? IE what's your feedback regarding ease-of-migration ?

jbrnzl (dismissed):
Clean! Nice work 🎉

lib/features/caching/clients/composer_hive_cache_client.dart
hoangdat: I still prefer keep provider into `presentation` layer, Riverpod only belong to presentation, `domain` or `data` layer not depend on Riverpod, as the current way we work with GetX.

dab246: Done. Removed the provider from the class file entirely — data layer is now Riverpod-free.
All data-layer providers (`composerHiveCacheClient`, `composerCacheDatasource`, `composerCacheRepository`) are consolidated in the presentation layer at:
`mailbox_dashboard/presentation/providers/composer_cache_data_providers.dart`

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
dab246: > Could we use ref.watch instead of ref.read in build method?

We removed the interactor fields from `build()` entirely.

Using `ref.watch` in `build()` would be incorrect: if the watched provider is invalidated, `build()` re-runs and the Notifier state resets to default (`hasRecoverableSnapshot = false`, etc.) — a silent data-loss bug for an in-progress composer session.
> Could we use ref.watch instead of ref.read in build method?

The interactors are one-shot command executors, not reactive state. The idiomatic Riverpod pattern is:
- `build()` → returns only initial state, no provider reads
- async methods → `ref.read(provider)` in event handlers (explicitly correct per Riverpod docs)

```dart
@override
ComposerAutoSaveState build(String composerId) => const ComposerAutoSaveState();

Future<void> clearCache(...) async {
  final result = await ref.read(removeAllComposerCacheProvider).execute(...);
  ...
}
```

This also eliminates the `avoid_manual_providers_as_generated_provider_dependency` lint warning entirely.

> Do we have riverpod_lint?

We didn't. Added `riverpod_lint: ^3.1.3` to `dev_dependencies` and registered it as an analyzer plugin in `analysis_options.yaml`. No `custom_lint` needed — `riverpod_lint` 3.x integrates directly via `analysis_server_plugin`. Future violations like `avoid_manual_providers_as_generated_provider_dependency` will be caught at IDE level.

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
dab246: fixed

lib/features/composer/presentation/providers/composer_auto_save_notifier.dart
dab246: Done.

```dart
(_) {
  if (!mounted) return;
  state = state.copyWith(hasRecoverableSnapshot: false);
  log('ComposerAutoSaveNotifier::clearCache: cleared');
},
```

lib/features/mailbox_dashboard/data/repository/composer_cache_repository_impl.dart
dab246: Done. Same approach as comment 1 — repository class is now Riverpod-free, provider lives in presentation layer.

hoangdat:
> does this change brings in your opinion a better code organisation / easier mechanics ? It's ok to say no or I do not know.

for me it not much more different, but we can move forward to have first step to apply riverpod. With these, we hope we can easily to develop new features with Riverpod than continue to depend on our ghost controller. But tbh, the very big challenge: all our UI depend on GetX (Obx), to completely apply Riverpod, we still have huge work to do.

> can it be generalized to other parts of the code base easily? Or will we have difficulties at some point? Or will some GetIT usage be harder to address? IE what's your feedback regarding ease-of-migration ?

We are trying to generalized the way to apply Riverpod. it is in progress, the first function we apply is https://github.com/linagora/tmail-flutter/pull/4576. 

- Move all the logic of clean all email from controller into riverpod provider 
- Add new logic of clean Trash's sub folder in riverpod also 
- Adapt the centralized error handling to riverpod 

We will discover more and more during the way we apply Riverpod, hope we can soon generalize it to other functions

chibenwa:
Thanks for the honest feedback

tddang-linagora (approved):

Issue #4479, opened 2026-05-01
Clear trash must remove subfolders


chibenwa:
Given I have a trash subfolder 

When I clear trash

Subfolder must also be deleted 

Not the case today

hoangdat:
- [ ] ADR how to not depend on controller::consumeState

PR #4576, opened 2026-06-02
TF-4479 Clear trash subfolders with Riverpod state management

dab246:
## Issue

#4479 

## Summary

- Add `DeleteTrashSubfoldersNotifier` (`@riverpod`, sealed state hierarchy) to handle async deletion of trash subfolders
- Add `MailboxProviderListenerWidget` at mailbox view scope to listen to notifier state (toast + refresh)
- Wire action to context menu, long-press, and empty-trash banner (thread view + dashboard web)
- Add l10n strings for success / partial-success / failure outcomes
- Add unit tests for `DeleteTrashSubfoldersNotifier`

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Enhanced trash folder clearing with support for removing nested subfolders
  * Improved confirmation dialogs for trash operations with better visual feedback
  * Added progress tracking during folder cleanup operations
  * Localized messages for subfolder deletion outcomes (success, partial, failure)

* **Bug Fixes**
  * Fixed trash banner visibility to correctly detect folder contents including subfolders
  * Improved handling of partial folder deletion scenarios


lib/features/mailbox/presentation/extensions/handle_delete_trash_subfolders_listener_extension.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  void _onDeleteFailed(BuildContext context, Object? exception) {
+    getBinding<AppToast>()?.showToastErrorMessage(
+      context,
+      AppLocalizations.of(context).clearTrashSubfoldersFailed,
+    );
+    _handleUrgentException(exception);
+  }
+
+  void _handleUrgentException(Object? exception) {
```
hoangdat: now the error handling is tighly couple to `base_controller`, how about to separate it into a specific instance then invoke it in everywhere we need it? WDYT?

dab246: Since many controllers depend on it, separating it would require significant refactoring and retesting of the logic to ensure no side effects. We should consider creating a separate PR for this.

lib/features/mailbox/presentation/widgets/mailbox_provider_listener_widget.dart
```diff
@@ -0,0 +1,18 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:tmail_ui_user/features/mailbox/presentation/extensions/handle_delete_trash_subfolders_listener_extension.dart';
+
+class MailboxProviderListenerWidget extends ConsumerWidget {
```
hoangdat: I still not very confident with this idea: create a widget + extension to handle the side effect in UI. What is the best practice in riverpod to handle the logic in UI when a state change? 
how about another provider listen to the logic provider and react on it? Fx: toast have a provider listen to the state, error handler also....? WDYT?

lib/features/mailbox/presentation/extensions/handle_delete_trash_subfolders_listener_extension.dart
hoangdat: I know, how about create an instance for it, and first use it here?

hoangdat:
How about to completely move all the logic of clear trash into a specific provider?

lib/features/mailbox/presentation/extensions/handle_delete_trash_subfolders_listener_extension.dart
dab246: Agree. I will extract a standalone `UrgentExceptionHandler` class, register it in the bindings, and use it in `MailboxProviderListenerWidget` first. This keeps the scope of the change limited to this PR without affecting other controllers. A follow-up PR can then progressively replace existing `BaseController.handleUrgentException()` calls.

lib/features/mailbox/presentation/widgets/mailbox_provider_listener_widget.dart
dab246: `ref.listen` inside `ConsumerWidget.build` is the Riverpod recommended pattern for handling UI side effects. Toasts require a BuildContext, which pure providers cannot access. Even with a dedicated "toast provider," a widget layer would still be required to actually display the toast, adding another level of indirection without reducing complexity. The listener widget serves exactly that purpose: it provides a clean bridge between provider state changes and UI actions.

dab246:
> How about to completely move all the logic of clear trash into a specific provider?

The execution logic is already fully contained in EmptyTrashNotifier. What remains in the widget are UI-only side effects: showing toasts (which require BuildContext) and dispatching GetX actions (which require the controller). These responsibilities cannot live in a pure Riverpod provider. The current separation, business logic in the provider and UI reactions in the widget, is both intentional and the correct architectural boundary.

lib/features/mailbox_dashboard/presentation/providers/empty_folder_provider.dart
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  Future<_ClearEmailResult> _emptyTrashByEmailDeletion(
+    Session session,
+    AccountId accountId,
+    PresentationMailbox trashMailbox,
+  ) async {
+    final progressCtrl = StreamController<Either<Failure, Success>>();
+    try {
+      final result = await getBinding<EmptyTrashFolderInteractor>()
```
hoangdat, line 216: we should address this point of coexisted DI in documents

lib/features/mailbox_dashboard/presentation/providers/empty_trash_provider.dart
```diff
@@ -0,0 +1,244 @@
...
+      return _ClearEmailFailure(exception: e);
+    }
+  }
+
+  Future<_ClearEmailResult> _emptyTrashByEmailDeletion(
+    Session session,
+    AccountId accountId,
+    PresentationMailbox trashMailbox,
+  ) async {
+    final progressCtrl = StreamController<Either<Failure, Success>>();
```
hoangdat: why we need it?

dab246: In the old logic, we needed a StreamController to update the progress status bar when cleaning the mailbox.

lib/features/mailbox/presentation/widgets/mailbox_provider_listener_widget.dart
```diff
@@ -0,0 +1,155 @@
...
+import 'package:tmail_ui_user/features/mailbox/presentation/action/mailbox_ui_action.dart';
+import 'package:tmail_ui_user/features/base/urgent_exception_handler.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/delete_emails_in_mailbox_extension.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/providers/empty_trash_provider.dart';
+import 'package:tmail_ui_user/main/error/capability_validator.dart';
+import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
+import 'package:tmail_ui_user/main/routes/route_navigation.dart';
+
+class MailboxProviderListenerWidget extends ConsumerWidget {
```
hoangdat: in mobile, this widget maybe not available when we click it in the banner. it may not work

dab246: Fixed

dab246:
@9clg6  Please take the time to review this PR for us. Thanks

lib/features/mailbox_dashboard/presentation/riverpod_widgets/mailbox_dashboard_provider_listener_widget.dart
```diff
@@ -0,0 +1,49 @@
...
+  });
+
+  @override
+  ConsumerState<MailboxDashboardProviderListenerWidget> createState() =>
+      _MailboxDashboardProviderListenerWidgetState();
+}
+
+class _MailboxDashboardProviderListenerWidgetState
+    extends ConsumerState<MailboxDashboardProviderListenerWidget> {
+  late final List<DashboardProviderListenerDelegate> _delegates;
```
hoangdat, line 26: So we maybe keep the providers alive through the app because the delegate always live? am i right?

lib/features/mailbox_dashboard/presentation/extensions/execute_empty_trash_extension.dart
```diff
@@ -0,0 +1,48 @@
...
+        appLocalizations.empty_trash_dialog_message,
+        appLocalizations.delete,
+        cancelTitle: appLocalizations.cancel,
+        onCloseButtonAction: popBack,
+        onConfirmAction: () {
+          popBack();
+          emptyTrashFolderAction(trashMailbox: trashMailbox);
+        },
+      );
+    }
```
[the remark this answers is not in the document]
dab246: We don't need to handle this here, because this function is called when the banner is clicked. And the logic for displaying the banner has already checked for subfolders and emails in the trash.

lib/features/mailbox_dashboard/presentation/riverpod_widgets/mailbox_dashboard_provider_listener_widget.dart
dab246: Yes, two providers, two roles:

- `emptyFolderRequestedProvider(tag)`:  a stream derived from the GetX controller’s `onEmptyFolderRequested` event. It is kept alive for the full widget lifecycle to ensure no incoming requests are missed.
- `emptyFolderProvider(mailboxId)`: a notifier responsible for executing the actual empty-folder operation. It is opened on demand per operation and closed on terminal states (Success / Failure), allowing it to be auto-disposed immediately after each run.

lib/features/mailbox_dashboard/presentation/providers/empty_folder_provider.dart
```diff
@@ -0,0 +1,263 @@
...
+
+  Future<void> execute(
+    Session session,
+    AccountId accountId,
+    PresentationMailbox mailbox,
+    List<MailboxId> childIds,
+    bool useJmapClear,
+  ) async {
+    if (_isExecuting) return;
+    state = const EmptyFolderLoading();
```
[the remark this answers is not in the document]
dab246: Is this comment is done?

lib/features/mailbox_dashboard/presentation/providers/empty_folder_provider.dart
```diff
@@ -0,0 +1,263 @@
...
+    AccountId accountId,
+    List<MailboxId> childIds,
+  ) async {
+    if (childIds.isEmpty) return SubfoldersDeleteStatus.none;
+    final result = await _deleteSubfolders(session, accountId, childIds);
+    return switch (result) {
+      _SubfoldersAllDeleted() => SubfoldersDeleteStatus.allDeleted,
+      _SubfoldersSomeDeleted() => SubfoldersDeleteStatus.someDeleted,
+      _SubfoldersDeleteFailed() => SubfoldersDeleteStatus.failed,
+    };
```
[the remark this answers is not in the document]
dab246: Is this comment is done?

lib/features/mailbox_dashboard/presentation/riverpod_widgets/mailbox_dashboard_provider_listener_widget.dart
```diff
@@ -0,0 +1,49 @@
...
+      delegate.dispose();
+    }
+    super.dispose();
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    for (final delegate in _delegates) {
+      delegate.listen(context, ref);
+    }
```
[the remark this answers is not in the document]
dab246: Is this comment is done?

dab246:
> ⚠️ Outside diff range comments (1)
> > lib/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart (1)> `1891-1893`: _⚠️ Potential issue_ | _🔴 Critical_ | _⚡ Quick win_
> > > **Resolve the selected team trash before falling back to personal trash.**
> > > This branch only accepts `selectedMailbox` when `isTrash == true`. But the same controller enables the empty-trash flow for `isTrashTeamMailbox` on Lines 2417-2430, and `DeleteActionType.all` reaches this method without passing `trashMailbox` on Line 1875. From a team-trash screen, this can fall through to `mapDefaultMailboxIdByRole[roleTrash]` and purge the personal trash instead of the selected team trash.

> 🧹 Nitpick comments (1)
> > test/features/mailbox/repository/mailbox_respository_test.dart (1)> `289-301`: _⚡ Quick win_
> > > **Strengthen this test to validate non-empty error passthrough.**
> > > The setup at Line 293-294 returns an empty network error map, so Line 300 only proves “empty stays empty.” It doesn’t validate the stated behavior (“original error map is returned”) when the map is non-empty.
> > > 


Fixed

tddang-linagora (dismissed):

hoangdat:
hi @jbrnzl @C. it is in better shape, can you take a look and give us the feedback. Thank you

lib/features/base/urgent_exception_handler.dart
```diff
@@ -0,0 +1,7 @@
+import 'package:core/presentation/state/failure.dart';
+
+abstract class UrgentExceptionHandler {
```
hoangdat: IMO, it is not in use, why not create a separate instance for it and use directly with `EmptyFolderProviderListenerDelegate`?

lib/features/mailbox_dashboard/presentation/riverpod_widgets/mailbox_dashboard_provider_listener_widget.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/delegates/dashboard_provider_listener_delegate.dart';
+
+class MailboxDashboardProviderListenerWidget extends ConsumerStatefulWidget {
```
hoangdat, line 5: with the idea of delegate, in the future `controller` will use as a place to keep the data, variable, ...?

lib/features/mailbox_dashboard/presentation/delegates/empty_folder_provider_listener_delegate.dart
```diff
@@ -0,0 +1,225 @@
...
+    dashboardController.syncViewStateMailboxActionProgress(
+      newState: Right(UIState.idle),
+    );
+  }
+
+  void _showEmptyFolderSuccessToast(
+    BuildContext context, {
+    required SubfoldersDeleteStatus subfoldersStatus,
+  }) {
+    final appToast = getBinding<AppToast>();
```
hoangdat: how about we should minimize using getBinding to use riverpod more? 
create AppToast with riverpod the use it here? wdyt?

lib/features/mailbox_dashboard/presentation/extensions/execute_empty_trash_extension.dart
```diff
@@ -0,0 +1,48 @@
...
+import 'package:flutter/material.dart';
+import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
+import 'package:model/mailbox/presentation_mailbox.dart';
+import 'package:tmail_ui_user/features/base/mixin/message_dialog_action_manager.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
+import 'package:tmail_ui_user/main/routes/route_navigation.dart';
+
+extension ExecuteEmptyTrashExtension on MailboxDashBoardController {
+  List<MailboxId> childMailboxIds(PresentationMailbox parent) => mapMailboxById
```
hoangdat: should move it as a extension of Map, no need to make extension of controller bigger.

lib/features/mailbox_dashboard/presentation/extensions/execute_empty_trash_extension.dart
```diff
@@ -0,0 +1,48 @@
...
+      .values
+      .where((m) => m.parentId == parent.id)
+      .map((m) => m.id)
+      .toList();
+
+  void requestEmptyTrash(
```
hoangdat: why need to define as a method in extension? we should find the way to reduce extension as much as possible.

how about to separate it as a dialog and have the callback? we can eliminate this extension?

lib/features/mailbox_dashboard/presentation/delegates/empty_folder_provider_listener_delegate.dart
```diff
@@ -0,0 +1,225 @@
...
+
+  void _executeEmptyFolder(
+    BuildContext context,
+    WidgetRef ref,
+    PresentationMailbox mailbox,
+    MailboxDashBoardController dashboardController,
+  ) {
+    final session = dashboardController.sessionCurrent;
+    final accountId = dashboardController.accountId.value;
+    if (session == null || accountId == null) return;
```
hoangdat: how about toast showing here?

lib/features/base/urgent_exception_handler.dart
dab246: It's still in use. It's used in the listener, and Riverpod uses it via GetX binding.

lib/features/mailbox_dashboard/presentation/riverpod_widgets/mailbox_dashboard_provider_listener_widget.dart
dab246: No. The controller will all be moved to Riverpod later.

lib/features/mailbox_dashboard/presentation/delegates/empty_folder_provider_listener_delegate.dart
dab246: AppToast is independent of any classes. Therefore, we can convert it to RiverPod for use. Fixed

lib/features/mailbox_dashboard/presentation/extensions/execute_empty_trash_extension.dart
dab246: reasonable, fixed

lib/features/mailbox_dashboard/presentation/extensions/execute_empty_trash_extension.dart
dab246: reasonable, fixed

lib/features/mailbox_dashboard/presentation/delegates/empty_folder_provider_listener_delegate.dart
dab246: toast is added

chibenwa:
Why `app_localizations.dart` goes down in code quality ?

I saw this in a couple of code scene PRs and it gets me curious...

dab246:
> Why `app_localizations.dart` goes down in code quality ?
> 
> I saw this in a couple of code scene PRs and it gets me curious...

I think we discussed this in a previous PR, perhaps you've forgotten (https://github.com/linagora/tmail-flutter/pull/4553#issuecomment-4533956109). Basically, it's because app_localizations.dart is the only file containing declarations for all Strings in the app, so the number of lines of code in this file is currently exceeding the Code Scene limit, leading to the error `Lines of code in a single file`.

lib/features/mailbox_dashboard/presentation/delegates/empty_folder_provider_listener_delegate.dart
```diff
@@ -0,0 +1,232 @@
...
+      dashboardController.handleDeleteEmailsInMailbox(
+        emailIds: clearedEmailIds,
+        affectedMailboxId: mailboxId,
+      );
+    }
+  }
+
+  void _handleUrgentException(
+    Object? exception,
+    MailboxDashBoardController dashboardController,
```
hoangdat: how about using interface instead `UrgentExceptionHandler`?

tddang-linagora (dismissed):

hoangdat (approved):

Issue #4528, opened 2026-05-16
Scribe: raw markdown is displayed and html seems to be doubled escaped


chibenwa:
Given

<img width="705" height="927" alt="Image" src="https://github.com/user-attachments/assets/ca4a36dd-970f-4516-96a1-00e00b760413" />

Then

<img width="1045" height="927" alt="Image" src="https://github.com/user-attachments/assets/b7b7b3af-0c43-44a5-a146-c87fa2756c00" />

It seems `'` is replaced by `&#39;`

Also it would be nice to show me rich text and not raw markdown

Crash--:
I just try if this was added by the llm or not. So yes the issue seems to come from the front-end code. 

<img width="590" height="233" alt="Image" src="https://github.com/user-attachments/assets/2066b17b-b590-4f6c-ad81-9cb46115a489" />

zatteo:
I will give a try next week.

If we start to render rich text/markdown, we need to be more careful about prompt injection because it can introduce misleading links and thing like these.

hoangdat:
I tested it worked in canary now @poupotte

PR #4578, opened 2026-06-02
fix: Do not double escape text in Scribe

zatteo:
After merging our final PRs, we ended with double escape of text which led to some characters being replaced by their HTML code like ' was replaced by &#39;

See https://github.com/linagora/tmail-flutter/issues/4528

## Summary by CodeRabbit

* **Refactor**
  * Simplified text processing in the AI Scribe composer for improved HTML conversion accuracy when inserting or updating text in the editor.

### DIFF
--- a/lib/features/composer/presentation/extensions/ai_scribe/handle_ai_scribe_in_composer_extension.dart
+++ b/lib/features/composer/presentation/extensions/ai_scribe/handle_ai_scribe_in_composer_extension.dart
@@ -42,8 +42,7 @@ extension HandleAiScribeInComposerExtension on ComposerController {
 
   Future<void> insertTextInEditor(String text) async {    
     try {
-      final escapedText = StringConvert.escapeTextContent(text);
-      final htmlContent = StringConvert.convertTextContentToHtmlContent(escapedText);
+      final htmlContent = StringConvert.convertTextContentToHtmlContent(text);
 
       if (PlatformInfo.isWeb) {
         await richTextWebController?.editorController.evaluateJavascriptWeb(
@@ -62,8 +61,7 @@ extension HandleAiScribeInComposerExtension on ComposerController {
 
   Future<void> setTextInEditor(String text) async {
     try {
-      final escapedText = StringConvert.escapeTextContent(text);
-      final htmlContent = StringConvert.convertTextContentToHtmlContent(escapedText);
+      final htmlContent = StringConvert.convertTextContentToHtmlContent(text);
       
       if (PlatformInfo.isWeb) {
         richTextWebController?.editorController.setText(htmlContent);


hoangdat:
@coderabbit review

chibenwa:
Oh thanks a lot for this !

chibenwa:
While we are at it if the text generation could skip user signature that would be a boomer! 

Is that the case here? Or does it needs a prompt refinement ?

lib/features/composer/presentation/extensions/ai_scribe/handle_ai_scribe_in_composer_extension.dart
```diff
@@ -42,8 +42,7 @@ extension HandleAiScribeInComposerExtension on ComposerController {
 
   Future<void> insertTextInEditor(String text) async {    
     try {
-      final escapedText = StringConvert.escapeTextContent(text);
-      final htmlContent = StringConvert.convertTextContentToHtmlContent(escapedText);
+      final htmlContent = StringConvert.convertTextContentToHtmlContent(text);
```
chibenwa, line 45: I'd propose we keep consider the LLM as an untrusted source of data and that we do sanitize it (as per HTML sanitization) prior integreting it into the resulting content.

chibenwa (commented):
The LLM seems to rely on markdown for its overall formatting eg code.

Should we ...
 - ask it to format straight away in HTML ?
 - Content markdown markup in HTML ourselves ?

(follow up issue maybe)

chibenwa:
Screenshot from 2026-06-02 09-29-07

Screenshot from 2026-06-02 09-29-20

Screenshot from 2026-06-02 09-29-26

paultranvan:
> The LLM seems to rely on markdown for its overall formatting eg code.
> 
> Should we ...
> 
>     * ask it to format straight away in HTML ?
> 
>     * Content markdown markup in HTML ourselves ?
> 
> 
> (follow up issue maybe)

I think we should stick to markdown generation by the LLM: it is much more reliable than HTML generation.
So I would advise a markdown -> HTML conversion made by tmail directly

lib/features/composer/presentation/extensions/ai_scribe/handle_ai_scribe_in_composer_extension.dart
zatteo: What do you mean? `StringConvert.escapeTextContent` is already called inside `StringConvert.convertTextContentToHtmlContent` actually

zatteo:
> The LLM seems to rely on markdown for its overall formatting eg code.
> 
> Should we ...
> 
>     * ask it to format straight away in HTML ?
> 
>     * Content markdown markup in HTML ourselves ?
> 
> 
> (follow up issue maybe)

This should be thought about and done in a follow up issue. We must especially be cautious about prompt injection as markdown can render link and image.

zatteo:
> While we are at it if the text generation could skip user signature that would be a boomer!
> 
> Is that the case here? Or does it needs a prompt refinement ?

What do you mean? Do not take it into account in prompt data source? Or don't touch the signature during insertion?

chibenwa:
in some scribe interaction, scribe did set again my signature in the model response leading me to a message with 2 signature.

Again that is very certainly another issue ;-)

zatteo:
> in some scribe interaction, scribe did set again my signature in the model response leading me to a message with 2 signature.
> 
> Again that is very certainly another issue ;-)

Agree, another issue, but maybe the answer is both: ensure signature is not passed to the model (I am not sure now)  and strengthen the prompt

hoangdat (approved):

tddang-linagora (approved):

dab246 (approved):

Issue #4598, opened 2026-06-10
[Sentry] Web - Refresh token failed in some cases


dab246:
### Desc

Some log show the refresh token failed. And I found in some cases we have status code 400 in refresh token.

https://errors.cozycloud.cc/organizations/cozycloud/issues/335570/?environment=tmail-linagora&project=83&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=3

```
ArgumentError
Invalid argument(s): Failed to get token: [error: token_failed, description: invalid_grant]

Crashed in non-app: org-dartlang-sdk:///dart-sdk/lib/_internal/js_runtime/lib/js_helper.dart in wrapException

… he/git/flutter_appauth_web-0952346bf781bd87b93b8c61619f81a264be52db/lib/flutter_appauth_web.dart in AppAuthWebPlugin.requestToken at line 171:7
In App

    final Map<String, dynamic> jsonResponse = jsonDecode(response.body);
    if (response.statusCode != 200) {
      print(jsonResponse["error"].toString());
      throw ArgumentError(_TOKEN_ERROR_MESSAGE_FORMAT.replaceAll("%2", jsonResponse["error"]?.toString() ?? response.reasonPhrase ?? "Unknow {snip}
```

### DoD
- [ ] understand what is the problem
- [ ] fix and increase the capability of handling refreshToken failed

PR #4604, opened 2026-06-11
TF-4598 Add log to tracing Sentry [Web] -  Refresh token invalid grant

dab246:
## Issue

#4598

## Problem

On web, when the OIDC token endpoint rejects a refresh request, `flutter_appauth_web` throws `ArgumentError` instead of a `DioException`. The original fix added a blanket `on ArgumentError catch` that always triggered logout — but this caused two regressions:

1. **Sentry crashes** — `super.onError(originalError.copyWith(error: ArgumentError(...)), handler)` propagated a raw `ArgumentError` through the Dio pipeline; `RemoteExceptionThrower` did not recognise it, causing unhandled exceptions in Sentry:
   - `ArgumentError: Failed to get token: [error: token_failed, description: invalid_request]`
   - `ArgumentError: Invalid or corrupted pad block`

2. **Over-eager logout** — non-standard codes such as `token_failed` have no confirmed HTTP status; logging out without a known 400/401 is incorrect.

3. **No tracing context** — Sentry events arrived with no `auth_error_type`, `oauth_error_code`, or `oauth_error_description`, making triage impossible.

## Root Cause

```
flutter_appauth_web.requestToken()         ← throws ArgumentError("token_failed/invalid_grant")
  _refreshTokenThenRetry                   ← on DioException: NO MATCH → falls through
    _handleRefreshErrorOnWeb               ← blanket ArgumentError → always logout
      super.onError(copyWith(ArgumentError)) ← raw ArgumentError propagated into Dio
        RemoteExceptionThrower             ← unrecognised type → CRASH → Sentry untagged error
```

### Classification strategy

Introduced `WebRefreshTokenErrorClassifier` (SRP — RFC 6749 §5.2 knowledge isolated from Dio infrastructure):

| Error type | `isServerRejection` | Action |
|---|---|---|
| `AccessTokenInvalidException` | `true` | logout |
| `DioException` with 400 or 401 response | `true` | logout |
| `ArgumentError` with RFC 6749 standard code (`invalid_grant`, `invalid_client`, `invalid_request`, `invalid_scope`, `unauthorized_client`) | `true` | logout |
| `ArgumentError` with non-standard code (e.g. `token_failed`) | `false` | keep session + `logError` to Sentry with `oauth_error_code` / `oauth_error_description` for tracing |
| `DioException` with 5xx / no response | `false` | keep session, `logWarning` only |

RFC 6749 §5.2 reference: https://datatracker.ietf.org/doc/html/rfc6749#section-5.2

### Sentry trace map (after fix)

**Logout path** (RFC 6749 standard code or 400/401):

| # | Location | Level | `auth_error_type` |
|---|---|---|---|
| 1 | `AuthorizationInterceptors._handleRefreshErrorOnWeb` | `logError` ✅ | `web_server_rejected_refresh` |
| 2 | `BaseController.handleRefreshTokenFailedException` | `logError` ✅ | `refresh_token_failed` |

**Keep-session path** (non-standard code — tracing only):

| # | Location | Level | extras |
|---|---|---|---|
| 1 | `AuthorizationInterceptors._handleRefreshErrorOnWeb` | `logError` ✅ | `oauth_error_code`, `oauth_error_description`, `auth_error_type=web_token_unknown_error` |

## Summary by CodeRabbit

* **New Features**
  * Added web refresh-token error classification to decide when to clear session state versus keep it, with consistent enriched error reporting.

* **Bug Fixes**
  * Improved web refresh failure handling: more accurate forced-logout behavior for specific OAuth/RFC server rejections, while transient/other errors now preserve session as expected.
  * Refined logging so refresh failures are reported correctly and avoid duplicate error events.

* **Tests**
  * Updated and expanded web interceptor and classifier tests to match the new logout/session-preservation rules and logging expectations.

### DIFF
--- a/lib/features/base/base_controller.dart
+++ b/lib/features/base/base_controller.dart
@@ -285,13 +285,9 @@ abstract class BaseController extends GetxController
   }
 
   void handleRefreshTokenFailedException() {
-    logError(
+    log(
       '$runtimeType::handleRefreshTokenFailedException: '
-      'auth_error_type=refresh_token_failed | will_logout=true | '
       'hasComposer=${twakeAppManager.hasComposer}',
-      exception: RefreshTokenFailedException(),
-      stackTrace: StackTrace.current,
-      extras: {'auth_error_type': 'refresh_token_failed'},
       webConsoleEnabled: true,
     );
     if (twakeAppManager.hasComposer) {
--- a/lib/features/base/base_controller.dart
+++ b/lib/features/base/base_controller.dart
@@ -201,9 +201,11 @@ abstract class BaseController extends GetxController
     } else if (exception is RefreshTokenFailedException) {
       handleRefreshTokenFailedException();
     } else {
-      logWarning(
+      logError(
         '$runtimeType::handleUrgentExceptionOnWeb(): NO branch matched — '
         'will NOT navigate to login. exception=${exception.runtimeType}',
+        exception: exception,
+        stackTrace: StackTrace.current,
         webConsoleEnabled: true,
       );
     }
@@ -285,8 +287,13 @@ abstract class BaseController extends GetxController
   }
 
   void handleRefreshTokenFailedException() {
-    log(
-      '$runtimeType::handleRefreshTokenFailedException: hasComposer=${twakeAppManager.hasComposer}',
+    logError(
+      '$runtimeType::handleRefreshTokenFailedException: '
+      'auth_error_type=refresh_token_failed | will_logout=true | '
+      'hasComposer=${twakeAppManager.hasComposer}',
+      exception: RefreshTokenFailedException(),
+      stackTrace: StackTrace.current,
+      extras: {'auth_error_type': 'refresh_token_failed'},
       webConsoleEnabled: true,
     );
     if (twakeAppManager.hasComposer) {
--- a/lib/features/base/base_controller.dart
+++ b/lib/features/base/base_controller.dart
@@ -201,11 +201,9 @@ abstract class BaseController extends GetxController
     } else if (exception is RefreshTokenFailedException) {
       handleRefreshTokenFailedException();
     } else {
-      logError(
+      logWarning(
         '$runtimeType::handleUrgentExceptionOnWeb(): NO branch matched — '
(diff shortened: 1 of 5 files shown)


hoangdat:
please also address the issue also https://errors.cozycloud.cc/organizations/cozycloud/issues/401700

```
Invalid argument(s): Invalid or corrupted pad block
```

lib/features/base/base_controller.dart
```diff
@@ -201,9 +201,11 @@ abstract class BaseController extends GetxController
     } else if (exception is RefreshTokenFailedException) {
       handleRefreshTokenFailedException();
     } else {
-      logWarning(
+      logError(
```
hoangdat: why need logError here?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -356,6 +359,26 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         ),
         handler,
       );
+    } on ArgumentError catch (e, st) {
```
hoangdat: IMO not a good idea, we should go to logout if and only if it is 400 or 401.

lib/features/base/base_controller.dart
dab246: Revert to logWarning

lib/features/login/data/network/interceptors/authorization_interceptors.dart
dab246: Removed. Handling log placement for tracing

dab246:
> please also address the issue also https://errors.cozycloud.cc/organizations/cozycloud/issues/401700
> 
> ```
> Invalid argument(s): Invalid or corrupted pad block
> ```

Added

hoangdat:
what is the root cause of this log: https://errors.cozycloud.cc/organizations/cozycloud/issues/375740
```
ClientException
ClientException: Failed to fetch, uri=https://sso.linagora.com/oauth2/token
```

hoangdat (approved):

tddang-linagora (approved):

PR #4605, opened 2026-06-11
add more log and try to write token before prune safety for update token 

hoangdat:
## Summary by CodeRabbit

* **Bug Fixes**
  * Safer account and token persistence: writes now occur before pruning to avoid temporary cache loss during updates and refreshes; redundant deletes during refresh flows removed.
  * Improved error logging for encryption/credential retrieval: emits structured classified errors (with reason and stack trace) and stops logging raw encryption key values.

* **Tests**
  * Added/updated tests covering cache prune/idempotency scenarios and token-persist failure handling in refresh flows.

### DIFF
--- a/lib/features/caching/config/hive_cache_config.dart
+++ b/lib/features/caching/config/hive_cache_config.dart
@@ -115,6 +115,7 @@ class HiveCacheConfig {
       logError(
         'HiveCacheConfig::getEncryptionKey(): '
         'encryption_key_unavailable=true | reason=cache_manager_binding_missing',
+        stackTrace: StackTrace.current,
       );
       return null;
     }
@@ -128,6 +129,7 @@ class HiveCacheConfig {
       logError(
         'HiveCacheConfig::getEncryptionKey(): '
         'encryption_key_unavailable=true | reason=key_cache_empty',
+        stackTrace: StackTrace.current,
       );
       return null;
     }
--- a/lib/features/caching/config/hive_cache_config.dart
+++ b/lib/features/caching/config/hive_cache_config.dart
@@ -109,17 +109,26 @@ class HiveCacheConfig {
   Future<Uint8List?> getEncryptionKey() async {
     final encryptionKeyCacheManager = getBinding<EncryptionKeyCacheManager>() ?? getBinding<EncryptionKeyCacheManager>(tag: BindingTag.isolateTag);
     if (encryptionKeyCacheManager == null) {
-      log('HiveCacheConfig::getEncryptionKey(): encryptionKeyCacheManager not found');
+      // No key → an encrypted box (incl. the OIDC token box) opens without a
+      // cipher and cannot be decrypted, surfacing later as a "no token" startup
+      // logout. Emit remotely so this silent root cause is observable.
+      logError(
+        'HiveCacheConfig::getEncryptionKey(): '
+        'encryption_key_unavailable=true | reason=cache_manager_binding_missing',
+      );
       return null;
     }
     var encryptionKeyCache = await encryptionKeyCacheManager.getEncryptionKeyStored();
 
     if (encryptionKeyCache != null) {
-      final encryptionKey = encryptionKeyCache.value;
-      log('HiveCacheConfig::getEncryptionKey(): encryptionKey: $encryptionKey');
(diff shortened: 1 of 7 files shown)


lib/features/caching/config/hive_cache_config.dart
```diff
@@ -109,17 +109,26 @@ class HiveCacheConfig {
...
     var encryptionKeyCache = await encryptionKeyCacheManager.getEncryptionKeyStored();
 
     if (encryptionKeyCache != null) {
-      final encryptionKey = encryptionKeyCache.value;
-      log('HiveCacheConfig::getEncryptionKey(): encryptionKey: $encryptionKey');
+      log('HiveCacheConfig::getEncryptionKey(): encryption key resolved');
       final encryptionKeyDecode = base64Decode(encryptionKeyCache.value);
       return encryptionKeyDecode;
     } else {
+      logError(
```
dab246, line 129: If use logError should add exception & stacktrace param for it

lib/features/login/data/local/account_cache_manager.dart
dab246, line 28: We should write unit test for this when has new logic

tddang-linagora (dismissed):

lib/features/login/data/local/token_oidc_cache_manager.dart
```diff
@@ -23,15 +23,23 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
...
   }
 
+  /// Removes every token except [keepKey] so the box keeps exactly one entry
+  Future<void> _removeStaleTokens({required String keepKey}) async {
+    final allItems = await _tokenOidcCacheClient.getMapItems();
+    final staleKeys = allItems.keys.where((key) => key != keepKey).toList();
+    if (staleKeys.isNotEmpty) {
+      log('TokenOidcCacheManager::_removeStaleTokens(): removing ${staleKeys.length} stale token(s)');
+      await _tokenOidcCacheClient.deleteMultipleItem(staleKeys);
+    }
```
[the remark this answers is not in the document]
hoangdat: Non-atomic — agreed. But both `persistOneTokenOidc` callers are already serialized (interceptor is QueuedInterceptorsWrapper; login is a separate phase), so concurrent different-key persists don't  occur. A per-instance mutex also can't cover the FCM background isolate. Empty-box was the kill-mid-write case —already fixed by  write-before-prune.

lib/features/login/data/local/account_cache_manager.dart
```diff
@@ -30,18 +30,35 @@ class AccountCacheManager extends CacheManagerInteraction {
...
+        newAccountCache.id,
+        ...otherAccounts.map((account) => account.id),
+      };
+      final staleKeys = allAccounts
+        .map((account) => account.id)
+        .where((id) => !keepKeys.contains(id))
+        .toList();
+      if (staleKeys.isNotEmpty) {
+        await _accountCacheClient.deleteMultipleItem(staleKeys);
       }
```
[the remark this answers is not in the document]
hoangdat: Non-atomic — agreed. But setCurrentAccount callers are login/sign-in/sign-up flows (sequential, one-time) plus the interceptor refresh (QueuedInterceptorsWrapper, serialized) — they don't run concurrently. A per-instance lock also can't cover the FCM background isolate. Split-brain needs concurrent calls that don't occur; write-before-insert already prevents the empty/no-selected state.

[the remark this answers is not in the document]
hoangdat: @coderabbit What do you think?

tddang-linagora (dismissed):

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -481,8 +481,9 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
   Future<PersonalAccount> _updateCurrentAccount({required TokenOIDC tokenOIDC}) async {
```
dab246, line 481: write unit test for this

hoangdat: added

dab246 (approved):
lgtm

Issue #4606, opened 2026-06-12
[Android] Cannot download attachment


hoangdat:
### Desc:

https://github.com/user-attachments/assets/b1b6552f-1ec2-4a27-8cf7-6e6c6992e54b

#### Sentry 

https://errors.cozycloud.cc/organizations/cozycloud/issues/404500

```
ArgumentError
RangeError (length): Invalid value: Not in inclusive range 0..1948: 1949
split_config.arm64_v8a.apk
+0x695a5c
AesEngine.decryptBlock (aes_engine.dart)
split_config.arm64_v8a.apk
+0x693df8
AesCbcPkcs7.decrypt (aes_cbc_pkcs7.dart:81)
split_config.arm64_v8a.apk
+0x693c40
HiveAesCipher.decrypt (hive_aes_cipher.dart:46)
split_config.arm64_v8a.apk
+0x691980
BinaryReaderImpl.readEncrypted (binary_reader_impl.dart:390)
split_config.arm64_v8a.apk
+0x691980
IsolatedBoxBaseImpl._readValue (isolated_box_impl_vm.dart:184)
split_config.arm64_v8a.apk
+0xf639d8
IsolatedBoxBaseImpl.get (isolated_box_impl_vm.dart:159)
<unknown>
<unknown> (<asynchronous suspension>)
split_config.arm64_v8a.apk
+0xdf997c
TokenOidcCacheManager.getTokenOidc (token_oidc_cache_manager.dart:16)
<unknown>
<unknown> (<asynchronous suspension>)
split_config.arm64_v8a.apk
+0xdf98f0
AuthenticationOIDCDataSourceImpl.getStoredTokenOIDC.<T> (authentication_oidc_datasource_impl.dart:81)
<unknown>
<unknown> (<asynchronous suspension>)
split_config.arm64_v8a.apk
+0xe171e0
ExportAttachmentInteractor.execute (export_attachment_interactor.dart:40)
<unknown>
<unknown> (<asynchronous suspension>)
split_config.arm64_v8a.apk
+0x64ba38
BaseController.onError (base_controller.dart:147)
```

Cc: @chibenwa

hoangdat:
please try and test in version v0.29.6

PR #4608, opened 2026-06-12
TF-4606 Fix cannot download attachment

dab246:
## Problem

#4606

Two symptoms were reported:

1. **Attachment download fails** — tapping "Download" shows a "download failed" toast.
2. **App restart logs user out** — after the failure, killing and reopening the app bounces straight to login.

Both stem from the same root cause: the Hive box storing the OIDC token becomes corrupted (cross-isolate race between the main app and the FCM background isolate), and the app had no recovery mechanism.

---

## Root Cause

```mermaid
flowchart TD
    A[FCM isolate + main app write to same Hive file concurrently]
    A --> B[Partial write — AES block boundary violated]
    B --> C[getStoredTokenOIDC throws RangeError]
    C --> D1[Download fails — toast shown]
    C --> D2[Box never cleared]
    D2 --> E[App restart → same error → logout loop]
```

**Download fails:** `ExportAttachmentInteractor` calls `getStoredTokenOIDC`. The `RangeError` propagates to `DownloadController` → "download failed" toast. Box is left corrupt.

**Logout loop on restart:** `GetStoredTokenOidcInteractor` reads the same corrupt box → `GetStoredTokenOidcFailure` → `isNotSignedIn()` → `clearDataAndGoToLoginPage()`. Every restart repeats.

**Secondary — `persistOneTokenOidc` propagates corruption:** When Dio refreshes an expired token, `_removeStaleTokens` calls `getMapItems()` → `toMap()`. A corrupt neighbour entry causes `toMap()` to crash before pruning completes, propagating `UnknownRemoteException` to the in-flight request.

**Minor — wrong argument to failure handlers:** `DownloadController` was passing the `ExportAttachmentFailure` wrapper instead of `failure.exception`, so `CancelDownloadFileException` was never matched — cancelled downloads showed a spurious "download failed" toast.

---

## Fix

Three defence layers applied in depth.

### Layer 1 — `TokenOidcCacheManager` clears the corrupted box on first detection

**`getTokenOidc`:** catches any non-`AppBaseException` (AES decryption error, `RangeError`, etc.), logs to Sentry, clears the box, throws `NotFoundStoredTokenException`.

**`persistOneTokenOidc`:** writes the new token first (crash-safe), then prunes stale entries. If pruning crashes on a corrupt neighbour entry, recovers by clearing the box and re-inserting only the new valid token.

→ Eliminates the logout loop on restart. Does **not** prevent the current-session download failure on its own.

---

### Layer 2 — Export interactors fall back to the in-memory token and repair storage

After Layer 1 clears the box, the valid OIDC token is still held in memory by `AuthorizationInterceptors._token`. Layer 2 uses it.

```mermaid
flowchart TD
    A[ExportAttachmentInteractor needs OIDC token] --> B{getStoredTokenOIDC accountCacheKey}
    B -- success --> C[Use storage token]
    B -- throws --> D{fallbackToken present?}
    D -- no --> E[Rethrow → download fails]
    D -- yes --> F[Use in-memory token from AuthorizationInterceptors]
    F --> G[persistTokenOIDCAt accountCacheKey — fire & forget]
    G -- success --> H[Box healed at correct key — token survives app restart]
    G -- fails --> I[logError to Sentry]
    F --> J[Download proceeds normally]
```

`ExportAttachmentRequest` and `ExportAllAttachmentsRequest` carry an optional `fallbackToken` field, populated from `AuthorizationInterceptors.currentToken` at the call site. Both interactors resolve the token via `_getTokenOidc(accountCacheKey, fallbackToken: request.fallbackToken)`.

**Why `persistTokenOIDCAt(accountCacheKey, ...)` matters:** `GetStoredTokenOidcInteractor` looks up the token at startup using `personalAccount.id` (`AccountCache.id`) as the Hive key. `persistTokenOIDC` stores under `fallbackToken.tokenIdHash`. These can diverge if a prior 401 token refresh updated `_token` in memory but `_updateCurrentAccount` failed before `setCurrentAccount` ran — leaving `_token.tokenIdHash ≠ currentAccount.id`. Repairing under `fallbackToken.tokenIdHash` would write to a key startup never reads, causing logout. `persistTokenOIDCAt` stores at the explicit `accountCacheKey` that startup will use.

`persistTokenOIDCAt` is fire-and-forget (not awaited) so the download is never blocked by a storage repair. On repair failure, `logError` with the exception is sent to Sentry.

→ Download succeeds in the current session even when storage is corrupt.  
→ Box is healed at the correct key so the token survives an app restart without requiring re-login.

---

### Layer 3 — `DownloadController` passes the correct exception to failure handlers

```dart
// Before
exportAttachmentFailureAction(failure);
// After
exportAttachmentFailureAction(failure.exception);
```

→ `CancelDownloadFileException` check works correctly; cancelled downloads no longer show a "download failed" toast.

---

## Outcome

| Scenario | Before | After |
|---|---|---|
| Box corrupt + download | Fails with toast, box left corrupt | **Succeeds** using in-memory token, box auto-repaired |
| Box corrupt + app restart | Logout loop (every restart) | **No logout** — token re-persisted at `accountCacheKey` |
| `_token` hash diverges from `currentAccount.id` + restart | Logout | **No logout** — repair writes at the key startup reads |
| Box corrupt + token refresh | `persistOneTokenOidc` throws → request fails | Token refresh succeeds, box recovered |
| User cancels download | Spurious "download failed" toast | Silent cancel (correct) |

---

## Summary by CodeRabbit

* **New Features**
  * Request-based export flows for single and all-attachments downloads with optional fallback token support.
  * New request models for export operations.
  * New explicit keyed token-persistence API.

* **Enhancements**
  * Improved token-cache resilience: detects corruption, clears/rebuilds cache, and forces clean re-authentication when needed.
  * Public accessor to read current in-memory OIDC token.
  * Export failures now forward the underlying exception for clearer reporting.

* **Tests**
  * Expanded coverage and new test helpers/stubs for cache recovery, fallback-token repair, and export flows.

### DIFF
--- a/lib/features/login/data/local/token_oidc_cache_manager.dart
+++ b/lib/features/login/data/local/token_oidc_cache_manager.dart
@@ -90,7 +90,12 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
     try {
       await _tokenOidcCacheClient.clearAllData();
     } catch (e) {
-      logWarning('TokenOidcCacheManager::_safelyClearBox(): failed to clear box | $e');
+      logWarning('TokenOidcCacheManager::_safelyClearBox(): clear failed, deleting box from disk | $e');
+      try {
+        await _tokenOidcCacheClient.deleteBox();
+      } catch (e2) {
+        logWarning('TokenOidcCacheManager::_safelyClearBox(): deleteBox also failed | $e2');
+      }
     }
   }
 
--- a/lib/features/login/data/local/token_oidc_cache_manager.dart
+++ b/lib/features/login/data/local/token_oidc_cache_manager.dart
@@ -68,7 +68,16 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
         stackTrace: stackTrace,
       );
       await _safelyClearBox();
-      await _tokenOidcCacheClient.insertItem(key, tokenOIDC.toTokenOidcCache());
+      try {
+        await _tokenOidcCacheClient.insertItem(key, tokenOIDC.toTokenOidcCache());
+      } catch (insertError, insertStackTrace) {
+        logError(
+          'TokenOidcCacheManager::_persistAtKey(): '
+          're-insert after clear also failed | error_type=${insertError.runtimeType}',
+          exception: insertError,
+          stackTrace: insertStackTrace,
+        );
+      }
     }
   }
 
--- a/lib/features/login/data/local/token_oidc_cache_manager.dart
+++ b/lib/features/login/data/local/token_oidc_cache_manager.dart
@@ -1,3 +1,4 @@
+import 'package:core/domain/exceptions/app_base_exception.dart';
 import 'package:core/utils/app_logger.dart';
 import 'package:model/oidc/token_oidc.dart';
 import 'package:tmail_ui_user/features/caching/clients/token_oidc_cache_client.dart';
@@ -13,22 +14,62 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
 
   Future<TokenOIDC> getTokenOidc(String tokenIdHash) async {
     log('TokenOidcCacheManager::getTokenOidc(): tokenIdHash: $tokenIdHash');
-    final tokenCache = await _tokenOidcCacheClient.getItem(tokenIdHash);
-    log('TokenOidcCacheManager::getTokenOidc(): tokenCache: $tokenCache');
-    if (tokenCache == null) {
-      throw NotFoundStoredTokenException();
-    } else {
+    try {
+      final tokenCache = await _tokenOidcCacheClient.getItem(tokenIdHash);
+      log('TokenOidcCacheManager::getTokenOidc(): tokenCache: $tokenCache');
+      if (tokenCache == null) {
+        throw NotFoundStoredTokenException();
+      }
       return tokenCache.toTokenOidc();
+    } on AppBaseException {
+      rethrow;
+    } catch (e, stackTrace) {
+      // AES-CBC decryption failed (e.g. non-block-aligned bytes from a partial
+      // write or cross-isolate race). Clear the corrupted box so the next read
+      // starts clean and triggers normal re-authentication instead of looping.
+      logError(
+        'TokenOidcCacheManager::getTokenOidc(): '
+        'token_box_corrupted=true | error_type=${e.runtimeType} | clearing box',
+        exception: e,
+        stackTrace: stackTrace,
+      );
+      await _safelyClearBox();
+      throw NotFoundStoredTokenException();
     }
   }
 
   Future<void> persistOneTokenOidc(TokenOIDC tokenOIDC) async {
-    // Crash-safe persist: write the new token FIRST, then prune stale entries, so
-    // the box is never empty if the process is killed mid-write → forced re-login.
     log('TokenOidcCacheManager::persistOneTokenOidc(): keyHash: ${tokenOIDC.tokenIdHash}');
-    await _tokenOidcCacheClient.insertItem(tokenOIDC.tokenIdHash, tokenOIDC.toTokenOidcCache());
-    await _removeStaleTokens(keepKey: tokenOIDC.tokenIdHash);
-    log('TokenOidcCacheManager::persistOneTokenOidc(): done');
+    try {
+      // Crash-safe persist: write the new token FIRST, then prune stale entries,
+      // so the box is never empty if the process is killed mid-write.
+      await _tokenOidcCacheClient.insertItem(tokenOIDC.tokenIdHash, tokenOIDC.toTokenOidcCache());
+      await _removeStaleTokens(keepKey: tokenOIDC.tokenIdHash);
+      log('TokenOidcCacheManager::persistOneTokenOidc(): done');
+    } on AppBaseException {
+      rethrow;
+    } catch (e, stackTrace) {
+      // _removeStaleTokens calls getMapItems() → toMap(), which iterates ALL entries
+      // and decrypts each one. If the box contains a corrupted entry (e.g. from a
+      // cross-isolate partial write), toMap() throws before prune completes.
+      // Recover by clearing and re-inserting only the new valid token.
+      logError(
+        'TokenOidcCacheManager::persistOneTokenOidc(): '
+        'box_corrupted=true | error_type=${e.runtimeType} | clearing and re-inserting token',
+        exception: e,
+        stackTrace: stackTrace,
+      );
+      await _recoverBoxWithToken(tokenOIDC);
+    }
+  }
+
+  Future<void> _recoverBoxWithToken(TokenOIDC tokenOIDC) async {
+    await _safelyClearBox();
+    try {
+      await _tokenOidcCacheClient.insertItem(tokenOIDC.tokenIdHash, tokenOIDC.toTokenOidcCache());
+    } catch (e) {
+      logWarning('TokenOidcCacheManager::_recoverBoxWithToken(): re-insert failed | $e');
+    }
   }
 
   /// Removes every token except [keepKey] so the box keeps exactly one entry
@@ -43,6 +84,14 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
 
   Future<void> clear() async {
     await _tokenOidcCacheClient.clearAllData();
+  }
+
+  Future<void> _safelyClearBox() async {
+    try {
+      await _tokenOidcCacheClient.clearAllData();
+    } catch (e) {
+      logWarning('TokenOidcCacheManager::_safelyClearBox(): failed to clear box | $e');
+    }
   }
 
   @override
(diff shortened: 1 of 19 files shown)


hoangdat:
How about we should support boxItem.flush() inside `insertItem` as a defensive to prevent mid-write broken?

```dart
  Future<void> insertItem(
    String key,
    T newObject, {
    bool isolated = true,
    bool flushAfterWrite = false,
  }) {
    log('$runtimeType::insertItem:encryption: $encryption - key = $key - isolated = $isolated');
    return Future.sync(() async {
      if (isolated) {
        final boxItem = await openIsolatedBox();
        await boxItem.put(key, newObject);
        // Flush pending writes to reduce the risk of losing recently written
        // encrypted data if the app is terminated immediately after a write.
        // Improves durability; does NOT guarantee atomic persistence. No-op on
        // web (IndexedDB writes are already transactional).
        if (flushAfterWrite) await boxItem.flush();
      } else {
        final boxItem = await openBox();
        await boxItem.put(key, newObject);
        if (flushAfterWrite) await boxItem.flush();
      }
    }).catchError((error) {
      throw error;
    });
  }
```

dab246:
> How about we should support boxItem.flush() inside `insertItem` as a defensive to prevent mid-write broken?
> 
> ```dart
>   Future<void> insertItem(
>     String key,
>     T newObject, {
>     bool isolated = true,
>     bool flushAfterWrite = false,
>   }) {
>     log('$runtimeType::insertItem:encryption: $encryption - key = $key - isolated = $isolated');
>     return Future.sync(() async {
>       if (isolated) {
>         final boxItem = await openIsolatedBox();
>         await boxItem.put(key, newObject);
>         // Flush pending writes to reduce the risk of losing recently written
>         // encrypted data if the app is terminated immediately after a write.
>         // Improves durability; does NOT guarantee atomic persistence. No-op on
>         // web (IndexedDB writes are already transactional).
>         if (flushAfterWrite) await boxItem.flush();
>       } else {
>         final boxItem = await openBox();
>         await boxItem.put(key, newObject);
>         if (flushAfterWrite) await boxItem.flush();
>       }
>     }).catchError((error) {
>       throw error;
>     });
>   }
> ```

After analyzing, `flush()` unfortunately doesn't address the corruption here, and could actually make one scenario worse.

### Why it doesn't fix cross-isolate race

The confirmed failure mode is two independent `IsolatedHive` worker isolates writing to the same `.hive` file concurrently — the FCM isolate spawns its own worker when the main app's worker isn't reachable via `FcmIsolateNameServer`. `flush()` only ensures worker A's data reaches disk sooner; it has no effect on worker B interleaving its writes into the same file at the same time. The two workers remain completely unaware of each other.

### Why it may worsen partial write on process kill

`flush()` addresses **durability** (page cache → disk), not **atomicity** (write is either complete or not). Hive CE writes a token as multiple sequential syscalls: `[length_prefix][...data bytes...]`.

| | Outcome on process kill |
|---|---|
| **Without `flush()`** | OS buffer may be discarded entirely → old valid entry remains intact on disk |
| **With `flush()` after each syscall** | Partial data committed to disk with certainty → `length_prefix=1952` on disk, only `1949` bytes of data written → **same corruption state** |

The proposed code's own comment acknowledges this: *"does NOT guarantee atomic persistence"*.

### API design concern

With `flushAfterWrite = false` as the default, existing callers get no benefit unless they explicitly opt in — so the defensive goal isn't achieved unless we also change call sites. Flipping the default to `true` would impose an `fsync` on every write across all cache clients (mailbox, email, session…) without addressing the root cause.

### Correct direction

The real fix requires either removing the shared-file write path altogether or using a storage backend with OS-level atomic writes. That's tracked as a follow-up PR — migrating the OIDC token to **Flutter Secure Storage** (iOS Keychain / Android Keystore), where each `write()` is atomic by the OS and there is no shared `.hive` file between isolates.

hoangdat:
How about we should have the fallback token with token which in the memory (interceptor)? It means if we failed to get token in Hive, we try one more time with token in memory to download?

lib/features/login/data/local/token_oidc_cache_manager.dart
```diff
@@ -13,22 +14,62 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
...
+    }
+  }
+
+  Future<void> _recoverBoxWithToken(TokenOIDC tokenOIDC) async {
+    await _safelyClearBox();
+    try {
+      await _tokenOidcCacheClient.insertItem(tokenOIDC.tokenIdHash, tokenOIDC.toTokenOidcCache());
+    } catch (e) {
+      logWarning('TokenOidcCacheManager::_recoverBoxWithToken(): re-insert failed | $e');
+    }
```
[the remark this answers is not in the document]
dab246: Good catch — you're right on both points.

On the try scope: wrapping `insertItem` was incorrect. Any non-`AppBaseException` from the initial write would have triggered `_safelyClearBox()`, wiping previously valid tokens, and then `_recoverBoxWithToken()` would have returned success even if the re-insert also failed. The recovery path is only meant for the scenario where the new token is already written but `_removeStaleTokens()` → `getMapItems()` → `toMap()` crashes on a corrupt pre-existing entry. Moving `insertItem` outside the `try` block enforces that contract precisely.

On `_recoverBoxWithToken()`: removing the suppression is the right call. If the re-insert fails after clearing the box, the exception should propagate so the caller knows the token was not actually persisted, rather than returning a false success.

Applied both changes and added a new test case:

* **WHEN** `insertItem` itself throws
* **THEN** the exception propagates immediately without touching the recovery path

The test also asserts `clearCalled == false` on the mock, confirming that `_safelyClearBox()` is never reached in this scenario.

dab246:
> How about we should have the fallback token with token which in the memory (interceptor)? It means if we failed to get token in Hive, we try one more time with token in memory to download?

Agree. applied

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -72,6 +72,8 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
   AuthenticationType get authenticationType => _authenticationType;
 
+  TokenOIDC? get currentToken => _token;
```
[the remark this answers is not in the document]
dab246: Fixed

lib/features/login/data/local/token_oidc_cache_manager.dart
```diff
@@ -13,25 +14,65 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
...
+
+  Future<void> persistOneTokenOidcAt(String key, TokenOIDC tokenOIDC) async {
+    await _persistAtKey(key, tokenOIDC);
+  }
+
+  Future<void> _persistAtKey(String key, TokenOIDC tokenOIDC) async {
+    log('TokenOidcCacheManager::_persistAtKey(): keyHash: $key');
+    // Crash-safe persist: write the new token FIRST, then prune stale entries,
+    // so the box is never empty if the process is killed mid-write.
+    await _tokenOidcCacheClient.insertItem(key, tokenOIDC.toTokenOidcCache());
```
hoangdat, line 53: why keep it outside try/catch?

hoangdat: ah got it

lib/features/login/data/local/token_oidc_cache_manager.dart
```diff
@@ -45,6 +86,14 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
     await _tokenOidcCacheClient.clearAllData();
   }
 
+  Future<void> _safelyClearBox() async {
+    try {
+      await _tokenOidcCacheClient.clearAllData();
```
hoangdat, line 100: in case of box corupted, this method still work as expected or not?

dab246: `clear()` is a write-only operation and does not read or decrypt existing entries, so it works correctly on a corrupted box. When corruption is detected, the box is already open in memory. `openIsolatedBox()` returns the existing in-memory reference without re-parsing the file, allowing `clear()` to proceed normally. The only edge case is when the box has already been closed and reopening fails (for example, due to corrupted magic bytes). In that scenario, `_safelyClearBox()` intentionally swallows the exception and emits a `logWarning`. The worst-case outcome is that the user is logged out on the next application restart, which is the same behavior that existed before this PR.

hoangdat: fixed in [cd95e69](https://github.com/linagora/tmail-flutter/pull/4608/commits/cd95e6916d849f4464e1cb80cbe90081963d298f)

lib/features/login/data/local/token_oidc_cache_manager.dart
```diff
@@ -13,25 +14,74 @@ class TokenOidcCacheManager extends CacheManagerInteraction {
...
+
+  Future<void> persistOneTokenOidcAt(String key, TokenOIDC tokenOIDC) async {
+    await _persistAtKey(key, tokenOIDC);
+  }
+
+  Future<void> _persistAtKey(String key, TokenOIDC tokenOIDC) async {
+    log('TokenOidcCacheManager::_persistAtKey(): keyHash: $key');
+    // Crash-safe persist: write the new token FIRST, then prune stale entries,
+    // so the box is never empty if the process is killed mid-write.
+    await _tokenOidcCacheClient.insertItem(key, tokenOIDC.toTokenOidcCache());
```
hoangdat, line 53: agree to keep it outside the try/catch, but if it still failed by box coruption, recover box was not executed. WDYT?

hoangdat (approved):

Issue #4590, opened 2026-06-05
Wrong filter applied to my search result


Crash--:
### Description

My inbox is filtered on "Unread emails". Then I make a search "from:XXX". The results of this search only display the "unread emails" from this person. But the "unread chip" is not selected. And if I click on it it changes nothing. Only unread messages are still displayed. 

I need to refresh my tab to get it back to the expected result. 

### Expected result

I don't know we have 2 choices: 
- If my inbox is on "unread message", maybe my search should be by default "unread" but I should be able to change the filter 
- Or, we don't care about the status of the filter on the inbox, and we're making search by default with no filter.

#### Reproduction Steps
See video 

### Acceptance criteria

The list of acceptance criteria (defined in the corresponding User Story or coming from your functional knowledge of the product)
These acceptance criteria must be met by the corrective implementation done by the developers.

### Context

tmail-canary / LNG / Chrome 

https://github.com/user-attachments/assets/432533ed-5ad8-4ca3-8a11-f71e6b1636d1

### Additional information

frequency, log files,...

Crash--:
Maybe it can be related to https://github.com/linagora/tmail-flutter/issues/4490 and maybe I've found the way to reproduce it! (you can see that I've set my UI in english now, it'll be easier to understand ^^).

hoangdat:
will address with #4421

Issue #4490, opened 2026-05-06
Filter applied but without knowing it


Crash--:
### Description

I filtered my email on unread ones. Then I used Tmail. 

But at some point, I was back on my Inbox with the filter active but no filter selected in the UI.

### Expected result

See the selected filter in the UI

### Current behavior
Nothing displayed

#### Reproduction Steps
I tried to reproduce without any success

### Context

Chrome / tmail canary. 

Here I only see my unread mails... 

<img width="1520" height="987" alt="Image" src="https://github.com/user-attachments/assets/b2ed9e1b-bda0-4a43-aa4b-bf83ecba92ef" />

hoangdat:
I still not totally get your issue, can you record a video?

Crash--:
You can see here

<img width="1520" height="987" alt="Image" src="https://github.com/user-attachments/assets/781c80d3-9cba-4a90-9968-a714bda80c28" />

that I only see my unread email. But the UX says there is no enabled filtrers. 

<img width="1520" height="987" alt="Image" src="https://github.com/user-attachments/assets/89e64d9c-38f6-4848-a2fa-ceceef9324bc" />

I would have expected to have this UX. 

<img width="762" height="155" alt="Image" src="https://github.com/user-attachments/assets/8d510a36-d69a-4cf3-a27e-ec2bd66d3844" />

Issue #4421, opened 2026-03-26
Bug: "From me" or "7 previous days" shortcuts


Crash--:
### Description
When I click on "from me" or "7 previous days", I'm expecting to see these "filters selected / assigned" in the "filters panel". 

My use case was: 
- I clicked on "from me" from the "shortcuts", then I said to myself "oh I'll add a keyword on the subject, so I'll click on the icon to display all the filters". I wrote my keyword and then click on search, and I discovered that my "from me" was not selected anymore...  

https://github.com/user-attachments/assets/7682a799-fdbd-44f6-a0dd-86a9b36324dc

### Expected result
I want my "shortcuts" to be assigned to my filters view. 

Currently on tmail-canary

dab246:
Twake Mail's current behavior is similar to Gmail. If you simply click on "From me" and don't confirm anything to search for the final result (pressing Enter or the Search icon), the search filter won't add "From me," but will only treat it as a temporary action and include it in the suggestion search list on the dropdown list. You can see Gmail's behavior in the video below. Additionally, if we feel this UX isn't user-friendly, we can revise it.

https://github.com/user-attachments/assets/a17ab3e2-8748-4beb-9df6-78a63c6f2e39

Crash--:
cc @poupotte

chibenwa:
Rather agree that quick filters (that I do not like BTW) should be refleced in the search options.

We should think of this in term of a single state with different "views" of it, quick filters being one of them. Advanced search being one of them.

Anything else IMO is SUPER confusing to the end user.

I'm thus with @Crash-- on this.

hoangdat:
### DoD
- [ ] remove all redundant search filter model, keep a single source of truth for search
- [ ] ADR
- [ ] E2E test

PR #4618, opened 2026-06-19
TF-4421 Add ADR-0093 Search filter single source of truth

dab246:
## Isssue

#4421 
#4590
#4490

## Summary by CodeRabbit

* **Documentation**
  * Added ADR-0093 introducing a unified “Search Filter Single Source of Truth” model with clear separation of committed vs. draft filter state, consistent search execution, and transient pagination handling.
  * Added ADR-0094 outlining an 11-step, buildable migration plan to transition existing search/filter controller logic to the new architecture, including update rules for advanced filters and message options.


docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+| Bug | Cause |
+|---|---|
+| #4421 | `listFilterOnSuggestionForm` never merged into `searchEmailFilter` until submit; mobile `SearchEmailController` starts fresh, unaware of any chip state from `SearchController` |
+| #4612 | `SearchEmailFilter.mailbox` always initialises to `null`; current folder context never seeded on search entry |
+| #4590 | `ThreadController._searchEmail()` ANDs `filterMessageOption` as `moreFilterCondition`; chip bar reads `SearchEmailFilter.unread` (always `false`) → chip deselected, results still filtered |
+| #4490 | `filterMessageOption` reset to `all` by unrelated code paths (e.g. `handleClearAdvancedSearchFilterEmail()`); cached email list still shows filtered results while chip shows deselected |
+
+## Decision
+
+### Chosen approach: Riverpod `SearchFilterNotifier` as SSOT (mobile path now, web bridge)
```
chibenwa: SSOT ? No acronym please that's unreadable

chibenwa (commented):
Otherwize the ADR seems to go in a really good direction IMO

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Core invariant
+
+> `SearchEmailFilter.copyWith()` is called in exactly one place: `SearchFilterNotifier.update()`.
+> No other controller or widget may call `copyWith()` on filter state through any other path.
+
+### 1. `SearchFilterNotifier` — the single source of truth
+
+```dart
+@Riverpod(keepAlive: true)
```
hoangdat, line 47: it will not reset in case of logout/login again, or switch account in the future, how about to have a way to reset this provider when a session end

Arsnael (commented):
Interesting read thanks :)

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: SSOT is single source of truth. I will update the ADR to make it more accurate and easier to understand.

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+      hasAttachment: hasAttachmentOption,
+      emailReceiveTimeType: emailReceiveTimeTypeOption,
+      notKeyword: notKeywordOption,
+      subject: subjectOption,
+      unread: unreadOption,
+    );
+  }
+
+  void clear() {
+    state = SearchEmailFilter.initial();
```
hoangdat: ensure the sort still be the option which user are using,

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+
+`SearchController.updateFilterEmail()` delegates writes to the notifier instead of calling `copyWith()` directly:
+
+```dart
+void updateFilterEmail({...}) {
+  appProviderContainer
+    .read(searchFilterNotifierProvider.notifier)
+    .update(...);
+}
+```
```
hoangdat: we still have method which change `searchEmailFilter.value` directly like `synchronizeSearchFilter`, `clearSearchFilter`, ... we should to notice here to remove it?

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+
+`clearAllFilterSearch()` and `disableAllSearchEmail()` call `notifier.clear()`.
+
+All existing web `Obx` widgets watching `searchController.searchEmailFilter.obs` continue to work without change.
+
+### 3. Eliminate `SearchEmailController.searchEmailFilter.obs` (the mobile duplicate)
+
+`SearchEmailController` drops:
```
hoangdat: in mobile we also have `emailReceiveTimeType` and `emailSortOrderType`, can we also remove it?

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+Introducing `SearchFilterNotifier` — a Riverpod `@keepAlive` notifier — as the single canonical owner of `SearchEmailFilter`. The mobile search view (`SearchEmailView`) is migrated to `ConsumerStatefulWidget` to consume it reactively. The web path (`SearchController`) keeps its existing `searchEmailFilter.obs` as a mirror bridged from the notifier via one documented `appProviderContainer.listen()` call, allowing progressive migration without breaking web widgets.
+
+This establishes a clean SSOT now, and positions web `Obx` widgets for incremental migration to `Consumer` in a follow-up PR.
+
+### Core invariant
+
+> `SearchEmailFilter.copyWith()` is called in exactly one place: `SearchFilterNotifier.update()`.
+> No other controller or widget may call `copyWith()` on filter state through any other path.
+
+### 1. `SearchFilterNotifier` — the single source of truth
```
hoangdat: what do you think we should split into 2 states: one for SSOT, one for draft. IMO it is easier for us to apply the different behavior in mobile or web
Idea to make web/mobile easier to maintain: split this into two states — `searchFilterProvider` (committed = what search runs with, keepAlive, the SSOT) and `searchFilterDraftProvider` (the advanced-overlay editing buffer, autoDispose, lives only while the panel is open).

docs/adr/0093-search-filter-single-source-of-truth.md
```diff
@@ -0,0 +1,299 @@
...
+
+```dart
+@Riverpod(keepAlive: true)
+class SearchFilterNotifier extends _$SearchFilterNotifier {
```
hoangdat, line 48: how we manage the lifecycle of SearchFilter? when we clear it?

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: You are right. `appProviderContainer` is application-scoped and lives for the process lifetime (ADR-0092), so `keepAlive: true` alone means filter state from Account A leaks into Account B after a switch.

The established pattern in this codebase is `appProviderContainer.invalidate(provider)` called from the owning controller's `onClose()` — the same approach used for `composerAutoSaveProvider` in `ComposerController.onClose()`.

Fix applied to the ADR:

- Added a **Notifier lifecycle** table covering Created / Updated / Cleared / Session-end events.
- Documented that `MailboxDashBoardController.onClose()` must call:

```dart
appProviderContainer.invalidate(searchFilterNotifierProvider);
```

`invalidate()` disposes the current state; the next `ref.read/watch` triggers `build()` and returns `SearchEmailFilter.initial()`.

- Added `MailboxDashBoardController` to the **Modified files** table accordingly.

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: Correct. `SearchEmailFilter.initial()` resets `sortOrderType` to `defaultSortOrder` (relevance). If the user has selected "Date: Newest first" and clears all filters, their sort preference should survive. `SearchEmailFilter.withSortOrder()` already exists for exactly this.

Fix applied to the ADR — `clear()` updated to:

```dart
void clear() {
  // Preserve the user's sort order preference; reset everything else.
  state = SearchEmailFilter.withSortOrder(state.sortOrderType);
}
```

Unit test coverage updated: `clear()` preserves `sortOrderType` and resets all other fields.

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: Good catch. The ADR only addressed `updateFilterEmail()`, `clearAllFilterSearch()`, and `disableAllSearchEmail()`. Two methods were missed:

- `SearchController.synchronizeSearchFilter(SearchEmailFilter)` — called from `AdvancedFilterController` when applying the advanced filter; mutates `.value` directly.
- `SearchController.clearSearchFilter({EmailSortOrderType?})` — called internally within `SearchController`; also mutates `.value` directly.

Both violate the core invariant ("copyWith in exactly one place"). Fix applied to the ADR:

- Added a **"Methods that bypass the notifier and must be removed"** subsection in Section 2.
- Added both to the **Deleted symbols** table.
- Updated **Modified files** entry for `SearchController`.

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: Yes. `SearchEmailController` declares:

```dart
final emailReceiveTimeType = EmailReceiveTimeType.allTime.obs;  // line 121
final emailSortOrderType = SearchEmailFilter.defaultSortOrder.obs;  // line 123
```

Both are duplicates of fields already present in `SearchEmailFilter`. With the notifier as single source of truth, all reads go through `ref.watch(searchFilterNotifierProvider)`, making these obsolete.

Fix applied to the ADR:

- Added both to the removal list in **Section 3**, with replacement reads via `ref.watch(searchFilterNotifierProvider).emailReceiveTimeType` / `.sortOrderType`.
- Added both to the **Deleted symbols** table.
- Updated unit test coverage: `emailReceiveTimeType` and `emailSortOrderType` obs fields are absent.

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: This is the right long-term direction. The split would be:

| Provider | Scope | Purpose |
|---|---|---|
| `searchFilterProvider` (keepAlive) | Committed | What the JMAP query runs with |
| `searchFilterDraftProvider` (autoDispose) | Panel-scoped | Editing buffer while the advanced filter panel is open |

`AdvancedFilterController._memorySearchFilter` is already an implicit draft — this would formalize it as a Riverpod provider. `autoDispose` would eliminate the need for manual `clearFilterSuggestion()`-style cleanup.

**Why deferred to ADR-0094:** This ADR's scope is fixing 4 specific bugs via a single write path. The draft/committed split requires rethinking the `AdvancedFilterController` apply/cancel flow, and is most naturally done together with the Obx→Consumer migration once the web path reads the notifier directly. Doing it now would widen scope without benefiting the bug fixes.

Added a **"Draft/committed state split (deferred to ADR-0094)"** subsection in the Future section documenting this rationale.

docs/adr/0093-search-filter-single-source-of-truth.md
dab246: The ADR did not document state transitions. Added a **Notifier lifecycle** table directly after the `SearchFilterNotifier` code block:

| Event | Trigger | Action |
|---|---|---|
| **Created** | First `ref.read/watch` after container start or after `invalidate()` | `build()` → `SearchEmailFilter.initial()` |
| **Updated** | Chip tap, form field edit, folder pre-fill on search entry | `update(...)` — partial field update, sort order preserved |
| **Cleared** | User clears all filters (exit or "clear all" button) | `clear()` — resets all fields, preserves `sortOrderType` |
| **Session end** | Logout or account switch | `appProviderContainer.invalidate(searchFilterNotifierProvider)` in `MailboxDashBoardController.onClose()` |

tddang-linagora (dismissed):

Crash--:
Global feedbacks: thanks! great move! This SSOT (single source of truth) is really needed for this use case.

Few things: 
- ADR-0092 frames appProviderContainer as a global-singleton anti-pattern to eliminate once its call sites are gone. This ADR adds new appProviderContainer.read/.listen usage across 5 controllers. Can we add a note on how this stays compatible with the eventual ProviderScope migration rather than entrenching the bridge? 
Or did I miss something? 

- The Invariants section (copyWith only in the notifiers, cursors never written to the SSOT, draft never read by the executor) is exactly what should be locked down by tests. The plan currently relies on "existing unit tests catch regressions." For a refactor whose whole purpose is preventing state divergence, could we add explicit invariant tests to the plan (ideally in steps 2/4, before the controllers are touched)?

- The doc says "no big-bang migration," but steps 5-9 rewrite the search bodies in three of the most heavily used controllers and migrate two views to ConsumerStatefulWidget. Can we confirm each step ships as its own PR so review and regression testing stay scoped?

chibenwa:
STOT -> ???

(Sorry to be naive)

hoangdat:
### Overview
- refresh intent
- draft notifier -> search filter notifier: suggestion (text or search or enter) - advanced (clear filter  or search or enter)
- UI -> Draft notifier write to Search filter notifier ?? (cursor store also in search filter)
- Transformers write to `Search filter notifier`
- life cycle of search; end search -> clear filter (draft + search), 

### Transformer
- SearchEmailFilter apply(SearchEmailFilter base) =>
      _transformers.fold(base, (acc, t) => t.transform(acc)); ????
- can order free of transformer
- add more transformer

dab246:
> * ADR-0092 frames appProviderContainer as a global-singleton anti-pattern to eliminate once its call sites are gone. This ADR adds new appProviderContainer.read/.listen usage across 5 controllers. Can we add a note on how this stays compatible with the eventual ProviderScope migration rather than entrenching the bridge?
> 

Good catch, added an explicit note in ADR-0093 → **Riverpod-First Direction → "Compatibility with the ProviderScope migration"**.

Short version: the 5 new usages **consolidate** ~30 scattered `searchEmailFilter.obs` write sites into a small, enumerable set of bridge points — net movement toward ProviderScope, not away. Since `UncontrolledProviderScope` (ADR-0092) already shares `appProviderContainer` with the widget tree, each bridge is a one-line swap to `ref.read/watch` when its controller becomes a `ConsumerWidget`. The ADR now lists each bridge + the exact condition that removes it; when all 5 are gone, `appProviderContainer` is deleted per ADR-0092.

> * The Invariants section (copyWith only in the notifiers, cursors never written to the SSOT, draft never read by the executor) is exactly what should be locked down by tests. The plan currently relies on "existing unit tests catch regressions." For a refactor whose whole purpose is preventing state divergence, could we add explicit invariant tests to the plan (ideally in steps 2/4, before the controllers are touched)?
> 

Agreed, done in ADR-0094. The invariants are now locked by explicit tests in **Steps 2–5, before any controller is touched**:

- **Step 3** (`SearchFilterNotifier`): `update()` leaves cursors untouched; `clear()` preserves `sortOrderType`.
- **Step 4** (`SearchFilterDraftNotifier`): containment — `search ⊆ draft` after `seedFrom` + edits.
- **Step 5** (`SearchEmailNotifier`) — core anti-regression suite: cursors/position resolved correctly per intent; committed SSOT unchanged after `execute()`; and **draft-isolation** (committed=`filterA`, draft=`filterB` → interactor receives `filterA`), which directly encodes "executor never reads draft".

(Step numbering shifted vs your note — the build order is now bottom-up, so all invariant tests land in Steps 2–5 ahead of the controller migration in Steps 6–10.)

> * The doc says "no big-bang migration," but steps 5-9 rewrite the search bodies in three of the most heavily used controllers and migrate two views to ConsumerStatefulWidget. Can we confirm each step ships as its own PR so review and regression testing stay scoped?
> 

Confirmed. This PR is ADR-only; implementation is **11 steps, each its own PR** (ADR-0094 → "Shipping strategy"):

- **Steps 1–5** (model + strategies + notifiers + executor): no controller touched — self-contained, new files + invariant tests only.
- **Steps 6–10**: one controller (or controller + its view) per PR — regression scope limited to that unit.
- **Step 11**: remove `position` from the model (cleanup, once no caller reads it).

So the heavy controller/view rewrites you flagged are split one-per-PR, each reviewable and regression-testable in isolation.

tddang-linagora (dismissed):

dab246:
## Proposed ticket breakdown for implementing ADR-0093 / ADR-0094

**7 subtasks — filed as issues [#4642](https://github.com/linagora/tmail-flutter/issues/4642)–[#4648](https://github.com/linagora/tmail-flutter/issues/4648).**

- [ ] **SSOT Part 1 — Add filter-state notifiers + pagination strategies (Riverpod foundation)** → [#4642](https://github.com/linagora/tmail-flutter/issues/4642)
- [ ] **SSOT Part 2 — Add central search executor + result model (Riverpod foundation)** → [#4643](https://github.com/linagora/tmail-flutter/issues/4643)
- [ ] **SSOT Part 3 — Migrate web search input to filter-state notifiers (SearchController + AdvancedFilterController)** → [#4644](https://github.com/linagora/tmail-flutter/issues/4644)
- [ ] **SSOT Part 4 — Migrate mobile search to the central executor (SearchEmailController)** → [#4645](https://github.com/linagora/tmail-flutter/issues/4645)
- [ ] **SSOT Part 5 — Migrate web results to the central executor (ThreadController)** → [#4646](https://github.com/linagora/tmail-flutter/issues/4646)
- [ ] **SSOT Part 6 — Wire dashboard search: filter-option mapping (MailboxDashBoardController)** → [#4647](https://github.com/linagora/tmail-flutter/issues/4647)
- [ ] **SSOT Part 7 — Remove `position` field from SearchEmailFilter model** → [#4648](https://github.com/linagora/tmail-flutter/issues/4648)

### Detail

| # | Issue | What it does | Relates | Pts | Diff |
|---|---|---|---|---|---|
| 1 | [#4642](https://github.com/linagora/tmail-flutter/issues/4642) | `copyWithOptions`, committed + draft notifiers, pagination intent/spec/strategies (additive, no controller touched) | — | 5 | ~585 |
| 2 | [#4643](https://github.com/linagora/tmail-flutter/issues/4643) | `SearchEmailNotifier` + result model + query params + anti-regression suite | — | 3 | ~390 |
| 3 | [#4644](https://github.com/linagora/tmail-flutter/issues/4644) | `SearchController` + `AdvancedFilterController` write to SSOT/draft; drop deferred staging | #4421 | 5 | ~460 |
| 4 | [#4645](https://github.com/linagora/tmail-flutter/issues/4645) | `SearchEmailController` delegates to the executor; results bridged into `.obs` | — | 3 | ~380 |
| 5 | [#4646](https://github.com/linagora/tmail-flutter/issues/4646) | `ThreadController` delegates; **removes the hidden `moreFilterCondition` AND** | #4590, #4490 | 3 | ~240 |
| 6 | [#4647](https://github.com/linagora/tmail-flutter/issues/4647) | map `filterMessageOption` into SSOT; invalidate on logout | #4490 | 2 | ~210 |
| 7 | [#4648](https://github.com/linagora/tmail-flutter/issues/4648) | type-enforce "cursors never in the SSOT" | — | 1 | ~35 |

### Merge order & dependencies

Arrows = "must merge before". Subtask **1** is the critical path; once its foundation is in, **3/4/5/6 run in
parallel**; **7** merges last (needs 4 + 5 to remove every `position` reader).

```mermaid
flowchart LR
    S1["1 · Notifiers + strategies<br/>(foundation)"]
    S2["2 · Central executor<br/>(foundation)"]
    S3["3 · Web search input<br/>SearchController + AdvancedFilter"]
    S4["4 · Mobile search<br/>SearchEmailController"]
    S5["5 · Web results<br/>ThreadController"]
    S6["6 · Dashboard wiring<br/>MailboxDashBoard"]
    S7["7 · Remove position<br/>(cleanup)"]

    S1 --> S2
    S1 --> S3
    S2 --> S4
    S2 --> S5
    S2 --> S6
    S4 --> S7
    S5 --> S7

    classDef foundation fill:#dbeafe,stroke:#3b82f6,color:#1e3a8a;
    classDef migration fill:#dcfce7,stroke:#22c55e,color:#14532d;
    classDef cleanup fill:#fef9c3,stroke:#eab308,color:#713f12;
    class S1,S2 foundation;
    class S3,S4,S5,S6 migration;
    class S7 cleanup;
```

hoangdat (dismissed):

tddang-linagora (approved):

PR #4626, opened 2026-06-24
Fix sort order not applied from search bar on restart

dab246:
## Reproduce

https://github.com/user-attachments/assets/b6aa5b73-49ec-4080-9374-5e98e2d6cd59

## Root cause

`setUpDefaultEmailSortOrder()` — called on app startup when the stored sort order is loaded from local storage — was updating `currentSortOrder` and dispatching `SynchronizeEmailSortOrderAction`, but **not** syncing `SearchController.searchEmailFilter`.

On web, the inline search bar (`SearchInputFormWidget`) triggers `searchEmailByQueryString()` → `StartSearchEmailAction` → `ThreadController._searchEmail()`, which reads `searchController.searchEmailFilter.sortOrderType` directly. Because that field was never updated on restart, every inline search after a cold start would silently fall back to the default sort order (`relevance`), ignoring the user's stored preference.

## Solution

```dart
// handle_store_email_sort_order_extension.dart
void setUpDefaultEmailSortOrder(EmailSortOrderType storedSortOrder) {
  currentSortOrder = storedSortOrder;
+ searchController.updateFilterEmail(sortOrderTypeOption: Some(storedSortOrder)); // ← fix
  dispatchAction(SynchronizeEmailSortOrderAction(currentSortOrder));
}
```

One line added: propagate the stored sort order into `searchController.searchEmailFilter` so both the restart path (`loadStoredEmailSortOrder`) and the user-change path (`storeEmailSortOrder`) keep `currentSortOrder` and `searchController` in sync.

## Resolved

https://github.com/user-attachments/assets/2d8ef4a3-1040-4637-b4d8-1b35e4dd6d52

## Summary by CodeRabbit

* **Bug Fixes**
  * Sort order preferences are now correctly persisted and synchronized with the search functionality after app restart.

* **Tests**
  * Added integration test scenario validating sort order persistence in search after app restart.
  * Added unit tests verifying sort order synchronization between email and search controllers.

### DIFF
--- a/integration_test/scenarios/search/sort_order_persisted_from_search_bar_on_restart_scenario.dart
+++ b/integration_test/scenarios/search/sort_order_persisted_from_search_bar_on_restart_scenario.dart
@@ -1,0 +1,63 @@
+
+import 'package:get/get.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/extensions/handle_store_email_sort_order_extension.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/model/search/email_sort_order_type.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart';
+
+import '../../base/base_test_scenario.dart';
+import '../../models/provisioning_email.dart';
+import '../../utils/test_timeouts.dart';
+import '../../utils/wait_for_condition.dart';
+
+class SortOrderPersistedFromSearchBarOnRestartScenario extends BaseTestScenario {
+  const SortOrderPersistedFromSearchBarOnRestartScenario(super.$, super.robots);
+
+  static const queryString = 'sort order restart';
+  static const listUsername = ['Alice', 'Brian', 'Charlotte', 'David', 'Emma'];
+
+  @override
+  Future<void> runTestLogic() async {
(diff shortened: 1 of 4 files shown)


dab246:
@coderabbitai full review

integration_test/scenarios/search/sort_order_persisted_from_search_bar_on_restart_scenario.dart
```diff
@@ -0,0 +1,63 @@
...
+    // Simulate app restart: load the stored sort order and wait for it to
+    // propagate into searchController.searchEmailFilter
+    dashboardController.loadStoredEmailSortOrder();
+    await waitForCondition(
+      () async =>
+          dashboardController.searchController.sortOrderFiltered ==
+          EmailSortOrderType.oldest,
+      timeout: TestTimeouts.medium,
+    );
+    await $.pump(const Duration(seconds: 1));
```
hoangdat, line 52: why not have a robot action to call to `web.window.location.reload();` in web?

dab246: `window.location.reload()` would kill the Patrol test driver since it runs inside the same Dart/Flutter runtime as the app — there's no reconnect mechanism. The simulation approach (reset state → call `loadStoredEmailSortOrder()`) correctly targets the bug site (the sync in `setUpDefaultEmailSortOrder`) without requiring a real page reload. Hive round-trip persistence is better covered by unit tests on the interactors.

hoangdat (approved):

PR #4659, opened 2026-07-01
TF-4643 SSOT Part 2 — Add central search executor + result model

### LINKED ISSUE #4643: SSOT Part 2 — Add central search executor + result model (Riverpod foundation)
## Problem / Why

Search execution is duplicated across `ThreadController` (web) and `SearchEmailController` (mobile), each writing
pagination cursors into the same object as user intent. This ticket adds the **single executor** that reads the
committed SSOT ([#4642](https://github.com/linagora/tmail-flutter/issues/4642)), resolves pagination via the strategies ([#4642](https://github.com/linagora/tmail-flutter/issues/4642)), and dispatches to the right
interactor. It is the anti-regression core — the ADR-0093 invariants become provable here, before any controller
delegates to it.

## Scope

- `SearchEmailState` (`emails`, `canLoadMore`) — replaces the scattered `emailList` + `canLoadMore` booleans.
- `SearchEmailQueryParams` — parameter object collapsing the three interactor argument lists.
- `SearchEmailNotifier` (`@Riverpod(keepAlive: true)`): `_strategies.firstWhere(appliesTo).apply`, build params,
  `switch(intent)` → interactor (resolved via `Get.find` — documented temporary GetX bridge); `_foldSearch`
  appends on load-more, replaces otherwise.

## Out of scope

- No controller calls the executor yet (tickets 4, 5). No view migration. No interactor signature changes.

## Definition of Done

- [x] `NewSearchIntent` clears cursors regardless of prior SSOT values; `LoadMoreIntent` → `position == count`
      (position sort) **or** exactly one date cursor (date sort), never both; `RefreshChangesIntent` →
      `limit == currentCount`, `position == 0`.
- [x] **Draft isolation:** with committed = `filterA`, draft = `filterB`, `execute` uses `filterA`.
- [x] Committed SSOT unchanged after any `execute`; load-more appends, new search replaces.
- [x] No existing controller/view modified.

## Relates to / Depends on

- **Depends on:** [#4642 · Part 1](https://github.com/linagora/tmail-flutter/issues/4642)
- **Blocks:** [#4645 · Part 4](https://github.com/linagora/tmail-flutter/issues/4645), [#4646 · Part 5](https://github.com/linagora/tmail-flutter/issues/4646), [#4647 · Part 6](https://github.com/linagora/tmail-flutter/issues/4647)

dab246:
Closes #4643

> Full scope, design rationale and DoD live in the issue: https://github.com/linagora/tmail-flutter/issues/4643

## Summary

Second foundational step of the Search Filter SSOT refactor. **Additive only** — new Riverpod primitives on top of Part 1; no controller/view touched, no interactor signature change, no regression surface. This is the **anti-regression core**: the ADR-0093 invariants become provable here before any controller delegates to it.

**Added**
- `SearchEmailState` (`emails`, `canLoadMore`) — result model replacing the scattered `emailList` + `canLoadMore` booleans.
- `SearchEmailQueryParams` — one parameter object for the three interactors' argument lists.
- `SearchEmailNotifier` (`@Riverpod(keepAlive: true)`): reads the committed SSOT, resolves pagination via the Part 1 strategies, dispatches by intent, and folds the result (`_foldSearch` — append on load-more, else replace).

**Hardening** (beyond DoD, so later tickets inherit a robust executor)
- **Out-of-order guard** — a request token drops a late-resolving earlier call so it can't clobber a newer result.
- **Error safety** — param-resolution throws, interactor stream errors, and `Either.Left` all fold into `AsyncError` (never stuck on `AsyncLoading`), and are reported to Sentry via `logError` (no filter/email content — privacy).
- **Pages preserved on failure** — the last loaded page list is kept outside `state`, so a retry after a failed load-more still appends instead of dropping earlier pages.
- **Lazy DI** — interactors resolve per call, so the provider can be built before the GetX bindings run; a missing interactor becomes `AsyncError`, not a `build()` throw.

## Notes for review
- The anti-regression suite is the contract later tickets build on — focus on `search_email_notifier_test.dart`: cursor resolution per `(intent, sort)`, draft isolation, committed-SSOT immutability, append-vs-replace, and the failure / stream-error / out-of-order / missing-interactor / retry-after-failure paths.
- `Get.find` is the documented temporary GetX bridge, replaced when interactors migrate to Riverpod.
- Run `scripts/prebuild.sh` — `search_email_notifier.g.dart` is generated, not committed.

## Release Notes

* **New Features**
  * Implemented email search pagination with load-more functionality to efficiently handle large result sets
  * Added search refresh capability to retrieve updated results without disrupting the current view
  * Enhanced search result state management with improved reliability, proper handling of concurrent requests, and protection against stale results


lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,215 @@
...
+        if (newEmails == null) return state;
+        final current = state.value?.emails ?? const <PresentationEmail>[];
+        final emails = append ? [...current, ...newEmails] : newEmails;
+        return AsyncData(SearchEmailState(
+          emails: emails,
+          canLoadMore: newEmails.isNotEmpty,
+        ));
+      },
+    );
+  }
```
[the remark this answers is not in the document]
dab246: fixed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,226 @@
...
+
+  /// Runs [intent] and updates [state]. The committed SSOT is never mutated.
+  Future<void> execute(
+    SearchExecutionIntent intent, {
+    required Session session,
+    required AccountId accountId,
+    required Properties properties,
+    required bool collapseThreads,
+    required Set<MailboxId>? trashSpamMailboxIds,
+    PresentationEmail? lastEmail,
```
tddang-linagora: can/should `lastEmail` be a `getter` pointing to `_loadedEmails`?

dab246: Why do we have to do that? I don't get your point

dab246: I've moved its id onto LoadMoreIntent as lastEmailId, next to currentCount/lastEmailDate, so all  three cursors are declared together. On making it a getter over _loadedEmails: that's the right end-state, but not safe yet. During the GetX transition the rendered list (listResultSearch) is authoritative and is mutated outside execute() (dedup/append, in-place sync, selection rebuilds), so _loadedEmails can drift from it — deriving the anchor internally would page from a stale row. The cursor stays caller-supplied (from the rendered list) until the notifier owns that list; then all three params collapse into internal derivation.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

tddang-linagora (dismissed):

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,223 @@
...
+            limit: params.limit,
+            position: params.position,
+            sort: params.sort,
+            filter: params.filter,
+            properties: params.properties,
+            collapseThreads: params.collapseThreads,
+          ).last,
+        );
+      case LoadMoreIntent():
+        await _applyResult(
```
hoangdat: on this case, the loadmore error will remove all the previous list?

lib/features/search/email/domain/model/search_email_state.dart
hoangdat, line 1: quite duplicate with the previous one, how about change it to `search_email_result.dart`

lib/features/search/email/domain/model/search_email_query_params.dart
```diff
@@ -0,0 +1,38 @@
...
+  final Filter? filter;
+  final Set<Comparator>? sort;
+  final Properties? properties;
+  final bool collapseThreads;
+  final UnsignedInt? limit;
+  final int? position;
+
+  /// Load-more only — excludes the cursor row from the appended page.
+  final EmailId? lastEmailId;
+  final bool needRefreshSearchState;
```
hoangdat: quite we dont use it ?

lib/features/search/email/domain/notifier/search_email_notifier.dart
dab246: fixed

lib/features/search/email/domain/model/search_email_state.dart
dab246: oke, rename to `search_email_result `

lib/features/search/email/domain/model/search_email_query_params.dart
dab246: removed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,253 @@
...
+      onPageLoaded: (page) => state = AsyncData(
+        SearchEmailResult(emails: page, canLoadMore: page.isNotEmpty),
+      ),
+      onFailure: (error, stackTrace) => state = AsyncError(error, stackTrace),
+    );
+  }
+
+  /// Load-more: mark the page in progress, then append it. On failure the loaded
+  /// page stays and [LoadMoreState.failure] lets the UI offer a retry.
+  Future<void> _runLoadMore(int requestId, SearchExecutionRequest request) {
```
hoangdat, line 122: what happen if `runLoadMore` is in progress then `refresh` comes by web-socket, loadmore state be stucked in `inProgress`?

hoangdat: please add the test

```dart
    test('a refresh failure clears a load-more spinner it superseded', () async {
      final container = containerForSort(EmailSortOrderType.relevance);
      stubSearch([emailWith('e1')]);
      await runExecute(container, const NewSearchIntent());

      // A slow load-more is in flight (spinner on) when a refresh supersedes it.
      final slowLoadMore = Completer<Either<Failure, Success>>();
      answerSearchMore((_) => Stream.fromFuture(slowLoadMore.future));
      final stale = runExecute(
        container,
        LoadMoreIntent(
          currentCount: 1,
          lastEmailDate: cursorDate,
          lastEmailId: EmailId(Id('e1')),
        ),
      );
      expect(container.read(searchEmailProvider).value?.loadMore,
          LoadMoreState.inProgress);

      // The refresh fails: it keeps the list but must not leave the dropped
      // load-more's spinner stuck on.
      when(refreshInteractor.execute(
        any,
        any,
        limit: anyNamed('limit'),
        position: anyNamed('position'),
        sort: anyNamed('sort'),
        filter: anyNamed('filter'),
        collapseThreads: anyNamed('collapseThreads'),
        properties: anyNamed('properties'),
      )).thenAnswer((_) => Stream.value(Left(SearchEmailFailure(Exception('boom')))));
      await runExecute(container, const RefreshChangesIntent(currentCount: 1));

      // The superseded load-more resolves last; its result is dropped as stale.
      slowLoadMore.complete(Right(SearchMoreEmailSuccess([emailWith('e2')])));
      await stale;

      final result = container.read(searchEmailProvider);
      expect(result.value?.emails.map((e) => e.id), [EmailId(Id('e1'))]);
      expect(result.value?.loadMore, LoadMoreState.idle);
    });
```

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,253 @@
...
+      )),
+      onFailure: (_, __) => state = AsyncData(
+        _currentResult.copyWith(loadMore: LoadMoreState.failure),
+      ),
+    );
+  }
+
+  /// Refresh: replace the list on success; on failure keep the current list, never
+  /// erroring the whole result.
+  Future<void> _runRefresh(int requestId, SearchExecutionRequest request) {
```
hoangdat, line 155: one more case, if you stay in empty mailbox, then new message come in this mailbox, refresh will not change the list, still empty? how to handle this case?

dab246: We are in search mode, not open mailbox mode

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,253 @@
...
+      NewSearchIntent() => _runNewSearch(requestId, request),
+      LoadMoreIntent() => _runLoadMore(requestId, request),
+      RefreshChangesIntent() => _runRefresh(requestId, request),
+    };
+  }
+
+  /// New search: show a first-page spinner, then replace the list or error (a
+  /// fresh search has no prior list to fall back on).
+  Future<void> _runNewSearch(int requestId, SearchExecutionRequest request) {
+    state = const AsyncLoading();
```
hoangdat, line 93: the same if `refresh` come. it will be stuck here

lib/features/search/email/domain/notifier/search_email_notifier.dart
dab246: Fixed. Refresh now resets `loadMore` to `LoadMoreState.idle` before running, so a superseded load-more spinner cannot stay stuck. Added a regression test for this case.

lib/features/search/email/domain/notifier/search_email_notifier.dart
dab246: Fixed. Refresh now uses `currentCount` only when it is greater than zero. If the current list is empty, it falls back to the default first-page limit, so the newly arrived message can be fetched and displayed. Added a test for this case.

lib/features/search/email/domain/notifier/search_email_notifier.dart
dab246: Fixed. When refresh supersedes an in-flight new search, it restores an `AsyncData` result before executing. If refresh fails, the provider no longer remains stuck in `AsyncLoading`, and the stale new-search response is still ignored. Added a regression test for this case

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -0,0 +1,255 @@
...
+      onPageLoaded: (page) => state = AsyncData(
+        SearchEmailResult(emails: page, canLoadMore: page.isNotEmpty),
+      ),
+      onFailure: (error, stackTrace) => state = AsyncError(error, stackTrace),
+    );
+  }
+
+  /// Load-more: mark the page in progress, then append it. On failure the loaded
+  /// page stays and [LoadMoreState.failure] lets the UI offer a retry.
+  Future<void> _runLoadMore(int requestId, SearchExecutionRequest request) {
```
tddang-linagora, line 122: Would this case happen, and if it would, how should the app handle it:
```
WHEN user is trying to load more
AND the state is LoadMoreState.inProgress
THEN web socket signals an update
THEN _runRefresh is triggered
THEN the result of _runRefresh is returned before _runLoadMore
```

dab246: Yes, it can happen, and it's handled by the monotonic `_latestRequestId`, not by `LoadMoreState`.

`execute()` increments the request ID synchronously as soon as a refresh is requested, so any in-flight load-more request is superseded immediately. When that load-more request eventually completes, `_runGuarded` detects that `requestId != _latestRequestId` and discards its result, regardless of which future finishes first.

As a result, whether the refresh completes before the load-more request or vice versa, the outcome is the same:
- The refreshed list is displayed.
- The stale load-more page is discarded.
- `loadMore` returns to the idle state.

This behavior is covered in `search_email_notifier_test.dart` by the test **"a succeeding refresh supersedes an in-flight load-more"** (just added), along with the nearby failure and new-search variants.

I also clarified the `_runLoadMore` comment to explicitly reference the request guard.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4659

hoangdat (approved):

tddang-linagora (approved):

Issue #4660, opened 2026-07-01
Loading state for exchanging token + get intent URL


hoangdat:
### Desc
- still take time to exchange token and intents url
- now it display with blank screen until intent ready come back

### DoD
- [ ] loading screen for user

https://github.com/user-attachments/assets/e598d6e8-2c3b-4333-8223-a2c44a2be13b

tddang-linagora:
https://github.com/user-attachments/assets/3ec0253e-9459-4f47-99db-4bb8e050bc0f

Can we have this for now?

hoangdat:
hi @poupotte any update for this, we are still waiting for the design

poupotte:
Loading screen : https://www.figma.com/design/XqLqMINlZw09BdEHI8yLb9/Teammail---1.1?node-id=10343-9204&t=AwOkkAcsr6GY4rNh-0

<img width="1440" height="813" alt="Image" src="https://github.com/user-attachments/assets/fed5458f-9dc1-467e-a093-1469e5d241da" />

hoangdat:
hi @poupotte because we want to keep it generic as much as possible, keep the twake drive icon in the loading screen is violate it? how we can adapt this? how we can show the loading screen for other drive integration?

hoangdat:
- split exchange token to a single step can be reuse in: openning composer, retry when refreshToken expired, ...
- handle refreshToken for exchanged token

PR #4662, opened 2026-07-01
TF-4660 Add temporarly drive loading

tddang-linagora:
## Issue
- #4660 

## Demo

https://github.com/user-attachments/assets/3246eda8-f070-4b8c-aded-576a2e8d4466

 

## Summary by CodeRabbit

* **New Features**
  * Added clearer drive-picking status updates, including loading and displaying states.
  * Improved the app’s handling of drive selection so users see more responsive feedback while content is loading.
* **Bug Fixes**
  * Prevents duplicate loading dialogs during drive picking, reducing flicker and inconsistent modal behavior.

### DIFF
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -1,5 +1,6 @@
 import 'package:core/presentation/extensions/composer_attachment_extension_registry.dart';
 import 'package:core/utils/app_logger.dart';
+import 'package:flutter_smart_dialog/flutter_smart_dialog.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 import 'package:tmail_ui_user/features/composer/presentation/composer_controller.dart';
 import 'package:tmail_ui_user/features/login/data/network/interceptors/authorization_interceptors.dart';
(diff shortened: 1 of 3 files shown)


lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,6 +20,11 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
       workplaceUri: uriNotifier,
       oidcTokenGetter: () => getBinding<AuthorizationInterceptors>()?.currentOidcIdToken,
       onPickState: (composerId, state) {
+        if (state is DrivePickLoading && !SmartDialog.checkExist()) {
```
dab246, line 23: What about if state is DrivePickFailure?

dab246:
IMO, Using that loading screen doesn't look very good, we should wait for the latest design.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: It will dismiss the loading

hoangdat:
waiting for design provide us the detail screen

Issue #4410, opened 2026-03-25
POC - Attach from Drive one file as a copy


poupotte:
# Summary

* [Related EPIC](#related-epic)
* [Definition](#definition)
* [Screenshots](#screenshots)
* [Misc](#misc)

## Related EPIC

https://github.com/linagora/tmail-flutter/issues/4376

## Definition

This task is to deliver a proof of concept for attaching a file from Drive to an email, based on the following Figma design: https://www.figma.com/design/XqLqMINlZw09BdEHI8yLb9/Teammail---1.1?node-id=6042-5651&t=uXNG22PdBnZxH6cM-0.
​

### Scope of the task
Add two button in the iframe ''Copy" and "Link"
if user select a file and click on "Copy", then we upload the file as an attachment in the mail composer
​
The user can select only one file at a time.

### Expected behavior
When the user selects a file, we upload the file as an attachment in the mail composer

[Back to Summary](#summary)

## Screenshots

None

[Back to Summary](#summary)

## Misc

None

[Back to Summary](#summary)

PR #4688, opened 2026-07-07
TF-4410 Propose ADR attach drive as attachment

tddang-linagora:
## Issue
- #4410

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record describing how Drive-picked documents are attached as real email attachments using `downloadLink` (instead of HTML links).
  * Defined precedence between `sharingLink` and `downloadLink`, including staged download→upload behavior across IO, OPFS (when available), and web fallbacks.
  * Documented per-file workflow with separate downloading/uploading progress, single per-file cancellation, bounded concurrency, size validation, cleanup guarantees, and streaming timeout/retry constraints.


docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,72 @@
...
+  FutureOr<bool> validate();
+}
+
+class CompositeAttachmentUploadValidator implements AttachmentUploadValidator {
+  final List<AttachmentUploadValidator> validators;
+  const CompositeAttachmentUploadValidator(this.validators);
+
+  @override
+  Future<bool> validate() async {
+    for (final v in validators) {
```
dab246, line 51: should return early  if validators is empty

tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Decision
+
+**Partitioning:** `DriveAttachmentHandler` splits picked `DriveDocument`s per-doc: `sharingLink != null` → HTML link (unchanged); `downloadLink != null` → download + attach (new).
+
+**Download (`workplace` package), streamed:** `DriveFileDatasourceImpl.openFileForUpload(doc)` branches on `kIsWeb`. **IO**: `Dio.get(responseType: stream)` piped directly as the upload request body (`FileInfo.byteStream`) — true zero-copy, no whole-file buffering, backpressure handled automatically by Dio. **Web**: `BrowserHttpClientAdapter` (XHR) can't stream a body, so web still buffers into `FileInfo.bytes`. `DownloadDriveFileInteractor.openFileForUpload` is a thin future delegate (no more stream-of-states). A bounded worker pool (`kMaxConcurrentDriveTransfers = 3`) runs per-file pipelines: file N+1 starts downloading as a slot frees, and each file starts uploading the instant its own stream is ready (no "download all, then upload all" batching). A failing file logs/removes its chip; others continue.
```
dab246: Is this value `kMaxConcurrentDriveTransfers = 3` described in the documentation?

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+
+**Upload (reused as-is):** downloaded/picked `FileInfo`s all go through the composer's existing single upload trigger (session/account resolution → JMAP upload URI → chunked upload). Drive attachments are indistinguishable from local ones past this point.
+
+**Module boundary:** intent protocol, `DriveDocument`, and the download chain stay in `workplace`; only `DriveAttachmentHandler` (main app) composes download orchestration with the shared validator abstraction and the composer's existing upload trigger.
+
+## Consequences
+- Drive attachments reuse the exact same size-limit UX and upload pipeline as every other attachment source; `FileInfo` gains an additive, transient `byteStream` field so the rest of the upload machinery (interactor/repository/`FileUploader`) is unchanged.
```
dab246: Stream upload is not compatible with the current non-web isolate upload path

The ADR says IO downloads will be streamed directly into the upload request body via `FileInfo.byteStream`, while the upload pipeline remains mostly unchanged.

Current code does not support that contract:

- `FileUploader.uploadAttachment` sends non-web uploads to a worker isolate when `Platform.numberOfProcessors > 1`.
- `UploadFileArguments` passes `FileInfo` across the worker boundary.
- `FileInfo` currently carries `filePath` or `bytes`, not a stream.
- A `Stream<List<int>>` from `Dio.get(responseType: stream)` is one-shot and not a safe value to serialize or pass through the isolate worker model.

Risk:

The ADR promises true zero-copy streaming while reusing the current upload machinery, but this is not implementable without changing the upload contract. A future implementation following the ADR literally is likely to fail on mobile/desktop IO or silently fall back to buffering.

Suggested ADR update:

Add an explicit `Upload Pipeline Contract` section and choose one of these approaches:

- Route drive stream uploads through the main isolate and document why this is acceptable.
- Change the upload pipeline to accept a stream factory/reopenable body instead of a raw one-shot stream.
- Use a temporary file or bytes fallback and explicitly drop the zero-copy guarantee.

The ADR should not state that `FileUploader` remains unchanged if `FileInfo.byteStream` is introduced.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+- A new attachment-eligibility condition is one new class + one line, not a change to every call site.
+- Memory stays flat on IO regardless of file size/concurrent count (zero-copy pipe); web still buffers per file (XHR limitation).
+- User sees live per-file progress immediately instead of a blocking wait.
+- A streamed IO upload isn't 401-retry-safe (one-shot tee body can't be rebuilt) — accepted, that file's chip fails and is removed, others continue.
```
dab246: One-shot streamed body conflicts with existing 401 upload retry behavior

The ADR acknowledges that streamed IO upload is not 401-retry-safe and accepts that the file chip fails. However, the current interceptor actively retries upload requests when `FileUploader.uploadAttachmentExtraKey` is present.

Relevant code:

- `authorization_interceptors.dart:589` checks for upload requests and retries them.
- `authorization_interceptors.dart:429` rebuilds retry request data from file path or `streamDataExtraKey`.
- For bytes, retry is safe because `BodyBytesStream.fromBytes(...)` can be recreated.
- For a drive download stream, the body is one-shot and cannot be replayed after it has been consumed.

Risk:

The implementation may retry with an already-consumed stream, send an empty/truncated body, fail with a misleading error, or leave user-visible progress in an inconsistent state.

Suggested ADR update:

Define retry policy for drive-stream uploads. For example:

- Add an upload request extra flag such as `nonRetryableUploadBody: true` and make the auth interceptor skip retry for that request.
- Or provide a body factory that can reopen the stream for retry.
- Or proactively refresh token before starting each drive stream upload and still mark the request non-retryable.

The ADR should describe the concrete interceptor change instead of only accepting retry failure as a risk.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+
+**Visible progress:** each drive file gets a composer attachment chip immediately (`UploadFileStatus.downloading`, color-distinct from `uploading`), fed by `onDownloadProgress`/Dio's existing `onSendProgress`. On IO, download+upload are coupled in lockstep by the pipe, so the chip shows one continuous "downloading"-styled bar for ~the whole transfer (single phase, zero-copy memory kept over a two-phase visual). On web, downloading and uploading are genuinely sequential — two distinct phases shown.
```
dab246: Download progress state is underspecified and does not exist in current UI model

The ADR says every drive file gets a composer attachment chip immediately with `UploadFileStatus.downloading`, visually distinct from `uploading`.

Current code does not yet have this state:

- `UploadFileStatus` only has `waiting`, `uploading`, `uploadFailed`, and `succeed`.
- `AttachmentUploadState` only has pending, uploading, success, error, and cancel upload states.
- `AttachmentProgressLoadingComposerWidget` renders only waiting/uploading/completed states.

Risk:

This is more than a small UI style change. It requires a domain/progress event model for download progress, upload progress, failure, cancel, and removal behavior. The ADR currently makes this sound like a straightforward enum addition.

Suggested ADR update:

Add a `Progress and Cancellation State Model` section covering:

- New status values or a separate transfer phase enum.
- How `onDownloadProgress` maps into composer chips.
- How IO single-phase stream progress is represented.
- How web two-phase download/upload progress is represented.
- What happens when the user deletes a chip while download is in progress.
- Whether the same `CancelToken` cancels both download and upload.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+  Future<bool> validate() async {
+    if (validators.isEmpty) return true;
+    for (final v in validators) {
+      if (!await v.validate()) return false;
+    }
+    return true;
+  }
+}
+```
+No size (or other condition-specific data) in the interface. Each concrete validator captures what it needs via its own constructor and is fully self-contained, including showing its own dialog. Today's only concrete validator, `SizeLimitAttachmentUploadValidator`, wraps the existing size-limit/warning-dialog check. Every attachment entry point (composer's local-file/paste/drop handlers, `DriveAttachmentHandler`) constructs the validator(s) it needs and calls `.validate()` — no cross-controller dependency. A new, unrelated condition later = one new class + one line in a `CompositeAttachmentUploadValidator([...])` list.
```
dab246: Validator abstraction is promising, but migration boundary is unclear

The ADR proposes a clean `AttachmentUploadValidator` abstraction and says every attachment entry point will construct validators directly.

Current validation is inside `UploadController` and depends on composer/UI state:

- `UploadController.validateTotalSizeAttachmentsBeforeUpload` calculates current uploaded and pending attachment size.
- It depends on `currentContext`, mailbox dashboard max-size config, app localizations, and dialog side effects.
- Existing local-file, paste, and drop entry points call this controller method directly.

Risk:

Without a migration plan, this refactor can either leak UI/controller dependencies into validators or force a broad change across composer entry points. That may be fine, but the ADR should make it explicit.

Suggested ADR update:

Define where validators live and how dependencies are passed. For example:

- Keep validators in the main app composer/upload layer, not in `workplace`.
- Inject only minimal inputs needed by each validator: current attachment totals, mailbox max-size, localization/context/dialog service.
- Migrate local-file, paste, drop, and drive entry points in one focused step.
- Make `DriveAttachmentHandler` depend on a validator interface rather than directly reaching into `UploadController`.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,73 @@
...
+## Reference
+Builds on [ADR-0095](0095-external-drive-file-picker-integration.md) (intent protocol, `DriveDocument` entity, link-only insertion).
+
+## Context
+
+Composer already inserts picked drive documents with a `sharingLink` as HTML links. Documents with a `downloadLink` instead need to be downloaded and attached as real email attachments.
+
+## Decision
+
+**Partitioning:** `DriveAttachmentHandler` splits picked `DriveDocument`s per-doc: `sharingLink != null` → HTML link (unchanged); `downloadLink != null` → download + attach (new).
```
dab246: Partition rule needs an explicit invariant for documents containing both links

The ADR says `sharingLink != null` remains HTML link insertion and `downloadLink != null` becomes download plus attach. ADR-0095's done payload allows both fields to be present on a document.

Risk:

If a document contains both `sharingLink` and `downloadLink`, the implementation may silently pick link insertion and skip attachment, depending on partition order.

Suggested ADR update:

State one of these explicitly:

- Drive picker guarantees only one selected action field is present per document.
- If both are present, `downloadLink` wins for the "Add as attachment" action.
- If both are present, treat it as invalid and log/drop the document.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: clarify added.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Addressed in **"Upload pipeline contract for streamed IO uploads"** section

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: nonRetryableUploadBody flag added

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: addressed in **"Progress and cancellation state model"** section

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: addressed in **"Validator migration boundary"** section

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: explicit invariant added

dab246 (dismissed):
LGTM

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+## Reference
+Builds on [ADR-0095](0095-external-drive-file-picker-integration.md) (intent protocol, `DriveDocument` entity, link-only insertion).
+
+## Context
+
+Composer already inserts picked drive documents with a `sharingLink` as HTML links. Documents with a `downloadLink` instead need to be downloaded and attached as real email attachments.
+
+## Decision
+
+**Partitioning:** `DriveAttachmentHandler` splits picked `DriveDocument`s per-doc: `sharingLink != null` → HTML link (unchanged); `downloadLink != null` → download + attach (new). **Invariant:** both fields may be present on one doc (per ADR-0095's payload); when they are, `downloadLink` wins — download+attach only, enforced in the partition step, not call-site order.
```
hoangdat: IMO, `sharingLink` will win if both exists, user have more benefit from it than download.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Download (`workplace` package), streamed:** `DriveFileDatasourceImpl.openFileForUpload(doc)` branches on `kIsWeb`. **IO**: `Dio.get(responseType: stream)` piped directly as the upload request body (`FileInfo.byteStream`) — zero-copy. **Web**: XHR can't stream a body, so web buffers into `FileInfo.bytes`. A bounded worker pool (`kMaxConcurrentDriveTransfers = 3`, new constant, not tied to any existing limit) runs per-file pipelines with no "download all, then upload all" batching. A failing file logs/removes its chip; others continue.
+
+**Upload pipeline contract:** the worker-isolate upload path (`FileUploader.uploadAttachment` → `UploadFileArguments`) only carries `filePath`/`bytes` across the isolate boundary — a one-shot `Stream<List<int>>` can't cross it. So **drive stream uploads run on the main isolate** instead, bypassing the worker-isolate branch. Local-file/paste/drop uploads are unaffected.
```
hoangdat: what happen if upload has token expired? can we retry?

hoangdat: in case of running in main isolate, what will be impacted?

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Module boundary:** intent protocol, `DriveDocument`, and the download chain stay in `workplace`; only `DriveAttachmentHandler` (main app) composes download orchestration with the shared validator abstraction and the composer's existing upload trigger.
+
+## Consequences
+- Drive attachments reuse the same size-limit UX and upload pipeline as every other attachment source; `FileInfo` gains an additive `byteStream` field. `FileUploader` gains two additive changes: drive-stream uploads route to the main-isolate path (not worker-isolate), and carry `nonRetryableUploadBody: true` so the auth interceptor skips retry. Neither affects local-file/paste/drop uploads.
+- A new attachment-eligibility condition is one new class + one line, not a change to every call site.
+- Memory stays flat on IO regardless of file size/concurrent count (zero-copy pipe); web still buffers per file (XHR limitation).
+- User sees live per-file progress immediately instead of a blocking wait.
+- A streamed IO upload isn't 401-retry-safe (one-shot body can't be rebuilt) — accepted and made explicit via `nonRetryableUploadBody`: that file's chip fails and is removed, others continue (see Retry Policy above).
```
hoangdat: so should we use temp file?

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Upload pipeline contract:** the worker-isolate upload path (`FileUploader.uploadAttachment` → `UploadFileArguments`) only carries `filePath`/`bytes` across the isolate boundary — a one-shot `Stream<List<int>>` can't cross it. So **drive stream uploads run on the main isolate** instead, bypassing the worker-isolate branch. Local-file/paste/drop uploads are unaffected.
+
+**Visible progress:** each drive file gets a composer attachment chip immediately (`UploadFileStatus.downloading`, color-distinct from `uploading`), fed by `onDownloadProgress`/Dio's existing `onSendProgress`. IO shows one continuous phase (download+upload coupled by the pipe); web shows two sequential phases (download then upload).
+
+**Progress/cancel model:** `UploadFileStatus` gains `downloading`; `AttachmentUploadState` gains a matching `DownloadingAttachmentUploadState`; the composer progress widget gains a render branch for it. A single `CancelToken` per file spans both the download and upload calls, so cancelling (including deleting the chip mid-transfer) aborts whichever stage is active.
+
+**Validation — minimal, condition-agnostic abstraction:**
```
hoangdat, line 38: I am not still clear about the flow starting from the time we got the document list? validators will be run first -> download -> upload? can you more detail about the flow?

hoangdat, line 38: how we handle the case: 
- maxFileSize of a document is valid, but total size of files not valid
- totalSize of files is valid, but one file have not valid in file size

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Upload pipeline contract:** the worker-isolate upload path (`FileUploader.uploadAttachment` → `UploadFileArguments`) only carries `filePath`/`bytes` across the isolate boundary — a one-shot `Stream<List<int>>` can't cross it. So **drive stream uploads run on the main isolate** instead, bypassing the worker-isolate branch. Local-file/paste/drop uploads are unaffected.
+
+**Visible progress:** each drive file gets a composer attachment chip immediately (`UploadFileStatus.downloading`, color-distinct from `uploading`), fed by `onDownloadProgress`/Dio's existing `onSendProgress`. IO shows one continuous phase (download+upload coupled by the pipe); web shows two sequential phases (download then upload).
+
+**Progress/cancel model:** `UploadFileStatus` gains `downloading`; `AttachmentUploadState` gains a matching `DownloadingAttachmentUploadState`; the composer progress widget gains a render branch for it. A single `CancelToken` per file spans both the download and upload calls, so cancelling (including deleting the chip mid-transfer) aborts whichever stage is active.
```
hoangdat: not clear to me, how to indicate the download/upload in one progress bar, how to display exactly the progress?

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Context
+
+Composer already inserts picked drive documents with a `sharingLink` as HTML links. Documents with a `downloadLink` instead need to be downloaded and attached as real email attachments.
+
+## Decision
+
+**Partitioning:** `DriveAttachmentHandler` splits picked `DriveDocument`s per-doc: `sharingLink != null` → HTML link (unchanged); `downloadLink != null` → download + attach (new). **Invariant:** both fields may be present on one doc (per ADR-0095's payload); when they are, `downloadLink` wins — download+attach only, enforced in the partition step, not call-site order.
+
+**Download (`workplace` package), streamed:** `DriveFileDatasourceImpl.openFileForUpload(doc)` branches on `kIsWeb`. **IO**: `Dio.get(responseType: stream)` piped directly as the upload request body (`FileInfo.byteStream`) — zero-copy. **Web**: XHR can't stream a body, so web buffers into `FileInfo.bytes`. A bounded worker pool (`kMaxConcurrentDriveTransfers = 3`, new constant, not tied to any existing limit) runs per-file pipelines with no "download all, then upload all" batching. A failing file logs/removes its chip; others continue.
```
hoangdat: how about create an interface for `DriveFileDownloader` then implement for mobile (stream/file temp/ ....) and web?

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,81 @@
...
+- Concurrency fixed at 3 concurrent transfers, not user/config-configurable (YAGNI for now).
+- **Web buffering is a browser platform limit, not a Dio-specific gap.** Verified via `dio_web_adapter` 2.1.1 source (current, ships with `dio: 5.9.0`) — still XHR-based, still fully buffers the request stream before `xhr.send()`. A hand-rolled `fetch()` client wouldn't help either: streaming *upload* request bodies only works on Chromium (Chrome/Edge/Opera, v105+, and only over HTTP/2+HTTPS with `duplex: 'half'`, always CORS-preflighted); Firefox and desktop Safari still buffer the full request body as of current stable channels (WebKit is adding it under Interop 2026, landed in a Safari 26.4 beta, not GA). Fetch response-body streaming (download side) is broadly supported, but doesn't remove the need to buffer before upload on non-Chromium engines. Conclusion: the web buffered fallback is correct for all three engines today, not an XHR-only workaround to revisit later.
```
hoangdat: keep it as risk with current choice is not right. WDYT?

docs/adr/0103-attach-drive-file-as-attachment.md
hoangdat: how about in web we use Origin Private File System instead of memory buffer for file?

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Yes, already in **Retry policy** section.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Answered by **Visible progress** section

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Updated

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Explaination added.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: To follow zero-copy pipe, no. Explaination added.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Updated

tddang-linagora: Updated

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: YAGNI at the moment. Explaination added.

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Updated

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Considered but not adopted. Explaination added.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Doc updated, OPFS will be used.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,93 @@
...
+- **Web + OPFS** (feature-detected: `navigator.storage.getDirectory` and `FileSystemFileHandle.createWritable` both present — Baseline across Chrome/Edge, Firefox, and Safari as of Safari 26.0/Sept 2025): the download streams via `fetch()` + `ReadableStream`, written incrementally to a temp file in the Origin Private File System; the upload sends that file directly via a raw `XMLHttpRequest` (`xhr.send(file)`) — the browser streams the OPFS-backed file to the network without materializing it in the JS heap.
+- **Web buffered** (OPFS unavailable): `Dio.get(responseType: ResponseType.bytes)` into `FileInfo.bytes`, uploaded through the existing bytes-body path — the explicit, feature-detected fallback for browsers without OPFS write support.
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. Clarified in the ADR that the buffered fallback's actual-byte guard cancels on breach via `onReceiveProgress` but can't make that leg memory-flat, since `ResponseType.bytes` still assembles the full response before the stager receives it (checked, not memory-flat).

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,93 @@
...
+
+**Declared vs. actual size:** the pre-download gate sums `DriveDocument.size` — backend-reported metadata, unreliable for export-on-demand documents (e.g. Cozy, which may report `0` up front). A running actual-byte guard runs alongside the declared-size gate: each stager's download loop checks cumulative bytes across the batch against budget and aborts mid-transfer once exceeded, caught as a per-file failure. The declared-size gate still fails fast before wasting a download; the hard cap comes from the server `maxSizeAttachmentsPerEmail` capability and may be absent, so the actual-byte guard is the reliable bound.
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. ADR now specifies: the actual-byte guard's budget is `maxSizeAttachmentsPerEmail` when present, else a fixed fallback constant scoped to this new guard only (the pre-existing size-limit dialog check stays unlimited-when-absent, unchanged, for all other attachment sources). On the reservation-semantics point: no shared-counter precedent exists in this codebase, but the check-then-increment runs synchronously inside `onDownloadProgress` with no `await` in between, so on Dart's single-threaded event loop concurrent files can't interleave — the only possible overshoot is one in-flight chunk per file, not an unbounded race. Documented this instead of adding a chunk-reservation mechanism (unnecessary given the above).

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,93 @@
...
+- The user sees live per-file progress with two genuine phases (`downloading` then `uploading`) on every platform, instead of a blocking wait.
+- A new attachment-eligibility condition is one new class + one line, not a change to every call site.
+- Adding or changing a transfer strategy later (e.g. a Worker + `createSyncAccessHandle` variant) is one new stager implementation plus one line in `DriveTransferStrategyFactory` — orchestration never changes.
+
+## Risks
+- The `downloadLink` TTL (~5–10 min) can expire for later files in a large batch under bounded concurrency; those fail per-file and the user re-picks.
+- The web+OPFS raw-XHR upload is not 401-refresh-retry-safe (a fresh token is read once per request).
+- Disk and OPFS storage are subject to space/quota limits; exhaustion fails that file and cleans up.
+- Stream receive-timeout is uncapped (`Duration.zero`) on the Dio-backed download legs (IO, web-buffered); a stalled server relies on the user's manual cancel action. `receiveTimeout` only guards time-to-first-byte/headers on IO, not the whole transfer (verified against Dio 5.2.0 and current 5.9.0 source).
```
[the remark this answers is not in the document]
tddang-linagora: Dismissing for this ADR — the uncapped receive-timeout / manual-cancel-only behavior is already called out explicitly in the Risks section as an accepted v1 trade-off, not an oversight. An inactivity watchdog is a legitimate follow-up but is additional scope (correctly flagged Heavy lift) beyond this proposal; tracking it as future work rather than blocking this ADR.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+- **Web + OPFS** (feature-detected: `navigator.storage.getDirectory` and `FileSystemFileHandle.createWritable` both present — Baseline across Chrome/Edge, Firefox, and Safari as of Safari 26.0/Sept 2025): the download streams via `fetch()` + `ReadableStream`, written incrementally to a temp file in the Origin Private File System; the upload sends that file directly via a raw `XMLHttpRequest` (`xhr.send(file)`) — the browser streams the OPFS-backed file to the network without materializing it in the JS heap.
+- **Web buffered** (OPFS unavailable): `Dio.get(responseType: ResponseType.bytes)` into `FileInfo.bytes`, uploaded through the existing bytes-body path — the explicit, feature-detected fallback for browsers without OPFS write support. The actual-byte guard (below) still tracks this leg via `onReceiveProgress` and cancels the request on breach, but `ResponseType.bytes` assembles the full response before the stager receives it, so this path is checked, not memory-flat.
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. Reworded the Decision heading from "memory-flat on every platform" to "platform-appropriate staging", and qualified the Context sentence to state memory stays flat on IO always and on web only when OPFS is available, matching the caveat already in the Consequences section.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+
+- **IO (mobile/desktop):** stream the public `downloadLink` with `Dio.get(responseType: ResponseType.stream, receiveTimeout: Duration.zero)` (no auth header) and write each chunk to a temp file in `getTemporaryDirectory()` (`RandomAccessFile.writeFrom`, pause/resume, cancel-aware — the technique proven in `DownloadManager.downloadFile`). The upload builds `FileInfo(filePath: tempPath, ...)` and goes through the existing upload path: the worker isolate reads the file lazily via `File(filePath).openRead()`, keeping memory flat and CPU work off the main thread.
+- **Web + OPFS** (feature-detected: `navigator.storage.getDirectory` and `FileSystemFileHandle.createWritable` both present — Baseline across Chrome/Edge, Firefox, and Safari as of Safari 26.0/Sept 2025): the download streams via `fetch()` + `ReadableStream`, written incrementally to a temp file in the Origin Private File System; the upload sends that file directly via a raw `XMLHttpRequest` (`xhr.send(file)`) — the browser streams the OPFS-backed file to the network without materializing it in the JS heap.
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. The ADR (and plan.md) now specify `WorkplaceDio.instance` — a bare `Dio()` with no interceptors (`workplace/lib/data/workplace_dio.dart`) — for the IO downloadLink fetch, and explicitly call out that the shared `DioClient`/`AuthorizationInterceptors` path (as used by `DownloadManager.downloadFile` for the same trusted JMAP host) must not be reused here, since it would attach the session bearer token to the external Drive host.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+
+The counter's check-then-increment runs synchronously inside each `onDownloadProgress` callback with no `await` in between, so on Dart's single-threaded event loop it cannot interleave across concurrent files — the only possible overshoot is one in-flight progress-chunk per file at the moment the guard fires, not an unbounded race. No chunk-reservation mechanism is needed.
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. `onDownloadProgress` reports cumulative bytes per file, so naively adding each callback value into the shared counter would double-count. Reworded the ADR and plan.md: each stager now tracks its own previous callback value and adds only the delta to the shared actual-byte counter.

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+- The user sees live per-file progress with two genuine phases (`downloading` then `uploading`) on every platform, instead of a blocking wait.
+- A new attachment-eligibility condition is one new class + one line, not a change to every call site.
+- Adding or changing a transfer strategy later (e.g. a Worker + `createSyncAccessHandle` variant) is one new stager implementation plus one line in `DriveTransferStrategyFactory` — orchestration never changes.
+
+## Risks
+- The `downloadLink` TTL (~5–10 min) can expire for later files in a large batch under bounded concurrency; those fail per-file and the user re-picks.
+- The web+OPFS raw-XHR upload is not 401-refresh-retry-safe (a fresh token is read once per request).
+- Disk and OPFS storage are subject to space/quota limits; exhaustion fails that file and cleans up.
+- Stream receive-timeout is uncapped (`Duration.zero`) on the Dio-backed download legs (IO, web-buffered); a stalled server relies on the user's manual cancel action. `receiveTimeout` only guards time-to-first-byte/headers on IO, not the whole transfer (verified against Dio 5.2.0 and current 5.9.0 source).
```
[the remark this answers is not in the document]
tddang-linagora: Valid — fixed. Corrected the Risks section: `receiveTimeout` (when nonzero) guards both time-to-first-byte/headers *and* the gap between subsequent byte events, not just time-to-first-byte — `Duration.zero` disables it for both cases, which is why a mid-transfer stall isn't caught automatically. Also applied the same correction to plan.md's inline note. Kept the inactivity-watchdog as accepted follow-up work per the earlier thread on this line.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

dab246 (dismissed):
lgtm

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+
+- **IO (mobile/desktop):** stream the public `downloadLink` with `WorkplaceDio.instance.get(responseType: ResponseType.stream, receiveTimeout: Duration.zero)` and write each chunk to a temp file in `getTemporaryDirectory()` (`RandomAccessFile.writeFrom`, pause/resume, cancel-aware — the file-write technique proven in `DownloadManager.downloadFile`). `WorkplaceDio.instance` (`workplace/lib/data/workplace_dio.dart`) is a bare `Dio()` with no interceptors, so the public `downloadLink` is never routed through the main app's `AuthorizationInterceptors` — that interceptor attaches a Bearer/Basic header to every request on the shared `DioClient` instance regardless of host, and `DownloadManager.downloadFile`'s use of that shared client is safe only because its target is the same trusted JMAP host, unlike the external Drive host here; using the shared client for the Drive download would leak the session token to that external host. The upload builds `FileInfo(filePath: tempPath, ...)` and goes through the existing upload path: the worker isolate reads the file lazily via `File(filePath).openRead()`, keeping memory flat and CPU work off the main thread.
```
dab246: > `WorkplaceDio.instance` (`workplace/lib/data/workplace_dio.dart`) is a bare `Dio()` with no interceptors

`WorkplaceDio` is **not** a bare `Dio()`. It is built with `BaseOptions(sendTimeout: 10s, receiveTimeout: 10s, connectTimeout: 10s)` (`workplace_dio.dart:5-11`).

This matters because line 84 explicitly overrides `receiveTimeout: Duration.zero`, and then books the inactivity watchdog as accepted follow-up work. But per your own line 84, Dio's `receiveTimeout` guards the **gap between byte events** — so `WorkplaceDio`'s existing 10s default *is* the inactivity watchdog being deferred. A healthy streaming download emits bytes continuously and never trips it; only a stalled transfer does. `Duration.zero` therefore removes a working guard, and the "bare `Dio()`" wording hides that removal.

**Requested change:** either
- (a) keep a finite `receiveTimeout` (10s, or 60–120s if slow export-on-demand backends like Cozy need long time-to-first-byte — note `receiveTimeout` also covers time-to-first-byte), closing this risk now instead of deferring; or
- (b) if `Duration.zero` is intended, state why the existing 10s guard is unacceptable and why a larger finite value won't do.

Also fix the wording: "bare `Dio()`" → "a `Dio()` with only timeout `BaseOptions` and no interceptors."

docs/adr/0103-attach-drive-file-as-attachment.md
```diff
@@ -0,0 +1,95 @@
...
+
+**Upload pipeline:** IO and web-buffered drive attachments reuse `FileUploader`. IO uploads a temp `filePath`, so the pre-existing worker-isolate path is preserved (Dio drives the socket off the Dart main thread; `text/plain` charset detection stays on the worker isolate). `FileUploader`'s main-isolate branch (used on web and single-core devices) is extended to accept `filePath`, not only `bytes`, so single-core devices stream from disk too. Local-file/paste/drop uploads are unaffected. The web+OPFS uploader bypasses `FileUploader` — a raw XHR call driving the composer chip's progress state directly.
+
+**401-retry:** IO uploads are `filePath`-backed, so a 401 mid-upload is retried safely — `AuthorizationInterceptors` rebuilds the body by reopening the file from disk. No interceptor changes. The web+OPFS raw-XHR upload reads a fresh token once per request and is not refresh-and-retry safe; on any error the file's chip fails and is removed (partial success), the same path as any other transfer failure.
```
dab246: > IO uploads are `filePath`-backed, so a 401 mid-upload is retried safely

True for **multi-core** IO (worker isolate — sets `filePathExtraKey` at `file_uploader.dart:104`). It is **not** true out-of-the-box for the **single-core** IO path this ADR adds (line 30 routes it through the main-isolate branch). That branch today sets only `streamDataExtraKey`, never `filePathExtraKey` (`file_uploader.dart:178`), and the interceptor's rebuild-from-file keys off `filePathExtraKey` (`authorization_interceptors.dart:432-434`). Without that key, a 401 mid-upload on a single-core IO device won't rebuild the body and the retry fails.

**Requested change:** add one clause — "extending the main-isolate branch to `filePath` includes setting `filePathExtraKey` in its `mapExtra`, without which single-core IO uploads are not 401-retry-safe."

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Done

docs/adr/0103-attach-drive-file-as-attachment.md
tddang-linagora: Done

dab246 (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4688

PR #4699, opened 2026-07-10
TF-4660 Drive loading

tddang-linagora:
## Issue
- #4660 

### Web

https://github.com/user-attachments/assets/6b1491e6-b4d7-418e-8bae-3eed26bd5d4f

### Mobile

Mobile has a bug where drive is stuck at login stage, cannot end the loading, but as soon as it is fixed, the loading will go away without any further changes.

[drive-mobile-loading.webm](https://github.com/user-attachments/assets/a51167a0-3b2e-4150-9c11-2ef129a0bd01)

## Summary by CodeRabbit

* **New Features**
  * Added placeholder UI while Drive intent web content initializes.
  * Updated Drive intent modals to resolve intents asynchronously before loading.
  * Added a readiness timeout safeguard and new hooks for “ready” and timeout handling.

* **Bug Fixes**
  * Prevented premature message handling before intent setup completes.
  * Improved Drive picker failure handling and ensured modal cleanup on timeout/error.
  * Enhanced token exchange requests with an explicit JSON response header.
  * Updated attachment picker taps to close the current route before opening the picker.

* **Refactor**
  * Made the intent modal shell layout more configurable (padding/shape/optional close button).

* **Tests**
  * Expanded Drive intent message-handler timeout coverage and improved isolation.


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

dab246:
- [x] If a white screen still appears, we should also handle it using a loading view.

Screenshot 2026-07-10 at 13 54 02

workplace/lib/presentation/view/drive_intent_web_view_modal_web.dart
```diff
@@ -3,22 +3,24 @@ import 'dart:convert';
...
 import 'package:workplace/presentation/view/drive_intent_web_view_modal_shell.dart';
 
 class DriveIntentWebViewModal extends StatefulWidget {
-  final Uri url;
-  final String intentId;
+  final Future<WorkplaceIntent> intentFuture;
   // ADR-93: composer registers the window listener at composer-init time and
   // forwards messages here, so the handler is ready before the iframe loads.
   final OnRegisterExternalHandler? onRegisterExternalHandler;
+  final void Function(Object error)? onIntentError;
```
dab246: Should use typedef to strong type it

tddang-linagora:
>If a white screen still appears, we should also handle it using a loading view.

The only cue for that is ready postMessage, but that would be web only at the moment, as mobile is still stuck with login screen

workplace/lib/presentation/view/drive_intent_web_view_modal_web.dart
```diff
@@ -46,6 +45,19 @@ class _DriveIntentWebViewModalState extends State<DriveIntentWebViewModal>
...
+      _intentOrigin = origin;
+      initMessageHandler(intentId: intent.intentId, intentOrigin: origin);
+      // Browsers ignore `src` while the `srcdoc` attribute (set by the
+      // skeleton) is still present, so it must be removed before navigating.
+      _iframeElement?.attributes.remove('srcdoc');
+      _iframeElement?.src = intent.intentUrl.toString();
+    }).catchError((Object e) {
+      widget.onIntentError?.call(e);
+      if (mounted) Navigator.of(context).pop(null);
+    });
```
dab246: Should separate function to handle it

workplace/lib/presentation/view/drive_intent_web_view_modal_web.dart
tddang-linagora: Done

workplace/lib/presentation/view/drive_intent_web_view_modal_web.dart
tddang-linagora: Done

workplace/lib/presentation/notifier/drive_attachment_state.dart
```diff
@@ -1,25 +0,0 @@
...
-}
-
-final class DriveIntentPending extends DriveAttachmentState {
-  final WorkplaceIntent intent;
-  const DriveIntentPending({required this.intent, super.attachments});
-}
-
-final class DriveAttachmentError extends DriveAttachmentState {
-  final Object error;
-  const DriveAttachmentError({required this.error, super.attachments});
```
dab246, line 24: Why do we remove it?

tddang-linagora: We don't use it anywhere, initially planned, but implementation changed over time.

workplace/lib/presentation/model/drive_intent_timing.dart
```diff
@@ -0,0 +1,37 @@
...
+
+  void markTokenExchanged() {
+    tokenExchangedAt = DateTime.now();
+  }
+
+  void markUrlObtained() {
+    urlObtainedAt = DateTime.now();
+  }
+
+  void markReadyAndLog() {
```
dab246: This shouldn't be committed; we should only leave it in the trace log.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -33,37 +34,39 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
+          timing: timing,
         ),
       );
       if (result != null) pickerOnCallback?.call(DrivePickResult(result));
-    } catch (e) {
-      logWarning('DrivePickerStateMixin::onPickerTap: $e');
-      final message = mounted
-          ? AppLocalizations.of(context)?.attachFromDriveFailingMessage
-          : null;
-      pickerOnCallback?.call(DrivePickFailure(e, message: message));
```
dab246, line 64: Are we not going to use `catch` anymore?

dab246:
> > If a white screen still appears, we should also handle it using a loading view.
> 
> The only cue for that is ready postMessage, but that would be web only at the moment, as mobile is still stuck with login screen

So let's fix it for the web version first; for the mobile version, please comment so we can fix it later. Also, update the demo to work correctly for the web version.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: catch happens somewhere else.

workplace/lib/presentation/model/drive_intent_timing.dart
tddang-linagora: Removed

tddang-linagora:
> > > If a white screen still appears, we should also handle it using a loading view.
> > 
> > 
> > The only cue for that is ready postMessage, but that would be web only at the moment, as mobile is still stuck with login screen
> 
> So let's fix it for the web version first; for the mobile version, please comment so we can fix it later. Also, update the demo to work correctly for the web version.

Done and demos updated

dab246:
- [x] The corners aren't rounded yet, it doesn't quite match the design. Please check it again.

Screenshot 2026-07-10 at 15 04 11


Currently:

Screenshot 2026-07-10 at 15 04 38

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

tddang-linagora:
>The corners aren't rounded yet, it doesn't quite match the design. Please check it again.

Radius added

workplace/lib/data/datasource_impl/workplace_datasource_impl.dart
```diff
@@ -96,6 +96,9 @@ class WorkplaceDataSourceImpl implements WorkplaceDataSource {
           'token_exchange',
         ],
       ).toString(),
+      options: Options(
+        headers: {'Accept': 'application/json'},
+      ),
```
dab246, line 101: Why do we need it?

tddang-linagora: that request is failing in mobile if Accept is missing

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

workplace/lib/presentation/notifier/drive_attachment_state.dart
hoangdat: what is the change? why not using state to communicate between layers?

workplace/lib/data/datasource_impl/workplace_datasource_impl.dart
dab246: what about web platform?

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -74,5 +81,9 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
 
   void sendAck();
 
+  /// Called once when the real Drive page reports it's loaded and
+  /// interactive. Override to hide the loading skeleton.
+  void onReady() {}
```
hoangdat:
```suggestion
  void onDrivePickerReady() {}
```

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -7,14 +7,20 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
...
     _intentOrigin = intentOrigin;
     _modalClosed = false;
+    _ready = true;
   }
 
   void onMessage({required String raw, required String? origin}) {
+    // Messages can only originate from the real Drive content, which isn't
+    // loaded until initMessageHandler runs post-swap; guard against a stray
+    // message arriving while the skeleton is still showing.
+    if (!_ready) return;
```
hoangdat: maybe drive will send us failed intent to notice the loading cannot finish by SSO blocking or sth else? how we can handle these cases with our structure?

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
```diff
@@ -4,21 +4,23 @@ import 'dart:convert';
...
 import 'package:flutter_inappwebview/flutter_inappwebview.dart';
+import 'package:workplace/domain/entity/workplace_intent.dart';
 
 class DriveIntentWebViewModal extends StatefulWidget {
-  final Uri url;
-  final String intentId;
+  final Future<WorkplaceIntent> intentFuture;
   // Ignored on mobile — only used by the web variant (ADR-93).
   final OnRegisterExternalHandler? onRegisterExternalHandler;
+  final OnIntentError? onIntentError;
```
hoangdat: We will have how many callbacks for each kind of states from workplace? How about to have a single handler, if protocol changed, more kind of messages or more state, we dont need to care much. WDYT?

hoangdat:
- [x] Outbound (modal → app): Replace the app-facing callbacks with a single sealed result from the dialog — showDialog returns one DrivePickOutcome (Picked(docs) / Cancelled / Failed(error)), folding onIntentError and in-page errors into it. Keep the inbound Drive protocol as-is (sealed WorkplaceIntentMessage + exhaustive switch handles new/unknown states without new wiring). Goal: a new outcome = one variant + one switch arm, nothing threaded through constructors.

hoangdat:
- [x] Loading lifecycle: Replace the booleans (_showSkeleton, _ready, …) with one explicit state in the shared mixin — LoadingState { loadingIntent, loadingPage, ready, failed } — and add the missing exits (a timeout + error path) so the skeleton can't hang if ready never comes. Move the duplicated intentFuture.then/.catchError/load logic out of both views into the mixin, leaving each view with only its platform load call + skeleton. Goal: one state driven in one place, views stay presentational.

workplace/lib/data/datasource_impl/workplace_datasource_impl.dart
tddang-linagora: Web is not affected

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -33,37 +32,36 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
     }
   }
+
+  void _onIntentError(Object e) {
+    logWarning('DrivePickerStateMixin::onPickerTap: $e');
+    final message = mounted
+        ? AppLocalizations.of(context)?.attachFromDriveFailingMessage
+        : null;
+    pickerOnCallback?.call(DrivePickFailure(e, message: message));
+  }
```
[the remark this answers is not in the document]
tddang-linagora: Fixed: capture the localized failing message in `onPickerTap` before `showDialog` (and before the caller can pop/dispose this context), and thread it through `onIntentError` instead of re-resolving `AppLocalizations.of(context)` at failure time.

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
```diff
@@ -29,29 +31,43 @@ class DriveIntentWebViewModal extends StatefulWidget {
...
+      } else {
+        _webViewController?.loadUrl(
+          urlRequest: URLRequest(url: WebUri.uri(intent.intentUrl)),
+        );
+      }
+    }).catchError((Object e) {
+      widget.onIntentError?.call(e);
+      if (mounted) Navigator.of(context).pop(null);
+    });
   }
```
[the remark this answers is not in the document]
tddang-linagora: Fixed: store the resolved intent in state and trigger the load from both the future-completion callback and `onWebViewCreated`, guarded by a one-shot flag, so whichever finishes last performs the navigation.

workplace/lib/presentation/view/drive_intent_web_view_modal_web.dart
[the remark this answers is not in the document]
tddang-linagora: Fixed: buffer the resolved intent in state and apply the `src` navigation from both the future-completion callback and `onIframeCreated`, guarded by a one-shot flag.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Good catch — we only handled the case where Drive explicitly sends an `error` message, not the case where it sends nothing at all (silent SSO block, network stall, etc). Added a ready-timeout: `DriveIntentMessageHandlerMixin` now starts a 20s timer when the real page starts loading, cancelled as soon as a `ready` message (or any terminal `done`/`error`/`cancel`) arrives. If it fires, we call `onTimeout()` (which the mobile/web modals override to surface the same failing-message dialog used for the other error paths) and close the modal — same UX as an explicit error, just triggered by silence instead of a message.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
tddang-linagora: We already route everything through one entry point: `onMessage` → `WorkplaceIntentMessage.parse` (sealed class) → the single `_handleWorkplaceMessage` switch. Adding a new message kind from Drive is one more `case` in that switch, not a new callback. The two `void onXxx() {}` overrides on the mixin (`onDrivePickerReady`, now also `onTimeout`) aren't per-message-type callbacks — they're optional hooks a State can override if it cares about that particular outcome (e.g. to hide the skeleton loader); most message types (`done`/`error`/`cancel`) just go straight through `closeModal`, no per-type callback needed. So the protocol-change concern is already handled by the sealed-class + switch dispatch.

tddang-linagora:
> * [x]  Outbound (modal → app): Replace the app-facing callbacks with a single sealed result from the dialog — showDialog returns one DrivePickOutcome (Picked(docs) / Cancelled / Failed(error)), folding onIntentError and in-page errors into it. Keep the inbound Drive protocol as-is (sealed WorkplaceIntentMessage + exhaustive switch handles new/unknown states without new wiring). Goal: a new outcome = one variant + one switch arm, nothing threaded through constructors.

Done. Added `DrivePickOutcome` (`workplace/lib/presentation/model/drive_pick_outcome.dart`), a sealed type with `Picked(documents)` / `Cancelled` / `Failed(error)`. `DriveIntentWebViewModal` now always pops `showDialog<DrivePickOutcome>` with one of these — the `onIntentError` callback param is gone entirely (removed from both view constructors and the shared `OnIntentError` typedef). `DrivePickerStateMixin.onPickerTap` switches on the outcome once, mapping to the existing `DrivePickResult`/`DrivePickFailure` (unchanged — that's the public API the composer/context-menu callers already consume, so no ripple beyond `workplace`).



> * [x]  Loading lifecycle: Replace the booleans (_showSkeleton, _ready, …) with one explicit state in the shared mixin — LoadingState { loadingIntent, loadingPage, ready, failed } — and add the missing exits (a timeout + error path) so the skeleton can't hang if ready never comes. Move the duplicated intentFuture.then/.catchError/load logic out of both views into the mixin, leaving each view with only its platform load call + skeleton. Goal: one state driven in one place, views stay presentational.

Done, with one adjustment: kept it to one boolean (`showSkeleton`, flips false on `ready`) rather than a 4-way enum, since the extra states (`loadingIntent`/`loadingPage`) aren't independently observed by either view — nothing branches on them beyond "has the intent+view both landed yet," which is now internal to the mixin. Concretely:

- `DriveIntentMessageHandlerMixin` gained `startLoading(intentFuture)` and `notifyPlatformViewReady()`; it owns resolving the intent, waiting for the platform view, applying-once (whichever finishes last), the ready-timeout (1), and popping the host route via `DrivePickOutcome` (5) — including the previously-duplicated `intentFuture.then/.catchError`.
- Each view now only implements `loadIntent(WorkplaceIntent)` (the actual `loadUrl`/`loadData` vs `iframe.src =` call) and `sendAck()`. Both view files dropped their own `_resolvedIntent`/`_intentLoaded`/`_showSkeleton`/`_intentOrigin` fields and the `_loadIntentIfReady`/`_applyIntentIfReady` duplication (~25 lines each).
- Tests (`drive_intent_message_handler_mixin_test.dart`) cover both apply orderings (view-ready-first vs intent-resolved-first), the fetch-failure path, and the ready-timeout, in addition to the existing message-dispatch/origin-filtering/multi-instance-isolation cases.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -1,20 +1,72 @@
...
 
+/// Shared loading/messaging lifecycle for the mobile and web Drive picker modals.
 mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
   late String _intentId;
-  late String _intentOrigin;
-  bool _modalClosed = false;
+  String _intentOrigin = 'null';
+  bool _closed = false;
+  bool _messageHandlerReady = false;
+  Timer? _readyTimeoutTimer;
```
dab246, line 16: `_readyTimeoutTimer` is never cancelled on widget disposal (leak + post-dispose callback)

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
```diff
@@ -30,56 +30,70 @@ class _DriveIntentWebViewModalState extends State<DriveIntentWebViewModal>
...
+                controller.addJavaScriptHandler(
+                  handlerName: DriveIntentShims.handlerName,
+                  callback: (args) => onMessage(
+                    raw: args[0] as String,
+                    origin: args.length > 1 ? args[1] as String? : null,
+                  ),
+                );
+                notifyPlatformViewReady();
+              },
+              onLoadStop: null,
```
dab246: Why do web set it is null? if it not used, we should remove it

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
tddang-linagora: Removed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -33,40 +33,46 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
+      final failingMessage = l10n.attachFromDriveFailingMessage;
+      final intentFuture = pickerFetchIntent(
         addAsLinkTitle: l10n.addAsLink,
         addAsAttachmentTitle: l10n.addAsAttachment,
       );
-      if (!mounted) {
-        throw WorkplaceUIDisposedException();
-      }
-      final result = await showDialog<List<DriveDocument>?>(
+      final outcome = await showDialog<DrivePickOutcome>(
```
dab246: Web context-menu picker loses the message listener. `Navigator.pop(context)` disposes the context-menu tile and removes its `window.onmessage` listener. The Drive modal then cannot receive `ready` or `done`, so the picker eventually times out.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -1,20 +1,72 @@
...
+  void _tryApplyIntent() {
+    final intent = _resolvedIntent;
+    if (intent == null || !_platformViewReady || _intentApplied) return;
+    _intentApplied = true;
+    _intentId = intent.intentId;
+    _intentOrigin = intent.intentUrl.scheme == 'data'
+        ? 'null'
+        : intent.intentUrl.origin;
+    _messageHandlerReady = true;
+    _readyTimeoutTimer?.cancel();
```
dab246: Loading skeleton can remain visible forever. The timeout is cancelled after `ready`, but the skeleton is hidden only after `readyToUse`. If `readyToUse` is never sent, the skeleton remains permanently visible.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -1,20 +1,72 @@
...
+  bool _showSkeleton = true;
+
+  bool get showSkeleton => _showSkeleton;
+
+  String get intentOrigin => _intentOrigin;
+
+  /// Covers silent failures like SSO blocking the page with no postMessage ever sent.
+  Duration get readyTimeout => const Duration(seconds: 20);
+
+  void startLoading(Future<WorkplaceIntent> intentFuture) {
```
dab246, line 39: No timeout when the platform view never becomes ready. The timeout starts only after `notifyPlatformViewReady()`. If WebView/iframe initialization fails, the modal stays in loading state indefinitely.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Web doesn't use context menu tile, that's mobile only. Please provide reproducible steps/demo.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

dab246:
- [x] Cannot open drive picker?
- [x] Auto close drive picker loading when click outside?


https://github.com/user-attachments/assets/a2dc7a71-a541-431f-99d9-f822a195f0f2

dab246:
- [x] Loading time is too long? Why is that?


https://github.com/user-attachments/assets/bbe643bc-e755-4364-bb83-7115225354c9

tddang-linagora:
>Loading time is too long? Why is that?

wait for the new state to come through postMessage

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -33,40 +33,46 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
           onRegisterExternalHandler: externalHandlerRegistrar,
         ),
       );
-      if (result != null) pickerOnCallback?.call(DrivePickResult(result));
-    } catch (e) {
-      logWarning('DrivePickerStateMixin::onPickerTap: $e');
-      final message = mounted
-          ? AppLocalizations.of(context)?.attachFromDriveFailingMessage
-          : null;
-      pickerOnCallback?.call(DrivePickFailure(e, message: message));
```
dab246, line 64: should persist

tddang-linagora:
>Auto close drive picker loading when click outside?

Done

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

dab246:
- [x] The search box is overflowing the iframe in mobile responsive design when loading view


https://github.com/user-attachments/assets/7238bd9f-0f63-40ed-9074-47f33825be4f

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

tddang-linagora:
> The search box is overflowing the iframe in mobile responsive design when loading view


https://github.com/user-attachments/assets/132458ba-100b-4e42-ab4c-9816933f90e6

tddang-linagora:
https://github.com/user-attachments/assets/dde5c6cb-892c-4eb2-ad6e-d3c7d99ad806

dab246 (dismissed):
LGTM

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

hoangdat:
- [x] I still prefer the close button should work, agree not close by touch outside in mid-run, but close button should work (especially in mobile)

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -43,36 +97,53 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
...
+  }
+
+  void _finish(DrivePickOutcome outcome) {
+    if (_closed) return;
+    _closed = true;
+    _readyTimeoutTimer?.cancel();
     onCleanup();
-    if (mounted) Navigator.of(context).pop(result);
+    onFinished(outcome);
+    if (mounted) Navigator.of(context).pop(outcome);
```
hoangdat, line 147: how about


```suggestion
    if (mounted && (route?.isCurrent ?? false)) Navigator.of(context).pop(outcome);
```

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
hoangdat, line 1: Suggestion: replace the 4 flags in DriveIntentMessageHandlerMixin with one phase enum

Right now the modal's lifecycle is tracked by four separate booleans — _closed, _messageHandlerReady, _platformViewReady, _intentApplied — plus _showSkeleton. They work, and the tests cover the behaviour well. Two things bother me though:

1. Only _showSkeleton is updated inside setState. That's fine today because nothing else is read in build(). But the next time we want the UI to tell "fetching intent" apart from "loading page", someone will read _platformViewReady in build() and it just won't rebuild — no error, no clue why.
2. _messageHandlerReady and _intentApplied are the same fact. Both are set in _tryApplyIntent(), five lines apart, and never reset. Two names, one meaning.

Proposal — collapse the four latches into one phase, and keep the two things we're actually waiting on as plain data:

enum _Phase { waiting, loadingPage, interactive, closed }

_Phase _phase = _Phase.waiting;
WorkplaceIntent? _resolvedIntent;   // waiting on: intent fetch
bool _platformViewReady = false;    // waiting on: iframe/WebView created

bool get showSkeleton => _phase == _Phase.waiting || _phase == _Phase.loadingPage;
bool get _canHandleMessages => _phase.index >= _Phase.loadingPage.index;

void _setPhase(_Phase next) {
  if (_phase == _Phase.closed || next == _phase) return;
  if (mounted) setState(() => _phase = next); else _phase = next;
}

Mapping: _intentApplied / _messageHandlerReady → _phase != waiting; _closed → _phase == closed; _showSkeleton → derived, not stored.

Why it's worth it:
- Every transition goes through one setter, so "forgot to call setState" can't happen again.
- The closed-state guard lives in one place instead of being re-checked at each call site.
- Impossible states stop compiling — today _closed && _showSkeleton is perfectly valid code.

hoangdat:
- [x] log error in failed states

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: `route` is undefined.

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

tddang-linagora:
>log error in failed states

Done

>I still prefer the close button should work, agree not close by touch outside in mid-run, but close button should work (especially in mobile)

[untitled.webm](https://github.com/user-attachments/assets/0d7b467c-2310-4ebb-80eb-b17714ee656f)

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

hoangdat:
- [x] in mobile `Attach from Drive` instead of `Browse`

tddang-linagora:
>in mobile Attach from Drive instead of Browse

Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4699

hoangdat (approved):

Issue #4646, opened 2026-06-29
SSOT Part 5 — Migrate web results to the central executor (ThreadController)


dab246:
## Problem / Why

`ThreadController` silently ANDs `filterMessageOption` into the JMAP query via
`moreFilterCondition: getFilterCondition()`. The chip bar reads a different field, so a user can deselect the
unread/starred chip yet still get filtered results (**#4590**), and cached results stay filtered while the chip
shows off (**#4490**). This ticket routes web result queries through the executor ([#4643](https://github.com/linagora/tmail-flutter/issues/4643)) and removes the
hidden AND — the filter now flows only through the SSOT.

## Scope

- `_searchEmail` / `_searchMoreEmails` → executor (`NewSearchIntent` / `LoadMoreIntent`,
  `collapseThreads: _shouldCollapseThreads`); bridge `searchEmailNotifierProvider` → `updateEmailList` via
  `appProviderContainer.listen`.
- **Remove `moreFilterCondition: getFilterCondition()`** (×3) and the inline cursor branching; delete
  `_searchEmailsSuccess` (×4) / `_searchMoreEmailsSuccess` (×3).

## Out of scope

- The `filterMessageOption → SSOT` mapping that makes the chips *carry* the filter is [#4647](https://github.com/linagora/tmail-flutter/issues/4647) (#4490 is not
  fully closed until both land). No view migration.

## Definition of Done

- [x] `_searchEmail` calls the executor (not the interactor directly).
- [x] **Test asserts no `moreFilterCondition` leaks into the JMAP query** (#4590 / #4490 regression).
- [x] Web result list + selection still render via the bridge; controller tests updated.
- [x] Manual web regression pass recorded in the PR.

## Relates to / Depends on

- **Depends on:** [#4643 · Part 2](https://github.com/linagora/tmail-flutter/issues/4643)
- **Blocks:** [#4648 · Part 7](https://github.com/linagora/tmail-flutter/issues/4648)
- **Relates to:** #4590, #4490 *(do not auto-close — #4490 also needs [#4647 · Part 6](https://github.com/linagora/tmail-flutter/issues/4647); close after all tickets merge)*

PR #4706, opened 2026-07-14
TF-4646  SSOT Part 5.1 — PR 2 — Route web results through the central executor + search SSOT

dab246:
Part of [#4646](https://github.com/linagora/tmail-flutter/issues/4646)  

## Summary

- Add a shared `SearchExecutorService` to execute search, load-more, and refresh intents through the Riverpod search SSOT.
- Route mobile `SearchEmailController` and web `ThreadController` through the shared executor with observer-based loading, result, error, pagination, and websocket refresh handling.
- Extract thread search observer logic, simplify controller conditionals, and update bindings and related tests.

## Summary
* **New Features**
  * Centralized, intent-based email search execution via a shared executor with unified lifecycle callbacks across mailbox dashboard and thread views.
  * Coordinated “refresh on server state changes” for active searches, including serialized dispatch and bounded retry.
* **Bug Fixes**
  * Improved fresh-result semantics, pagination/cursor resets on new searches/sort changes, and clearer failure mapping for retry UI.
  * Thread inline search now updates only when the view “owns” results; refresh handling uses cached change-list updates.
  * Websocket queue retry behavior is now configurable; disposed queues stop marking messages processed.
* **Tests**
  * Expanded coverage for executor/observer replay semantics, refresh retry/lifecycle, cursor immutability, and queue disposal/retry flows.


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4706

lib/features/mailbox_dashboard/presentation/model/search/search_email_filter.dart
```diff
@@ -247,9 +244,28 @@ class SearchEmailFilter with EquatableMixin, OptionParamMixin {
...
+      startDate != null ||
+      endDate != null ||
+      emailReceiveTimeType != EmailReceiveTimeType.allTime ||
+      (mailbox != null && mailbox?.isUnifiedMailbox != true) ||
+      label != null ||
+      hasAttachment ||
+      isContainFlagged ||
+      unread ||
+      notIncludeEvents;
+
```
tddang-linagora: Compare these 2 boolean getters.

lib/features/thread/presentation/thread_controller.dart
```diff
@@ -357,7 +374,7 @@ class ThreadController extends BaseController with EmailActionController {
         if (listEmailController.hasClients) {
           listEmailController.jumpTo(0);
         }
-        canSearchMore = true;
+        appProviderContainer.read(searchEmailProvider.notifier).clearResult();
```
hoangdat: why need to call `clearResult` here? why not depend on the action (int)

lib/features/thread/presentation/extensions/handle_search_email_execution_extension.dart
```diff
@@ -0,0 +1,206 @@
+part of '../thread_controller.dart';
+
+/// Web search executor bridge for [ThreadController].
+/// Delegates search intents and mirrors [SearchEmailResult] into GetX state.
+extension HandleSearchEmailExecutionExtension on ThreadController {
```
hoangdat: why we need using extension, do we have better approach for this, a specific class handle logic of search. We can instance and use it inside thread controller?

lib/features/thread/presentation/extensions/handle_search_email_execution_extension.dart
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    if (!needRefreshSearchState && listEmailController.hasClients) {
+      isListEmailScrollViewJumping = true;
+      listEmailController.jumpTo(0);
+    }
+    if (!needRefreshSearchState) {
+      mailboxDashBoardController.emailsInCurrentMailbox.clear();
+    }
+    loadingMoreStatus.value = LoadingMoreStatus.idle;
```
hoangdat: I still not get the idea, we refactor to use riverpod, but still touch to a lot of state variable, how we can make it clean, no need to control the state in center. I mean states should listen and react base on the intent or sth else.

lib/features/thread/presentation/extensions/list_presentation_email_extensions.dart
```diff
@@ -12,8 +13,40 @@ import 'package:tmail_ui_user/main/routes/app_routes.dart';
 import 'package:tmail_ui_user/main/routes/navigation_router.dart';
 import 'package:tmail_ui_user/main/routes/route_utils.dart';
 
+/// Inputs for per-row mailbox, selection, and web-route sync.
+class PresentationEmailSyncContext {
```
hoangdat: why the `Sync`?

lib/features/thread/presentation/widgets/thread_compose_floating_button.dart
```diff
@@ -0,0 +1,72 @@
...
+    required this.isSelectModeActive,
+    required this.hasSelectedEmails,
+    required this.onTapCompose,
+  });
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    if (responsiveUtils.isWebDesktop(context)) {
+      return const SizedBox.shrink();
+    }
```
hoangdat: why we hide it in Web Desktop?

lib/features/thread/presentation/extensions/list_presentation_email_extensions.dart
dab246: "Sync" = the `syncPresentationEmail` op it feeds, which reconciles a fetched page with
derived per-row state (mailbox, carried-over selection, web route). Kept the name
(mirrors the method) and expanded the doc comment to say this.

lib/features/thread/presentation/widgets/thread_compose_floating_button.dart
dab246: Pre-existing behavior, carried verbatim from `ThreadView._buildFloatingButtonCompose`.
Web desktop composes via the persistent `ComposeButtonWidget` in the left panel, so the
FAB is mobile/tablet only. No behavioral change.

lib/features/thread/presentation/extensions/handle_search_email_execution_extension.dart
dab246: Deleted the `part` extension. Search logic now lives in **`WebSearchService`** (`newSearch` / `loadMore` / `refreshChanges`; dispatches intents, holds **no** view state) and a port **`SearchLifecycleObserver`**. `ThreadController implements SearchLifecycleObserver` and registers itself; the service references no consumer.

lib/features/thread/presentation/extensions/handle_search_email_execution_extension.dart
dab246: Deleted

lib/features/thread/presentation/thread_controller.dart
dab246: removed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4706

test/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller_test.dart
```diff
@@ -433,12 +433,12 @@ void main() {
         refreshAllMailboxInteractor);
       mailboxController.onReady();
 
+      Get.put<SearchEmailInteractor>(searchEmailInteractor);
+      Get.put<SearchMoreEmailInteractor>(searchMoreEmailInteractor);
```
hoangdat, line 437: why we need this?

lib/features/thread/presentation/thread_search_execution_observer.dart
```diff
@@ -0,0 +1,72 @@
...
+    _controller.canSearchMore = true;
+    _controller.loadingMoreStatus.value = LoadingMoreStatus.idle;
+    _controller.searchController.activateSimpleSearch();
+  }
+
+  @override
+  void onSearchLoading() => _controller.dispatchState(Right(SearchingState()));
+
+  @override
+  void onSearchResult(
```
hoangdat, line 39: in web small screen, search will be dispatch from search_email_controller. So in this case it will trigger this method in thread? any problem with it? regression?

test/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller_test.dart
dab246: These interactors were previously injected into `ThreadController` through its constructor.

After search execution was moved to the central `SearchEmailNotifier`, the notifier resolves them using `Get.find` because it currently has no constructor-based dependency injection.

* **Production:** Registered in `MailboxDashboardBindings` (`lines 383–384`).
* **Tests:** Must also be registered so `Get.find` can resolve them when a search is dispatched.

Removing these registrations causes the test to fail on the first search.

lib/features/thread/presentation/thread_search_execution_observer.dart
dab246: Confirmed and fixed.

Previously, `ThreadSearchExecutionObserver` received all web search events. On small screens, searches initiated by `SearchEmailController` also reached the thread, causing:

* Redundant mapping into the off-screen thread list.
* Duplicate failure toasts.

The observer is now gated with `responsiveUtils.isWebDesktop(context)`, matching the layout logic:

* **Desktop:** The thread owns search results through inline search on both `thread` and `threadDetailed` routes.
* **Small screens:** The dedicated `SearchEmailView` owns the results, so the thread ignores executor events.

Screen size is used instead of the current route because a WebSocket-triggered search refresh may arrive while an email is open on the `threadDetailed` route. In the desktop layout, the underlying thread list must still be updated.

Widget tests were added for both desktop and small-screen layouts.

lib/features/search/email/presentation/service/search_executor_service.dart
```diff
@@ -0,0 +1,120 @@
...
+/// [searchEmailProvider] result out to registered observers. Holds no view
+/// state and does not know how many observers exist. See ADR-0092/0093.
+class SearchExecutorService {
+  SearchExecutorService(this._container);
+
+  final ProviderContainer _container;
+  final Set<SearchExecutionObserver> _observers = <SearchExecutionObserver>{};
+  ProviderSubscription<AsyncValue<SearchEmailResult>>? _subscription;
+  bool _hasDispatchedSearch = false;
+  SearchExecutionIntent? _lastDispatchedIntent;
```
hoangdat, line 20: maybe different controller run thier own refresh in web socket, we should to handle it better. maybe the same guard check in thread execution.

IMO, at the time, we have only one active observer, how about we manage it to make it easy to dispatch without changing the other? WDYT?

lib/features/search/email/presentation/service/search_executor_service.dart
```diff
@@ -0,0 +1,120 @@
...
+  void _closeSubscription() {
+    _subscription?.close();
+    _subscription = null;
+  }
+}
+
+/// App-lifetime singleton bridging every search consumer to the executor.
+@Riverpod(keepAlive: true)
+SearchExecutorService searchExecutorService(Ref ref) {
+  final service = SearchExecutorService(appProviderContainer);
```
hoangdat: how about using `ref.container` instead of `appProviderContainer`?

lib/features/search/email/presentation/service/search_executor_service.dart
```diff
@@ -0,0 +1,120 @@
...
+    for (final observer in _observers.toList()) {
+      action(observer);
+    }
+  }
+
+  void _notifyCurrentState(SearchExecutionObserver observer) {
+    final current = _container.read(searchEmailProvider);
+    if (current.isLoading) {
+      observer.onSearchLoading();
+    } else if (current.hasError) {
```
hoangdat, line 123: why not prevent urgent error handling here instead of handling in each observer?

hoangdat: moreover, if the error is urgent, why not handle it directly in InteractorConsumer?

lib/features/search/email/presentation/service/search_executor_service.dart
dab246: oke. fixed.

lib/features/search/email/presentation/service/search_executor_service.dart
dab246: fixed

lib/features/search/email/presentation/service/search_executor_service.dart
dab246: oke, we wil trigger ownership in the executor SearchRefreshCoordinator

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4706

test/features/search/email/presentation/service/search_executor_service_test.dart
tddang-linagora, line 1: Would this test valid
```dart
test(
      'a refresh dispatched while a search is loading must not blank the '
      'results, and the search must not lose its real response',
      () async {
    // Reproduces the critical race from PR review: a websocket-triggered
    // RefreshChangesIntent lands while a NewSearchIntent is still in flight.
    final (:service, :observer) = registeredService();

    final searchPending = Completer<Either<Failure, Success>>();
    stubSearch((_) => Stream.fromFuture(searchPending.future));

    final search = service.dispatch(const NewSearchIntent(), context);
    await pumpEventQueue();
    expect(observer.loading, 1);
    expect(observer.results, isEmpty);

    final refreshPending = Completer<Either<Failure, Success>>();
    stubRefresh((_) => Stream.fromFuture(refreshPending.future));
    final refresh = service.dispatch(
      const RefreshChangesIntent(currentCount: 0),
      context,
    );
    await pumpEventQueue();

    // BUG: before the refresh's own network call resolves, observers are
    // immediately fanned an empty result (SearchExecutorService._notifier's
    // _currentResult falls back to SearchEmailResult.empty() because the
    // in-flight AsyncLoading discarded the previous data). No observer
    // should ever see an empty result while the original search is still
    // legitimately pending and hasn't failed.
    expect(
      observer.results,
      isEmpty,
      reason: 'a refresh racing a loading search must not emit an empty '
          'placeholder result before either request has actually resolved',
    );

    // The original search now completes with real data...
    searchPending.complete(Right(SearchEmailSuccess([email])));
    await search;
    await pumpEventQueue();

    // ...but PR #4706 as written discards it as stale, because dispatching
    // the refresh already bumped _latestRequestId past the search's id.
    expect(
      observer.results.any((r) => r.emails.contains(email)),
      isTrue,
      reason: 'the search that actually found results must not be silently '
          'dropped just because a refresh was dispatched while it was '
          'in flight',
    );

    refreshPending.complete(
      Right(RefreshChangesSearchEmailSuccess(const [])),
    );
    await refresh;
    await pumpEventQueue();
  });
```

dab246: The test is not valid as written. `observer.results` stores emitted result batches, so an empty-email batch still makes the list non-empty; the assertion therefore fails before Claim 2 is reached.

Claim 1 may expose a real empty-state flash, but it should be verified for reachability and tested separately against the emitted emails.

Claim 2 is not a bug: dropping the older search result after a newer refresh is the intended `_latestRequestId` supersession behavior and is already covered by an existing test. If the refresh is dispatched too early, the fix should be in the coordinator, not by weakening last-write-wins.

lib/features/search/email/presentation/providers/search_refresh_providers.dart
```diff
@@ -0,0 +1,63 @@
...
+@riverpod
+SearchRefreshCoordinator? searchRefreshCoordinator(Ref ref) {
+  final stateSource = ref.watch(dashboardEmailStateChangeSourceProvider);
+  if (stateSource == null) return null;
+
+  final coordinator = SearchRefreshCoordinator(
+    executor: ref.read(searchExecutorServiceProvider),
+    stateSource: stateSource,
+    callbacks: SearchRefreshCoordinatorCallbacks(
+      resolveDispatchContext: () => ref.read(searchDispatchContextProvider),
```
hoangdat: I still not understand about `searchDispatchContextProvider`? why need it? because I still see we construct it new in ThreadController?

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
```diff
@@ -0,0 +1,160 @@
...
+
+  /// Subscribes to [stateSource] and serializes refresh dispatches.
+  SearchRefreshCoordinator({
+    required SearchExecutorService executor,
+    required EmailStateChangeSource stateSource,
+    required SearchRefreshCoordinatorCallbacks callbacks,
+  })  : _executor = executor,
+        _stateSource = stateSource,
+        _callbacks = callbacks {
+    _queue = WebSocketQueueHandler(
```
hoangdat, line 56: it quite queue be recreated every new state?

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
```diff
@@ -0,0 +1,160 @@
...
+    if (_retryTimers.containsKey(stateId)) return true;
+    if (_queue.isMessageQueued(stateId)) return true;
+    return _queue.isMessageProcessed(stateId);
+  }
+
+  /// Dispatches one refresh for a queued state change.
+  Future<void> _processMessage(WebSocketMessage message) async {
+    _inFlightStateIds.add(message.id);
+    try {
+      if (!_canProcessMessage()) return;
```
hoangdat, line 95: maybe missing set search active for the case of mobile/large tablet ?

hoangdat:
- [x] quick search filter not update in web


https://github.com/user-attachments/assets/49dd7ee7-eeaa-4b73-bc62-c4cea0bc22b2

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
dab246: Correct. isSearchActive (searchStatus == ACTIVE) is only set through enableSearch() on web desktop. Mobile and narrow tablets open search through activateSimpleSearch(), so isSearchActive = false and refresh is blocked. I changed the gate to isSearchEmailRunning, which works correctly across all platforms and screen widths.

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
dab246: No. The coordinator and queue are created once per session. New states are only enqueued into the existing queue. The fragile point was that if dashboardEmailStateChangeSourceProvider recomputed and returned a new instance, it would rebuild the coordinator and lose deduplication. I added == and hashCode to DashboardEmailStateChangeSource, so the same dashboard does not trigger a rebuild.

lib/features/search/email/presentation/providers/search_refresh_providers.dart
dab246: searchDispatchContextProvider removed

tddang-linagora (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4706

lib/features/search/email/presentation/service/search_executor_service.dart
```diff
@@ -0,0 +1,136 @@
...
+import 'package:tmail_ui_user/features/search/email/domain/execution/search_execution_result.dart';
+import 'package:tmail_ui_user/features/search/email/domain/model/search_email_result.dart';
+import 'package:tmail_ui_user/features/search/email/domain/notifier/search_email_notifier.dart';
+import 'package:tmail_ui_user/features/search/email/presentation/service/search_dispatch_context.dart';
+import 'package:tmail_ui_user/features/search/email/presentation/service/search_execution_observer.dart';
+
+/// Single owner of the search executor: dispatches intents and fans each
+/// [searchEmailProvider] result out to registered observers. Holds no view
+/// state and does not know how many observers exist. See ADR-0092/0093.
+class SearchExecutorService {
```
hoangdat, line 13: add test file to prevent regression for `gotoSearchView`

```dart
import 'package:flutter_riverpod/flutter_riverpod.dart';
import 'package:flutter_test/flutter_test.dart';
import 'package:get/get.dart';
import 'package:jmap_dart_client/jmap/account_id.dart';
import 'package:jmap_dart_client/jmap/core/session/session.dart';
import 'package:jmap_dart_client/jmap/core/state.dart' as jmap;
import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
import 'package:mockito/mockito.dart';
import 'package:tmail_ui_user/features/email/presentation/action/email_ui_action.dart';
import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/notifier/search_view_state_notifier.dart';
import 'package:tmail_ui_user/features/search/email/domain/execution/search_execution_intent.dart';
import 'package:tmail_ui_user/features/search/email/domain/execution/search_execution_result.dart';
import 'package:tmail_ui_user/features/search/email/presentation/providers/search_executor_provider.dart';
import 'package:tmail_ui_user/features/search/email/presentation/providers/search_refresh_providers.dart';
import 'package:tmail_ui_user/features/search/email/presentation/service/email_state_change_source.dart';
import 'package:tmail_ui_user/features/search/email/presentation/service/search_dispatch_context.dart';
import 'package:tmail_ui_user/features/search/email/presentation/service/search_executor_service.dart';

import '../../../../../fixtures/account_fixtures.dart';
import '../../../../../fixtures/session_fixtures.dart';

/// Supplies only what the refresh wiring reads: the action stream the state
/// source adapts, plus the session fields the dispatch context is built from.
class _FakeDashboard extends Mock implements MailboxDashBoardController {
  // Get.put drives the GetX lifecycle, which reads these directly rather than
  // through noSuchMethod.
  @override
  final InternalFinalCallback<void> onStart =
      InternalFinalCallback<void>(callback: () {});

  @override
  final InternalFinalCallback<void> onDelete =
      InternalFinalCallback<void>(callback: () {});

  @override
  final Rxn<EmailUIAction> emailUIAction = Rxn<EmailUIAction>();

  @override
  final Rxn<AccountId> accountId = Rxn<AccountId>();

  @override
  Session? sessionCurrent;

  @override
  Set<MailboxId>? get trashSpamMailboxIds => null;
}

class _RecordingExecutor extends SearchExecutorService {
  _RecordingExecutor(this._ownContainer) : super(_ownContainer);

  final ProviderContainer _ownContainer;

  int dispatchCount = 0;

  @override
  Future<SearchExecutionResult> dispatch(
    SearchExecutionIntent intent,
    SearchDispatchContext context,
  ) async {
    dispatchCount++;
    return SearchExecutionResult.success;
  }

  void disposeContainer() {
    dispose();
    _ownContainer.dispose();
  }
}

/// Covers the production refresh wiring end to end — the real gate predicate
/// reading the real [searchViewStateProvider], and the real dispatch-context
/// resolution off the bound dashboard.
///
/// search_refresh_coordinator_test.dart injects the gate as a stub, so it can
/// only prove the coordinator honours whatever predicate it is handed. It
/// cannot catch the wiring handing over the wrong one — which is exactly the
/// defect these tests exist to pin.
void main() {
  late _FakeDashboard dashboard;
  late _RecordingExecutor executor;
  late ProviderContainer container;

  setUp(() {
    dashboard = _FakeDashboard()
      ..sessionCurrent = SessionFixtures.aliceSession
      ..accountId.value = AccountFixtures.aliceAccountId;
    // _resolveDispatchContext looks the dashboard up through getBinding.
    Get.put<MailboxDashBoardController>(dashboard);

    executor = _RecordingExecutor(ProviderContainer());
    container = ProviderContainer(overrides: [
      dashboardEmailStateChangeSourceProvider
          .overrideWithValue(DashboardEmailStateChangeSource(dashboard)),
      searchExecutorServiceProvider.overrideWithValue(executor),
    ]);
    // Reading the provider is what builds the coordinator and subscribes it.
    container.read(searchRefreshCoordinatorProvider);
  });

  tearDown(() {
    container.dispose();
    executor.disposeContainer();
    Get.reset();
  });

  void emitStateChange(String state) {
    dashboard.emailUIAction.value =
        RefreshChangeEmailAction(newState: jmap.State(state));
  }

  test(
    'refreshes on the SearchEmailView entry path — simple search activated '
    'without the dashboard search bar ever being enabled',
    () async {
      // goToSearchView() activates simple search and routes to SearchEmailView.
      container.read(searchViewStateProvider.notifier).activateSimpleSearch();

      emitStateChange('state-1');
      await pumpEventQueue();

      expect(executor.dispatchCount, 1);
    },
  );

  test('refreshes while advanced search is running', () async {
    container.read(searchViewStateProvider.notifier).activateAdvancedSearch();

    emitStateChange('state-1');
    await pumpEventQueue();

    expect(executor.dispatchCount, 1);
  });

  test('does not refresh when no search is running', () async {
    emitStateChange('state-1');
    await pumpEventQueue();

    expect(executor.dispatchCount, 0);
  });

  test('stops refreshing once the search is deactivated', () async {
    container.read(searchViewStateProvider.notifier).activateSimpleSearch();
    emitStateChange('state-1');
    await pumpEventQueue();

    container.read(searchViewStateProvider.notifier).deactivateSimpleSearch();
    emitStateChange('state-2');
    await pumpEventQueue();

    expect(executor.dispatchCount, 1);
  });

  test('does not refresh once the dashboard binding is gone', () async {
    container.read(searchViewStateProvider.notifier).activateSimpleSearch();
    Get.reset();

    emitStateChange('state-1');
    await pumpEventQueue();

    expect(executor.dispatchCount, 0);
  });
}
```

lib/features/search/email/presentation/search_email_controller.dart
```diff
@@ -1170,14 +1084,13 @@ class SearchEmailController extends BaseController
...
-    emailUIActionWorker.dispose();
     dashBoardActionWorker.dispose();
     viewStateWorker.dispose();
-    _searchEmailResultSubscription?.close();
-    _webSocketQueueHandler?.dispose();
+    _searchService.unregister(this);
     onKeyboardShortcutDispose();
     // Release keepAlive search state with the binding, including the committed
     // filter, so no stale filter leaks into the next search session.
+    _searchService.resetSession();
```
hoangdat: Search invoice on a wide browser → click a mailbox to leave search → narrow the window → tap search.
The old invoice results are already on screen before you type anything.

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
```diff
@@ -0,0 +1,161 @@
...
+
+  /// Adds a state change to the serialized queue.
+  void _enqueue(jmap.State state) {
+    if (_disposed) return;
+    final message = WebSocketMessage(newState: state);
+    if (_isStateKnown(message.id)) return;
+    _queue.enqueue(message);
+  }
+
+  bool _isStateKnown(String stateId) {
```
hoangdat, line 84: New mail arrives → app re-runs your search even when it already knew about that change. Results are right; it just burns an extra server round-trip every time.
The old currentEmailState != newState guard was dropped and not replaced — _isStateKnown only dedups by state id.

lib/features/search/email/domain/notifier/search_email_notifier.dart
```diff
@@ -192,6 +202,10 @@ class SearchEmailNotifier extends _$SearchEmailNotifier with InteractorConsumer
       },
       onFailure: onFailure,
     );
+    if (succeeded) return SearchExecutionResult.success;
```
hoangdat, line 205: A search fails at the same moment you start a new one → reported as superseded, so no retry. That's correct (the new search is running), but the code re-checks staleness after the await and doesn't say it's on purpose. One comment.

lib/features/search/email/presentation/service/search_executor_service.dart
dab246: added

lib/features/search/email/presentation/search_email_controller.dart
dab246: Fixed. Leaving search now clears the result SSOT via `resetSearchResultSession` in `SearchController.disableAllSearchEmail`, so results no longer survive on the mailbox-select exit path (which never disposes `SearchEmailController`).

lib/features/search/email/presentation/service/search_refresh_coordinator.dart
dab246: Fixed. Restored the guard: `_enqueue` now skips when the new state equals `currentAppliedState` (exposed via `EmailStateChangeSource`), so no refresh fires for a change already synced.

lib/features/search/email/domain/notifier/search_email_notifier.dart
dab246: Intentional, the post-`await` re-check of `requestId != _latestRequestId` means a newer search owns the refresh, so the stale failure must not retry. Kept as-is; the request-id check makes the intent clear. Comment added

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4706

hoangdat (approved):

hoangdat:
@Coderabbit review

PR #4708, opened 2026-07-14
Reduce flaky test not rely on toast 

hoangdat:
## Summary by CodeRabbit

* **Tests**
  * Improved draft-save integration flows by waiting for confirmed draft-save completion via view-state changes, with failure handling tied to the confirmation dialog.
  * Strengthened email send scenarios to verify completion through state changes and sent mailbox contents, removing reliance on transient success notifications.
  * Enhanced test utilities for conditional waiting and view-state validation, including improved timeout behavior and clearer error reporting.

### DIFF
--- a/integration_test/scenarios/composer/base_save_draft_then_reopen_scenario.dart
+++ b/integration_test/scenarios/composer/base_save_draft_then_reopen_scenario.dart
@@ -1,7 +1,11 @@
-import 'package:flutter_test/flutter_test.dart';
+import 'package:get/get.dart';
+import 'package:tmail_ui_user/features/composer/domain/state/save_email_as_drafts_state.dart';
+import 'package:tmail_ui_user/features/composer/domain/state/update_email_drafts_state.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 
 import '../../robots/abstract/abstract_composer_robot.dart';
+import '../../utils/wait_for_view_state.dart';
 import 'base_save_and_reopen_scenario.dart';
 
 abstract class BaseSaveDraftThenReopenScenario extends BaseSaveAndReopenScenario {
@@ -18,9 +22,10 @@ abstract class BaseSaveDraftThenReopenScenario extends BaseSaveAndReopenScenario
     await composerRobot.tapCloseComposer();
     await $.pumpAndTrySettle();
     await expectViewVisible($(#confirm_dialog_action));
+
+    final saved = _waitForDraftSaved();
     await composerRobot.tapSaveButtonOnSaveDraftConfirmDialog(l10n);
-    await $.pumpAndTrySettle();
-    await _expectSaveDraftSuccessToast(l10n);
+    await saved;
     await $.pumpAndTrySettle();
   }
 
@@ -29,12 +34,15 @@ abstract class BaseSaveDraftThenReopenScenario extends BaseSaveAndReopenScenario
     AbstractComposerRobot composerRobot,
     AppLocalizations l10n,
   ) async {
+    final saved = _waitForDraftSaved();
     await composerRobot.tapSaveAsDraftButton();
-    await $.pumpAndTrySettle();
-    await _expectSaveDraftSuccessToast(l10n);
+    await saved;
   }
 
-  Future<void> _expectSaveDraftSuccessToast(AppLocalizations l10n) async {
-    await expectViewVisible($(l10n.drafts_saved));
-  }
+  Future<void> _waitForDraftSaved() => waitForViewState(
+    Get.find<MailboxDashBoardController>(),
+    matcher: (state) =>
+        state is SaveEmailAsDraftsSuccess || state is UpdateEmailDraftsSuccess,
+    description: 'SaveEmailAsDraftsSuccess/UpdateEmailDraftsSuccess',
+  );
 }
--- a/integration_test/scenarios/composer/base_save_draft_then_reopen_scenario.dart
+++ b/integration_test/scenarios/composer/base_save_draft_then_reopen_scenario.dart
@@ -40,6 +40,7 @@ abstract class BaseSaveDraftThenReopenScenario extends BaseSaveAndReopenScenario
     AppLocalizations l10n,
   ) async {
     await _saveDraftAndWaitForOutcome(composerRobot.tapSaveAsDraftButton);
+    await $.pumpAndTrySettle();
   }
 
   /// Runs [saveAction], then returns once the save has actually resolved: either the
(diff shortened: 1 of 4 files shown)


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

integration_test/utils/wait_for_condition.dart
```diff
@@ -2,16 +2,24 @@ import 'dart:async';
...
-  await Future.doWhile(() async {
-    if (await condition()) {
-      return false;
+  final elapsed = Stopwatch()..start();
+
+  while (true) {
+    if (await condition()) return;
+
+    if (elapsed.elapsed >= timeout) {
+      throw TimeoutException('waitForCondition timed out', timeout);
```
dab246: `wait_for_condition.dart` no longer bounds a hanging `condition()`

The old body wrapped the whole loop in `.timeout(timeout)`, so even a `condition()` that never returned was bounded. The rewrite only checks `elapsed` *between* awaits:

```dart
while (true) {
  if (await condition()) return;        // if this never completes, no timeout fires
  if (elapsed.elapsed >= timeout) throw TimeoutException(...);
  await Future.delayed(interval);
}
```

For current callers `condition` is a synchronous visibility check, so this is theoretical. But the signature is `FutureOr<bool> Function()`, and a future caller passing a hanging async probe would now hang forever instead of timing out. Cheap to keep the guarantee — e.g. `await condition().timeout(remainingBudget)` or document that `condition` must resolve promptly.

integration_test/utils/wait_for_view_state.dart
```diff
@@ -0,0 +1,64 @@
...
+  required bool Function(Success state) matcher,
+  required String description,
+  Duration timeout = TestTimeouts.long,
+}) {
+  final completer = Completer<Success>();
+  late final StreamSubscription<Either<Failure, Success>> subscription;
+
+  subscription = controller.viewState.stream.listen((state) {
+    state.fold(
+      (_) {},
```
dab246: Also handle failure avoid `wait_for_view_state.dart` silently ignores failure states → slow, misleading timeout

tddang-linagora (dismissed):

tddang-linagora (dismissed):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

integration_test/scenarios/send_email_scenario.dart
```diff
@@ -27,12 +29,15 @@ class SendEmailScenario extends BaseTestScenario {
...
-  }
-
-  Future<void> _expectSendEmailSuccessToast() async {
-    await expectViewVisible(
-      $(find.text(AppLocalizations().message_has_been_sent_successfully)),
+    await waitForCondition(
+      () => $(ThreadView).visible || $(#confirm_dialog_action).visible,
+      timeout: TestTimeouts.long,
+    );
+    expect(
```
dab246: `send_email_scenario` can pass before send completion. 

Problem:

`ThreadView` is not a reliable signal that send completed. On web, the composer is rendered as an overlay above the dashboard, and `ThreadView` remains in the widget tree behind it. On tablet or other non-opaque flows, a similar visibility ambiguity may exist. Because `waitForCondition` uses `ThreadView.visible` as one of the success conditions, the wait can finish immediately after `robots.composerRobot().send()`.

Risk:

The test can pass even when:

- `SendingMessageDialogView` is still running.
- `SendEmailSuccess` has not been emitted.
- The send operation later fails.
- The failure confirm dialog appears after the test already passed.

This weakens the test from "email sent successfully" to "thread list exists somewhere in the UI tree".

integration_test/scenarios/send_email_scenario.dart
```diff
@@ -27,12 +29,15 @@ class SendEmailScenario extends BaseTestScenario {
...
+      $(#confirm_dialog_action).visible,
+      isFalse,
+      reason: 'Email failed to send: a confirm dialog is shown instead of returning to the thread list.',
```
dab246: The guard `expect($(#confirm_dialog_action).visible, isFalse)` is meant to catch "send was blocked by a dialog". But because the `waitForCondition` OR-clause is satisfied immediately on Web, the guard is evaluated *before* any such dialog would have had time to appear. A blocking dialog that surfaces a few frames later would slip past. Adopting the state-stream approach (W1) makes this guard unnecessary, a real block produces `SendEmailFailure`, which `failureMatcher` reports immediately with the actual failure.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

integration_test/scenarios/send_email_scenario.dart
hoangdat: done

integration_test/scenarios/send_email_scenario.dart
hoangdat: done

integration_test/scenarios/composer/base_save_draft_then_reopen_scenario.dart
```diff
@@ -29,12 +34,19 @@ abstract class BaseSaveDraftThenReopenScenario extends BaseSaveAndReopenScenario
...
 
-  Future<void> _expectSaveDraftSuccessToast(AppLocalizations l10n) async {
-    await expectViewVisible($(l10n.drafts_saved));
-  }
+  Future<void> _waitForDraftSaved() => waitForViewState(
+    Get.find<MailboxDashBoardController>(),
+    ViewStateExpectation(
+      matcher: (state) =>
+          state is SaveEmailAsDraftsSuccess || state is UpdateEmailDraftsSuccess,
+      failureMatcher: (state) =>
```
dab246: Draft `failureMatcher` can never fire; a failed draft save hangs to full timeout. 

Both save paths in `composer_controller.dart` call `consumeState(...)` on **success**, but on **failure** they surface a confirm dialog (`_showConfirmDialogWhenSaveMessageToDraftsFailure`) and **never** `consumeState` the failure. So `SaveEmailAsDraftsFailure`/`UpdateEmailDraftsFailure` are never emitted on `MailboxDashBoardController.viewState`, and this `failureMatcher` is effectively dead. 

Consequence: on a genuinely failed draft save, the wait can't fast-fail — it burns the entire `TestTimeouts.long` and then reports a *timeout* ("no ...Success emitted"), exactly the "burn the whole timeout to report an answer it already knew" problem the `waitForViewState` doc warns against. The **send** scenario already solves this by racing `#confirm_dialog_action`; the draft scenario should mirror it.

dab246 (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4708

PR #4723, opened 2026-07-22
feat(mail): support shared mailboxes across JMAP accounts

VasilisNtovolos:
## Summary

Add support for shared mailboxes exposed through additional mail-capable
accounts in the JMAP session.

Previously, mailbox loading and message fetching were primarily tied to the
main account. This change allows Twake Mail to discover additional JMAP mail
accounts, display their mailboxes in the mailbox tree, and use the correct
account when loading messages from a shared mailbox.

The implementation was tested against Cyrus JMAP 3.12.3.

## Changes

- Add account information to `PresentationMailbox`
- Distinguish personal mailboxes, shared mailboxes, and synthetic account roots
- Load mailboxes from additional JMAP mail accounts
- Display each additional account as an expandable group under Team Mailboxes
- Preserve the parent/child hierarchy of shared mailboxes
- Use account-aware mailbox keys to avoid mailbox ID collisions across accounts
- Use the selected mailbox account when fetching messages
- Prevent synthetic account roots from being opened as real mailboxes

## Testing

Verified that:

- The personal Inbox continues to load normally
- Team Mailboxes is displayed
- Additional JMAP mail accounts appear as expandable groups
- Shared mailbox hierarchy is preserved
- Messages are loaded from the selected shared mailbox account
- Synthetic account roots only expand and collapse
- Returning to the personal Inbox continues to work
- Static analysis passes with no issues

## Notes

The Cyrus test environment also required `List-Unsubscribe` to be requested as
`header:List-Unsubscribe:asURLs`.

That compatibility fix was used only for local runtime testing and is not
included in this PR because it is unrelated to shared mailbox account support.

Mailbox actions such as move, delete, archive, spam, bulk operations, and
read/starred state changes across shared accounts have not yet been fully
validated.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading, displaying, and managing shared mailboxes from other accounts.
  * Shared mailboxes appear in clearly identified account sections with consistent icons.
  * Mailbox navigation and URLs preserve the owning account context.
  * Mailbox visibility and subscription controls now support shared accounts.

* **Bug Fixes**
  * Improved mailbox selection, paths, counts, and actions when accounts reuse folder IDs.
  * Preserved mailbox permissions and sharing details across actions.
  * Corrected team mailbox empty-state detection.
  * Prevented unsupported cross-account moves with clear messages.

### DIFF
--- a/lib/features/mailbox/presentation/mailbox_controller.dart
+++ b/lib/features/mailbox/presentation/mailbox_controller.dart
@@ -14,6 +14,7 @@ import 'package:jmap_dart_client/jmap/core/session/session.dart';
 import 'package:jmap_dart_client/jmap/core/state.dart' as jmap;
 import 'package:jmap_dart_client/jmap/mail/email/email.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
+import 'package:jmap_dart_client/jmap/mail/mailbox/namespace.dart';
 import 'package:model/model.dart';
 import 'package:rxdart/transformers.dart';
 import 'package:tmail_ui_user/features/base/base_mailbox_controller.dart';
@@ -122,6 +123,11 @@ class MailboxController extends BaseMailboxController
   final SubaddressingInteractor _subaddressingInteractor;
   final CreateDefaultMailboxInteractor _createDefaultMailboxInteractor;
   final MoveFolderContentInteractor _moveFolderContentInteractor;
+  final GetAllMailboxInteractor _sharedMailboxGetAllMailboxInteractor;
+
+  final Map<AccountId, List<PresentationMailbox>> _sharedMailboxesByAccount = {};
+
+  bool _didLoadSharedMailboxes = false;
 
   IOSSharingManager? _iosSharingManager;
   late MailboxActionReactor mailboxActionReactor;
@@ -161,11 +167,11 @@ class MailboxController extends BaseMailboxController
     VerifyNameInteractor verifyNameInteractor,
     GetAllMailboxInteractor getAllMailboxInteractor,
     RefreshAllMailboxInteractor  refreshAllMailboxInteractor,
-  ) : super(
+  ) : _sharedMailboxGetAllMailboxInteractor = getAllMailboxInteractor, super(
     treeBuilder,
     verifyNameInteractor,
     getAllMailboxInteractor: getAllMailboxInteractor,
-    refreshAllMailboxInteractor: refreshAllMailboxInteractor
+    refreshAllMailboxInteractor: refreshAllMailboxInteractor,
   );
 
   @override
@@ -173,7 +179,7 @@ class MailboxController extends BaseMailboxController
     _registerObxStreamListener();
     _initWebSocketQueueHandler();
     mailboxActionReactor = MailboxActionReactor(
-      _moveFolderContentInteractor,
+      _moveFolderContentInteractor
     );
     super.onInit();
   }
@@ -188,7 +194,7 @@ class MailboxController extends BaseMailboxController
         _handleOpenMailbox(event.buildContext, event.presentationMailbox);
       });
     _initCollapseMailboxCategories();
-    mailboxListScrollController.addListener(_mailboxListScrollControllerListener);
+    mailboxListScrollController.addListener(_mailboxListScrollControllerListener,);
     super.onReady();
   }
 
@@ -211,9 +217,9 @@ class MailboxController extends BaseMailboxController
     } else if (success is CreateNewMailboxSuccess) {
       _createNewMailboxSuccess(success);
     } else if (success is DeleteMultipleMailboxAllSuccess) {
-      _deleteMultipleMailboxSuccess(success.listMailboxIdDeleted, success.currentMailboxState);
+      _deleteMultipleMailboxSuccess(success.listMailboxIdDeleted, success.currentMailboxState,);
     } else if (success is DeleteMultipleMailboxHasSomeSuccess) {
-      _deleteMultipleMailboxSuccess(success.listMailboxIdDeleted, success.currentMailboxState);
+      _deleteMultipleMailboxSuccess(success.listMailboxIdDeleted, success.currentMailboxState,);
     } else if (success is RenameMailboxSuccess) {
       _renameMailboxSuccess(success);
     } else if (success is MoveMailboxSuccess) {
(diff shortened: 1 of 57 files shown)


tddang-linagora:
Hi @VasilisNtovolos . Thank you for the PR. There are few questions regarding this:
- Skimming through, the number of changes is quite small for a seemingly big additional feature promised. There's no update to the UI whatsoever, so I assume the shared mailboxes sit in the same section with the main account mailboxes?
- If you're so kindly add a demo of this work, and/or even better
- Provide the team with test account and steps to try this feature locally (maybe with Cyrus test server?)

chibenwa:
Hello @VasilisNtovolos 

Thanks for the PR

I would like to drag the attention of the team ( cc @tddang-linagora ) that this work could imensly benefit us for "delegation of accounts" ie minister@linagora.com grant full access to his account to secretary@linagora.com I would thus like you to test this change. I did grant team member delegation access on twake-calendar-dev account.

chibenwa:
image

dab246:
Hi @VasilisNtovolos . Thank you for the PR. There are a number of issues we need to consider and fix:

1. Mailbox identity is not consistently account-scoped. `TreeBuilder` uses `accountId + mailboxId`, but several lookup, selection, routing and update flows still use only `MailboxId`, which can target the wrong mailbox when IDs overlap.

2. Mailbox actions still use the primary account ID. Delete, rename, move, subscribe/unsubscribe and subaddressing should use `selectedMailbox.accountId` and pass it through all action flows.

3. Primary mailbox refresh removes shared mailboxes from the UI because `_handleGetAllMailboxSuccess()` rebuilds the tree from primary mailboxes only. The primary/shared mailbox merge must also be applied during refresh and websocket updates.

4. The synthetic shared-account root can still receive email drops. It must be excluded from `DragTarget`, destination selection and all mailbox actions.

Please also add regression tests for account-scoped IDs, correct account usage in actions, shared-mailbox preservation after refresh, and synthetic-root protection.

zisisvakras:
Quick heads-up on what this does and the JMAP reasoning behind it.

### The RFC bit: James vs the standard

When you have access to someone else's mailboxes, a server can expose that in two completely different ways:

- The RFC way (RFC 8620): the other person shows up as a *separate account* in the session (`session.accounts`), each with its own id, name and read-only flag. This is what Cyrus does.
- The James way: James does not add an extra account. It drops the other user's mailbox into *your own* account and tags it with a `namespace` field like `Delegated[alice@example.com]`. That `namespace` field is a James extension (`urn:apache:james:params:jmap:mail:shares`). It is **not** part of RFC 8621 at all.

The original code was built around that James `namespace` trick (it even faked a namespace and dumped everything into Team Mailboxes), which is why it didn't really work against a plain RFC backend like Cyrus.

### What this changes

This moves us onto the standard model:

- Mailbox identity is now (accountId, mailboxId) instead of just `mailboxId`, because RFC 8620 explicitly says ids can clash across accounts.
- "Other users" come straight from `session.accounts` (the RFC mechanism), filtered to the accounts that actually support Mail.
- Every action (rename, delete, subscribe, mark read, empty...) is sent to the account that *owns* the mailbox, not always the primary one.

To keep the UX the same on both backends, delegated mailboxes still render exactly like James delegation: flat rows under Team Mailboxes with the owner's email underneath. We get that by synthesizing the James-style `Delegated[owner]` namespace purely for display, while the real identity and routing use the account. So Cyrus and James look identical to the user, but the plumbing underneath is now RFC-correct.

### How this addresses the review comments

- @tddang-linagora (no UI / same section): shared mailboxes now actually show up, as their own rows under Team Mailboxes with the owner shown.
- @dab246 1 (identity not account-scoped): fixed by keying everything on `(accountId, mailboxId)`.
- @dab246 2 (actions use the primary account): every action now routes to the owning account.
- @dab246 3 (refresh drops shared mailboxes): one rebuild path, so a primary refresh or a websocket change no longer wipes other users' mailboxes.
- @dab246 4 (synthetic root can be dropped on): there is no synthetic root anymore. Delegated mailboxes are real mailboxes folded into Team Mailboxes, so there is nothing fake to drop on, open or route to.
- @dab246 5 (no tests): added regression tests for the account-scoped identity and the tree routing.
- @chibenwa (validate against delegation): this is the account-delegation path, tested against Cyrus.

### Tested against

Checked this end to end against the latest cyrus-imapd (the RFC-standard account model) and the latest tmail-backend (the James namespace model), so both delegation styles work. If anyone wants to try it on a Cyrus setup themselves, we are happy to share the compilation and configuration steps.

lib/features/mailbox/presentation/mailbox_controller.dart
```diff
@@ -272,37 +297,240 @@ class MailboxController extends BaseMailboxController
...
+  /// Loads each delegated account's mailboxes concurrently, rebuilding the
+  /// sidebar as each one arrives so the section fills in progressively.
+  Future<void> _loadOtherUserMailboxes(
+    Session session,
+    AccountId primary,
+  ) async {
+    final accountIds =
+        MailboxUtils.resolveOtherUserAccountIds(session, primary);
+
+    await Future.wait(
```
dab246, line 335: Unbounded concurrent delegated fetches and rebuilds can cause startup/network pressure.  `Future.wait` starts one `Mailbox/get` per non-primary mail-capable session account and rebuilds the complete mailbox tree after every account response. A user with many delegated accounts can generate a large request burst and repeated full-tree rebuilds during startup or account changes.

model/lib/mailbox/presentation_mailbox.dart
```diff
@@ -49,6 +50,9 @@ class PresentationMailbox with EquatableMixin {
   static final roleActionRequired = Role(actionRequiredRole);
 
   final MailboxId id;
+  final AccountId? accountId;
+  final bool isSharedAccount;
+  final bool isSharedAccountRoot;
```
dab246, line 55: `isSharedAccountRoot` is dead model and tree-builder surface. The final implementation deliberately flattens delegated folders into the Team Mailboxes tree. No production code creates a mailbox with `isSharedAccountRoot: true`; the test suite explicitly asserts that no synthetic root exists. The field, selectable-filter branch, copying boilerplate, and sorting comment are therefore unused and describe a model that the PR no longer builds.

dab246 (commented):
Hi @zisisvakras , thanks you for your continued contributions to this PR. I’ve done a quick review and left some comments below, please take a look and address them.


1. Delegated email mutations still default to the primary account

`ThreadController` now reads a delegated mailbox with `selectedMailbox.accountId`, but the mutation entry points used by the thread toolbar/context menu still use c`MailboxDashBoardController.accountId.value` (the primary account). Examples include move-to-folder, move to Trash/Spam, mark read/unread, star/unstar, permanent delete, and their undo flows.

The new drag path is correctly account-aware, but it is only one of several ways to mutate an email. For example, `moveEmailsToMailbox()` opens the picker with the primary account and then calls `moveEmailsToFolder()` with the primary account even when the selected source mailbox is delegated.

**Impact:** `Email/set` can be sent to the wrong account. On servers where IDs collide, this can update or remove a primary-account email; otherwise it will usually fail with `notFound`. Both outcomes are unacceptable for a delegated-mailbox feature.

2. Delegated deep links race against asynchronous account loading and lose account context on replacement


The primary mailbox request and `_loadOtherUserMailboxes()` start concurrently. `_handleDataFromNavigationRouter()` resolves a delegated URL as soon as the primary request completes. If the requested delegated account is still loading, `findMailboxNodeByKey()` returns `null`, the router is cleared, and the user is sent to the unknown-route page. The later delegated load rebuilds the tree but cannot retry the discarded route.

There is a second persistence issue: `_openMailboxFromLocationBar()` rewrites the URL with a new `NavigationRouter` but does not include `mailboxAccountId`. Reloading that rewritten URL falls back to the primary account and can open the wrong same-ID mailbox.

3. Mailbox counter reconciliation re-scopes delegated action results to the primary account


The updated controller converts every successful email action back to `primaryMailboxKey(...)`. This is correct only for primary-account actions. The success states for move/read/delete/empty flows carry a `MailboxId` but not an `AccountId`, so the controller cannot update the delegated tree after an action that was correctly sent to its owner account.

**Impact:** after a delegated action, unread/total counters can remain stale. If the primary account has the same mailbox ID, the primary counter is updated instead.

4. Empty-folder operations are only partially account-scoped

The request itself now uses the mailbox owner account, but child-folder discovery reads `mapMailboxById`, which is intentionally limited to the primary account. Emptying a delegated Trash with child folders therefore passes an empty child list and can leave delegated child folders undeleted.

The Riverpod provider family is also keyed only by `MailboxId`. Two accounts with the same mailbox ID share notifier identity, so loading/progress/completion state can be conflated across accounts.

dab246:
Furthermore, this PR is quite large; we should break it down into smaller parts to make reviewing and merging easier.

dab246:
Thanks everyone for the detailed review. I created [#4761](https://github.com/linagora/tmail-flutter/issues/4761) to track the remaining work for RFC JMAP shared mailbox delegation, including account-scoped mutations, navigation, counter reconciliation, and regression coverage. These items will be addressed in follow-up PRs.

hoangdat:
we tried to run this PR with our delegation settings. It mixed the mailboxes of delegation account inside Team Mailbox

Screenshot 2026-08-18 at 4 15 15 PM

hoangdat:
Hi @VasilisNtovolos, @zisisvakras 

Thanks so much for sticking with this through multiple review rounds — the account-scoped MailboxKey and request-routing foundation is genuinely solid work. 🙏

Looking at @dab246's review https://github.com/linagora/tmail-flutter/pull/4723#pullrequestreview-4893412317, we think it's actually pointing us toward a structural answer rather than more individual fixes: even after several rounds, primary-account defaults are still leaking through in new places — thread toolbar mutations, deep-link resolution, counter reconciliation, empty-folder child lookup. It's not any single mistake, it's that the "merge everything into one shared UI/state and thread accountId through every call site" approach has a lot of surface area, and each round of fixes tends to surface the next spot that was missed.

That's a good reason to move toward the **account switcher** direction for our delegation roadmap: instead of blending accounts into one tree/state and needing every action, deep link, and counter update to carry the right accountId, you'd operate against one account at a time — the class of bug the review keeps finding mostly goes away by construction, not by chasing each call site.

You can reference the idea here.
image

chibenwa:
IMO delegation set up and management shall be out of scope for this work.

We shall focus solely on **account switcher** which is a common ground we have though the session object with StallWart.

PR #93, opened 2021-09-15
TF-78 Mark as email read/unread

dab246:
- [x] Open mail -> read
- [x] Action in ThreadView -> read/unread
- [x] EmailView -> click to icon to unread

https://user-images.githubusercontent.com/80730648/133536687-8b5548e8-c376-42ac-9838-e7c927bc2d40.mov

PR #4725, opened 2026-07-27
TF-4720 Fix signature is not well displayed

### LINKED ISSUE #4720: Signature is not well displayed
### Description
I received an email, and this is how the signature looks like. 

Image

### Expected result
Have well displayed signature. Here is the display on an other client. 

Image

### Context

tmail-canary

I tried to create an eml file without all the content. If it's not working, ask me I'll send you the source on tchat. 

[element-signature-repro.eml](https://github.com/user-attachments/files/30226495/element-signature-repro.eml)

### Additional information

frequency, log files,...

dab246:
Closes #4720 

## Context

Received signatures (e.g. Gmail/Element) ship inline styles like `line-height:0.1` on each `<p>`. A line-height below `1` collapses the line box under the font size, so consecutive lines overlap and the signature is unreadable — even though other clients render it correctly.

## Root cause

`NormalizeLineHeightInStyleTransformer` only stripped two hard-coded values (`1px`, `100%`) and missed general degenerate values such as unitless `0.1`. Generalizing the match naively is unsafe: a regex that splits on `;` corrupts complex inline CSS (quoted values, `url()`, comments, escapes), and unbounded quantifiers risk ReDoS on attacker-controlled email content.

## Solution

- Split inline styles with a linear, fail-open scanner that finds only top-level `;`; semicolons inside quotes, comments, `url()`/`calc()`, and CSS `\` escapes are never treated as boundaries. Malformed CSS (unterminated quote/comment, unbalanced parenthesis, trailing backslash) is left untouched.
- Remove a `line-height` declaration only when its value is degenerate: unitless/`em`/`rem` `< 1`, `%` `<= 100`, `px`/`pt` `<= 1`. Normal values (`1`, `1.5`, `150%`, `20px`, `normal`, `calc(...)`, negatives) are kept. Removing `1px`/`100%` is intentional, pre-existing behavior (TF-3964).
- Strip a trailing `!important` (any whitespace) with a linear helper, then match the number with a whitespace-free, `^`-anchored, bounded regex — O(n), no ReDoS.

## Demo

https://github.com/user-attachments/assets/8697719f-1842-4620-827a-24067f1b6c78

## Summary by CodeRabbit

- **New Features**
  - Composer identity signatures are now normalized before insertion for consistent rendering.
  - Added HTML document-to-editor fragment conversion that preserves `<head>` styles and extracts body content.

- **Bug Fixes**
  - Improved normalization/removal of ineffective `line-height` in inline styles with case-insensitive and unit-aware handling.
  - More reliably preserves `line-height` when it’s meaningful, including `line-height: 0`, and avoids touching custom properties, quoted content, function values, and malformed CSS.
  - Removes the entire `style` attribute when it would otherwise only contain ineffective `line-height`.

- **Performance**
  - Added safeguards and broader coverage to keep style processing fast even with pathological inputs.

### DIFF
--- a/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
+++ b/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
@@ -12,13 +12,24 @@ import 'package:html/dom.dart';
 class NormalizeLineHeightInStyleTransformer extends DomTransformer {
   const NormalizeLineHeightInStyleTransformer();
 
-  // Removal bounds. Unitless/em/rem is exclusive, so a normal `line-height:1` is
-  // kept, while `%` and `px`/`pt` are inclusive: `100%` and `1px` are stripped
-  // deliberately (TF-3964 — those exact values broke email-preview rendering).
-  // The unitless-vs-`%` asymmetry is intentional, not an oversight.
-  static const _unitlessRemovalThreshold = 1.0; // unitless / em / rem multiplier
-  static const _percentRemovalThreshold = 100.0; // %
-  static const _absoluteRemovalThreshold = 1.0; // px / pt
+  // Removal bounds. Major mail clients (Gmail, Outlook, Apple Mail) never
+  // rewrite `line-height`, so intervene as narrowly as possible — two classes
+  // of removal only:
+  //  - Degenerate: element-relative values small enough to visibly overlap
+  //    text lines (< 0.8× font-size; the UA default is ~1.2×). Values in
+  //    [0.8, 1) such as `90%`/`0.9em` are tight-but-legitimate typography and
+  //    are preserved for fidelity — especially since reply/draft pipelines
+  //    save and send the transformed style.
+  //  - TF-3964 legacy: exactly `100%` (and `1em`, the same computed length),
+  //    plus absolute values up to `1px`/`1pt`. These shipped as removals and
+  //    must not silently come back.
+  //  `rem` is root-relative, so overlap cannot be judged from the number
+  //  alone (`0.9rem` ≈ 14.4px is fine leading for 12px text) — never removed.
+  static const _degenerateRatioBound = 0.8; // unitless & em, exclusive
+  static const _degeneratePercentBound = 80.0; // %, exclusive
+  static const _legacyRatio = 1.0; // TF-3964: `1em` ≡ `100%`
+  static const _legacyPercent = 100.0; // TF-3964
+  static const _absoluteRemovalThreshold = 1.0; // px / pt, inclusive
 
   /// A CSS number + optional unit. `!important` is stripped beforehand by
   /// [_stripImportant], so the pattern needs no whitespace: keywords,
@@ -120,7 +131,10 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
       } else if (char == '\\') {
         if (next.isEmpty) return null; // trailing backslash → fail-open
         i++; // escaped char is literal data (e.g. `\;`), never a boundary
-      } else if (char == '/' && next == '*') {
+      } else if (char == '/' && next == '*' && parenDepth == 0) {
+        // Inside parentheses `/*` is data, not a comment — an unquoted
+        // `url(https://x/a/*/b.png)` must not poison the whole attribute
+        // (fail-open would keep a degenerate line-height → TF-3964 returns).
         inComment = true;
         i++;
       } else if (char == "'") {
@@ -186,12 +200,16 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
 
     switch (match.group(2)?.toLowerCase()) {
       case '%':
-        return number <= _percentRemovalThreshold;
+        return number < _degeneratePercentBound || number == _legacyPercent;
+      case 'em':
+        return number < _degenerateRatioBound || number == _legacyRatio;
+      case 'rem':
+        return false; // root-relative: overlap cannot be judged from the number
       case 'px':
       case 'pt':
         return number <= _absoluteRemovalThreshold;
-      default: // unitless / em / rem multiplier
-        return number < _unitlessRemovalThreshold;
+      default: // unitless multiplier
+        return number < _degenerateRatioBound;
     }
   }
 }
--- a/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
+++ b/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
@@ -7,6 +7,8 @@ import 'package:html/dom.dart';
 /// Strips `line-height` values small enough to overlap text lines (e.g. some
 /// signatures ship `line-height:0.1`), letting the renderer fall back to its
 /// default leading. Removed rather than clamped, matching correct clients.
+/// `line-height:0` is left untouched: it is the intentional image-gap/spacer
+/// trick, not an overlap bug.
 class NormalizeLineHeightInStyleTransformer extends DomTransformer {
   const NormalizeLineHeightInStyleTransformer();
 
@@ -176,6 +178,11 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
 
     final number = double.tryParse(match.group(1) ?? '');
     if (number == null || !number.isFinite) return false;
+
+    // `line-height:0` (any unit) is the intentional image-gap/spacer trick used
+    // to collapse whitespace around images; preserve it. Only strip the small
+    // non-zero values that overlap text.
+    if (number <= 0) return false;
 
     switch (match.group(2)?.toLowerCase()) {
       case '%':
--- a/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
+++ b/core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
@@ -4,20 +4,32 @@ import 'package:core/utils/app_logger.dart';
 import 'package:flutter/widgets.dart' show visibleForTesting;
 import 'package:html/dom.dart';
 
+/// Strips `line-height` values small enough to overlap text lines (e.g. some
+/// signatures ship `line-height:0.1`), letting the renderer fall back to its
+/// default leading. Removed rather than clamped, matching correct clients.
 class NormalizeLineHeightInStyleTransformer extends DomTransformer {
   const NormalizeLineHeightInStyleTransformer();
 
-  static const _removableLineHeights = ['1px', '100%'];
-  static final _multipleSpacesRegex = RegExp(r' {2,}');
-  static final _lineHeightPattern = RegExp(
-    r'line-height\s*:\s*(?:' +
-        _removableLineHeights.map(RegExp.escape).join('|') +
-        r')\s*;?',
+  // Removal bounds. Unitless/em/rem is exclusive, so a normal `line-height:1` is
+  // kept, while `%` and `px`/`pt` are inclusive: `100%` and `1px` are stripped
+  // deliberately (TF-3964 — those exact values broke email-preview rendering).
+  // The unitless-vs-`%` asymmetry is intentional, not an oversight.
+  static const _unitlessRemovalThreshold = 1.0; // unitless / em / rem multiplier
+  static const _percentRemovalThreshold = 100.0; // %
+  static const _absoluteRemovalThreshold = 1.0; // px / pt
+
+  /// A CSS number + optional unit. `!important` is stripped beforehand by
+  /// [_stripImportant], so the pattern needs no whitespace: keywords,
+  /// `calc(...)`, negatives, and >9-digit values simply do not match and are
+  /// kept. No whitespace + bounded `\d{1,9}` + `^` anchor make matching
+  /// constant-work — ReDoS-safe (ADR 0069-redos-vulnerability-mitigation).
+  static final _numericValuePattern = RegExp(
+    r'^\+?(\d{1,9}(?:\.\d{0,9})?|\.\d{1,9})(px|pt|em|rem|%)?$',
     caseSensitive: false,
   );
 
   @visibleForTesting
-  static RegExp get lineHeightPattern => _lineHeightPattern;
+  static RegExp get numericValuePattern => _numericValuePattern;
 
   @override
   Future<void> process({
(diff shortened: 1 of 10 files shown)


core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
```diff
@@ -4,20 +4,32 @@ import 'package:core/utils/app_logger.dart';
...
-    r'line-height\s*:\s*(?:' +
-        _removableLineHeights.map(RegExp.escape).join('|') +
-        r')\s*;?',
+  // Removal bounds. Unitless/em/rem is exclusive, so a normal `line-height:1` is
+  // kept, while `%` and `px`/`pt` are inclusive: `100%` and `1px` are stripped
+  // deliberately (TF-3964 — those exact values broke email-preview rendering).
+  // The unitless-vs-`%` asymmetry is intentional, not an oversight.
+  static const _unitlessRemovalThreshold = 1.0; // unitless / em / rem multiplier
+  static const _percentRemovalThreshold = 100.0; // %
+  static const _absoluteRemovalThreshold = 1.0; // px / pt
```
hoangdat: Tested with the issue's .eml — fix works. But `line-height:0` now gets stripped too, which breaks the image-gap/spacer trick, and `em` under 1 is new. Both intentional?

core/test/utils/redos_vulnerability_test.dart
```diff
@@ -93,91 +105,102 @@ void main() {
...
+        expect(element, isNotNull, reason: 'No styled element found in: $html');
+        return (
+          elapsedMs: stopwatch.elapsedMilliseconds,
+          style: element!.attributes['style'] ?? '',
+        );
+      }
 
-        expect(stopwatch.elapsedMilliseconds, lessThan(100));
+      // 500ms: catches ReDoS (which would be seconds) without CI flakiness.
+      const performanceBudgetMs = 500;
```
hoangdat: 500ms looks unneeded here — measured 13ms for the 120KB style attr and 5ms for the 10k-space one on this implementation. The other budgets are 100ms; a 5x gap would hide a real perf regression.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
dab246: > `em`/`rem < 1`:
>
intentional — same multiplier semantics as unitless, so they must share the threshold (otherwise `0.1em` would still overlap while `0.1` is fixed).

> `line-height:0`
>
unintended regression — now preserved (any unit). It's the deliberate image-gap/spacer trick, not an overlap bug. Added tests for `0`/`0.0`/`0px`/`0em`/`0%` and a doc note.

core/test/utils/redos_vulnerability_test.dart
dab246: Fixed. Lowered to 100ms to match the other budgets and keep a real perf regression visible. Tests still pass with wide margin.

core/test/utils/redos_vulnerability_test.dart
```diff
@@ -433,33 +429,23 @@ void main() {
         // Test with a very long string that looks like base64 but isn't
         final longInvalidBase64 = 'A' * 10000 + '!'; // Not valid base64
 
-        final stopwatch = Stopwatch()..start();
-        StringConvert.extractEmailAddress(longInvalidBase64);
-        stopwatch.stop();
-
-        expect(stopwatch.elapsedMilliseconds, lessThan(100));
+        expectFastResult(() => StringConvert.extractEmailAddress(longInvalidBase64));
       });
```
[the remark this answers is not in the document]
dab246: Fixed. Root cause: input length was 10001, so the `length % 4 == 0` gate in `extractStrings` skipped t
         he base64 regex entirely. Changed to length 10000 (`'A' * 9999 + '!'`) so `_base64ValidationRegex.hasMat
         ch` actually runs.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4725

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4725

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
hoangdat: Zero fix looks good — checked `0`, `0px`, `0em`, `0%` etc. all preserved now, and the `font-size:0;line-height:0` spacer survives intact. Thanks!

On the `em` half I was wrong — `em` inherits as a computed length like `%`, so stripping `0.9em` is right. Left a separate note about `1em` though.

core/test/utils/redos_vulnerability_test.dart
hoangdat: Thanks — 100ms is good. Also checked the refactor kept every other budget the same as master, and the fix to assert timing only after a successful return is a nice catch.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
```diff
@@ -26,29 +40,158 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
...
+    if (number <= 0) return false;
+
+    switch (match.group(2)?.toLowerCase()) {
+      case '%':
+        return number <= _percentRemovalThreshold;
+      case 'px':
+      case 'pt':
+        return number <= _absoluteRemovalThreshold;
+      default: // unitless / em / rem multiplier
+        return number < _unitlessRemovalThreshold;
```
hoangdat: Small inconsistency: `1em` is kept but `100%` is removed, even though they compute to the same length and inherit the same way. `em`/`rem` sit in the unitless bucket here, but they inherit as a length like `%`, not as a multiplier. So if the `100%` case from TF-3964 was real, `1em` is the same thing in another unit and still slips through. Not a blocker.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
```diff
@@ -26,29 +40,158 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
...
+    if (number == null || !number.isFinite) return false;
+
+    // `line-height:0` (any unit) is the intentional image-gap/spacer trick used
+    // to collapse whitespace around images; preserve it. Only strip the small
+    // non-zero values that overlap text.
+    if (number <= 0) return false;
+
+    switch (match.group(2)?.toLowerCase()) {
+      case '%':
+        return number <= _percentRemovalThreshold;
```
hoangdat: **Line-height stripping got much wider**

We now strip any line-height under 1x, not just the two old values. Emails using 90% or 0.9em suddenly render ~20% taller, and on Reply/Draft the stripped style gets saved and sent, so the sender's original formatting is lost.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
```diff
@@ -26,29 +40,158 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
...
+      } else if (inDoubleQuote) {
+        if (char == '\\') {
+          i++;
+        } else if (char == '"') {
+          inDoubleQuote = false;
+        }
+      } else if (char == '\\') {
+        if (next.isEmpty) return null; // trailing backslash → fail-open
+        i++; // escaped char is literal data (e.g. `\;`), never a boundary
+      } else if (char == '/' && next == '*') {
```
hoangdat: **TF-3964 partially comes back**

The new CSS parser bails on the whole style attribute if it sees anything malformed — even a `/*` inside a `url()`. When it bails, `line-height:1px` survives and text overlaps again. That's TF-3964 returning on some emails.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
```diff
@@ -26,29 +40,158 @@ class NormalizeLineHeightInStyleTransformer extends DomTransformer {
...
+    // non-zero values that overlap text.
+    if (number <= 0) return false;
+
+    switch (match.group(2)?.toLowerCase()) {
+      case '%':
+        return number <= _percentRemovalThreshold;
+      case 'px':
+      case 'pt':
+        return number <= _absoluteRemovalThreshold;
+      default: // unitless / em / rem multiplier
```
hoangdat: **rem values wrongly stripped**

We treat rem like a multiplier, but rem is root-relative. So `line-height:0.9rem` (~14px, totally fine) gets stripped too. Rare in email HTML, low impact, but the classification is just wrong.

hoangdat (commented):
Regression pass over `a8eaf6e64..HEAD`. Tests are green (68 in the two changed test files, 52 in `html_transform_text_html_test.dart`), and the `!important` handling + quote/paren-aware splitting are real improvements over the old regex. Three things I'd like the team's read on, left inline.

#2 is a one-line fix and I think it should go in regardless, since it silently reopens a shipped fix. #1 needs a decision — was the "under 1x" rule intentional, or did we over-generalize from the `0.1` case in the ticket?

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
dab246: Fixed. `em`/`rem` now use the same inclusive `<= 1` bound as `%`, since they inherit as computed lengths rather than multipliers. As a result, `1em` and `1rem` are stripped like `100%`, while unitless values keep the exclusive bound. Tests updated.

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
dab246: fixed

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
dab246: fixed

core/lib/presentation/utils/html_transformer/dom/normalize_line_height_in_style_transformer.dart
dab246: fixed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4725

hoangdat:
this test raise `red`

```
test(
  'Should keep the signature stylesheet when inserting into the editor\n'
  'When the transformed document carries a <style> block in <head>',
() async {
  // A signature whose first node is a <style> block parses with that block
  // hoisted into <head> — this is the exact document the
  // forComposerSignature() pipeline returns for such a signature.
  const transformedDocument =
      '<html><head><style>.sig{color:#093}</style></head>'
      '<body><p class="sig">Alice</p></body></html>';
  composerController?.richTextMobileTabletController =
      mockRichTextMobileTabletController;
  when(mockRichTextMobileTabletController.htmlEditorApi)
      .thenReturn(mockHtmlEditorApi);
  when(mockTransformHtmlEmailContentInteractor.execute(any, any))
      .thenAnswer((_) => Stream.value(
          Right(TransformHtmlEmailContentSuccess(transformedDocument))));

  await composerController?.applySignature(rawSignature);

  final inserted = verify(mockHtmlEditorApi.insertSignature(
    captureAny,
    allowCollapsed: false,
  )).captured.single as String;
  expect(
    inserted,
    contains('<style>.sig{color:#093}</style>'),
    reason: 'Dropping <head> discards the signature stylesheet',
  );
  expect(inserted, contains('<p class="sig">Alice</p>'));
});
```

hoangdat:
also `red`

```
test(
  'Should not flatten a table-fragment signature\n'
  'When the signature is a bare table row',
() async {
  const transformedDocument =
      '<html><head></head><body><tr><td style="padding:4px">Jane Doe</td></tr></body></html>';
  composerController?.richTextMobileTabletController =
      mockRichTextMobileTabletController;
  when(mockRichTextMobileTabletController.htmlEditorApi)
      .thenReturn(mockHtmlEditorApi);
  when(mockTransformHtmlEmailContentInteractor.execute(any, any))
      .thenAnswer((_) => Stream.value(
          Right(TransformHtmlEmailContentSuccess(transformedDocument))));

  await composerController?.applySignature(rawSignature);

  final inserted = verify(mockHtmlEditorApi.insertSignature(
    captureAny, allowCollapsed: false,
  )).captured.single as String;
  expect(inserted, contains('<td'), reason: 'Re-serialising drops stray table tags');
});
```

dab246:
> also `red`
> 
> ```
> test(
>   'Should not flatten a table-fragment signature\n'
>   'When the signature is a bare table row',
> () async {
>   const transformedDocument =
>       '<html><head></head><body><tr><td style="padding:4px">Jane Doe</td></tr></body></html>';
>   composerController?.richTextMobileTabletController =
>       mockRichTextMobileTabletController;
>   when(mockRichTextMobileTabletController.htmlEditorApi)
>       .thenReturn(mockHtmlEditorApi);
>   when(mockTransformHtmlEmailContentInteractor.execute(any, any))
>       .thenAnswer((_) => Stream.value(
>           Right(TransformHtmlEmailContentSuccess(transformedDocument))));
> 
>   await composerController?.applySignature(rawSignature);
> 
>   final inserted = verify(mockHtmlEditorApi.insertSignature(
>     captureAny, allowCollapsed: false,
>   )).captured.single as String;
>   expect(inserted, contains('<td'), reason: 'Re-serialising drops stray table tags');
> });
> ```

This fixture is unreachable because `forComposerSignature()` parses the signature with the HTML5 parser before any transformer runs, which already removes bare `<tr>/<td>` elements at the body level, so the test fails on its own mock rather than `_normalizeSignature`; browser `innerHTML` behaves the same way, while complete `<table>` signatures remain intact, so no change is needed.

dab246:
> this test raise `red`
> 
> ```
> test(
>   'Should keep the signature stylesheet when inserting into the editor\n'
>   'When the transformed document carries a <style> block in <head>',
> () async {
>   // A signature whose first node is a <style> block parses with that block
>   // hoisted into <head> — this is the exact document the
>   // forComposerSignature() pipeline returns for such a signature.
>   const transformedDocument =
>       '<html><head><style>.sig{color:#093}</style></head>'
>       '<body><p class="sig">Alice</p></body></html>';
>   composerController?.richTextMobileTabletController =
>       mockRichTextMobileTabletController;
>   when(mockRichTextMobileTabletController.htmlEditorApi)
>       .thenReturn(mockHtmlEditorApi);
>   when(mockTransformHtmlEmailContentInteractor.execute(any, any))
>       .thenAnswer((_) => Stream.value(
>           Right(TransformHtmlEmailContentSuccess(transformedDocument))));
> 
>   await composerController?.applySignature(rawSignature);
> 
>   final inserted = verify(mockHtmlEditorApi.insertSignature(
>     captureAny,
>     allowCollapsed: false,
>   )).captured.single as String;
>   expect(
>     inserted,
>     contains('<style>.sig{color:#093}</style>'),
>     reason: 'Dropping <head> discards the signature stylesheet',
>   );
>   expect(inserted, contains('<p class="sig">Alice</p>'));
> });
> ```

Fixed.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4725

hoangdat (approved):

tddang-linagora (approved):

PR #4729, opened 2026-07-27
TF-4726 Composable validator

### LINKED ISSUE #4726: Add composable attachment-upload size validator
### Background

Per [ADR-0103](../../docs/adr/0103-attach-drive-file-as-attachment.md#decision) ("Validation — minimal, condition-agnostic abstraction"), attachment upload needs a size check that runs *before* any upload starts, and the codebase needs this check to be extensible: today `UploadController.validateTotalSizeAttachmentsBeforeUpload` is called ad-hoc from several composer call sites, and there's no way to add a new upload-eligibility condition without touching every call site.

The ADR specifies a minimal validator abstraction:

```dart
abstract class AttachmentUploadValidator {
  FutureOr<bool> validate();
}

class CompositeAttachmentUploadValidator implements AttachmentUploadValidator {
  final List<AttachmentUploadValidator> validators;
  const CompositeAttachmentUploadValidator(this.validators);

  @override
  Future<bool> validate() async {
    if (validators.isEmpty) return true;
    for (final v in validators) {
      if (!await v.validate()) return false;
    }
    return true;
  }
}
```

No size (or other condition-specific data) belongs in the interface itself — each concrete validator captures what it needs via its own constructor and is fully self-contained, including showing its own dialog if it needs one.

### Scope

1. Add `AttachmentUploadValidator` and `CompositeAttachmentUploadValidator` (short-circuit on first `false`).
2. Add `SizeLimitAttachmentUploadValidator` wrapping the existing `UploadController.validateTotalSizeAttachmentsBeforeUpload` size/warning check — move its body in, keep `UploadController` with a thin wrapper.
3. **Fix the callback shape first** — today `validateTotalSizeAttachmentsBeforeUpload` only fires `onValidationSuccess` on the allow path; on hard-limit-exceeded, null-context, and warning-dialog cancel/dismiss it just returns without calling anything. A validator wired only to the success callback via a `Completer` would hang forever on any of those paths. Add an `onValidationFailure` callback (or return `Future<bool>` directly) and make sure it fires on every one of those exits — the hard-limit dialog, both null-context returns, and the warning dialog's cancel/close actions. `validate()` must resolve on every exit path, not only the happy path.
4. Update every composer call site that currently calls `validateTotalSizeAttachmentsBeforeUpload` directly to instead construct a `SizeLimitAttachmentUploadValidator` and gate its action on `if (await validator.validate()) { ...existing action... }`. Each call site keeps its own distinct action (e.g. `uploadAttachmentsAction(...)`, `initializeUploadAttachments(...)`) — this is a control-flow refactor at each site, not a dedup of identical code.
5. Leave the *inline* image-attachment validation path (`validateTotalSizeInlineAttachmentsBeforeUpload`) untouched — different shape, out of scope (YAGNI).

### Acceptance criteria

- [x] `AttachmentUploadValidator` / `CompositeAttachmentUploadValidator` / `SizeLimitAttachmentUploadValidator` exist and are unit tested, including the composite's short-circuit behavior.
- [x] `validate()` resolves `false` on hard-limit-exceeded, null-context, and warning-dialog-cancel; resolves `true` only when upload is actually allowed. No path leaves the returned future unresolved.
- [x] Every existing attachment call site (local file pick, drag/drop, paste) goes through the validator and preserves its prior user-visible behavior (same dialogs, same thresholds).
- [x] Adding a new, unrelated upload-eligibility condition later requires only one new class + one line in a `CompositeAttachmentUploadValidator` list — no call-site changes.
- [x] `flutter analyze` and `flutter test` pass for touched files.

### Out of scope

- Drive-specific attachment handling (downloading/uploading drive documents) — that consumes this validator but is not part of this issue.
- The inline-image validation path.

tddang-linagora:
## Issue
- #4726

## Summary
Extracts attachment size-limit validation out of `UploadController` into a composable validator pipeline (`domain/validator/`), fixes a dialog-dismiss deadlock, and routes all composer attachment/inline-image entry points through one validation hook — groundwork for ADR-0103 (drive attachment download/upload).

## Changes
- Add generic `ValidationRule<Request, Failure, Prompt>` / `ValidationPipeline` / `ValidationDecision` (`ValidationAllowed` / `ValidationRejected` / `ValidationConfirmationRequired`) kernel in `domain/validator/`.
- Add `AttachmentUploadRequest` (sizes + limits), `AttachmentSizeLimitPolicy` (pure size-comparison helpers), `AttachmentSizeLimitRule` implementing the hard-limit + warning-limit checks previously duplicated across composer call sites, and `AttachmentUploadRequestFactory` to build requests from picked files or raw byte counts.
- Add `AttachmentUploadGate` (presentation) to run the validator against a request and delegate to `AttachmentValidationFeedback` (implemented by `FlutterAttachmentValidationFeedback`) for failure/confirmation dialogs — exposed via a single overridable `validateAttachmentSize` hook on `ComposerController`, replacing 5 duplicated call sites (pick file/image, drop zone, paste, share).
- `UploadController`: remove `validateTotalSizeAttachmentsBeforeUpload` / `validateTotalSizeInlineAttachmentsBeforeUpload` and their private dialog helpers entirely; validation now lives in the gate + rule, not the controller.
- Dialog fix: `UploadSizeWarningDialogPresenter`'s close button now calls `popBack()` before `cancelAction`, and `outsideDismissible: false`, so `AttachmentConfirmationPresenter`'s awaited result always resolves instead of hanging.
- Docs: update ADR-0103 and the drive-attachment-refactor plan to match the new validator/gate boundary.

## Summary by CodeRabbit

* **New Features**
  * Added centralized attachment-size validation for file selection, drag-and-drop, paste, and sending.
  * Uploads exceeding server limits are blocked with a clear explanation.
  * Large regular attachments prompt for confirmation before uploading.
  * Validation accounts for existing attachments, inline content, concurrent uploads, and current limits.
  * Added consistent dialogs for size errors and warnings.

* **Bug Fixes**
  * Canceled or dismissed dialogs now reliably prevent uploads.

* **Tests**
  * Expanded coverage for validation, limits, dialogs, confirmations, and upload flows.

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -89,6 +89,7 @@ import 'package:tmail_ui_user/features/composer/presentation/model/prefix_recipi
 import 'package:tmail_ui_user/features/composer/presentation/model/saved_composing_email.dart';
 import 'package:tmail_ui_user/features/composer/presentation/model/screen_display_mode.dart';
 import 'package:tmail_ui_user/features/composer/presentation/styles/composer_style.dart';
+import 'package:tmail_ui_user/features/composer/presentation/validator/composer_attachment_upload_state_source.dart';
 import 'package:tmail_ui_user/features/composer/presentation/view/editor_view_mixin.dart';
 import 'package:tmail_ui_user/features/composer/presentation/widgets/mobile/from_composer_bottom_sheet_builder.dart';
 import 'package:tmail_ui_user/features/composer/presentation/widgets/saving_message_dialog_view.dart';
@@ -119,7 +120,6 @@ import 'package:tmail_ui_user/features/network_connection/presentation/network_c
 import 'package:tmail_ui_user/features/server_settings/domain/usecases/get_server_setting_interactor.dart';
 import 'package:tmail_ui_user/features/upload/domain/exceptions/pick_file_exception.dart';
 import 'package:tmail_ui_user/features/upload/domain/extensions/file_info_extension.dart';
-import 'package:tmail_ui_user/features/upload/domain/extensions/list_file_info_extension.dart';
 import 'package:tmail_ui_user/features/upload/domain/extensions/list_file_upload_extension.dart';
 import 'package:tmail_ui_user/features/upload/domain/model/upload_task_id.dart';
 import 'package:tmail_ui_user/features/upload/domain/state/attachment_upload_state.dart';
@@ -128,6 +128,7 @@ import 'package:tmail_ui_user/features/upload/domain/state/local_image_picker_st
 import 'package:tmail_ui_user/features/upload/domain/usecases/local_file_picker_interactor.dart';
 import 'package:tmail_ui_user/features/upload/domain/usecases/local_image_picker_interactor.dart';
 import 'package:tmail_ui_user/features/upload/presentation/controller/upload_controller.dart';
+import 'package:tmail_ui_user/features/upload/presentation/validator/attachment_upload_validation_service.dart';
 import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
@@ -186,6 +187,14 @@ class ComposerController extends BaseController
   final String? autoSaveComposerId;
   final ComposerArguments? composerArgs;
   final SaveTemplateEmailInteractor _saveTemplateEmailInteractor;
+
+  late AttachmentUploadValidationService attachmentUploadValidationService =
+      AttachmentUploadValidationService(
+        stateSource: ComposerAttachmentUploadStateSource.fromServerCapability(
+          uploadController: uploadController,
+          maxSizeAttachmentsPerEmail: mailboxDashBoardController.maxSizeAttachmentsPerEmail?.value,
+        ),
+      );
 
   GetAllAutoCompleteInteractor? _getAllAutoCompleteInteractor;
   GetAutoCompleteInteractor? _getAutoCompleteInteractor;
@@ -860,7 +869,7 @@ class ComposerController extends BaseController
       return;
     }
 
-    if (uploadController.isExceededMaxSizeAttachmentsPerEmail()) {
+    if (attachmentUploadValidationService.isExceededMaxSizeAttachmentsPerEmail()) {
       MessageDialogActionManager().showConfirmDialogAction(
         context,
         appLocalizations.message_dialog_send_email_exceeds_maximum_size(
@@ -1241,18 +1250,17 @@ class ComposerController extends BaseController
     }
   }
 
-  void _handlePickFileSuccess(LocalFilePickerSuccess success) {
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
-      totalSizePreparedFiles: success.pickedFiles.totalSize,
-      onValidationSuccess: () => uploadAttachmentsAction(pickedFiles: success.pickedFiles)
-    );
+  Future<void> _handlePickFileSuccess(LocalFilePickerSuccess success) {
+    return attachmentUploadValidationService.validateFiles(
+      files: success.pickedFiles,
+      onAllowed: () => uploadAttachmentsAction(pickedFiles: success.pickedFiles));
   }
 
-  void _handlePickImageSuccess(LocalImagePickerSuccess success) {
-    uploadController.validateTotalSizeInlineAttachmentsBeforeUpload(
-      totalSizePreparedFiles: success.fileInfo.fileSize,
-      onValidationSuccess: () => uploadAttachmentsAction(pickedFiles: [success.fileInfo.withInline()])
-    );
+  Future<void> _handlePickImageSuccess(LocalImagePickerSuccess success) {
+    final inlineFile = success.fileInfo.withInline();
+    return attachmentUploadValidationService.validateFiles(
+      files: [inlineFile],
+      onAllowed: () => uploadAttachmentsAction(pickedFiles: [inlineFile]));
   }
 
   void uploadAttachmentsAction({required List<FileInfo> pickedFiles}) {
@@ -2047,12 +2055,12 @@ class ComposerController extends BaseController
 
   void setSubjectEmail(String subject) => subjectEmail.value = subject;
 
-  void onAttachmentDropZoneListener(Attachment attachment) {
+  Future<void> onAttachmentDropZoneListener(BuildContext context, Attachment attachment) {
     log('ComposerController::onAttachmentDropZoneListener: attachment = $attachment');
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
-      totalSizePreparedFiles: attachment.size?.value ?? 0,
-      onValidationSuccess: () => uploadController.initializeUploadAttachments([attachment])
-    );
+    return attachmentUploadValidationService.validateAttachment(
+      context: context,
+      attachment: attachment,
+      onAllowed: () => uploadController.initializeUploadAttachments([attachment]));
   }
 
   Future<void> onChangeIdentity(Identity? newIdentity) async {
@@ -2189,7 +2197,9 @@ class ComposerController extends BaseController
 
     final listFileInfo = await onDragDone(context: context, details: details);
 
-    if (listFileInfo.isEmpty && context.mounted) {
+    if (!context.mounted) return;
+
+    if (listFileInfo.isEmpty) {
       appToast.showToastErrorMessage(
         context,
         AppLocalizations.of(context).can_not_upload_this_file_as_attachments
@@ -2197,11 +2207,10 @@ class ComposerController extends BaseController
       return;
     }
 
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
-      totalSizePreparedFiles: listFileInfo.totalSize,
-      totalSizePreparedFilesWithDispositionAttachment: listFileInfo.listAttachmentFiles.totalSize,
-      onValidationSuccess: () => uploadAttachmentsAction(pickedFiles: listFileInfo)
-    );
+    await attachmentUploadValidationService.validateFiles(
+      context: context,
+      files: listFileInfo,
+      onAllowed: () => uploadAttachmentsAction(pickedFiles: listFileInfo));
   }
 
   void _handleSaveMessageToDraft(BuildContext context) async {
@@ -2473,16 +2482,16 @@ class ComposerController extends BaseController
     required BuildContext context,
     required double maxWidth,
     required List<FileUpload> listFileUpload
-  }) {
+  }) async {
     log('ComposerController::handleOnPasteImageSuccessAction: listFileUpload = ${listFileUpload.length}');
     _setUpMaxWidthInlineImage(context: context, maxWidth: maxWidth);
 
     final listFileInfo = listFileUpload.toListFileInfo();
 
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
-      totalSizePreparedFiles: listFileInfo.totalSize,
-      onValidationSuccess: () => uploadAttachmentsAction(pickedFiles: listFileInfo)
-    );
+    await attachmentUploadValidationService.validateFiles(
+      context: context,
+      files: listFileInfo,
+      onAllowed: () => uploadAttachmentsAction(pickedFiles: listFileInfo));
   }
 
   void handleOnPasteImageFailureAction({
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1295,6 +1295,20 @@ class ComposerController extends BaseController
     );
   }
 
+  /// Mirrors the bytes reservation [AttachmentUploadValidationService] made
+  /// for a re-attached [attachment] whose bytes are transferred to
+  /// [uploadController]'s state synchronously by
+  /// [initializeUploadAttachments], so there is no later settlement event to
+  /// release it on.
+  void _releaseReservedBytesForAttachment(Attachment attachment) {
+    final attachmentBytes = (attachment.size?.value ?? 0).toInt();
+    attachmentUploadValidationService.releaseReservedBytes(
+      allAttachmentBytes: attachmentBytes,
+      regularAttachmentBytes:
+          attachment.isDispositionInlined() ? 0 : attachmentBytes,
+    );
+  }
+
   void deleteAttachmentUploaded(UploadTaskId uploadId) {
     uploadController.deleteFileUploaded(uploadId);
   }
@@ -2073,7 +2087,10 @@ class ComposerController extends BaseController
     return attachmentUploadValidationService.validateAttachment(
       context: context,
       attachment: attachment,
-      onAllowed: () => uploadController.initializeUploadAttachments([attachment]));
+      onAllowed: () {
+        uploadController.initializeUploadAttachments([attachment]);
+        _releaseReservedBytesForAttachment(attachment);
+      });
   }
 
   Future<void> onChangeIdentity(Identity? newIdentity) async {
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -192,7 +192,7 @@ class ComposerController extends BaseController
       AttachmentUploadValidationService(
         stateSource: ComposerAttachmentUploadStateSource.fromServerCapability(
           uploadController: uploadController,
-          maxSizeAttachmentsPerEmail: mailboxDashBoardController.maxSizeAttachmentsPerEmail?.value,
+          maxSizeAttachmentsPerEmail: () => mailboxDashBoardController.maxSizeAttachmentsPerEmail?.value,
         ),
       );
 
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1272,41 +1272,14 @@ class ComposerController extends BaseController
         uploadController.justUploadAttachmentsAction(
           uploadFiles: pickedFiles,
           uploadUri: uploadUri,
-          onFileSettled: _releaseReservedBytesForFile,
         );
       } catch (e) {
         logWarning('ComposerController::uploadAttachmentsAction: $e');
         uploadController.consumeState(Stream.value(Left(UploadAttachmentFailure(e, pickedFiles[0]))));
-        pickedFiles.forEach(_releaseReservedBytesForFile);
       }
     } else {
       logWarning('ComposerController::uploadAttachmentsAction: SESSION OR ACCOUNT_ID is NULL');
-      pickedFiles.forEach(_releaseReservedBytesForFile);
     }
-  }
-
-  /// Mirrors the bytes reservation [AttachmentUploadValidationService] made
-  /// for [file] when its upload was allowed, now that it has actually
-  /// settled (or will never start).
-  void _releaseReservedBytesForFile(FileInfo file) {
-    attachmentUploadValidationService.releaseReservedBytes(
-      allAttachmentBytes: file.fileSize,
-      regularAttachmentBytes: file.isInline == true ? 0 : file.fileSize,
-    );
-  }
-
-  /// Mirrors the bytes reservation [AttachmentUploadValidationService] made
-  /// for a re-attached [attachment] whose bytes are transferred to
-  /// [uploadController]'s state synchronously by
-  /// [initializeUploadAttachments], so there is no later settlement event to
-  /// release it on.
-  void _releaseReservedBytesForAttachment(Attachment attachment) {
-    final attachmentBytes = (attachment.size?.value ?? 0).toInt();
-    attachmentUploadValidationService.releaseReservedBytes(
-      allAttachmentBytes: attachmentBytes,
-      regularAttachmentBytes:
-          attachment.isDispositionInlined() ? 0 : attachmentBytes,
-    );
   }
 
   void deleteAttachmentUploaded(UploadTaskId uploadId) {
@@ -2087,10 +2060,7 @@ class ComposerController extends BaseController
     return attachmentUploadValidationService.validateAttachment(
       context: context,
       attachment: attachment,
-      onAllowed: () {
-        uploadController.initializeUploadAttachments([attachment]);
-        _releaseReservedBytesForAttachment(attachment);
-      });
+      onAllowed: () => uploadController.initializeUploadAttachments([attachment]));
   }
 
   Future<void> onChangeIdentity(Identity? newIdentity) async {
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1272,14 +1272,27 @@ class ComposerController extends BaseController
         uploadController.justUploadAttachmentsAction(
           uploadFiles: pickedFiles,
           uploadUri: uploadUri,
+          onFileSettled: _releaseReservedBytesForFile,
         );
       } catch (e) {
         logWarning('ComposerController::uploadAttachmentsAction: $e');
         uploadController.consumeState(Stream.value(Left(UploadAttachmentFailure(e, pickedFiles[0]))));
+        pickedFiles.forEach(_releaseReservedBytesForFile);
       }
     } else {
       logWarning('ComposerController::uploadAttachmentsAction: SESSION OR ACCOUNT_ID is NULL');
+      pickedFiles.forEach(_releaseReservedBytesForFile);
     }
+  }
+
+  /// Mirrors the bytes reservation [AttachmentUploadValidationService] made
+  /// for [file] when its upload was allowed, now that it has actually
+  /// settled (or will never start).
+  void _releaseReservedBytesForFile(FileInfo file) {
+    attachmentUploadValidationService.releaseReservedBytes(
+      allAttachmentBytes: file.fileSize,
+      regularAttachmentBytes: file.isInline == true ? 0 : file.fileSize,
+    );
   }
 
   void deleteAttachmentUploaded(UploadTaskId uploadId) {
--- a/lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
+++ b/lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
@@ -12,27 +12,29 @@ class ComposerAttachmentUploadStateSource implements AttachmentUploadStateSource
 
   @override
   final int warningLimitBytes;
-  @override
-  final int? hardLimitBytes;
+  final num? Function() maxSizeAttachmentsPerEmailProvider;
 
   const ComposerAttachmentUploadStateSource({
     required this.uploadController,
     required this.warningLimitBytes,
-    required this.hardLimitBytes,
+    required this.maxSizeAttachmentsPerEmailProvider,
   });
 
   /// Resolves the limits from the raw server capability value.
   factory ComposerAttachmentUploadStateSource.fromServerCapability({
     required UploadController uploadController,
-    required num? maxSizeAttachmentsPerEmail,
+    required num? Function() maxSizeAttachmentsPerEmail,
   }) {
     return ComposerAttachmentUploadStateSource(
       uploadController: uploadController,
       warningLimitBytes: AppConfig.warningAttachmentFileSizeInMegabytes * 1024 * 1024,
-      hardLimitBytes: AttachmentUploadRequestFactory.normalizeServerLimitBytes(
-          maxSizeAttachmentsPerEmail),
+      maxSizeAttachmentsPerEmailProvider: maxSizeAttachmentsPerEmail,
     );
   }
+
+  @override
+  int? get hardLimitBytes => AttachmentUploadRequestFactory
+      .normalizeServerLimitBytes(maxSizeAttachmentsPerEmailProvider());
 
   @override
   int get currentAllAttachmentBytes =>
--- a/lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
+++ b/lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
@@ -1,0 +1,45 @@
+
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_request_factory.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_state_source.dart';
+import 'package:tmail_ui_user/features/upload/presentation/controller/upload_controller.dart';
+import 'package:tmail_ui_user/main/utils/app_config.dart';
+
+/// Reads the composer's current upload state for the attachment upload
+/// validator.
+class ComposerAttachmentUploadStateSource implements AttachmentUploadStateSource {
+  /// Live reference: byte totals below are read from it on every access.
+  final UploadController uploadController;
+
+  @override
+  final int warningLimitBytes;
+  @override
+  final int? hardLimitBytes;
+
+  const ComposerAttachmentUploadStateSource({
+    required this.uploadController,
+    required this.warningLimitBytes,
+    required this.hardLimitBytes,
+  });
+
+  /// Resolves the limits from the raw server capability value.
+  factory ComposerAttachmentUploadStateSource.fromServerCapability({
+    required UploadController uploadController,
+    required num? maxSizeAttachmentsPerEmail,
+  }) {
+    return ComposerAttachmentUploadStateSource(
+      uploadController: uploadController,
+      warningLimitBytes: AppConfig.warningAttachmentFileSizeInMegabytes * 1024 * 1024,
+      hardLimitBytes: AttachmentUploadRequestFactory.normalizeServerLimitBytes(
+          maxSizeAttachmentsPerEmail),
+    );
+  }
+
+  @override
+  int get currentAllAttachmentBytes =>
+      uploadController.regularAttachmentsTotalBytes +
+      uploadController.inlineAttachmentsTotalBytes;
+
+  @override
+  int get currentRegularAttachmentBytes =>
+      uploadController.regularAttachmentsTotalBytes;
+}
--- a/lib/features/upload/domain/validator/attachment_size_limit_policy.dart
+++ b/lib/features/upload/domain/validator/attachment_size_limit_policy.dart
@@ -1,0 +1,19 @@
+
+class AttachmentSizeLimitPolicy {
+  const AttachmentSizeLimitPolicy._();
+
+  static bool isExceededMaxSizeAttachmentsPerEmail({
+    required int allAttachmentBytes,
+    required int? hardLimitBytes,
+  }) {
+    if (hardLimitBytes == null) return false;
+    return allAttachmentBytes > hardLimitBytes;
+  }
+
+  static bool isExceededWarningAttachmentFileSizeInComposer({
+    required int regularAttachmentBytes,
+    required int warningLimitBytes,
+  }) {
+    return regularAttachmentBytes > warningLimitBytes;
+  }
+}
--- a/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
+++ b/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
@@ -17,9 +17,10 @@ final class AttachmentSizeLimitRule implements AttachmentUploadRule {
       return ValidationRejected(MaxEmailAttachmentSizeExceeded(hardLimit!));
     }
 
-    if (AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer(
-        regularAttachmentBytes: request.sizes.regularAttachmentBytes,
-        warningLimitBytes: request.limits.warningLimitBytes)) {
+    if (request.sizes.proposedRegularAttachmentBytes > 0 &&
+        AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer(
+            regularAttachmentBytes: request.sizes.regularAttachmentBytes,
+            warningLimitBytes: request.limits.warningLimitBytes)) {
       return ValidationConfirmationRequired(const [LargeRegularAttachmentWarning()]);
     }
 
--- a/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
+++ b/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
@@ -1,0 +1,28 @@
+
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_size_limit_policy.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_failure.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_prompt.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_request.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/validation_decision.dart';
+
+final class AttachmentSizeLimitRule implements AttachmentUploadRule {
+  const AttachmentSizeLimitRule();
+
+  @override
+  AttachmentUploadDecision validate(AttachmentUploadRequest request) {
+    final hardLimit = request.limits.hardLimitBytes;
+    if (AttachmentSizeLimitPolicy.isExceededMaxSizeAttachmentsPerEmail(
+        allAttachmentBytes: request.sizes.allAttachmentBytes,
+        hardLimitBytes: hardLimit)) {
+      return ValidationRejected(MaxEmailAttachmentSizeExceeded(hardLimit!));
+    }
+
+    if (AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer(
+        regularAttachmentBytes: request.sizes.regularAttachmentBytes,
+        warningLimitBytes: request.limits.warningLimitBytes)) {
+      return ValidationConfirmationRequired(const [LargeRegularAttachmentWarning()]);
+    }
+
+    return const ValidationAllowed();
+  }
+}
--- a/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
+++ b/lib/features/upload/domain/validator/attachment_size_limit_rule.dart
@@ -1,4 +1,5 @@
 
+import 'package:core/utils/app_logger.dart';
 import 'package:tmail_ui_user/features/upload/domain/validator/attachment_size_limit_policy.dart';
 import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_failure.dart';
 import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_prompt.dart';
@@ -14,6 +15,9 @@ final class AttachmentSizeLimitRule implements AttachmentUploadRule {
     if (AttachmentSizeLimitPolicy.isExceededMaxSizeAttachmentsPerEmail(
         allAttachmentBytes: request.sizes.allAttachmentBytes,
         hardLimitBytes: hardLimit)) {
+      logWarning('AttachmentSizeLimitRule::validate: rejected, '
+          'allAttachmentBytes = ${request.sizes.allAttachmentBytes} | '
+          'hardLimitBytes = $hardLimit');
       return ValidationRejected(MaxEmailAttachmentSizeExceeded(hardLimit!));
     }
 
@@ -21,6 +25,9 @@ final class AttachmentSizeLimitRule implements AttachmentUploadRule {
         AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer(
             regularAttachmentBytes: request.sizes.regularAttachmentBytes,
             warningLimitBytes: request.limits.warningLimitBytes)) {
+      log('AttachmentSizeLimitRule::validate: confirmation required, '
+          'regularAttachmentBytes = ${request.sizes.regularAttachmentBytes} | '
+          'warningLimitBytes = ${request.limits.warningLimitBytes}');
       return ValidationConfirmationRequired(const [LargeRegularAttachmentWarning()]);
     }
 
--- a/lib/features/upload/domain/validator/attachment_upload_request.dart
+++ b/lib/features/upload/domain/validator/attachment_upload_request.dart
@@ -1,0 +1,19 @@
+
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_failure.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_limits.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_prompt.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_size_snapshot.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/validation_decision.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/validation_pipeline.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/validation_rule.dart';
+
+final class AttachmentUploadRequest {
+  final AttachmentUploadSizeSnapshot sizes;
+  final AttachmentUploadLimits limits;
+
+  const AttachmentUploadRequest({required this.sizes, required this.limits});
+}
+
+typedef AttachmentUploadDecision = ValidationDecision<AttachmentUploadFailure, AttachmentUploadPrompt>;
+typedef AttachmentUploadRule = ValidationRule<AttachmentUploadRequest, AttachmentUploadFailure, AttachmentUploadPrompt>;
+typedef AttachmentUploadValidator = ValidationPipeline<AttachmentUploadRequest, AttachmentUploadFailure, AttachmentUploadPrompt>;
--- a/lib/features/upload/domain/validator/attachment_upload_request_factory.dart
+++ b/lib/features/upload/domain/validator/attachment_upload_request_factory.dart
@@ -1,0 +1,71 @@
+
+import 'package:model/upload/file_info.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_limits.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_request.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_size_snapshot.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_state_source.dart';
+
+final class AttachmentUploadRequestFactory {
+  const AttachmentUploadRequestFactory();
+
+  AttachmentUploadRequest fromProposedFiles({
+    required Iterable<FileInfo> files,
+    required AttachmentUploadStateSource state,
+  }) {
+    final candidate = List<FileInfo>.unmodifiable(files);
+    final proposedAllBytes = candidate.fold<int>(0, (total, file) => total + _byteCount(file));
+    final proposedRegularBytes = candidate
+        .where((file) => file.isInline != true)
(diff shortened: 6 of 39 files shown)


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1240,11 +1241,26 @@ class ComposerController extends BaseController
     }
   }
 
-  void _handlePickFileSuccess(LocalFilePickerSuccess success) {
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
+  void _handlePickFileSuccess(LocalFilePickerSuccess success) async {
```
dab246: Should return Future

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -2013,12 +2029,11 @@ class ComposerController extends BaseController
 
   void setSubjectEmail(String subject) => subjectEmail.value = subject;
 
-  void onAttachmentDropZoneListener(Attachment attachment) {
+  void onAttachmentDropZoneListener(Attachment attachment) async {
```
dab246: idem

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1240,11 +1241,26 @@ class ComposerController extends BaseController
...
+  }
+
+  Future<void> _runIfAttachmentSizeAllowed(
+    {
+      required num totalSizePreparedFiles,
+      num? totalSizePreparedFilesWithDispositionAttachment,
+      required VoidCallback onAllowed
+    }
+  ) async {
+    final validator = SizeLimitAttachmentUploadValidator(
```
hoangdat: we added `CompositeAttachmentUploadValidator`, but it quite not in use?

test/features/upload/domain/validator/composite_attachment_upload_validator_test.dart
```diff
@@ -0,0 +1,52 @@
...
+      final composite = CompositeAttachmentUploadValidator([first, second]);
+
+      final result = await composite.validate();
+
+      expect(result, isTrue);
+      expect(first.callCount, 1);
+      expect(second.callCount, 1);
+    });
+
+    test('should short-circuit and return false on first failing validator', () async {
```
hoangdat: please test also the case of: `first(true), second(false)`

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -323,22 +323,26 @@ class UploadController extends BaseController {
   void validateTotalSizeAttachmentsBeforeUpload({
```
hoangdat: @dab246 where we validate `maxSizeUpload`?

hoangdat: @tddang-linagora @dab246 how about we also move out the logic of validation from the upload controller?

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -396,7 +402,10 @@ class UploadController extends BaseController {
       appLocalizations.continueAction,
       cancelTitle: appLocalizations.cancel,
       alignCenter: true,
+      outsideDismissible: false,
       onConfirmAction: confirmAction,
+      onCancelAction: cancelAction,
+      onCloseButtonAction: cancelAction,
```
dab246: The close button does not dismiss the dialog because `cancelAction` is passed directly to `onCloseButtonAction` without calling `popBack()`, so the validator resolves to `false` while the dialog remains open, and clicking **Cancel** or **Continue** afterward completes the same `Completer` again and may throw an unhandled `StateError`.

dab246: Also write unit test for this case

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: ~It's a poc for when we need to combine validators~
Updating

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

hoangdat:
- how to handle error for each validator?
- how to show toast/dialog in case of a validator failed?

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

test/features/upload/domain/validator/composite_attachment_upload_validator_test.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart

tddang-linagora:
> * how to handle error for each validator?
> 
>     * how to show toast/dialog in case of a validator failed?

Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -302,104 +301,29 @@ class UploadController extends BaseController {
...
       }
 
-      _showConfirmDialogWhenExceededMaxSizeAttachmentsPerEmail(context: currentContext!);
+      MaxSizeAttachmentsDialogPresenter.show(
+        context: currentContext!,
+        maxSizeAttachmentsPerEmail: _mailboxDashBoardController.maxSizeAttachmentsPerEmail?.value);
       return;
     }
 
     onValidationSuccess?.call();
```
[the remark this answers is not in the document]

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

hoangdat:
I propose the idea

## The abstraction

```dart
abstract class FileUploadRequest {}
```

```dart
// What we're checking.
class AttachmentUploadRequest extend FileUploadRequest { ...files, current sizes, server cap... }

// What a rule can answer.
sealed class FileValidationResult {}
class ValidationPassed            extends FileValidationResult {}
class ValidationRejected          extends FileValidationResult { final FileValidationFailure failure; }
class ValidationNeedsConfirmation extends AttachmentValidationResult { final AttachmentConfirmationPrompt prompt; }

// A rule. Pure, sync, boring on purpose.
abstract class FileValidationRule {
  AttachmentValidationResult validate(FileUploadRequest request);
}

// The only thing call sites touch.
class FileUploadValidator {
  final List<FileValidationRule> rules;
  AttachmentValidationResult validate(FileUploadRequest request);
}
```

Failures carry data, not words — so the domain never imports `AppLocalizations`:

```dart
sealed class FileValidationFailure {}
class MaxSizePerEmailExceeded extends FileValidationFailure { final num? maxSize; }
class UnsupportedFileType     extends FileValidationFailure { final List<FileInfo> rejectedFiles; }
```

## How the composite runs them

```dart
for (final rule in rules) {
  switch (rule.validate(request)) {
    case ValidationRejected(:final failure): return ValidationRejected(failure);  // hard stop
    case ValidationNeedsConfirmation(:final p): pendingPrompt ??= p;              // remember, keep going
    case ValidationPassed(): break;
  }
}
return pendingPrompt == null ? ValidationPassed() : ValidationNeedsConfirmation(pendingPrompt);
```

## Controller 

```dart
switch (validator.validate(request)) {
  case ValidationPassed():
    onAllowed();                                    // upload starts here, and only here
  case ValidationRejected(:final failure):
    FailurePresenter.present(context, failure);     // dialog or toast, one place decides
  case ValidationNeedsConfirmation(:final prompt):
    if (await ConfirmationPresenter.ask(context, prompt)) onAllowed();
}
```

tddang-linagora:
> I propose the idea
> ## The abstraction
> 
> ```dart
> abstract class FileUploadRequest {}
> ```
> 
> ```dart
> // What we're checking.
> class AttachmentUploadRequest extend FileUploadRequest { ...files, current sizes, server cap... }
> 
> // What a rule can answer.
> sealed class FileValidationResult {}
> class ValidationPassed            extends FileValidationResult {}
> class ValidationRejected          extends FileValidationResult { final FileValidationFailure failure; }
> class ValidationNeedsConfirmation extends AttachmentValidationResult { final AttachmentConfirmationPrompt prompt; }
> 
> // A rule. Pure, sync, boring on purpose.
> abstract class FileValidationRule {
>   AttachmentValidationResult validate(FileUploadRequest request);
> }
> 
> // The only thing call sites touch.
> class FileUploadValidator {
>   final List<FileValidationRule> rules;
>   AttachmentValidationResult validate(FileUploadRequest request);
> }
> ```
> 
> Failures carry data, not words — so the domain never imports `AppLocalizations`:
> 
> ```dart
> sealed class FileValidationFailure {}
> class MaxSizePerEmailExceeded extends FileValidationFailure { final num? maxSize; }
> class UnsupportedFileType     extends FileValidationFailure { final List<FileInfo> rejectedFiles; }
> ```
> 
> ## How the composite runs them
> 
> ```dart
> for (final rule in rules) {
>   switch (rule.validate(request)) {
>     case ValidationRejected(:final failure): return ValidationRejected(failure);  // hard stop
>     case ValidationNeedsConfirmation(:final p): pendingPrompt ??= p;              // remember, keep going
>     case ValidationPassed(): break;
>   }
> }
> return pendingPrompt == null ? ValidationPassed() : ValidationNeedsConfirmation(pendingPrompt);
> ```
> 
> ## Controller
> 
> ```dart
> switch (validator.validate(request)) {
>   case ValidationPassed():
>     onAllowed();                                    // upload starts here, and only here
>   case ValidationRejected(:final failure):
>     FailurePresenter.present(context, failure);     // dialog or toast, one place decides
>   case ValidationNeedsConfirmation(:final prompt):
>     if (await ConfirmationPresenter.ask(context, prompt)) onAllowed();
> }
> ```

Done

lib/features/upload/domain/validator/attachment_size_limit_policy.dart
```diff
@@ -0,0 +1,23 @@
+
+import 'package:tmail_ui_user/main/utils/app_config.dart';
+
+class AttachmentSizeLimitPolicy {
+  const AttachmentSizeLimitPolicy._();
+
+  static bool isExceededMaxSizeAttachmentsPerEmail({
+    required num currentPickedTotalSize,
+    required num totalSizePreparedFiles,
+    required num? maxSizeAttachmentsPerEmail,
```
hoangdat:
```suggestion
    required num? serverMaxSizeAttachmentsPerEmail,
```

lib/features/upload/domain/validator/attachment_upload_request.dart
```diff
@@ -0,0 +1,18 @@
+
+import 'package:tmail_ui_user/features/upload/domain/validator/file_upload_request.dart';
+
+class AttachmentUploadRequest extends FileUploadRequest {
+  final num totalSizePreparedFiles;
+  final num? totalSizePreparedFilesWithDispositionAttachment;
+  final num currentPickedTotalSize;
+  final num currentAttachmentsPickedTotalSize;
+  final num? maxSizeAttachmentsPerEmail;
```
hoangdat:
```suggestion
  final num? serverMaxSizeAttachmentsPerEmail;
```

lib/features/upload/domain/validator/attachment_size_limit_policy.dart
tddang-linagora: Done

lib/features/upload/domain/validator/attachment_upload_request.dart
tddang-linagora: Done

lib/features/upload/presentation/validator/attachment_upload_validation_dispatcher.dart
```diff
@@ -0,0 +1,27 @@
+import 'package:flutter/material.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/file_upload_request.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/file_upload_validator.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/file_validation_result.dart';
+import 'package:tmail_ui_user/features/upload/presentation/dialog/attachment_confirmation_presenter.dart';
+import 'package:tmail_ui_user/features/upload/presentation/dialog/attachment_validation_failure_presenter.dart';
+import 'package:tmail_ui_user/main/routes/route_navigation.dart';
+
+Future<void> runIfAttachmentUploadAllowed({
```
hoangdat: maybe write it the same style other presenter, private constructor and static function? WDYT?

test/features/upload/presentation/dialog/attachment_validation_failure_presenter_test.dart
```diff
@@ -0,0 +1,35 @@
...
+    testWidgets('should show the max-size dialog for MaxSizePerEmailExceeded', (tester) async {
+      await tester.pumpWidget(WidgetFixtures.makeTestableWidget(child: const SizedBox.shrink()));
+      await tester.pump();
+
+      AttachmentValidationFailurePresenter.present(
+        tester.element(find.byType(SizedBox)),
+        const MaxSizePerEmailExceeded(1000000));
+      await tester.pump();
+
+      expect(find.byKey(const Key('confirm_dialog_action')), findsOneWidget);
```
hoangdat, line 37: how about find the title and message?

test/features/upload/presentation/dialog/max_size_attachments_dialog_presenter_test.dart
```diff
@@ -0,0 +1,38 @@
...
+      await tester.pumpWidget(WidgetFixtures.makeTestableWidget(
+        child: const SizedBox.shrink()));
+      await tester.pump();
+
+      MaxSizeAttachmentsDialogPresenter.show(
+        context: tester.element(find.byType(SizedBox)),
+        maxSizeAttachmentsPerEmail: 1000000);
+      await tester.pump();
+
+      expect(find.byKey(const Key('confirm_dialog_action')), findsOneWidget);
```
hoangdat, line 40: idem

test/features/upload/presentation/dialog/attachment_confirmation_presenter_test.dart
```diff
@@ -0,0 +1,58 @@
...
+        const AttachmentSizeWarningPrompt());
+      await tester.pump();
+
+      final closeButton = find.byWidgetPredicate((widget) =>
+        widget is TMailButtonWidget && widget.icon == ImagePaths().icCloseDialog);
+      await tester.tap(closeButton);
+      await tester.pumpAndSettle();
+
+      // Regression: the dialog's barrier must be non-dismissible and the close
+      // button must resolve the completer, or this future would hang forever.
```
hoangdat: how we address it?

lib/features/upload/presentation/dialog/attachment_confirmation_presenter.dart
```diff
@@ -0,0 +1,21 @@
...
+  static Future<bool> ask(BuildContext context, AttachmentConfirmationPrompt prompt) {
+    final completer = Completer<bool>();
+    switch (prompt) {
+      case AttachmentSizeWarningPrompt():
+        UploadSizeWarningDialogPresenter.showExceededWarningSize(
+          context: context,
+          confirmAction: () => completer.complete(true),
+          cancelAction: () => completer.complete(false));
+    }
+    return completer.future;
```
dab246: Confirmation Future can remain pending. `Navigator/back/programmatic` overlay dismissal can close the dialog without invoking these callbacks, leaving `runIfAttachmentUploadAllowed` suspended and retaining its upload closure.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1240,18 +1276,16 @@ class ComposerController extends BaseController
...
   }
 
-  void _handlePickImageSuccess(LocalImagePickerSuccess success) {
-    uploadController.validateTotalSizeInlineAttachmentsBeforeUpload(
+  Future<void> _handlePickImageSuccess(LocalImagePickerSuccess success) {
+    return validateAttachmentSize(
       totalSizePreparedFiles: success.fileInfo.fileSize,
-      onValidationSuccess: () => uploadAttachmentsAction(pickedFiles: [success.fileInfo.withInline()])
-    );
+      onAllowed: () => uploadAttachmentsAction(pickedFiles: [success.fileInfo.withInline()]));
```
dab246: Inline uploads now show warning. This path previously used inline-only validation, which enforced the server hard cap but did not require the regular-attachment warning confirmation. Give inline uploads a separate rule set/scope to preserve behavior.

lib/features/upload/domain/validator/size_limit_attachment_validation_rule.dart
```diff
@@ -0,0 +1,33 @@
...
+import 'package:tmail_ui_user/features/upload/domain/validator/file_validation_failure.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/file_validation_result.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/file_validation_rule.dart';
+
+class SizeLimitAttachmentValidationRule implements FileValidationRule {
+  const SizeLimitAttachmentValidationRule();
+
+  @override
+  FileValidationResult validate(FileUploadRequest request) {
+    request as AttachmentUploadRequest;
```
dab246: A rule accepts any FileUploadRequest while the concrete attachment rule casts it at runtime. The generic result also imports an attachment prompt, so the abstraction is neither type-safe nor reusable. Prefer attachment-specific contracts until a second real upload context exists.

lib/features/upload/presentation/validator/attachment_upload_validation_dispatcher.dart
tddang-linagora: Done

test/features/upload/presentation/dialog/attachment_validation_failure_presenter_test.dart
tddang-linagora: Done

test/features/upload/presentation/dialog/max_size_attachments_dialog_presenter_test.dart
tddang-linagora: Done

test/features/upload/presentation/dialog/attachment_confirmation_presenter_test.dart
tddang-linagora: Done

dab246 (commented):
I propose upload-validation the following architecture. WDYT?


## End-to-end upload flow

```mermaid
flowchart TD
  A["User supplies file(s)<br/>picker, paste, drop, share, or Drive"] --> B["Classify candidate<br/>regular attachment or inline image"]
  B --> C["AttachmentUploadRequestFactory<br/>snapshot current attachments, selected files, and limits"]
  C --> D["AttachmentUploadGate.permits(request)"]
  D --> E["ValidationPipeline<br/>run ordered pure rules"]
  E --> F{"Validation decision"}

  F -->|Rejected| G["Feedback shows localized failure"]
  G --> H["Stop<br/>no upload task is created"]

  F -->|Confirmation required| I["Feedback presents every confirmation prompt"]
  I --> J{"All prompts explicitly confirmed?"}
  J -->|No, dismissed, Back, pop, or unmounted| H
  J -->|Yes| K["Composer starts upload"]

  F -->|Allowed| K
  K --> L["UploadController creates task(s)<br/>waiting → uploading → progress"]
  L --> M{"Upload terminal state"}
  M -->|Success| N["Store attachment state and update composer"]
  M -->|Failure or cancellation| O["Mark/remove failed task and present upload feedback"]
```

`AttachmentUploadGate` is only a pre-upload decision point. `UploadController` remains responsible for upload task lifecycle, progress, cancellation, retry behavior, and final state.

## Classes and contracts to add

### 1. Generic validation kernel

```dart
abstract interface class ValidationRule<Request, Failure, Prompt> {
  ValidationDecision<Failure, Prompt> validate(Request request);
}

sealed class ValidationDecision<Failure, Prompt> {
  const ValidationDecision();
}

final class ValidationAllowed<Failure, Prompt> extends ValidationDecision<Failure, Prompt> {
  const ValidationAllowed();
}

final class ValidationRejected<Failure, Prompt> extends ValidationDecision<Failure, Prompt> {
  const ValidationRejected(this.failure);

  final Failure failure;
}

final class ValidationConfirmationRequired<Failure, Prompt> extends ValidationDecision<Failure, Prompt> {
  ValidationConfirmationRequired(Iterable<Prompt> prompts) : prompts = List.unmodifiable(prompts);

  final List<Prompt> prompts;
}

final class ValidationPipeline<Request, Failure, Prompt> {
  ValidationPipeline(Iterable<ValidationRule<Request, Failure, Prompt>> rules) : _rules = List.unmodifiable(rules);

  final List<ValidationRule<Request, Failure, Prompt>> _rules;

  ValidationDecision<Failure, Prompt> validate(Request request) {
    final prompts = <Prompt>[];
    for (final rule in _rules) {
      switch (rule.validate(request)) {
        case ValidationAllowed():
          break;
        case ValidationRejected(:final failure):
          return ValidationRejected(failure);
        case ValidationConfirmationRequired(prompts: final rulePrompts):
          prompts.addAll(rulePrompts);
      }
    }
    return prompts.isEmpty ? const ValidationAllowed() : ValidationConfirmationRequired(prompts);
  }
}
```

### 2. Attachment domain model and size-limit rule

```dart
final class AttachmentUploadLimits {
  const AttachmentUploadLimits({required this.warningLimitBytes, this.hardLimitBytes});

  final int warningLimitBytes;
  final int? hardLimitBytes;
}

final class AttachmentUploadSizeSnapshot {
  const AttachmentUploadSizeSnapshot({
    required this.currentAllAttachmentBytes,
    required this.proposedAllAttachmentBytes,
    required this.currentRegularAttachmentBytes,
    required this.proposedRegularAttachmentBytes,
  });

  final int currentAllAttachmentBytes;
  final int proposedAllAttachmentBytes;
  final int currentRegularAttachmentBytes;
  final int proposedRegularAttachmentBytes;

  int get allAttachmentBytes => currentAllAttachmentBytes + proposedAllAttachmentBytes;
  int get regularAttachmentBytes => currentRegularAttachmentBytes + proposedRegularAttachmentBytes;
}

final class AttachmentUploadRequest {
  const AttachmentUploadRequest({required this.sizes, required this.limits});

  final AttachmentUploadSizeSnapshot sizes;
  final AttachmentUploadLimits limits;
}

sealed class AttachmentUploadFailure {
  const AttachmentUploadFailure();
}

final class MaxEmailAttachmentSizeExceeded extends AttachmentUploadFailure {
  const MaxEmailAttachmentSizeExceeded(this.maximumBytes);

  final int maximumBytes;
}

sealed class AttachmentUploadPrompt {
  const AttachmentUploadPrompt();
}

final class LargeRegularAttachmentWarning extends AttachmentUploadPrompt {
  const LargeRegularAttachmentWarning();
}

typedef AttachmentUploadDecision = ValidationDecision<AttachmentUploadFailure, AttachmentUploadPrompt>;
typedef AttachmentUploadRule = ValidationRule<AttachmentUploadRequest, AttachmentUploadFailure, AttachmentUploadPrompt>;
typedef AttachmentUploadValidator = ValidationPipeline<AttachmentUploadRequest, AttachmentUploadFailure, AttachmentUploadPrompt>;

final class AttachmentSizeLimitRule implements AttachmentUploadRule {
  const AttachmentSizeLimitRule();

  @override
  AttachmentUploadDecision validate(AttachmentUploadRequest request) {
    final hardLimit = request.limits.hardLimitBytes;
    if (hardLimit != null && request.sizes.allAttachmentBytes > hardLimit) {
      return ValidationRejected(MaxEmailAttachmentSizeExceeded(hardLimit));
    }
    if (request.sizes.regularAttachmentBytes > request.limits.warningLimitBytes) {
      return ValidationConfirmationRequired([const LargeRegularAttachmentWarning()]);
    }
    return const ValidationAllowed();
  }
}
```

The request must represent an immutable byte snapshot. Inline images contribute to `proposedAllAttachmentBytes`, but use `proposedRegularAttachmentBytes = 0`, so they obey the hard limit without triggering the regular-attachment warning.

### 3. Application request factory

```dart
final class AttachmentUploadRequestFactory {
  const AttachmentUploadRequestFactory();

  AttachmentUploadRequest fromProposedFiles({
    required Iterable<FileInfo> files,
    required int currentAllAttachmentBytes,
    required int currentRegularAttachmentBytes,
    required AttachmentUploadLimits limits,
  }) {
    final candidate = List<FileInfo>.unmodifiable(files);
    final proposedAllBytes = candidate.fold<int>(0, (total, file) => total + _byteCount(file));
    final proposedRegularBytes = candidate.where((file) => file.isInline != true).fold<int>(0, (total, file) => total + _byteCount(file));
    return AttachmentUploadRequest(
      sizes: AttachmentUploadSizeSnapshot(
        currentAllAttachmentBytes: _requireNonNegative(currentAllAttachmentBytes, 'currentAllAttachmentBytes'),
        proposedAllAttachmentBytes: proposedAllBytes,
        currentRegularAttachmentBytes: _requireNonNegative(currentRegularAttachmentBytes, 'currentRegularAttachmentBytes'),
        proposedRegularAttachmentBytes: proposedRegularBytes,
      ),
      limits: limits,
    );
  }

  int _byteCount(FileInfo file) {
    return _requireNonNegative(file.fileSize, 'file.fileSize');
  }

  int _requireNonNegative(int value, String name) {
    if (value < 0) throw ArgumentError.value(value, name, 'must be non-negative');
    return value;
  }
}
```

The caller supplies the current totals from `UploadController` and a limit snapshot built from `AppConfig` plus the server capability. The composition boundary must normalize the current `num` server-cap API to a non-negative integral `int?` before it creates `AttachmentUploadLimits`.

### 4. Application gate, feedback, and dialog adapters

```dart
abstract interface class AttachmentValidationFeedback {
  Future<void> showFailure(AttachmentUploadFailure failure);
  Future<bool> confirmAll(List<AttachmentUploadPrompt> prompts);
}

final class FlutterAttachmentValidationFeedback implements AttachmentValidationFeedback {
  const FlutterAttachmentValidationFeedback(this._context);

  final BuildContext _context;

  @override
  Future<void> showFailure(AttachmentUploadFailure failure) async {
    if (!_context.mounted) return;
    AttachmentValidationFailurePresenter.present(_context, failure);
  }

  @override
  Future<bool> confirmAll(List<AttachmentUploadPrompt> prompts) async {
    for (final prompt in prompts) {
      if (!_context.mounted) return false;
      if (!await AttachmentConfirmationPresenter.ask(_context, prompt)) return false;
    }
    return true;
  }
}

final class AttachmentUploadGate {
  const AttachmentUploadGate(this._validator);

  final AttachmentUploadValidator _validator;

  Future<bool> permits({required AttachmentUploadRequest request, required AttachmentValidationFeedback feedback}) async {
    switch (_validator.validate(request)) {
      case ValidationAllowed():
        return true;
      case ValidationRejected(:final failure):
        await feedback.showFailure(failure);
        return false;
      case ValidationConfirmationRequired(:final prompts):
        return feedback.confirmAll(prompts);
    }
  }
}
```

`FlutterAttachmentValidationFeedback` is created per user action and is passed directly to the gate, so it has no global context provider, singleton state, or unused dialog abstraction. The interface remains necessary because the gate is unit-tested with a fake feedback implementation rather than Flutter dialogs.

### 5. Update the two existing presentation adapters

```dart
final class AttachmentValidationFailurePresenter {
  const AttachmentValidationFailurePresenter._();

  static void present(BuildContext context, AttachmentUploadFailure failure) {
    switch (failure) {
      case MaxEmailAttachmentSizeExceeded(:final maximumBytes):
        MaxSizeAttachmentsDialogPresenter.show(
          context: context,
          maxSizeAttachmentsPerEmail: maximumBytes,
        );
    }
  }
}

final class AttachmentConfirmationPresenter {
  const AttachmentConfirmationPresenter._();

  static Future<bool> ask(BuildContext context, AttachmentUploadPrompt prompt) async {
    var confirmed = false;
    switch (prompt) {
      case LargeRegularAttachmentWarning():
        await UploadSizeWarningDialogPresenter.showExceededWarningSize(
          context: context,
          confirmAction: () => confirmed = true,
        );
    }
    return confirmed;
  }
}

final class UploadSizeWarningDialogPresenter {
  const UploadSizeWarningDialogPresenter._();

  static Future<void> showExceededWarningSize({required BuildContext context, required VoidCallback confirmAction}) {
    final localizations = AppLocalizations.of(context);
    return MessageDialogActionManager().showConfirmDialogAction(
      context,
      localizations.warningMessageWhenExceedGenerallySizeInComposer,
      localizations.continueAction,
      title: '',
      cancelTitle: localizations.cancel,
      alignCenter: true,
      outsideDismissible: false,
      onConfirmAction: confirmAction,
      onCloseButtonAction: popBack,
    );
  }
}
```

No `Completer` is required. `MessageDialogActionManager.showConfirmDialogAction` already completes when its dialog route closes; the local `confirmed` flag is `false` for Cancel, close, Back, programmatic pop, navigator cleanup, and an unmounted screen. It is `true` only when the user explicitly presses Continue.

### 6. Composition and call site

```dart
// These are private fields of the existing ComposerController.
final _attachmentUploadValidator = AttachmentUploadValidator([const AttachmentSizeLimitRule()]);
final _attachmentUploadGate = AttachmentUploadGate(_attachmentUploadValidator);
static const _attachmentUploadRequestFactory = AttachmentUploadRequestFactory();

Future<void> validateAndUpload({required BuildContext context, required List<FileInfo> files}) async {
  final request = _attachmentUploadRequestFactory.fromProposedFiles(
    files: files,
    currentAllAttachmentBytes: uploadController.attachmentsPicked.totalSize + uploadController.inlineAttachmentsPicked.totalSize,
    currentRegularAttachmentBytes: uploadController.attachmentsPicked.totalSize,
    limits: AttachmentUploadLimits(
      warningLimitBytes: AppConfig.warningAttachmentFileSizeInMegabytes * 1024 * 1024,
      hardLimitBytes: _serverLimitBytes(mailboxDashBoardController.maxSizeAttachmentsPerEmail?.value),
    ),
  );
  final allowed = await _attachmentUploadGate.permits(
    request: request,
    feedback: FlutterAttachmentValidationFeedback(context),
  );
  if (!allowed || !context.mounted) return;
  uploadAttachmentsAction(pickedFiles: files);
}

static int? _serverLimitBytes(num? value) {
  if (value == null) return null;
  if (value is double && !value.isFinite) throw ArgumentError.value(value, 'value', 'must be finite');
  if (value < 0 || value != value.truncate()) throw ArgumentError.value(value, 'value', 'must be a non-negative integer');
  return value.toInt();
}
```

`uploadController` and `mailboxDashBoardController` are existing `ComposerController` dependencies. The view forwards its own mounted `BuildContext` to `validateAndUpload`; the context is retained only for the current validation/dialog operation and is never stored in a singleton or controller field. Every entry point calls `validateAndUpload`: local picker, inline-image picker, paste, drag-and-drop, share intent, and Drive. No entry point calls a validation rule or dialog presenter directly.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/upload/presentation/dialog/attachment_confirmation_presenter.dart
tddang-linagora: Fixed with proposed architecture

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Fixed with proposed architecture

lib/features/upload/domain/validator/size_limit_attachment_validation_rule.dart
tddang-linagora: Fixed with proposed architecture

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/upload/presentation/validator/flutter_attachment_validation_feedback.dart
```diff
@@ -0,0 +1,28 @@
+
+import 'package:flutter/material.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_failure.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_prompt.dart';
+import 'package:tmail_ui_user/features/upload/presentation/dialog/attachment_confirmation_presenter.dart';
+import 'package:tmail_ui_user/features/upload/presentation/dialog/attachment_validation_failure_presenter.dart';
+import 'package:tmail_ui_user/features/upload/presentation/validator/attachment_validation_feedback.dart';
+
+final class FlutterAttachmentValidationFeedback implements AttachmentValidationFeedback {
```
hoangdat:
```suggestion
final class AttachmentValidationFeedbackImp implements AttachmentValidationFeedback {
```

test/features/upload/presentation/validator/attachment_upload_gate_test.dart
```diff
@@ -0,0 +1,110 @@
...
+      await tester.pump();
+
+      final context = tester.element(find.byType(SizedBox));
+
+      final future = gate.permits(
+        request: _request(proposedAllAttachmentBytes: 200, hardLimitBytes: 100),
+        feedback: FlutterAttachmentValidationFeedback(context));
+      await tester.pump();
+
+      expect(find.byKey(const Key('confirm_dialog_action')), findsOneWidget);
```
hoangdat, line 73: do we have better expect to see the message of dedicate dialog?

lib/features/upload/domain/validator/attachment_upload_request_factory.dart
```diff
@@ -0,0 +1,69 @@
+
+import 'package:model/upload/file_info.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_limits.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_request.dart';
+import 'package:tmail_ui_user/features/upload/domain/validator/attachment_upload_size_snapshot.dart';
+
+final class AttachmentUploadRequestFactory {
```
hoangdat, line 8: should add test for this class 

```
fromProposedFiles
- empty list → all snapshot fields 0
- isInline: null → counted in All and Regular
- isInline: false → counted in All and Regular
- isInline: true → counted in All, excluded from Regular ← the important one
- mixed inline + regular → proposedAll = total, proposedRegular = subset
- current* args forwarded unchanged
- negative currentAllAttachmentBytes → ArgumentError

fromProposedBytes — two are enough, it's a pass-through:
- values land in the matching snapshot fields
- negative arg → ArgumentError
```

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -313,6 +332,38 @@ class ComposerController extends BaseController
...
+  AttachmentUploadRequest buildAttachmentUploadRequest(List<FileInfo> files) {
+    return _attachmentUploadRequestFactory.fromProposedFiles(
+      files: files,
+      currentAllAttachmentBytes: (uploadController.attachmentsPicked.totalSize +
+              uploadController.inlineAttachmentsPicked.totalSize)
+          .toInt(),
+      currentRegularAttachmentBytes: uploadController.attachmentsPicked.totalSize.toInt(),
+      limits: _attachmentUploadLimits,
+    );
+  }
```
dab246: Attachment size validation ignores existing attachments.

Reproduce:

- Open a draft containing an existing 19 MB attachment.
- The server hard limit is 20 MB.
- Select an additional 2 MB file.
- The snapshot reports the current attachment size as 0, so the 2 MB file is incorrectly allowed.
- The send-time validation at composer_controller.dart:914 also fails to account for the existing 19 MB attachment.

Risk:

The composer allows attachments to exceed the server hard limit, with the failure occurring only during the upload. Attachments imported from another email are also excluded from subsequent size validation.

test/features/composer/presentation/composer_controller_test.dart
```diff
@@ -1199,6 +1011,69 @@ void main() {
       });
     });
 
+    group('attachment size validation is injectable:', () {
```
dab246, line 1204: The current tests inject a callback and only verify that `initializeUploadAttachments()` is invoked. This bypasses the default validator and does not assert the contents of the generated request.

**Missing test cases:**

* Allow requests when both `BuildContext` and `currentContext` are `null`.
* Validate scenarios where existing attachments are initialized from `Attachment` rather than `FileInfo`.
* Reject uploads when the combined size of current and proposed attachments exceeds the server hard limit.
* Validate mixed inline and regular attachments.
* Verify cumulative size validation when two attachments are dragged consecutively.
* Cover `AttachmentUploadRequestFactory` edge cases, including negative, fractional, or infinite server limits, as well as current/proposed size aggregation.
* Validate send-time checks for draft and restored attachments.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -313,6 +332,38 @@ class ComposerController extends BaseController
...
+    required VoidCallback onAllowed,
+  }) async {
+    final resolvedContext = context ?? currentContext;
+    if (resolvedContext == null) return;
+
+    final allowed = await _attachmentUploadGate.permits(
+      request: request,
+      feedback: FlutterAttachmentValidationFeedback(resolvedContext),
+    );
+    if (allowed) onAllowed();
```
dab246: ## Problem

`ComposerController._defaultValidateAttachmentSize()` checks for a `BuildContext` before running validation:

```dart
final resolvedContext = context ?? currentContext;
if (resolvedContext == null) return;
```

As a result, `onAllowed()` is not called even when the request is valid and no UI is required. The attachment can be silently skipped, especially in file picker or share-intent flows that do not provide a `BuildContext`.

## Solution

Always run validation first and resolve the `BuildContext` only when the decision requires failure feedback or user confirmation. `AttachmentUploadGate` should receive a lazily evaluated feedback factory:

```dart
Future<bool> permits({
  required AttachmentUploadRequest request,
  required AttachmentValidationFeedback? Function() feedbackFactory,
}) async {
  switch (_validator.validate(request)) {
    case ValidationAllowed():
      return true;

    case ValidationRejected(:final failure):
      final feedback = feedbackFactory();
      if (feedback == null) return false;
      await feedback.showFailure(failure);
      return false;

    case ValidationConfirmationRequired(:final prompts):
      final feedback = feedbackFactory();
      if (feedback == null) return false;
      return feedback.confirmAll(prompts);
  }
}
```

In `ComposerController`, create Flutter feedback only when the gate requests it:

```dart
final allowed = await _attachmentUploadGate.permits(
  request: request,
  feedbackFactory: () {
    final resolvedContext = context ?? currentContext;
    return resolvedContext == null
        ? null
        : FlutterAttachmentValidationFeedback(resolvedContext);
  },
);

if (allowed) onAllowed();
```

This ensures that:

- Valid requests are uploaded without requiring a `BuildContext`.
- Requests requiring a dialog are safely rejected when no context is available.
- Validation runs only once, and the presentation dependency is created only when needed.

## Required Test

Add a regression test verifying that `ValidationAllowed` returns `true`, invokes `onAllowed()`, and does not call `feedbackFactory` when no `BuildContext` is available.

lib/features/upload/presentation/validator/flutter_attachment_validation_feedback.dart
tddang-linagora: Done

test/features/upload/presentation/validator/attachment_upload_gate_test.dart
tddang-linagora: Done

lib/features/upload/domain/validator/attachment_upload_request_factory.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

test/features/composer/presentation/composer_controller_test.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/upload/domain/validator/attachment_size_limit_rule.dart
```diff
@@ -0,0 +1,28 @@
...
+  @override
+  AttachmentUploadDecision validate(AttachmentUploadRequest request) {
+    final hardLimit = request.limits.hardLimitBytes;
+    if (AttachmentSizeLimitPolicy.isExceededMaxSizeAttachmentsPerEmail(
+        allAttachmentBytes: request.sizes.allAttachmentBytes,
+        hardLimitBytes: hardLimit)) {
+      return ValidationRejected(MaxEmailAttachmentSizeExceeded(hardLimit!));
+    }
+
+    if (AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer(
```
dab246: Inline-only uploads can still trigger the regular attachment warning. `AttachmentSizeLimitPolicy.isExceededWarningAttachmentFileSizeInComposer()` validates the total `regularAttachmentBytes`, including both existing and newly proposed attachments. If the existing regular attachments already exceed the warning threshold, adding only an inline image with `proposedRegularAttachmentBytes == 0` still returns `ValidationConfirmationRequired`. This is a regression from the previous behavior, where inline uploads were checked only against the hard limit.

https://github.com/user-attachments/assets/db22b22d-15d0-40cc-a943-fd5c037df169

lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
```diff
@@ -0,0 +1,45 @@
...
+/// Reads the composer's current upload state for the attachment upload
+/// validator.
+class ComposerAttachmentUploadStateSource implements AttachmentUploadStateSource {
+  /// Live reference: byte totals below are read from it on every access.
+  final UploadController uploadController;
+
+  @override
+  final int warningLimitBytes;
+  @override
+  final int? hardLimitBytes;
```
dab246: **Hard limit is snapshotted and does not update with a new session.** `hardLimitBytes` is stored as a `final` value when the validation service is first accessed, while `MailboxDashBoardController.maxSizeAttachmentsPerEmail` reads the capability directly from the current session. If a reconnect or session refresh changes the capability:
* The validator may allow uploads that exceed the new hard limit.
* The validator may reject valid uploads based on the old hard limit.
* Send-time validation may use the stale limit while the dialog displays the new limit.

test/features/upload/presentation/validator/attachment_upload_validation_service_test.dart
```diff
@@ -0,0 +1,286 @@
...
+    required AttachmentUploadStateSource stateSource,
+    required AttachmentValidationFeedback feedback,
+  }) {
+    return AttachmentUploadValidationService(
+      stateSource: stateSource,
+      feedbackBuilder: (_) => feedback,
+    );
+  }
+
+  group('AttachmentUploadValidationService::validateFiles', () {
```
dab246, line 90: Concurrent validations can both pass and exceed the hard cap. This test case is fail:

```dart
class _MutableStateSource implements AttachmentUploadStateSource {
  @override
  int currentAllAttachmentBytes;

  @override
  int currentRegularAttachmentBytes;

  @override
  final int warningLimitBytes;

  @override
  final int? hardLimitBytes;

  _MutableStateSource({
    this.currentAllAttachmentBytes = 0,
    this.currentRegularAttachmentBytes = 0,
    this.warningLimitBytes = 10000,
    this.hardLimitBytes,
  });
}

test('Should not allow concurrent uploads to exceed the hard limit', () async {
  final stateSource = _MutableStateSource(
    warningLimitBytes: 10000,
    hardLimitBytes: 1000,
  );
  final service = buildService(
    stateSource: stateSource,
    feedback: _RecordingFeedback(),
  );
  var allowedUploads = 0;

  void registerUpload() {
    allowedUploads++;
    stateSource.currentAllAttachmentBytes += 600;
    stateSource.currentRegularAttachmentBytes += 600;
  }

  final firstValidation = service.validateFiles(
    files: [_file(fileSize: 600)],
    onAllowed: registerUpload,
  );
  final secondValidation = service.validateFiles(
    files: [_file(fileSize: 600)],
    onAllowed: registerUpload,
  );

  await Future.wait([firstValidation, secondValidation]);

  expect(allowedUploads, 1);
  expect(stateSource.currentAllAttachmentBytes, 600);
});
```

lib/features/composer/presentation/validator/composer_attachment_upload_state_source.dart
tddang-linagora: Done

test/features/upload/presentation/validator/attachment_upload_validation_service_test.dart
tddang-linagora: Done

lib/features/upload/domain/validator/attachment_size_limit_rule.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
```diff
@@ -0,0 +1,113 @@
...
+    return previous.then((_) async {
+      try {
+        final allowed = await _gate.permits(
+          request: requestBuilder(),
+          feedbackFactory: () {
+            final resolvedContext = context ?? currentContext;
+            return resolvedContext == null ? null : _feedbackBuilder(resolvedContext);
+          },
+        );
+        if (allowed) onAllowed();
```
dab246: Validation lock releases too early. The validation lock is released immediately after `onAllowed()` returns, but the production callback only starts the upload asynchronously. Since attachment bytes are added to `UploadController` later, the next request can still use the old total and exceed the hard limit. Please keep the lock until pending upload bytes are registered or introduce a reservation mechanism.

lib/features/upload/presentation/validator/attachment_validation_feedback_impl.dart
```diff
@@ -0,0 +1,28 @@
...
+
+final class AttachmentValidationFeedbackImpl implements AttachmentValidationFeedback {
+  final BuildContext _context;
+
+  const AttachmentValidationFeedbackImpl(this._context);
+
+  @override
+  Future<void> showFailure(AttachmentUploadFailure failure) async {
+    if (!_context.mounted) return;
+    AttachmentValidationFailurePresenter.present(_context, failure);
```
dab246: Failure dialog is not awaited. `showFailure()` completes immediately because the presenters return void and ignore the dialog Future. This releases the validation lock while the dialog is still open, allowing another request to display an overlapping dialog. Please return and await the dialog Future through the presenter chain.

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
tddang-linagora: Done

lib/features/upload/presentation/validator/attachment_validation_feedback_impl.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -2047,12 +2068,12 @@ class ComposerController extends BaseController
...
   void setSubjectEmail(String subject) => subjectEmail.value = subject;
 
-  void onAttachmentDropZoneListener(Attachment attachment) {
+  Future<void> onAttachmentDropZoneListener(BuildContext context, Attachment attachment) {
     log('ComposerController::onAttachmentDropZoneListener: attachment = $attachment');
-    uploadController.validateTotalSizeAttachmentsBeforeUpload(
-      totalSizePreparedFiles: attachment.size?.value ?? 0,
-      onValidationSuccess: () => uploadController.initializeUploadAttachments([attachment])
-    );
+    return attachmentUploadValidationService.validateAttachment(
```
dab246, line 2060: `validateAttachment` does not release the reservation after a successful upload. Validation reserves attachment capacity, then `initializeUploadAttachments()` adds the attachment to the actual state without releasing the reservation, causing the attachment size to be counted twice. For example, with a 10 MB hard limit, re-attaching a valid 6 MB file results in the system accounting for 12 MB, which can incorrectly block sending the email or uploading additional attachments. This was reproduced with a unit test: 

```dart
test(
    'Should not double count an allowed re-attached attachment after it is registered',
    () async {
  final stateSource = _MutableStateSource(
    warningLimitBytes: 10000,
    hardLimitBytes: 1000,
  );
  final service = buildService(
    stateSource: stateSource,
    feedback: _RecordingFeedback(),
  );
  final attachment = Attachment(
    blobId: Id('regular-attachment'),
    size: UnsignedInt(600),
    disposition: ContentDisposition.attachment,
  );

  await service.validateAttachment(
    attachment: attachment,
    onAllowed: () {
      // Simulates UploadController.initializeUploadAttachments().
      stateSource.currentAllAttachmentBytes += 600;
      stateSource.currentRegularAttachmentBytes += 600;
    },
  );

  expect(stateSource.currentAllAttachmentBytes, 600);
  expect(service.isExceededMaxSizeAttachmentsPerEmail(), isFalse);
});
```

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
```diff
@@ -0,0 +1,168 @@
...
+      try {
+        final request = requestBuilder();
+        final allowed = await _gate.permits(
+          request: request,
+          feedbackFactory: () {
+            final resolvedContext = context ?? currentContext;
+            return resolvedContext == null ? null : _feedbackBuilder(resolvedContext);
+          },
+        );
+        if (allowed) {
```
dab246: Reservation leaks when onAllowed() throws. At here, the reservation counters are incremented before invoking the callback, but the finally block only releases the lock and does not roll back the reservation if `onAllowed()` throws. As a result, subsequent validations may be incorrectly blocked by stale reserved capacity until the composer is closed.

Unit test fail:

```dart
test('Should roll back reserved bytes when onAllowed throws', () async {
  final stateSource = _MutableStateSource(
    warningLimitBytes: 10000,
    hardLimitBytes: 1000,
  );
  final service = buildService(
    stateSource: stateSource,
    feedback: _RecordingFeedback(),
  );

  await expectLater(
    service.validateFiles(
      files: [_file(fileSize: 600)],
      onAllowed: () => throw StateError('Upload registration failed'),
    ),
    throwsA(isA<StateError>()),
  );

  var secondUploadAllowed = false;

  await service.validateFiles(
    files: [_file(fileSize: 500)],
    onAllowed: () => secondUploadAllowed = true,
  );

  expect(stateSource.currentAllAttachmentBytes, 0);
  expect(secondUploadAllowed, isTrue);
});
```

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Addressed

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
tddang-linagora: Addressed

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

dab246 (approved):
lgtm

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4729

hoangdat:
@coderabbit review

tddang-linagora (commented):
LGTM

PR #4737, opened 2026-07-28
Fix Drive picker stuck 20s on readyTimeout when token_exchange or page load fails

dab246:
## Context

"Attach from Drive" opens a modal with a skeleton loader while the application exchanges the OIDC token, creates a Workplace intent, and loads the Drive picker page into a WebView or iframe. On staging, the CORS preflight for `POST /auth/token_exchange` returns 403, but the modal keeps showing the skeleton for the full 20-second `readyTimeout` before displaying the error toast instead of failing immediately.

https://github.com/user-attachments/assets/a38dbdce-3813-48f4-b489-9df4033c4855

## Root cause

The intent `Future` was created **before** the modal was mounted, while the modal attached its failure handler later in `initState`. A CORS preflight rejection can fail almost immediately, so the `Future` completed with an error before the modal started observing it. The error was reported as an **uncaught zone error**, the modal's cleanup path (`_finish`) never ran, and the 20-second timeout became the only remaining exit.

```mermaid
sequenceDiagram
    participant User
    participant Picker as onPickerTap
    participant TokenX as token_exchange
    participant Modal

    User->>Picker: tap "Attach from Drive"
    Picker->>TokenX: create intentFuture (eager)
    TokenX--xPicker: CORS 403 (instant)
    Note over Picker: no handler attached yet → uncaught error
    Picker->>Modal: showDialog (next frame)
    Modal->>Modal: attach .catchError — too late
    Note over Modal: skeleton shown… readyTimeout fires after 20s
```

The audit found the same "error with no handler → wait for timeout" pattern in three additional mobile failure paths: WebView page-load failures were not observed, and rejected futures from `loadUrl` and `evaluateJavascript` escaped synchronous error handling.

## Solution

Every observable failure between the button tap and the interactive picker now reaches a single funnel that records the failing stage, exception, and stack trace in Sentry before closing the modal immediately.

```mermaid
flowchart LR
    A[intent loader] --> F
    B[loadIntent into platform view] --> F
    C[main-frame page load error] --> F
    D[ack / message handling] --> F
    E[ready timeout - last resort] --> F
    F[_failWith: Sentry log + close modal + error toast]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Drive picker reliability by lazily loading intents and standardizing modal failure handling across lifecycle/handshake stages.
  * Prevented modal hangs/crashes from failing cleanup/finish hooks and handled async acknowledgement failures.
  * Treats only main-frame web page-load errors as fatal while ignoring subframe noise; tightened “ready” timeout behavior.
* **New Features**
  * Page-load failures now include a specific reason for easier diagnosis.
* **Refactor**
  * Simplified attachment picker/context-menu widgets by consolidating mobile/web state handling.
* **Tests**
  * Expanded async failure-mode, lifecycle, and real window message transport coverage for the picker flow.

### DIFF
--- a/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
+++ b/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
@@ -202,13 +202,20 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
     _readyTimeoutTimer?.cancel();
     // Once closed, the timeout is cancelled and _finish won't run again, so
     // the pop must happen even if a subclass hook throws — otherwise the
-    // modal hangs with no fallback.
+    // modal hangs with no fallback. Each hook is guarded independently so a
+    // throwing onCleanup still runs onFinished, and neither escapes into the
+    // unawaited futures that call _finish as an unobserved async error.
     try {
       onCleanup();
-      onFinished(outcome);
-    } finally {
-      if (mounted) Navigator.of(context).pop(outcome);
+    } catch (e, s) {
+      logError('driveIntent: onCleanup failed', exception: e, stackTrace: s);
     }
+    try {
+      onFinished(outcome);
+    } catch (e, s) {
+      logError('driveIntent: onFinished failed', exception: e, stackTrace: s);
+    }
+    if (mounted) Navigator.of(context).pop(outcome);
   }
 
   /// Async so ack-delivery failures (e.g. evaluateJavascript) reject the
--- a/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
+++ b/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
@@ -9,6 +9,9 @@ import 'package:workplace/presentation/model/drive_pick_outcome.dart';
 
 enum _Phase { waiting, loadingPage, interactive, closed }
 
+/// Lazy so the modal starts the request only after its error handling is live.
+typedef DriveIntentLoader = Future<WorkplaceIntent> Function();
+
 /// Shared loading/messaging lifecycle for the mobile and web Drive picker modals.
 mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
   late String _intentId;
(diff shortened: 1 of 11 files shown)


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4737

hoangdat:
- [x] please ensure when modal close the listener should be dispose

hoangdat:
- should add some tests relate to the life cycle of message listener. We should take care some cases relate to listener and modal lifetime. Maybe we should keep it in part 2 ?

[drive_picker_web_message_listener_lifecycle_test.dart.txt](https://github.com/user-attachments/files/30488685/drive_picker_web_message_listener_lifecycle_test.dart.txt)

dab246:
> * [x]  please ensure when modal close the listener should be dispose

The modal now owns the listener throughout its lifecycle: `startWindowMessageListener` runs in `initState` before `loadIntent()` navigates the iframe, and `stopWindowMessageListener` runs in `onCleanup` when closing, guaranteeing disposal by construction. The composer-side indirection (`DrivePickerWebStateMixin`, `onRegisterExternalHandler`, `_webModalHandler`, `externalHandlerRegistrar`) was removed because it is now dead code and would cause duplicate event delivery, allowing the two entry-point widgets to share the same web/mobile flow.


> * should add some tests relate to the life cycle of message listener. We should take care some cases relate to listener and modal lifetime. Maybe we should keep it in part 2 ?
> 

Added

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4737

tddang-linagora (dismissed):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4737

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -142,12 +200,27 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
...
+      onCleanup();
+    } catch (e, s) {
+      logError('driveIntent: onCleanup failed', exception: e, stackTrace: s);
+    }
+    try {
+      onFinished(outcome);
+    } catch (e, s) {
+      logError('driveIntent: onFinished failed', exception: e, stackTrace: s);
+    }
     if (mounted) Navigator.of(context).pop(outcome);
```
hoangdat: how about 
```suggestion
    try {
      if (mounted) Navigator.of(context).pop(outcome);
    } catch (e, s) {
      logError('driveIntent: pop failed', exception: e, stackTrace: s);
    }
```

workplace/test/presentation/mixin/drive_picker_state_mixin_test.dart
hoangdat, line 1: add these tests 

```
      testWidgets('pick delivered even though the opener was disposed mid-flight', (tester) async {
        final state = await _pumpPicker(tester);
        final completer = Completer<DrivePickOutcome?>();
        state.modalStub = () => completer.future;

        final tap = state.onPickerTap();
        await tester.pumpWidget(const SizedBox.shrink());
        expect(state.mounted, isFalse, reason: 'sanity: opener really is gone');

        completer.complete(DrivePickOutcomePicked([_doc()]));
        await tap;

        expect(state.pickStates, hasLength(1));
        final result = state.pickStates.single as DrivePickResult;
        expect(result.documents.single.id, 'doc1');
      });

      // Pins the up-front capture of the toast text: localizations need a live
      // context, so reading them after the await would lose the message.
      testWidgets('failure after opener disposal keeps the localized message', (tester) async {
        final state = await _pumpPicker(tester);
        final completer = Completer<DrivePickOutcome?>();
        state.modalStub = () => completer.future;

        final tap = state.onPickerTap();
        await tester.pumpWidget(const SizedBox.shrink());

        completer.complete(DrivePickOutcomeFailed(StateError('intent failed')));
        await tap;

        expect(state.pickStates, hasLength(1));
        final failure = state.pickStates.single as DrivePickFailure;
        expect(failure.message, isNotNull);
      });
```

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
dab246: That makes sense, added.

workplace/test/presentation/mixin/drive_picker_state_mixin_test.dart
dab246: added

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4737

tddang-linagora (approved):

hoangdat (approved):

PR #4740, opened 2026-07-29
TF-4738 Fix response view while drive picker loading display prevent drive picker

### LINKED ISSUE #4738: Response view while drive picker loading display prevent drive picker
### Desc:
- during the loading screen display
- resize the browser from web -> mobile 
- app be hang in loading screen

### Expect
- [ ] loading screen should have also close button
- [ ] handle the responsive to still display file picker

https://github.com/user-attachments/assets/08f90ff6-5fc4-4026-837c-ae432fb2b7ac

dab246:
Closes #4738 

## Context

Resizing the browser across the desktop/mobile breakpoint while the Drive picker iframe was loading left it permanently stuck on the loading skeleton (only the 20s timeout eventually broke it). Related gaps: the desktop loading state had no close button, and once the picker was interactive a backdrop tap over the composer editor iframe was swallowed instead of dismissing the dialog.

## Root cause

Three independent issues, all rooted in web platform-view (iframe) behaviour:

1. **Unstable iframe ancestry → remount.** Crossing the breakpoint swapped a `SafeArea` in/out and restructured the `Column` (close button), changing the iframe's ancestor chain. Flutter then unmounted and recreated the `HtmlElementView`. The new iframe reported ready, but the loading lifecycle only applied the intent URL during its initial `waiting` phase, so the replacement stayed blank until the ready timeout fired.
2. **Listener owned by a transient widget.** The `window` message listener was registered by the Drive toolbar/context-menu widget (`DrivePickerWebStateMixin`). A responsive composer rebuild can dispose that widget while the modal is still open, orphaning the listener so `ready`/`readyToUse` never reach the picker.
3. **No pointer interceptor in the modal route.** The composer editor is itself a platform view, so backdrop taps over it were consumed by the browser instead of reaching Flutter. The loading close button also sat over the Drive iframe, which swallowed its clicks.

## Solution

- **Keep the iframe mounted.** Make the `SafeArea` unconditional (toggle `top` only), give the iframe's content slot a stable key, and keep a stable root `Stack` — the ancestry no longer changes across layouts, so the platform view is preserved.
- **Self-heal if it still remounts.** If the iframe is genuinely recreated *during loading*, reapply the resolved intent URL and restart the ready timeout. Never reapply after `readyToUse`, so an active picker session is never reloaded.
- **Own the listener in the modal.** Register the `window` message listener in the web modal's `initState` (before the iframe is created — no race) and remove it on close/dispose. Uses `addEventListener`, so it coexists with the editor's own listener. This removes `DrivePickerWebStateMixin` and the platform-specific entry-widget State subclasses.
- **Pointer interception.** A full-screen `PointerInterceptor` behind the dialog routes backdrop taps to Flutter while the Drive iframe stays directly interactive; the loading overlay is wrapped in a `PointerInterceptor` above the iframe so its close button is clickable.
- **Always-closable loading.** A 40×40 close button is shown during loading on both desktop and mobile.

## Demo

https://github.com/user-attachments/assets/db9ccff4-14ec-4ae9-a6e9-af68c625c6c6

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of workplace intent loading, including correct handling of platform-view recreation, ready-timeout restarts, acknowledgements, and safer modal cleanup.
  - Kept the embedded iframe mounted across responsive breakpoint changes, preventing unnecessary remounts.
  - Refined web modal pointer/tap handling and improved close-button placement/behavior, including during loading overlays.

- **Tests**
  - Expanded Chrome/web widget coverage for browser window messaging, modal shell overlays, iframe reuse on resize, and additional intent-loading/timeout scenarios.

### DIFF
--- a/workplace/test/presentation/view/drive_intent_web_view_modal_shell_test.dart
+++ b/workplace/test/presentation/view/drive_intent_web_view_modal_shell_test.dart
@@ -4,19 +4,6 @@ import 'package:flutter_test/flutter_test.dart';
 import 'package:pointer_interceptor/pointer_interceptor.dart';
 import 'package:workplace/presentation/view/drive_intent_web_view_modal_shell.dart';
 
-// Represents the iframe/platform view across resize.
-class _ProbeChild extends StatefulWidget {
-  const _ProbeChild({super.key});
-
-  @override
-  State<_ProbeChild> createState() => _ProbeChildState();
-}
-
-class _ProbeChildState extends State<_ProbeChild> {
-  @override
-  Widget build(BuildContext context) => const SizedBox.shrink();
-}
-
 const _closeIcon = 'assets/close.svg';
 
 Widget _shell({
@@ -37,32 +24,6 @@ Widget _shell({
   );
 }
 
-Widget _responsiveShell({required bool isWideScreen}) {
-  return MaterialApp(
-    home: Stack(
-      children: [
-        Positioned.fill(
-          child: PointerInterceptor(child: const SizedBox.expand()),
-        ),
-        SafeArea(
-          top: !isWideScreen,
-          left: false,
-          right: false,
-          bottom: false,
-          child: _modalShell(
-            haveCloseButton: !isWideScreen,
-            constraints: isWideScreen
-                ? const BoxConstraints(maxWidth: 800, maxHeight: 677)
-                : const BoxConstraints.expand(),
-            insetPadding:
-                isWideScreen ? const EdgeInsets.all(24) : EdgeInsets.zero,
-          ),
-        ),
-      ],
-    ),
-  );
-}
-
 Widget _modalShell({
   required bool haveCloseButton,
   required BoxConstraints constraints,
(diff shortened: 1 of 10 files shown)


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4740

tddang-linagora (dismissed):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4740

workplace/test/presentation/view/drive_intent_web_view_modal_shell_test.dart
```diff
@@ -0,0 +1,176 @@
...
+      haveCloseButton: haveCloseButton,
+      constraints: constraints,
+      insetPadding: insetPadding,
+      loadingWidget: loadingWidget,
+      onClose: onClose,
+    ),
+  );
+}
+
+Widget _responsiveShell({required bool isWideScreen}) {
```
hoangdat: it rebuilds the modal's tree by hand instead of pumping DriveIntentWebViewModal. So the test only proves the copy is correct — the real build() in ..._modal_web.dart is never exercised.

Can we pump the real modal instead? It's web-only, so it needs a --platform chrome test + a line in CHROME_TEST_PATHS.

dab246: Agree. The resize test currently mirrors the modal composition and only exercises the shell. I’ll keep shell-level tests there, but move the resize regression coverage to a Chrome-only test that pumps the real DriveIntentWebViewModal, then add it to CHROME_TEST_PATHS.

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4740

Issue #4739, opened 2026-07-29
Drive Picker in mobile


hoangdat:
### Desc 
- be blocked in login 
- so need to add `force_session_id = true` query parameter to `POST /intents`
- bypass mobile `origin check` in `onMessage` in handler mixin
- put value of `client` in response of `POST /intents` for the `sendAck()`s payload

### DoD
- drive picker in mobile can bypass login step and work

PR #4741, opened 2026-07-30
TF-4739 Enable mobile drive picker

tddang-linagora:
## Issue
- #4739 

## Resolved

[untitled.webm](https://github.com/user-attachments/assets/b0ac6ad3-49fe-4087-8d43-8ad07c745e0e)

## Summary by CodeRabbit

* **New Features**
  * Enhanced Drive intent handling by capturing an optional client identifier and exposing it as part of intent metadata.
  * Added session enforcement when creating workplace intents.

* **Bug Fixes**
  * Prevented unintended back navigation during the composer HTML insertion flow.
  * Improved web view acknowledgement by validating and correctly serializing the resolved client/origin value.
  * When no intent client is available, acknowledgement now fails with a clear error; mismatched or null-origin messages are dropped.

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1011,7 +1011,6 @@ class ComposerController extends BaseController
           if (PlatformInfo.isWeb) {
             richTextWebController?.editorController.insertHtml(html);
           } else {
-            popBack();
             await richTextMobileTabletController?.restoreMobileEditorFocus();
             await htmlEditorApi?.insertHtml(html);
             await SchedulerBinding.instance.endOfFrame;
--- a/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
+++ b/workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 import 'package:workplace/domain/entity/workplace_intent.dart';
 import 'package:workplace/domain/exceptions/workplace_exceptions.dart';
 import 'package:workplace/domain/message/workplace_intent_message.dart';
+import 'package:workplace/presentation/model/drive_origin_validator.dart';
 import 'package:workplace/presentation/model/drive_pick_outcome.dart';
 
 enum _Phase { waiting, loadingPage, interactive, closed }
@@ -16,6 +17,7 @@ typedef DriveIntentLoader = Future<WorkplaceIntent> Function();
 mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
   late String _intentId;
   String _intentOrigin = 'null';
+  String? _intentClient;
   Timer? _readyTimeoutTimer;
 
   _Phase _phase = _Phase.waiting;
@@ -35,6 +37,7 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
   }
 
   String get intentOrigin => _intentOrigin;
+  String? get intentClient => _intentClient;
 
   /// Covers silent failures like SSO blocking the page with no postMessage ever sent.
   Duration get readyTimeout => const Duration(seconds: 20);
@@ -101,6 +104,7 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
     _intentOrigin = intent.intentUrl.scheme == 'data'
         ? 'null'
         : intent.intentUrl.origin;
+    _intentClient = intent.client;
     _setPhase(_Phase.loadingPage);
     unawaited(_applyIntent(intent));
   }
@@ -165,13 +169,15 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
     }
   }
 
-  bool _isValidOrigin(String? origin) {
-    if (origin == null) return false;
-    if (origin == _intentOrigin) return true;
-    // On mobile the JS shim forwards targetOrigin as the origin arg. Data URIs
-    // have opaque 'null' origin but the shim receives '*' from postMessage.
-    return _intentOrigin == 'null' && origin == '*';
-  }
+  bool _isValidOrigin(String? origin) => originValidator.isValid(
+        origin,
+        intentOrigin: _intentOrigin,
+        intentClient: _intentClient,
+      );
+
+  /// Platform-specific origin matching rules (web vs mobile JS channel).
+  @protected
+  DriveOriginValidator get originValidator;
 
   Future<void> _handleWorkplaceMessage(WorkplaceIntentMessage msg) async {
     switch (msg) {
--- a/workplace/lib/presentation/model/drive_origin_validator.dart
+++ b/workplace/lib/presentation/model/drive_origin_validator.dart
@@ -1,0 +1,41 @@
+import 'package:flutter/foundation.dart';
+
+bool isValidDriveClient(String? client) =>
+    client != null && client.trim().isNotEmpty && client != '*';
+
+/// Validates a postMessage/JS-channel origin against the resolved intent.
+abstract class DriveOriginValidator {
+  const DriveOriginValidator();
+
+  /// Shared rule set; platform variants only override [matchesClient].
+  bool isValid(
+    String? origin, {
+    required String intentOrigin,
+    required String? intentClient,
+  }) {
+    if (origin == null) return false;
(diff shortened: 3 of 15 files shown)


workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
```diff
@@ -155,8 +158,8 @@ mixin DriveIntentMessageHandlerMixin<T extends StatefulWidget> on State<T> {
   bool _isValidOrigin(String? origin) {
     if (origin == null) return false;
     if (origin == _intentOrigin) return true;
-    // On mobile the JS shim forwards targetOrigin as the origin arg. Data URIs
-    // have opaque 'null' origin but the shim receives '*' from postMessage.
+    // Dev/fake intents (data: URI, no server-provided client) — the shim/
+    // postMessage sends '*' as targetOrigin in that case.
     return _intentOrigin == 'null' && origin == '*';
```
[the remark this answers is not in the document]
tddang-linagora: Test updated

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

workplace/lib/presentation/view/drive_intent_web_view_modal_mobile.dart
```diff
@@ -134,10 +134,11 @@ class _DriveIntentWebViewModalState extends State<DriveIntentWebViewModal>
     // so `event.data` in the page ends up a real object, not a JSON string —
     // Drive's getFilePickerConfig never calls JSON.parse on it.
     final payload = jsonEncode(widget.filePickerConfig.toJson());
+    final origin = jsonEncode(intentClient ?? intentOrigin);
     await _webViewController?.evaluateJavascript(source: '''
       window.dispatchEvent(new MessageEvent('message', {
         data: $payload,
-        origin: '$intentOrigin',
+        origin: $origin,
```
hoangdat: should prevent sending ack if no client in mobile case

tddang-linagora: Done

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
dab246: Validate mobile messages against intent client. The mobile shim forwards the page's `postMessage` targetOrigin, and cozy-interapp sends that target as `intent.attributes.client`, not the Drive service origin. This comparison therefore drops every valid mobile `ready` message when client and Drive are on different origins, so the picker never ACKs and times out. Accept `intentClient` for the native shim while retaining `intentOrigin` validation for web.

Real-world example:

```text
Drive service origin: https://drive.example.com
Intent client origin: https://tmail.example.com
```

The Drive service sends postMessage(..., "https://tmail.example.com"). The shim forwards this value to Flutter, where _isValidOrigin validates it against https://drive.example.com. Since the origins do not match, the message is discarded. As a result, the ready event is never received, no ACK is sent, and the picker eventually times out.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

workplace/lib/presentation/mixin/drive_intent_message_handler_mixin.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

hoangdat (dismissed):

workplace/lib/presentation/model/drive_origin_validator.dart
```diff
@@ -0,0 +1,38 @@
...
+class WebDriveOriginValidator extends DriveOriginValidator {
+  const WebDriveOriginValidator();
+}
+
+/// Mobile JS channel additionally allows the intent's declared client id.
+class MobileDriveOriginValidator extends DriveOriginValidator {
+  const MobileDriveOriginValidator();
+
+  @override
+  bool matchesClient(String origin, String? intentClient) =>
```
dab246, line 39: **Wildcard client weakens message validation on mobile.** A `client` value of `'*'` (or empty) is currently not rejected. As a result, the mobile client accepts messages with a wildcard origin and sends an ACK using the wildcard origin even when the intent origin is HTTPS. The validation should reject `null`, blank, or `'*'` before performing origin matching or sending an ACK, and unit tests should be added to cover these cases.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1011,7 +1011,6 @@ class ComposerController extends BaseController
           if (PlatformInfo.isWeb) {
             richTextWebController?.editorController.insertHtml(html);
           } else {
-            popBack();
```
dab246, line 1014: **Missing regression test for the mobile composer flow.**. Removing `popBack()` is the correct behavioral change, but there is no regression test verifying that selecting a file from Drive on mobile keeps the composer open, restores focus, and successfully inserts the HTML content. A widget or integration test should be added to cover this end-to-end flow.

workplace/lib/presentation/model/drive_origin_validator.dart
dab246: ## Summary

This is a hardening concern at the backend-to-WebView boundary. It is not evidence that the current backend returns `"*"` for `client`.

On mobile, Flutter does not receive a browser-trusted `event.origin`. The WebView shim forwards the `targetOrigin` supplied by the Drive page as `origin`, so that value must be validated carefully.

## Current Behaviour

Mobile messages are accepted when their origin matches the intent client:

```dart
origin == intentClient
```

If the backend returns `client: "*"`, the Drive page can send `postMessage(..., "*")`. The shim forwards `"*"` to Flutter, and the mobile validator accepts it because `origin == intentClient`. The ACK is then dispatched with `origin: "*"` as well.

This bypasses the intended rule that wildcard origins are allowed only for fake/development `data:` intents, never for production HTTPS intents.

## Recommendation

Reject a missing, blank, or wildcard client before validating a mobile message or sending the ACK:

```dart
bool isValidClient(String? client) =>
    client != null && client.isNotEmpty && client != '*';
```

Use this check in both `MobileDriveOriginValidator.matchesClient` and before `sendAck()`. If the backend contract guarantees that `client` is an origin URL, validate it more strictly as a valid HTTPS `Uri.origin`.

## Proposed Fix

Apply the validation in two layers:

1. Do not allow a wildcard or blank client to match an incoming mobile message.

   ```dart
   class MobileDriveOriginValidator extends DriveOriginValidator {
     const MobileDriveOriginValidator();

     bool _isValidClient(String? client) =>
         client != null && client.trim().isNotEmpty && client != '*';

     @override
     bool matchesClient(String origin, String? intentClient) =>
         _isValidClient(intentClient) && origin == intentClient;
   }
   ```

2. Do not dispatch an ACK when the client is missing, blank, or a wildcard.

   ```dart
   @override
   Future<void> sendAck() async {
     final client = intentClient;
     if (client == null || client.trim().isEmpty || client == '*') {
       throw WorkplaceNoIntentClientException();
     }

     final payload = jsonEncode(widget.filePickerConfig.toJson());
     await _webViewController?.evaluateJavascript(source: '''
       window.dispatchEvent(new MessageEvent('message', {
         data: $payload,
         origin: ${jsonEncode(client)},
         source: window
       }));
     ''');
   }
   ```

This is a small, low-compatibility-risk fix. If the API contract guarantees that `client` is an origin URL, harden it further by accepting only a valid HTTPS origin.

## Tests to Add

- An HTTPS intent with `client: "*"` must reject an incoming `"*"` message.
- A blank or whitespace-only client must be rejected.
- A valid client must still accept a `ready` message and send the ACK.

## Risk Assessment

If the backend guarantees that `client` is never blank or `"*"`, the current implementation does not have a runtime failure. The validation is still recommended so that an invalid or compromised backend response cannot weaken the origin-validation contract.

tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

dab246 (dismissed):
LGTM

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4741

Issue #4593, opened 2026-06-06
[Sentry] Mobile - Refresh token failed in some cases


hoangdat:
### Desc

Some log show the refresh token failed. And I found in some cases we have status code 400 in refresh token.

1. https://errors.cozycloud.cc/organizations/cozycloud/issues/328751/?environment=tmail-linagora&project=83&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=23

```
PlatformException
PlatformException(token_failed, Failed to get token: [error: null, description: Network error], Unable to resolve host "sso.linagora.com": No address associated with hostname, null)
```

2. https://errors.cozycloud.cc/organizations/cozycloud/issues/306580/?environment=tmail-linagora&project=83&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d&stream_index=1

```
MailboxDashBoardController::_performReconnection: forcing logout (urgent auth failure, no composer to save)
```

### DoD
- [ ] understand what is the problem
- [ ] fix and increase the capability of handling refreshToken failed

dab246:
We should split this into two issues for processing. Log 1 is for the web version, and logs 2-3 are for the mobile version.

PR #4745, opened 2026-08-08
Fix mobile session stuck forever when the refresh token is rejected

h7brasil:
## Problem

When the OIDC provider rejects the refresh token — revoked or expired, for example after the account password is changed — the token endpoint answers `400 invalid_grant`. On mobile the app never recovers from this:

- the mailbox spins forever,
- a generic *"Unknown error occurred, please try again"* toast appears,
- **Retry can never help**, and
- only clearing the app data unblocks it.

Observed against a self-hosted OIDC provider (Stalwart), reproduced end to end: `/jmap/` returns 401, the app calls `/auth/token` about ten times in eight seconds, and stays in that loop indefinitely without ever returning the user to the login screen.

## Root cause

The logout path already exists, but it is unreachable on mobile.

`_refreshTokenThenRetry` handles the rejection inside `on DioException catch`, gated on `refreshError.response?.statusCode == 400`. On mobile the refresh does **not** go through Dio — it goes through `_appAuth.token()` (flutter_appauth, native, `authentication_client_mobile.dart`). So:

1. flutter_appauth throws `FlutterAppAuthPlatformException`
2. `handleException` maps it to `OAuthAuthorizationError`
3. that type matches neither the `DioException` catch nor the `PlatformException` catch
4. it falls through to the outer catch in `onError` and reaches `_handleRefreshErrorOnMobile`
5. which only rewraps and propagates — **no `clear()`**

The dead session is kept and every subsequent request retries forever. In effect, the recovery behaviour described in ADR-0035 never executes on this platform.

## Fix

Classify `OAuthAuthorizationError` using the same RFC 6749 §5.2 codes the web path already relies on. A confirmed rejection calls `clear()` and rejects with `RefreshTokenFailedException`, which takes the user back to login.

Two deliberate constraints:

- **Network tolerance from #4593 is preserved.** Only confirmed bad-grant codes log out. `server_error` and `temporarily_unavailable` still keep the session, so a flaky mobile connection does not sign the user out. That issue covered the *opposite* case — `token_failed` with `Network error` — and this change does not regress it.
- **Web behaviour is unchanged.** `_handleRefreshErrorOnWeb` only delegates to the mobile handler after it has already concluded the error is not a rejection, so the new check is always false there.

The `token_failed` wrapper is honoured the same way as on web: when the plugin reports its own code, the real RFC 6749 code is read from the description.

## Tests

8 new cases, all in `test/features/interceptor/authorization_interceptor_test.dart`:

- `invalid_grant`, `invalid_client`, `unauthorized_client`, `invalid_scope` → forced logout, OIDC state cleared
- `token_failed` with `invalid_grant` in the description → forced logout
- `ServerError`, `TemporarilyUnavailable`, and `token_failed` with a non-RFC description → **session kept**

One existing test asserted the old behaviour (`invalid_grant` propagating without logout) and was updated to the corrected expectation.

```
flutter test test/features/interceptor/authorization_interceptor_test.dart \
             test/features/login/data/network/authentication_client/web_refresh_token_error_classifier_test.dart
→ 120 tests passed
flutter analyze on both changed files → no issues
```

## Scope

Three files, +155 / −2. No dependency, API, or UI changes.

This affects any deployment using its own OIDC provider, not just the one where it was found.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of authentication refresh failures across web and mobile platforms.
  * Confirmed authorization rejections now securely sign users out and clear authentication state.
  * Temporary network or server issues no longer trigger unnecessary sign-outs, allowing existing sessions to continue.
  * Improved recognition of common OAuth authorization errors, including invalid or expired refresh credentials.
  * Reduced duplicate error reporting for classified authentication failures.

### DIFF
--- a/lib/features/login/data/network/authentication_client/web_refresh_token_error_classifier.dart
+++ b/lib/features/login/data/network/authentication_client/web_refresh_token_error_classifier.dart
@@ -1,75 +1,47 @@
-import 'package:dio/dio.dart';
-import 'package:tmail_ui_user/features/login/domain/exceptions/authentication_exception.dart'
-    show AccessTokenInvalidException;
+import 'package:tmail_ui_user/features/login/data/network/authentication_client/refresh_token_error_classifier.dart';
 
-/// Classifies web token-refresh errors according to RFC 6749 §5.2.
-/// https://datatracker.ietf.org/doc/html/rfc6749#section-5.2
+/// Classifies the refresh failures thrown by flutter_appauth_web.
 ///
-/// Single responsibility: answer two questions —
-///   1. Is this error a definitive server rejection (400/401 equivalent)?
-///   2. What structured extras should go to Sentry for tracing?
-///
-/// Intentionally has NO dependency on Dio handler operations or interceptor
-/// state; callers act on the result (logout, keep session, log).
-class WebRefreshTokenErrorClassifier {
-  /// Standard RFC 6749 §5.2 token-endpoint error codes that map to a
-  /// definitive 400/401 rejection — logout is safe.
-  /// Non-standard codes (e.g. "token_failed") are absent intentionally;
-  /// they are logged to Sentry until their HTTP status is confirmed.
-  static const _badGrantCodes = {
-    'invalid_grant',          // refresh token expired or revoked (400)
-    'invalid_client',         // client authentication failed (400 or 401)
-    'invalid_request',        // request was malformed (400)
-    'invalid_scope',          // requested scope exceeds original grant (400)
-    'unauthorized_client',    // client not authorized for this grant type (400)
-    'unsupported_grant_type', // grant type not supported by the server (400)
-  };
+/// The web plugin reports every non-200 token response as an [ArgumentError]
+/// whose message carries the OAuth2 fields:
+/// `Failed to get token: [error: X, description: Y]`.
+class WebRefreshTokenErrorClassifier extends RefreshTokenErrorClassifier {
+  @override
+  String get rejectedAuthErrorType => 'web_server_rejected_refresh';
 
-  /// Returns `true` when [error] represents a confirmed server-side rejection
-  /// that maps to HTTP 400 or 401, making logout the correct response.
-  bool isServerRejection(Object error) {
-    if (error is AccessTokenInvalidException) return true;
-    if (error is ArgumentError) return _isArgumentErrorRejection(error);
-    if (error is DioException) return _isDioRejection(error);
-    return false;
-  }
+  @override
+  String get unclassifiedAuthErrorType => 'web_token_unknown_error';
 
-  bool _isArgumentErrorRejection(ArgumentError error) {
+  @override
+  bool isPlatformRejection(Object error) {
+    if (error is! ArgumentError) return false;
     final code = _parseOAuthCode(error);
-    if (code != null && _badGrantCodes.contains(code)) return true;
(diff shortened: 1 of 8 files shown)


test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -2963,6 +2971,109 @@ void main() {
...
+  group('onError: mobile refresh rejected by token endpoint', () {
+    for (final rejection in const <OAuthAuthorizationError>[
+      OAuthAuthorizationError(
+        error: 'invalid_grant',
+        errorDescription: 'The refresh token has been revoked',
+      ),
+      OAuthAuthorizationError(error: 'invalid_client'),
+      OAuthAuthorizationError(error: 'unauthorized_client'),
+      OAuthAuthorizationError(error: 'invalid_scope'),
+    ]) {
```
dab246, line 3000: Two OAuth rejection codes are missing from mobile tests. `WebRefreshTokenErrorClassifier` considers six RFC 6749 codes to be server rejections, but the new mobile test matrix directly covers only four:

- Covered: `invalid_grant`, `invalid_client`, `unauthorized_client`,
  `invalid_scope`.
- Missing: `invalid_request`, `unsupported_grant_type`.
- `token_failed` only covers the `invalid_grant` description.

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -2167,7 +2167,8 @@ void main() {
 
       test(
         'WHEN refresh throws OAuthAuthorizationError (e.g. invalid_grant)\n'
-        'THEN propagated DioException still has NO HTTP response',
+        'THEN the session is torn down and RefreshTokenFailedException is raised\n'
+        'AND the propagated DioException still has NO HTTP response',
```
dab246, line 2179: The native AppAuth → OAuthAuthorizationError boundary is untested. The interceptor test mocks `OAuthAuthorizationError` directly. It therefore does not prove the production mapping from `FlutterAppAuthPlatformException.platformErrorDetails.error` to that error the boundary on which this fix relies.

lib/features/login/data/network/authentication_client/web_refresh_token_error_classifier.dart
dab246: The classifier name no longer reflects its scope. `WebRefreshTokenErrorClassifier` is now also a mobile dependency. `RefreshTokenErrorClassifier` would better communicate its cross-platform
responsibility. This is a maintainability improvement only, it does not affect behavior.

dab246 (commented):
Hi @h7brasil, thanks for creating the PR to address this issue. I’ve done a quick review and left some comments, please take a look and fix them.

test/features/interceptor/authorization_interceptor_test.dart
hoangdat: added

test/features/interceptor/authorization_interceptor_test.dart
hoangdat: added

lib/features/login/data/network/authentication_client/web_refresh_token_error_classifier.dart
hoangdat: fixed

dab246 (approved):

tddang-linagora (approved):

PR #4731, opened 2026-07-27
TF-4727 [Part-1] Drive-transfer sealed types, stager interfaces, buffered stager

tddang-linagora:
## Issue
- #4727 

## Summary
Foundation for TF-4727 (drive-file-as-attachment download/upload): sealed `StagedDriveFile` result types, `DriveFileStager`/`DriveTransferStrategy` interfaces, and the buffered-web strategy as the first concrete implementation proving the interfaces out.

Isolated to `workplace/`, additive only, no main-app callers yet.

## Summary by CodeRabbit

- **New Features**
  - Added browser-based Drive downloads with in-memory staging and upload support.
  - Added download and upload progress reporting, cancellation, and a 60-second timeout.
  - Preserved downloaded file names, sizes, and MIME types.
  - Added platform-appropriate temporary storage options and automatic cleanup.
  - Added secure download-link validation and clear errors for missing, insecure, or empty downloads.

- **Bug Fixes**
  - Improved handling of slow, cancelled, failed, and interrupted Drive transfers.

Issue #4727, opened 2026-07-27
Add platform drive-file download stagers in `workplace`


tddang-linagora:
### Background

Per [ADR-0103](../../docs/adr/0103-attach-drive-file-as-attachment.md#decision), drive documents that carry a `downloadLink` (a public, short-lived URL, ~5–10 min TTL, no JMAP auth) must eventually be downloaded and attached as real email attachments. Because the client runs on memory-constrained mobile devices as well as web, downloading must keep memory flat wherever the platform allows it: always on IO, and on web whenever the Origin Private File System (OPFS) is available.

This issue is the **isolated, platform-facing piece**: given a drive document's `downloadLink`, stream it into platform-appropriate temporary storage and hand back a result that a (separate, not-yet-existing) upload step can consume. This work lives entirely in the `workplace` package/module and has no main-app runtime effect on its own — it is a new, self-contained capability, not yet wired to anything.

### Scope

Define a sealed result type bridging "downloaded" → "ready to upload":

```dart
sealed class StagedDriveFile {
  String get fileName;
  int get fileSize;
  String? get mimeType;
  Future<void> dispose();
}

class FileBackedStagedFile extends StagedDriveFile { final String filePath; } // IO temp file
class OpfsStagedFile extends StagedDriveFile { final JSObject fileHandle; }    // web OPFS
class BytesStagedFile extends StagedDriveFile { final Uint8List bytes; }       // web buffered fallback
```

And the staging interface:

```dart
abstract class DriveFileStager {
  Future<StagedDriveFile> stage({
    required DriveDocument doc,
    required void Function(int received, int total) onDownloadProgress,
    required CancelToken cancelToken,
  });
}
```

Implement one stager per platform capability:

1. **IO (mobile/desktop):** stream the public `downloadLink` via a plain `Dio()` instance carrying **no interceptors** and no auth headers — this is critical, since the shared app `Dio` client attaches a Bearer/Basic token to every request and the `downloadLink` targets an external, untrusted Drive host, not the JMAP host. Use `responseType: ResponseType.stream` with a 60s `receiveTimeout` (tolerates slow-to-first-byte export-on-demand backends, still trips on a genuinely stalled connection) and write each chunk to a temp file in `getTemporaryDirectory()` via `RandomAccessFile.writeFrom`, supporting pause/resume/cancel. Yields `FileBackedStagedFile`.
[...]

PR #4746, opened 2026-08-10
TF-4727 [Part-2] Add web OPFS drive file stager and strategy selection

tddang-linagora:
## Issue
- #4727

## Summary
TF-4727: streams a drive document via `fetch()`/`ReadableStream` into an Origin Private File System temp file, uploaded via raw XHR (the only upload path `FileUploader` can't do). `drive_transfer_strategy_factory_web.dart` picks OPFS vs. the buffered strategy based on cached feature detection.

## Dependencies
- #4731

## Summary by CodeRabbit

* **New Features**
  * Added browser-based OPFS support for large Drive file transfers.
  * Streams downloads to temporary files and uploads them with progress and cancellation support.
  * Automatically selects the best transfer method based on browser capabilities.
  * Cleans up stale temporary files and preserves transfer metadata.

* **Bug Fixes**
  * Improved handling of cancellations, network failures, invalid responses, incomplete downloads, and interrupted transfers.

* **Tests**
  * Added comprehensive Chrome coverage for transfers, cleanup, capability detection, cancellation, and failure scenarios.


workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,155 @@
...
+    var received = 0;
+    while (true) {
+      if (cancelToken.isCancelled) {
+        throw StateError('OpfsDriveFileStager: transfer cancelled');
+      }
+      final chunk = await _bindings.readChunk(reader);
+      if (chunk == null) break;
+      await _bindings.writeChunk(writable, chunk);
+      received += chunk.length;
+      onDownloadProgress(received, fetchStream.contentLength);
```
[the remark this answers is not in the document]
tddang-linagora: Done

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
```diff
@@ -0,0 +1,170 @@
...
+        doc: doc,
+        onDownloadProgress: (_, __) {},
+        cancelToken: CancelToken(),
+      ),
+      throwsA(isA<StateError>()),
+    );
+
+    // The temp entry was created before the failure, so cleanup must remove it.
+    expect(bindings.removeTempFileCount, 1);
+  });
```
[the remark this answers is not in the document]
tddang-linagora: Done

hoangdat:
@coderabbit please review

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
```diff
@@ -0,0 +1,189 @@
...
+
+    await expectLater(
+      OpfsDriveFileStager(isReleaseMode: true).stage(
+        doc: doc,
+        onDownloadProgress: (_, __) {},
+        cancelToken: CancelToken(),
+      ),
+      throwsA(isA<DriveDownloadInsecureLinkException>()),
+    );
+  });
```
hoangdat, line 261: please add these tests 

```dart 
  test('fails the transfer when cancelled while a chunk read is in flight',
      () async {
    final bindings = _ControllableStreamBindings(contentLength: 16);
    final cancelToken = CancelToken();

    final doc = DriveDocument(
      id: 'doc-opfs-7',
      name: 'cancelled-mid-read.txt',
      size: 16,
      mimeType: 'text/plain',
      downloadLink: Uri.parse('https://drive.example/cancelled-mid-read.txt'),
    );

    final staging = OpfsDriveFileStager(bindings: bindings).stage(
      doc: doc,
      onDownloadProgress: (_, __) {},
      cancelToken: cancelToken,
    );

    // Cancel only once the first chunk is on disk and the *second* read is
    // pending, so this exercises cancellation arriving mid-read rather than
    // between iterations.
    await bindings.firstChunkWritten;
    cancelToken.cancel();

    await expectLater(staging, throwsA(isA<DioException>()));
  });

  test('fails the transfer when the body ends short of content-length',
      () async {
    final bindings =
        _ControllableStreamBindings(contentLength: 16, closeAfterFirstChunk: true);

    final doc = DriveDocument(
      id: 'doc-opfs-8',
      name: 'truncated.txt',
      size: 16,
      mimeType: 'text/plain',
      downloadLink: Uri.parse('https://drive.example/truncated.txt'),
    );

    await expectLater(
      OpfsDriveFileStager(bindings: bindings).stage(
        doc: doc,
        onDownloadProgress: (_, __) {},
        cancelToken: CancelToken(),
      ),
      throwsA(anything),
    );
  });
}

/// Serves the body from a real `ReadableStream` that emits one 3-byte chunk
/// and then either stalls or closes. Only the *source* of the bytes is
/// substituted: `read`, `cancel`, and every OPFS write below them run
/// against the browser, so the streams semantics under test are Chrome's,
/// not the fake's.
class _ControllableStreamBindings extends OpfsJsBindings {
  _ControllableStreamBindings({
    required this.contentLength,
    this.closeAfterFirstChunk = false,
  });

  final int contentLength;
  final bool closeAfterFirstChunk;

  final _firstChunkWritten = Completer<void>();

  /// Resolves once the first chunk has been written to the OPFS temp file,
  /// i.e. once the loop is parked inside its second `read()`.
  Future<void> get firstChunkWritten => _firstChunkWritten.future;

  @override
  Future<OpfsFetchStream> fetchStream(Uri url, {Future<void>? cancelSignal}) async {
    final source = JSObject();
    source.setProperty(
      'start'.toJS,
      ((web.ReadableStreamDefaultController controller) {
        controller.enqueue(Uint8List.fromList([1, 2, 3]).toJS);
        if (closeAfterFirstChunk) controller.close();
      }).toJS,
    );
    final stream = web.ReadableStream(source);
    return OpfsFetchStream(
      reader: stream.getReader() as web.ReadableStreamDefaultReader,
      contentLength: contentLength,
    );
  }

  @override
  Future<void> writeChunk(
      web.FileSystemWritableFileStream stream, Uint8List chunk) async {
    await super.writeChunk(stream, chunk);
    if (!_firstChunkWritten.isCompleted) _firstChunkWritten.complete();
  }
```

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
```diff
@@ -0,0 +1,237 @@
...
+
+  Future<web.FileSystemDirectoryHandle> _opfsRoot() =>
+      web.window.navigator.storage.getDirectory().toDart;
+
+  /// True when `navigator.storage.getDirectory` is present. `createWritable`
+  /// ships together with it (Baseline across Chrome/Edge, Firefox, and
+  /// Safari as of Safari 26.0), so a single, side-effect-free property read
+  /// is enough — no probe file, no `await`, so callers can cache this once
+  /// per session synchronously.
+  bool isOpfsSupported() {
```
hoangdat: this checks only whether `navigator.storage.getDirectory` exists. However, our OPFS implementation actually requires `FileSystemFileHandle.createWritable()`.

This matters especially for Safari: `getDirectory()` was available before `createWritable()`, so older Safari versions can return true from our current check but then fail when openWritable() is called. That also means the buffered fallback is never selected.

hoangdat: how about to use 

```dart
/// The global `FileSystemFileHandle` constructor, nullable because it is
/// absent wherever OPFS is: reading an undefined property off `globalThis`
/// yields null rather than throwing.
@JS('FileSystemFileHandle')
external _FileSystemFileHandleCtorProbe? get _fileSystemFileHandleCtor;

extension type _FileSystemFileHandleCtorProbe(JSObject _) implements JSObject {
  external _FileSystemFileHandleProtoProbe? get prototype;
}

/// Reads `createWritable` off the prototype without invoking it, so probing
/// for it creates no writable stream and no temp entry.
extension type _FileSystemFileHandleProtoProbe(JSObject _) implements JSObject {
  external JSAny? get createWritable;
}

// .......

  bool isOpfsSupported() {
    final storage =
        (web.window.navigator as _NavigatorStorageProbe).storage;
    if (storage == null) return false;
    if ((storage as _StorageManagerFeatureProbe).getDirectory == null) {
      return false;
    }
    return _fileSystemFileHandleCtor?.prototype?.createWritable != null;
  }
```

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
hoangdat, line 1: add this test also

```dart
@TestOn('chrome')
library;

import 'dart:js_interop';
import 'dart:js_interop_unsafe';

import 'package:flutter_test/flutter_test.dart';
import 'package:workplace/data/datasource/drive_transfer/opfs_js_bindings.dart';

/// `FileSystemFileHandle.prototype`, the object `isOpfsSupported` probes.
JSObject get _fileHandlePrototype =>
    (globalContext['FileSystemFileHandle'] as JSObject)['prototype'] as JSObject;

void main() {
  group('OpfsJsBindings.isOpfsSupported', () {
    test('is true on a browser exposing getDirectory and createWritable', () {
      expect(OpfsJsBindings().isOpfsSupported(), isTrue);
    });

    test('is false when createWritable is missing, as on Safari before 26', () {
      // Chrome ships both halves, so the only way to exercise the
      // createWritable branch is to take it away: hide the prototype method
      // for the duration of this test, then put it back.
      final prototype = _fileHandlePrototype;
      final original = prototype['createWritable'];
      addTearDown(() => prototype['createWritable'] = original);
      prototype['createWritable'] = null;

      expect(OpfsJsBindings().isOpfsSupported(), isFalse);
    });
  });
}

```

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
```diff
@@ -0,0 +1,237 @@
...
+    final xhr = web.XMLHttpRequest();
+    xhr.open('POST', request.uploadUri.toString());
+    xhr.setRequestHeader('Authorization', request.authHeader);
+    // An OPFS-created `File` carries an empty type, so without this the
+    // request would go out with no usable content type.
+    final mimeType = request.mimeType;
+    if (mimeType != null && mimeType.isNotEmpty) {
+      xhr.setRequestHeader('Content-Type', mimeType);
+    }
+    xhr.timeout = driveTransferReceiveTimeout.inMilliseconds;
```
hoangdat: Why we need the timeout here? how about to remove it?

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,54 @@
...
+
+    final upload = _bindings.uploadFile(XhrUploadFileRequest(
+      file: file,
+      uploadUri: request.uploadUri,
+      authHeader: request.authHeader,
+      mimeType: request.mimeType,
+      onUploadProgress: request.onUploadProgress,
+    ));
+
+    final cancelSubscription =
```
hoangdat: Same user action, two exception types: cancel before the XHR exists → DioException(cancel) (line 21), cancel after → abort() → StateError('OPFS upload cancelled'). Download leg has a third case — cancel mid-read returns a truncated file with no error at all.

So the orchestrator can't tell "user pressed ✕" from "upload failed" without string-matching messages → error toasts on deliberate cancels, and Sentry noise.

Buffered strategy already throws DioException throughout. Suggest normalizing here too: cancelToken.cancelError for cancel, DioException(badResponse) for HTTP errors. Note tests at opfs_drive_file_uploader_test.dart:72 / :99 currently pin the inconsistency.

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
hoangdat, line 1: add this test too 
```dart
  test('leaks one OPFS entry per transfer when dispose() is never called',
      () async {
    const content = 'orphan me';
    final bindings = _NameRecordingBindings();
    addTearDown(() async {
      for (final name in bindings.createdNames) {
        try {
          await bindings.removeTempFile(name);
        } catch (_) {
          // Already gone; nothing to undo.
        }
      }
    });

    final doc = DriveDocument(
      id: 'doc-opfs-orphan',
      name: 'orphan.txt',
      size: content.length,
      mimeType: 'text/plain',
      downloadLink: Uri.dataFromString(content, mimeType: 'text/plain'),
    );

    // Three successful transfers of the same document. dispose() is never
    // called on any of them — the shape of a tab closed or crashed after
    // staging, and of any orchestrator that forgets the try/finally.
    for (var i = 0; i < 3; i++) {
      await OpfsDriveFileStager(bindings: bindings).stage(
        doc: doc,
        onDownloadProgress: (_, __) {},
        cancelToken: CancelToken(),
      );
    }

    // Each transfer took a distinct entry name, so they stack rather than
    // overwrite one another.
    expect(bindings.createdNames.toSet(), hasLength(3));

    // Resolved from a fresh OPFS root, holding no reference to anything the
    // transfers returned: the bytes are still on disk, and nothing in the
    // package will ever remove them.
    for (final name in bindings.createdNames) {
      expect(await _opfsEntryExists(name), isTrue,
          reason: '$name should still be in OPFS');
    }
  });

Future<bool> _opfsEntryExists(String name) async {
  final root = await web.window.navigator.storage.getDirectory().toDart;
  try {
    await root.getFileHandle(name).toDart;
    return true;
  } catch (_) {
    return false;
  }
}

/// Captures the generated temp-entry names, which are otherwise private to
/// the stager, so a test can look them up after the fact.
class _NameRecordingBindings extends OpfsJsBindings {
  final createdNames = <String>[];

  @override
  Future<web.FileSystemFileHandle> createTempFile(String fileName) {
    createdNames.add(fileName);
    return super.createTempFile(fileName);
  }
}
```

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,222 @@
...
+        fetchStream: fetchStream,
+        writable: writable,
+        cancelToken: cancelToken,
+        onDownloadProgress: onDownloadProgress,
+      );
+      await _bindings.closeWritable(writable);
+      completed = true;
+      return OpfsStagedFile(
+        fileHandle: handle,
+        removeEntry: (_) => _bindings.removeTempFile(tempFileName),
```
hoangdat: Temp entries are created in the OPFS root and removed in exactly two places: dispose() and the failure-cleanup path. Neither runs if the tab is closed or crashes mid-transfer, or if the orchestrator (not written yet) forgets a try/finally. OPFS is persistent storage — the bytes stay against the origin quota across restarts, and nothing sweeps them.

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
hoangdat, line 1: how about split this file into 3 files: feature detection, OPFS file ops, fetch streaming, and XHR upload

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_js_bindings.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,256 @@
...
+  Future<int> _streamToFile(_StreamToFileRequest request) async {
+    final received = await _readAndWriteAll(request);
+    _verifyComplete(request, received);
+    return received;
+  }
+
+  Future<int> _readAndWriteAll(_StreamToFileRequest request) async {
+    final reader = request.fetchStream.reader;
+    var received = 0;
+    while (true) {
```
dab246, line 168: Why do we need? It always comes in here.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,256 @@
...
+      if (request.cancelToken.isCancelled) {
+        throw request.cancelToken.cancelError!;
```
dab246: What about if `request.cancelToken.cancelError == null`?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,256 @@
...
+    }
+    return received;
+  }
+
+  /// `cancelReader` resolves an in-flight `read()` with done:true rather than
+  /// rejecting, so a mid-read cancellation exits the loop normally — without
+  /// the re-check here the truncated file would pass as a successful staging.
+  void _verifyComplete(_StreamToFileRequest request, int received) {
+    if (request.cancelToken.isCancelled) {
+      throw request.cancelToken.cancelError!;
```
dab246: idem

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,256 @@
...
+  /// `<opfsTempFilePrefix><micros>_<random>_<id>_<name>`: the timestamp ages
+  /// the entry for `sweepStaleTempFiles`, the random field keeps simultaneous
+  /// transfers of one [DriveDocument] off each other's entry.
+  ///
+  /// Separators are stripped because `getFileHandle` rejects them; always
+  /// prefixed, so the name can never come out as `.` or `..`.
+  static String _tempFileName(DriveDocument doc) {
+    final unique =
+        '${DateTime.now().microsecondsSinceEpoch}_${_random.nextInt(1000000)}';
+    final tail = '${doc.id}_${doc.name}'.replaceAll(RegExp(r'[/\\]'), '_');
```
dab246: should create const for Regex `RegExp(r'[/\\]')`

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  Future<int> _readAndWriteAll(_StreamToFileRequest request) async {
+    final reader = request.fetchStream.reader;
+    var received = 0;
+    while (true) {
+      if (request.cancelToken.isCancelled) {
+        throw request.cancelToken.cancelError!;
+      }
+      final chunk = await _bindings.readChunk(reader);
```
dab246: A response that stalls after headers can wait forever. The 60-second timer is cancelled as soon as `fetch()` resolves with headers. Every later `await reader.read()` has no timeout. If a server sends headers and then stops delivering body bytes, the staging loop stays pending indefinitely unless the user manually cancels. This conflicts with the intended receive-timeout behavior of the Dio-backed paths and can leave a composer chip permanently in the downloading state. It is especially important for export-on-demand Drive links and unreliable mobile networks.

.github/workflows/analyze-test.yaml
```diff
@@ -15,6 +15,9 @@ env:
   # "core"). Add more lines here as needed.
   CHROME_TEST_PATHS: |
     core/test/data/network/download/download_manager_test.dart
+    workplace/test/data/datasource/drive_transfer/drive_transfer_strategy_factory_web_test.dart
+    workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
+    workplace/test/data/datasource/drive_transfer/opfs_drive_file_uploader_test.dart
```
dab246, line 20: Missing `opfs_js_bindings_test.dart`

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,94 @@
...
+          .fetch(url.toString().toJS, web.RequestInit(signal: controller.signal))
+          .toDart;
+    } catch (e) {
+      if (controller.signal.aborted) {
+        throw DioException.connectionError(
+          requestOptions: requestOptions,
+          reason: 'the download request was aborted',
+        );
+      }
+      rethrow;
```
dab246: `fetchStream()` promises that callers can branch on `DioExceptionType`, but it does not provide that contract:

- Cancelling before response headers aborts the controller and enters the `controller.signal.aborted` branch, which throws `DioException.connectionError` instead of `DioExceptionType.cancel`.
- A CORS, DNS, TLS, or browser network failure reaches `rethrow`, commonly as a browser `TypeError`/DOM exception rather than any `DioException`.

The buffered strategy propagates Dio cancellation in the expected form, while the OPFS strategy gives the same user action a different error shape. A caller that removes a chip or suppresses an error only for `DioExceptionType.cancel` will incorrectly render a user cancellation as a transfer failure.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

.github/workflows/analyze-test.yaml
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

hoangdat:
Keep `OpfsDriveFileStager.stage()`'s `catch -> _cleanupAfterFailure() -> rethrow` (`opfs_drive_file_stager.dart:78-89`) through that rewrite. It is the only thing that removes a half-written OPFS entry when the download fails or is cancelled: `transfer()` disposes only a staged file that `stage()` **returned**, so anything written before a throw has no handle anywhere and would sit in the user's origin quota across reloads.

tddang-linagora:
Kept — `stage()` still does `catch -> cleanupAfterFailure() -> rethrow` (`opfs_drive_file_stager.dart:86-89`). The rewrite only moved the three acquired resources into `_OpfsStagingScope`, which owns the reader-cancel / writable-abort / `removeTempFile` sequence; the invariant you describe is unchanged.

While re-checking it against your reasoning, one hole was left: the cleanup was fire-and-forget at every step, so a failed `abortWritable` was followed by a `removeEntry` that *cannot* succeed — a still-open writable locks the entry, and Chrome throws `NoModificationAllowedError`. Both failures were only logged, so the half-written entry survived anyway, exactly the outcome the cleanup exists to prevent.

Cleanup now falls back to `closeWritable` when the abort fails, which releases the lock so the removal can land (the partial bytes it commits are deleted immediately after). Covered by a new test that fails the abort for real against Chrome's OPFS and asserts the entry is gone from a freshly resolved root.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
dab246: I still don't understand why a loop with this always-true value is needed.

workplace/lib/data/datasource/drive_transfer/opfs_feature_detection.dart
```diff
@@ -0,0 +1,49 @@
...
+  /// `getDirectory` several releases before `createWritable` (Baseline only
+  /// with Safari 26.0), so checking the former alone lets those versions
+  /// through and then fails at `openWritable`, fallback never selected.
+  ///
+  /// Property reads only, so callers can cache this once per session.
+  bool isOpfsSupported() {
+    final storage =
+        (web.window.navigator as _NavigatorStorageProbe).storage;
+    if (storage == null) return false;
+    if ((storage as _StorageManagerFeatureProbe).getDirectory == null) {
```
dab246: If `storage` is not of type `_StorageManagerFeatureProbe`, this will throw an exception.

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
```diff
@@ -0,0 +1,122 @@
...
+        );
+      }
+      // Spelled out rather than via `DioException.connectionError`, which
+      // hardcodes `error: null` and would drop the browser's own failure.
+      throw DioException(
+        type: DioExceptionType.connectionError,
+        requestOptions: requestOptions,
+        error: e,
+        message: 'The connection errored: the download request failed',
+      );
```
dab246, line 76: IMO, we should have a shared handler function for this catch logic.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: A more detailed doc comments updated explaining the existent of this `while (true)` loop

workplace/lib/data/datasource/drive_transfer/opfs_feature_detection.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
tddang-linagora: There's only one call site for this mapping. The upload leg's errors come from XHR event
callbacks — `onerror` → `connectionError`, `onabort` → `requestCancelled`, `onload` →
`badResponse` (`opfs_xhr_upload.dart:66-91`) — with no cancel/timeout flags and no
try/catch, so a shared handler would have to abstract over two shapes that differ in every
branch and would end up as a parameter list describing which case to produce. Happy to
extract one the moment a third site needs the same mapping.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,61 @@
...
+        request.cancelToken.whenCancel.then((_) => upload.abort()).catchError(
+      (error) {
+        logWarning('BrowserOpfsDriveFileUploader: failed to abort upload of ${request.fileName}: $error');
+      },
+    );
+    try {
+      final json = await upload.response;
+      final uploadResponse = UploadResponse.fromJson(json);
+      return uploadResponse.toAttachment(
+          nameFile: request.fileName, charset: null);
```
dab246: why set charset is null?

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

workplace/lib/data/datasource/drive_transfer/drive_transfer_strategy_factory_web.dart
```diff
@@ -0,0 +1,62 @@
...
+  /// Reclaims OPFS entries orphaned by a tab that died mid-transfer. Fire and
+  /// forget: it only touches entries hours older than anything this session
+  /// creates, so no transfer waits on it — or fails with it.
+  static void _sweepStaleTempFilesOnce() {
+    if (_swept) return;
+    _swept = true;
+    unawaited(OpfsJsBindings.instance.sweepStaleTempFiles().catchError((error) {
+      logWarning('DriveTransferStrategyFactory: OPFS sweep failed: $error');
+    }));
+  }
```
hoangdat: should wrap in try/catch?

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
dab246: If there are multiple handling scenarios here, this method will become bloated, resulting in excessive complexity; we should extract that logic into a separate function.

.github/workflows/analyze-test.yaml
dab246: Still missing file `opfs_js_bindings_test.dart` at here

workplace/test/data/datasource/drive_transfer/opfs_js_bindings_test.dart
hoangdat, line 1: need to it into `CHROME_TEST_PATHS`?

workplace/test/data/datasource/drive_transfer/drive_transfer_strategy_factory_web_test.dart
```diff
@@ -0,0 +1,68 @@
...
+  required Uri uploadUri,
+  required OnFileProcessedProgress onUploadProgress,
+  required CancelToken cancelToken,
+}) =>
+    throw UnimplementedError();
+
+void main() {
+  setUp(DriveTransferStrategyFactory.resetCache);
+
+  group('DriveTransferStrategyFactory (web)', () {
```
dab246, line 59: Current uploader tests mock `OpfsJsBindings.uploadFile()`, so they do not execute the real `XMLHttpRequest` event wiring. Missing coverage includes:
- request headers;
- progress handling / unknown totals;
- `onerror`;
- `onabort`;
- non-2xx responses;
- malformed JSON in successful responses;
- synchronous exceptions from `open`, `setRequestHeader`, or `send`.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,313 @@
...
+        fileName: doc.name,
+        fileSize: received,
+        mimeType: doc.mimeType,
+      );
+    } catch (e) {
+      completed = true;
+      await scope.cleanupAfterFailure();
+      rethrow;
+    } finally {
+      if (cancelSubscription != null) unawaited(cancelSubscription);
```
dab246: `unawaited(cancelSubscription)` does not unsubscribe the `whenCancel` callback.

After a successful transfer, `CancelToken.whenCancel` may remain pending and retain its callback. That closure still captures the `ReadableStreamDefaultReader` (and, in the uploader, the XHR handle), even though `completed = true` prevents further cancellation work.

This is only a leak risk when the `CancelToken` outlives the transfer, but it is worth clearing the captured active reader/XHR in `finally` and releasing the reader lock (`releaseLock()`) once the stream has completed or been cancelled.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,61 @@
...
+    } catch (_) {
+      // A cancelled token means the user pressed ✕, whatever the abort
+      // surfaced as. Makes cancelling after the XHR exists indistinguishable
+      // from cancelling before it did (above).
+      _throwIfCancelled(request.cancelToken);
+      rethrow;
+    } finally {
+      unawaited(cancelSubscription);
```
dab246: idem

workplace/lib/data/datasource/drive_transfer/opfs_xhr_upload.dart
```diff
@@ -0,0 +1,147 @@
...
+    xhr.send(request.file);
+
+    return XhrUploadHandle(
+      response: completer.future,
+      abort: () => xhr.abort(),
+    );
+  }
+
+  void _applyHeaders(web.XMLHttpRequest xhr, XhrUploadFileRequest request) {
+    xhr.setRequestHeader('Authorization', request.authHeader);
```
dab246, line 131: Missing headear: `Accept: application/json;jmapVersion=rfc-8621`

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,122 @@
...
+    if (cancelSignal != null) {
+      unawaited(cancelSignal.then((_) {
+        cancelled = true;
+        controller.abort();
+      }));
+    }
+    final headersDeadline = Timer(driveTransferReceiveTimeout, () {
+      timedOut = true;
+      controller.abort();
+    });
```
hoangdat: remove timeout

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Add charset detector

.github/workflows/analyze-test.yaml
tddang-linagora: Done

workplace/test/data/datasource/drive_transfer/drive_transfer_strategy_factory_web_test.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_xhr_upload.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,122 @@
...
+  /// rather than anything Dio-shaped) as `connectionError` carrying the
+  /// original error.
+  Future<OpfsFetchStream> fetchStream(Uri url, {Future<void>? cancelSignal}) async {
+    final requestOptions = RequestOptions(path: url.toString());
+    final controller = web.AbortController();
+    // Both aborts go through the one controller, so the reason has to be
+    // recorded here to tell a user cancellation from a stalled server. Set
+    // before `abort()`, hence observable by the time the fetch rejects.
+    var cancelled = false;
+    var timedOut = false;
```
hoangdat: remove timeout

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,122 @@
...
+    if (body == null) {
+      throw DioException.connectionError(
+        requestOptions: requestOptions,
+        reason: 'the download response has no body',
+      );
+    }
+    final contentLength =
+        int.tryParse(response.headers.get('content-length') ?? '') ?? -1;
+    final reader = body.getReader() as web.ReadableStreamDefaultReader;
+    return OpfsFetchStream(reader: reader, contentLength: contentLength);
```
hoangdat: split it into another method

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,146 @@
...
+  ///
+  /// Failures come out as [DioException], the shape
+  /// `BufferedWebDriveFileStager` produces, so callers branch on
+  /// `DioExceptionType` instead of matching messages. Cancellation surfaces as
+  /// `cancel` — the same type Dio gives the buffered leg for the same user
+  /// action — a headers stall as `receiveTimeout`, and any other transport
+  /// failure (CORS, DNS, TLS, offline, which reject with a browser `TypeError`
+  /// rather than anything Dio-shaped) as `connectionError` carrying the
+  /// original error.
+  Future<OpfsFetchStream> fetchStream(Uri url, {Future<void>? cancelSignal}) async {
```
hoangdat: this quite confuse in dart world. So how about we change it to `openDownload` because we did not consume anything here

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,146 @@
...
+import 'dart:typed_data';
+
+import 'package:dio/dio.dart';
+import 'package:web/web.dart' as web;
+import 'package:workplace/data/datasource/drive_transfer/drive_file_stager.dart';
+
+/// Result of [OpfsFetchStreaming.fetchStream]: a locked reader plus the
+/// declared content-length (-1 when the header is absent/unparseable, the
+/// same unknown-total sentinel Dio's `onReceiveProgress` uses).
+class OpfsFetchStream {
```
hoangdat: How about 
```suggestion
class FetchDownloadHandle {
```

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
```diff
@@ -0,0 +1,146 @@
...
+  final web.ReadableStreamDefaultReader reader;
+  final int contentLength;
+
+  const OpfsFetchStream({required this.reader, required this.contentLength});
+}
+
+/// The download leg: `fetch` driven one chunk at a time, so a document never
+/// has to fit in the JS heap.
+mixin OpfsFetchStreaming {
```
hoangdat:
```suggestion
mixin OpfsFetchDownload {
```

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
hoangdat, line 1: `opfs_fetch_download.dart`

workplace/test/data/datasource/drive_transfer/opfs_js_bindings_test.dart
tddang-linagora: Added in previous commit

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
```diff
@@ -0,0 +1,146 @@
...
+    final controller = web.AbortController();
+    // Both aborts go through the one controller, so the reason has to be
+    // recorded here to tell a user cancellation from a stalled server. Set
+    // before `abort()`, hence observable by the time the fetch rejects.
+    var cancelled = false;
+    var timedOut = false;
+    if (cancelSignal != null) {
+      unawaited(cancelSignal.then((_) {
+        cancelled = true;
+        controller.abort();
```
hoangdat, line 52: it will throw DomException, how we will handle it?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+      return OpfsStagedFile(
+        fileHandle: handle,
+        removeEntry: (_) => _bindings.removeTempFile(tempFileName),
+        fileName: doc.name,
+        fileSize: received,
+        mimeType: doc.mimeType,
+      );
+    } catch (e) {
+      await scope.cleanupAfterFailure();
+      rethrow;
```
hoangdat: where the code spends 99% of its time, and rethrow at the end of the catch sends whatever the browser threw straight up — `DomException` on cancel or connection drop, `TypeError/QuotaExceededError` on a failed write. The temp file is cleaned up correctly in every one of those cases; it's only the exception type that's wrong, which is why cancelling shows an error toast.

workplace/lib/data/datasource/drive_transfer/drive_transfer_strategy_factory_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_streaming.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
tddang-linagora: added

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+    } catch (cleanupError) {
+      logWarning('OpfsDriveFileStager: failed to $step for $tempFileName: $cleanupError');
+      return false;
+    }
+  }
+}
+
+/// Web+OPFS strategy: flat memory on both legs (streamed download to OPFS,
+/// streamed upload from OPFS via raw XHR).
+class OpfsDriveTransferStrategy extends DriveTransferStrategy<OpfsStagedFile> {
```
hoangdat: why not keep it in separate file?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+    }
+  }
+
+  Future<int> _streamToFile(_StreamToFileRequest request) async {
+    final received = await _readAndWriteAll(request);
+    _verifyComplete(request, received);
+    return received;
+  }
+
+  /// Pumps the response body into the OPFS temp file one chunk at a time.
```
hoangdat, line 126: too long comments, try to reduce it and other comments in code

workplace/test/data/datasource/drive_transfer/opfs_xhr_upload_test.dart
```diff
@@ -0,0 +1,317 @@
+@TestOn('chrome')
```
hoangdat, line 1: need on `CHROME_TEST_PATHS`

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+      final writable = scope.writable = await _bindings.openWritable(handle);
+
+      // Reads the reader off the scope rather than capturing it: a bare
+      // `whenCancel.then(...)` has no unsubscribe, so this listener outlives
+      // stage() for as long as the token does. `scope.release()` below empties
+      // what it can still reach, which both stops it acting on a finished
+      // transfer and drops its hold on the reader.
+      cancelSubscription = cancelToken.whenCancel.then((_) async {
+        final activeReader = scope.downloadHandle?.reader;
+        if (activeReader != null) await _bindings.cancelReader(activeReader);
```
hoangdat: on `cancel`, `cancelReader` be call 2 times: here and inside `cleanupAfterFailure`?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+      } catch (_) {
+        // The fetch abort controller stays live after the headers, so a
+        // cancellation landing mid-read surfaces as a transport failure. The
+        // token is the only thing that can tell the two apart, and callers
+        // branch on `DioExceptionType.cancel`.
+        _throwIfCancelled(request.cancelToken);
+        rethrow;
+      }
+      if (chunk == null) break;
+      await _bindings.writeChunk(request.writable, chunk);
```
hoangdat: why not wrap it in try/catch (same with readChunk)?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+    final reader = request.downloadHandle.reader;
+    var received = 0;
+    while (true) {
+      // Bails out at the iteration boundary. `cancelReader` is async, so
+      // without this the loop would keep reading and writing chunks until the
+      // cancellation propagated through the JS reader.
+      _throwIfCancelled(request.cancelToken);
+      final Uint8List? chunk;
+      try {
+        chunk = await _bindings.readChunk(reader);
```
hoangdat: what happen if `readChunk` return Uint8List(0)? infinite loop?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+        // cancellation landing mid-read surfaces as a transport failure. The
+        // token is the only thing that can tell the two apart, and callers
+        // branch on `DioExceptionType.cancel`.
+        _throwIfCancelled(request.cancelToken);
+        rethrow;
+      }
+      if (chunk == null) break;
+      await _bindings.writeChunk(request.writable, chunk);
+      received += chunk.length;
+      request.onDownloadProgress(received, request.downloadHandle.contentLength);
```
hoangdat: should wrap it try/catch to prevent break the logic

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+        received: received,
+        expected: expected,
+      );
+    }
+  }
+
+  /// The `??` fallback is unreachable in practice — Dio defines `isCancelled`
+  /// as `cancelError != null` — and only keeps a cancelled transfer from being
+  /// reported as a success should that ever stop holding.
+  static void _throwIfCancelled(CancelToken cancelToken) {
```
hoangdat: Cancel used to be silent (cancelReader → done:true), so the loop asked "cancelled?" at every step. Now abort() errors the stream — cancel throws, can't be missed.

Proposal: drop the 3 _throwIfCancelled calls, ask once in stage()'s catch → throw _asDioFailure(e, cancelToken).

Bonus: writeChunk failures (disk full) slip past all 3 checks today and escape as raw JS errors. One catch covers them too — fewer checks, better coverage.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,346 @@
...
+      // a locked entry fails — so the half-written file the abort was meant to
+      // discard would survive instead. `close()` commits the partial bytes,
+      // but the removal below deletes them, and it is the only way left to
+      // release the lock.
+      if (!aborted) {
+        await _attempt('close writable after failed abort',
+            () => bindings.closeWritable(openWritable));
+      }
+    }
+    if (handle != null) {
```
hoangdat: IMO, we can remove `if (handle != null) {`, we should call _attemp to remove it, error will be caught inside

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
```diff
@@ -0,0 +1,113 @@
...
+
+  /// Removes staging entries left behind by a tab that died mid-transfer:
+  /// OPFS is persistent, and neither `OpfsStagedFile.dispose()` nor the
+  /// stager's failure cleanup runs on a crash.
+  ///
+  /// [olderThan] keeps the sweep safe while another tab of the same origin is
+  /// mid-transfer — its entries are minutes old, far inside the window.
+  /// Per-entry failures are logged, never rethrown.
+  Future<void> sweepStaleTempFiles({
+    Duration olderThan = const Duration(hours: 24),
```
hoangdat: how about keep it shorter? 4 hours?

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,87 @@
...
+
+  /// Only text/plain carries one, the same rule `FileUploader` applies to
+  /// attachments picked from disk. Sniffed from a prefix rather than the whole
+  /// body; detection failures already fall back to utf-8 inside [FileUtils].
+  Future<String?> _resolveCharset(web.File file, String? mimeType) async {
+    if (mimeType != FileUtils.TEXT_PLAIN_MIME_TYPE) return null;
+    final prefix = await _bindings.readFilePrefix(file, _charsetSniffByteLimit);
+    final charset = await _fileUtils.getCharsetFromBytes(prefix);
+    return charset.toLowerCase();
+  }
```
hoangdat, line 144:
```suggestion
Future<String?> _resolveCharset(web.File file, String? mimeType) async {
  if (mimeType != FileUtils.TEXT_PLAIN_MIME_TYPE) return null;
  try {
    final prefix = await _bindings.readFilePrefix(file, _charsetSniffByteLimit);
    return (await _fileUtils.getCharsetFromBytes(prefix)).toLowerCase();
  } catch (e) {
    logWarning('BrowserOpfsDriveFileUploader: charset sniff failed: $e');
    return null;   // same outcome as a non-text file
  }
}
```

workplace/lib/data/datasource/drive_transfer/opfs_xhr_upload.dart
```diff
@@ -0,0 +1,190 @@
...
+
+    // `total` is 0 when the length is unknown, but the download leg reports an
+    // unknown total as -1. Normalized here so one progress consumer doesn't
+    // have to know two sentinels — and doesn't divide by zero.
+    xhr.upload.onprogress = ((web.ProgressEvent event) {
+      request.onUploadProgress(
+        event.loaded,
+        event.lengthComputable ? event.total : -1,
+      );
+    }).toJS;
```
hoangdat, line 92: try/catch

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,87 @@
...
+    final cancelSubscription = request.cancelToken.whenCancel
+        .then((_) => activeUpload?.abort())
+        .catchError(
+      (error) {
+        logWarning('BrowserOpfsDriveFileUploader: failed to abort upload of ${request.fileName}: $error');
+      },
+    );
+    try {
+      final json = await response;
+      final uploadResponse = UploadResponse.fromJson(json);
```
hoangdat: it maybe throw `TypeError` not DioException,

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,87 @@
...
+      authHeader: request.authHeader,
+      mimeType: request.mimeType,
+      onUploadProgress: request.onUploadProgress,
+    ));
+    final response = activeUpload.response;
+
+    final cancelSubscription = request.cancelToken.whenCancel
+        .then((_) => activeUpload?.abort())
+        .catchError(
+      (error) {
```
hoangdat: a lot of error can throw here, but not catch

hoangdat:
I want to go without mixin. I propose to use interface instead. WDYT? 

```dart
abstract interface class DriveDownloadSource {   // no OPFS in it — and now the name says so
  Future<FetchDownloadHandle> openDownload(Uri url, {Future<void>? cancelSignal});
  Future<Uint8List?> readChunk(ReadableStreamDefaultReader reader);
  Future<void> cancelReader(ReadableStreamDefaultReader reader);
  void releaseReaderLock(ReadableStreamDefaultReader reader);
}

abstract interface class OpfsStore { /* the 9 storage methods */ }
abstract interface class DriveUploadTransport { XhrUploadHandle uploadFile(...); }
abstract interface class OpfsCapability { bool isOpfsSupported(); }

OpfsDriveFileStager({required DriveDownloadSource download, required OpfsStore store})
BrowserOpfsDriveFileUploader({required OpfsStore store, required DriveUploadTransport transport})
```

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: `readChunk` only returns `Uint8List(0)` when `result.done` is false and `result.value` is null (opfs_fetch_download.dart:139-141), and each turn of the loop awaits a fresh `reader.read()` rather than re-polling a buffer — so an empty chunk costs one iteration and the stream still has to resolve `done: true` to terminate. A stream that yielded null values forever would park the browser side with or without this loop; a body that ends early is caught by the content-length check in `_verifyComplete`.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/test/data/datasource/drive_transfer/opfs_xhr_upload_test.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_xhr_upload.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

tddang-linagora:
> I want to go without mixin. I propose to use interface instead. WDYT?
> 
> ```dart
> abstract interface class DriveDownloadSource {   // no OPFS in it — and now the name says so
>   Future<FetchDownloadHandle> openDownload(Uri url, {Future<void>? cancelSignal});
>   Future<Uint8List?> readChunk(ReadableStreamDefaultReader reader);
>   Future<void> cancelReader(ReadableStreamDefaultReader reader);
>   void releaseReaderLock(ReadableStreamDefaultReader reader);
> }
> 
> abstract interface class OpfsStore { /* the 9 storage methods */ }
> abstract interface class DriveUploadTransport { XhrUploadHandle uploadFile(...); }
> abstract interface class OpfsCapability { bool isOpfsSupported(); }
> 
> OpfsDriveFileStager({required DriveDownloadSource download, required OpfsStore store})
> BrowserOpfsDriveFileUploader({required OpfsStore store, required DriveUploadTransport transport})
> ```

Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

workplace/lib/data/datasource/drive_transfer/drive_transfer_strategy_factory_web.dart
```diff
@@ -0,0 +1,82 @@
...
+
+/// Web branch of the conditional export in
+/// `drive_transfer_strategy_factory.dart`: OPFS detection runs once per
+/// session and is cached (module-level field), so repeated `create()` calls
+/// across a batch never re-probe.
+class DriveTransferStrategyFactory {
+  const DriveTransferStrategyFactory._();
+
+  static bool? _opfsSupported;
+  static bool _swept = false;
```
dab246: IMO, There is no need to use a static variable here; doing so only makes it easier for unit tests to access the variable, without providing any performance benefit for the production code.

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
```diff
@@ -0,0 +1,147 @@
...
+        await root.removeEntry(name).toDart;
+      } catch (e) {
+        logWarning('OpfsFileOps: failed to sweep stale temp file $name: $e');
+      }
+    }
+  }
+
+  /// Names are `<prefix><microsSinceEpoch>_...`. A name that doesn't parse is
+  /// left alone.
+  static bool _isStaleTempFile(String name, DateTime cutoff) {
```
dab246: Why do this function use `static`?

workplace/lib/data/datasource/drive_transfer/drive_transfer_strategy_factory_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
tddang-linagora: Updated

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,118 @@
...
+    // Cleared in `finally` so the un-unsubscribable `whenCancel` listener
+    // below stops holding the XHR once this call is over — `whenCancel` has no
+    // removal, and the token can outlive the transfer.
+    XhrUploadHandle? activeUpload;
+    Future<void>? cancelSubscription;
+    try {
+      final file = await _store.getFile(request.fileHandle);
+      _throwIfCancelled(request.cancelToken);
+      final charset = await _resolveCharset(file, request.mimeType);
+      _throwIfCancelled(request.cancelToken);
```
dab246: Can we use `Future.wait` to execute these two logic processes simultaneously?

workplace/lib/data/datasource/drive_transfer/opfs_feature_detection.dart
```diff
@@ -0,0 +1,54 @@
...
+  /// True when both halves of the write path are present. Safari shipped
+  /// `getDirectory` several releases before `createWritable` (Baseline only
+  /// with Safari 26.0), so checking the former alone lets those versions
+  /// through and then fails at `openWritable`, fallback never selected.
+  ///
+  /// Property reads only, so callers can cache this once per session.
+  @override
+  bool isOpfsSupported() {
+    final storage =
+        (web.window.navigator as _NavigatorStorageProbe).storage;
```
dab246: if `web.window.navigator` is not type _NavigatorStorageProbe. Can this throw exception?

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
```diff
@@ -0,0 +1,169 @@
...
+abstract interface class DriveDownloadSource {
+  Future<FetchDownloadHandle> openDownload(Uri url, {Future<void>? cancelSignal});
+
+  Future<Uint8List?> readChunk(web.ReadableStreamDefaultReader reader);
+
+  Future<void> cancelReader(web.ReadableStreamDefaultReader reader);
+
+  void releaseReaderLock(web.ReadableStreamDefaultReader reader);
+}
+
```
dab246: Why don't we move these datasource abstractions into separate files? That would help avoid creating large files.

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
```diff
@@ -0,0 +1,147 @@
...
+      web.FileSystemWritableFileStream stream, Uint8List chunk);
+
+  Future<void> closeWritable(web.FileSystemWritableFileStream stream);
+
+  Future<void> abortWritable(web.FileSystemWritableFileStream stream);
+
+  Future<void> removeTempFile(String fileName);
+
+  Future<void> sweepStaleTempFiles({Duration olderThan});
+}
```
dab246: idem

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
```diff
@@ -0,0 +1,301 @@
...
+      received += chunk.length;
+      request.onDownloadProgress(received, request.downloadHandle.contentLength);
+    }
+    return received;
+  }
+
+  /// A server that closes early ends the stream normally, so without this the
+  /// truncated file would pass as a successful staging. Only checkable when the
+  /// response declared a length.
+  void _verifyComplete(_StreamToFileRequest request, int received) {
```
hoangdat, line 194: in some cases, content-length is unknown: CORS, compress... , so maybe we should remove this verification? or have better reliable check

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,118 @@
...
+      if (cancelSubscription != null) unawaited(cancelSubscription);
+    }
+  }
+
+  /// [OpfsXhrUpload] already shapes every transport failure; this only covers
+  /// what runs outside it — the OPFS read and the response decoding.
+  static Object _asDioFailure(Object error, Uri uploadUri) {
+    if (error is DioException) return error;
+    return DioException(
+      type: DioExceptionType.badResponse,
```
hoangdat: every error map to `badResponse` is not a good idea

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
hoangdat: I still not found the try/catch on `onDownloadProgress`? we dont want it break the main logic

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_feature_detection.dart
tddang-linagora: try/catch added

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_file_ops.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_stager.dart
tddang-linagora: Done

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
tddang-linagora: Done

workplace/test/data/datasource/drive_transfer/opfs_drive_file_stager_test.dart
hoangdat, line 1: add these tests

Group chunked download (new) — all four need _ScriptedDownload to enqueue ≥3 chunks:
- ⊕ accumulates progress across chunks instead of reporting each chunk on its own
- ⊕ writes each chunk as it arrives rather than buffering the whole body
- ⊕ stages the concatenation of every chunk, in order
- ⊕ returns the total received byte count as the staged file size

Group content-length (new) — last two need _NameRecordingStore:
- ⊕ succeeds when the body overruns content-length, the way a compressed response does
- ⊕ skips the completeness check when the response declares no content-length
- ⤴ carries the received and expected counts on the incomplete exception
- ⊕ removes the truncated temp entry when the body ends short of content-length

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

workplace/lib/data/datasource/drive_transfer/opfs_fetch_download.dart
hoangdat, line 1: should add tests it on chrome 

Group openDownload — fixtures: blob: URL for the 200 + content-length cases, a relative missing path on the test server for the non-2xx:
- ⊕ maps a non-2xx download response to a badResponse carrying the status
- ⊕ maps a response with no body to a connection error
- ⊕ reads content-length off the response headers
- ⊕ reports an absent or unparseable content-length as -1
- ⊕ maps a cancel landing before the headers to a cancellation, not a connection error

workplace/test/data/datasource/drive_transfer/opfs_drive_file_uploader_test.dart
hoangdat, line 1: add tests

- ⊕ hands the staged file, upload uri, auth header, mime type and progress callback to the transport unchanged
- ⤴ maps a failure to open the staged entry to DioExceptionType.unknown
- ⤴ maps an upload transport failure to DioExceptionType.unknown
- ⊕ does not abort the upload when the token is cancelled after it already completed
- ⊕ sniffs at most 64KB of a large text document rather than reading it whole

workplace/test/data/datasource/drive_transfer/drive_transfer_strategy_factory_web_test.dart
hoangdat, line 1: - ⊕ falls back to the buffered strategy when the OPFS probe throws
- ⊕ sweeps stale temp files once across repeated create() calls
- ⊕ does not sweep when OPFS is unavailable

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,151 @@
...
+      if (cancelSubscription != null) unawaited(cancelSubscription);
+    }
+  }
+
+  /// The one failure on this path that really is the response's: a 2xx whose
+  /// JSON parsed but doesn't match the upload schema. Classified here rather
+  /// than in [_asDioFailure] so that catch-all doesn't have to guess a type it
+  /// cannot know — a `TypeError` off the decode and a `DOMException` off the
+  /// OPFS read arrive there indistinguishable.
+  static UploadResponse _parseUploadResponse(
```
dab246: IMO, we don't need static here.

workplace/lib/data/datasource/drive_transfer/opfs_drive_file_uploader_web.dart
```diff
@@ -0,0 +1,151 @@
...
+    }
+  }
+
+  /// [OpfsXhrUpload] already shapes every transport failure and
+  /// [_parseUploadResponse] the decode, so what reaches here is the OPFS read
+  /// and anything else unforeseen — browser errors with no HTTP meaning.
+  /// [DioExceptionType.unknown] is what that is, the same call
+  /// `OpfsDriveFileStager._asDioFailure` makes for its own OPFS leg; claiming
+  /// `badResponse` would point callers at a server that answered fine.
+  static Object _asDioFailure(Object error, Uri uploadUri) {
```
dab246: idem

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4746

Issue #4728, opened 2026-07-27
Download and upload drive attachments in composer


tddang-linagora:
### Background

Per [ADR-0103](../../docs/adr/0103-attach-drive-file-as-attachment.md), the composer today inserts picked drive documents that carry a `sharingLink` as plain HTML links (see ADR-0095). Documents that instead carry a `downloadLink` need to be downloaded and attached as real, uploaded email attachments — visibly, with progress, without blocking the composer, and without exceeding size limits.

This issue is the end-to-end integration: partitioning picked documents, running a bounded-concurrency download→upload pipeline per file, and showing live progress in the composer UI. It assumes:
- a size-limit validator abstraction is available to gate the batch before any download starts (an `AttachmentUploadValidator`-shaped check — see ADR-0103's "Validation" section for the exact contract expected);
- a per-platform way to download a drive document into temporary storage and get back a result ready to upload (an interface matching ADR-0103's `DriveFileStager` / sealed staged-file description).

If either of those doesn't exist yet, build a minimal version inline rather than blocking this issue — but prefer the shapes described in the ADR so future consolidation is possible.

### Scope

1. **Partition.** When the composer receives picked drive documents: docs with `sharingLink != null` keep the existing HTML-link behavior unchanged; docs with `sharingLink == null && downloadLink != null` go through the new download+upload path. When a doc has both, `sharingLink` wins (a live editable link beats a static copy) — the two sets must be mutually exclusive by construction.
2. **Size gate before any download.** Sum the declared size (`DriveDocument.size`) of the downloadable set and run it through the size-limit validator *before* starting any download.
3. **Actual-byte guard.** Declared size is unreliable for some backends (export-on-demand documents may report `0` up front). Track a shared running total of actually-received bytes across all in-flight downloads, updated by *delta* per progress callback (progress callbacks report cumulative bytes per file, so each file's own previous value must be subtracted, not summed raw — otherwise the guard double-counts and trips early). Cancel a file's transfer once the shared total exceeds budget; this is a per-file failure, not a batch abort. Budget = server `maxSizeAttachmentsPerEmail` capability when present, else a fixed client-side fallback constant (independent of the size-limit dialog's own "absent capability = unlimited" behavior, which stays unchanged).
[...]

PR #4751, opened 2026-08-13
TF-4728 [Part 1] Attach drive files as real attachments on web

tddang-linagora:
## Issue
- #4728 

## Description
Drive documents picked with only a `downloadLink` are downloaded into OPFS (or buffered, when
OPFS is unavailable) and uploaded as real email attachments, instead of falling back to the
"not available yet" message.

- `DriveAttachmentHandler` partitions a pick by link kind — `sharingLink` wins when a document
  carries both — and hands the download-only half to the new `DriveTransferPipeline`.
- `DriveTransferPipeline` selects a platform strategy through `DriveTransferStrategyFactory`
  and runs one bounded-concurrency pipeline per batch (OPFS 3 at a time, buffered 2). A `null`
  strategy means the platform has no staging support, so the caller keeps link-only behaviour;
  the picker's "Add as attachment" action is gated to web for the same reason, leaving mobile
  behaviour unchanged.
- Each file gets its chip immediately and fills a single bar across both legs: the new
  `downloading` status maps onto 0–50, `uploading` onto 50–100. A failing file drops its own
  chip only; its siblings continue.
- The batch is gated on its declared total through the existing attachment validator before any
  download starts (`AttachmentUploadValidationService.validateBytes`).
- `AuthorizationInterceptors` now exposes `currentAuthorizationHeader`, so the OPFS raw-XHR
  upload builds the same Basic/Bearer header `onRequest` does — built in one place.

## Demo

https://github.com/user-attachments/assets/53b55b9b-b81e-4a1b-91e5-6b1d4ac39b3e

## Summary by CodeRabbit

* **New Features**
  * Added web support for downloading Drive documents and attaching them directly to composed content.
  * Drive transfers show placeholders with download and upload progress.
  * Multiple documents can transfer concurrently.
  * Drive sharing links continue to insert directly into the composer.
  * Added localized success messages showing attached file counts.
  * Unsupported environments retain link-only behavior.

* **Bug Fixes**
  * Improved attachment validation, cancellation, failure cleanup, authorization, and invalid-link handling.
  * Progress indicators remain accurate and capped at 100%.

### DIFF
--- a/integration_test/scenarios/search/search_email_with_tag_scenario.dart
+++ b/integration_test/scenarios/search/search_email_with_tag_scenario.dart
@@ -50,8 +50,6 @@ class SearchEmailWithTagScenario extends BaseTestScenario
 
       await commonRobot.selectContextMenuItemByName(labelDisplayName);
       await searchRobot.expectEmailListCountAtLeast(emailCount);
-
-      await $.pumpAndSettle(duration: const Duration(seconds: 1));
     }
   }
 
--- a/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
+++ b/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
@@ -1,0 +1,24 @@
+import 'dart:async';
+
+/// Runs [task] over [items] with at most [maxConcurrent] in flight.
+///
+/// Workers pull from one shared iterator, so a slot frees the instant its
+/// current item finishes — no "start a batch, wait for all of it, start the
+/// next" stalling. [task] is expected to handle its own failures; an error it
+/// lets escape aborts that worker only, and the others keep draining.
+Future<void> runWithConcurrency<T>(
+  Iterable<T> items,
+  int maxConcurrent,
+  Future<void> Function(T item) task,
+) async {
+  final iterator = items.iterator;
+
+  Future<void> worker() async {
+    while (iterator.moveNext()) {
+      await task(iterator.current);
+    }
+  }
+
+  final workerCount = maxConcurrent < 1 ? 1 : maxConcurrent;
+  await Future.wait(List.generate(workerCount, (_) => worker()));
+}
--- a/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
+++ b/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
@@ -6,23 +6,32 @@ import 'dart:async';
 /// current item finishes — no "start a batch, wait for all of it, start the
 /// next" stalling. [task] is expected to handle its own failures; an error it
 /// lets escape aborts that worker only, and the others keep draining.
+///
+/// A worker is only spawned once there is an item for it to own, so an
+/// oversized [maxConcurrent] costs nothing: the worker count never exceeds the
+/// item count.
 Future<void> runWithConcurrency<T>(
   Iterable<T> items,
   int maxConcurrent,
   Future<void> Function(T item) task,
 ) {
   final iterator = items.iterator;
-  final workerCount = maxConcurrent < 1 ? 1 : maxConcurrent;
-  return Future.wait(
-    List.generate(workerCount, (_) => _drainQueue(iterator, task)),
-  );
+  final maxWorkers = maxConcurrent < 1 ? 1 : maxConcurrent;
+  final workers = <Future<void>>[];
+  while (workers.length < maxWorkers && iterator.moveNext()) {
+    workers.add(_drainQueue(iterator.current, iterator, task));
+  }
+  return Future.wait(workers);
 }
 
-/// One worker: pulls from the shared [iterator] until it is exhausted.
+/// One worker: runs [first], then pulls from the shared [iterator] until it is
+/// exhausted.
 Future<void> _drainQueue<T>(
+  T first,
   Iterator<T> iterator,
   Future<void> Function(T item) task,
 ) async {
+  await task(first);
   while (iterator.moveNext()) {
     await task(iterator.current);
   }
--- a/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
+++ b/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
@@ -10,18 +10,34 @@ import 'dart:async';
 /// A worker is only spawned once there is an item for it to own, so an
 /// oversized [maxConcurrent] costs nothing: the worker count never exceeds the
 /// item count.
+///
+/// [items] is expected to never throw while spawning (pass a materialised
+/// list, not a lazy source with side effects); if it does, the workers
+/// already started are still drained — not orphaned — before the spawn error
+/// is rethrown.
 Future<void> runWithConcurrency<T>(
   Iterable<T> items,
   int maxConcurrent,
   Future<void> Function(T item) task,
-) {
+) async {
   final iterator = items.iterator;
   final maxWorkers = maxConcurrent < 1 ? 1 : maxConcurrent;
   final workers = <Future<void>>[];
-  while (workers.length < maxWorkers && iterator.moveNext()) {
-    workers.add(_drainQueue(iterator.current, iterator, task));
+  try {
+    while (workers.length < maxWorkers && iterator.moveNext()) {
+      workers.add(_drainQueue(iterator.current, iterator, task));
+    }
+  } catch (spawnError) {
+    // Only one error can surface; a worker's own failure is swallowed here,
+    // not left to escape as an unhandled zone error.
+    try {
+      await Future.wait(workers);
+    } catch (_) {
+      // Swallowed: the spawn error below is what the caller sees.
+    }
+    rethrow;
   }
-  return Future.wait(workers);
+  await Future.wait(workers);
 }
 
 /// One worker: runs [first], then pulls from the shared [iterator] until it is
--- a/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
+++ b/lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
@@ -10,15 +10,20 @@ Future<void> runWithConcurrency<T>(
   Iterable<T> items,
   int maxConcurrent,
   Future<void> Function(T item) task,
-) async {
+) {
   final iterator = items.iterator;
-
-  Future<void> worker() async {
-    while (iterator.moveNext()) {
-      await task(iterator.current);
-    }
-  }
-
   final workerCount = maxConcurrent < 1 ? 1 : maxConcurrent;
-  await Future.wait(List.generate(workerCount, (_) => worker()));
+  return Future.wait(
+    List.generate(workerCount, (_) => _drainQueue(iterator, task)),
+  );
+}
+
+/// One worker: pulls from the shared [iterator] until it is exhausted.
+Future<void> _drainQueue<T>(
+  Iterator<T> iterator,
+  Future<void> Function(T item) task,
+) async {
+  while (iterator.moveNext()) {
+    await task(iterator.current);
+  }
 }
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -51,23 +51,26 @@ class DriveAttachmentHandler {
         .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
         .toList();
 
+    // A pick with any link doc is handled as a link pick: mixed batches are
+    // not supported.
     if (linkDocs.isNotEmpty) {
       await insertDriveLinkHtml(
         linkDocs,
         insertHtml: insertHtml,
         appLocalizations: appLocalizations,
       );
+      return;
     }
 
-    final transfersStarted = downloadableDocs.isNotEmpty &&
-        await transferDriveDocuments(downloadableDocs);
-
-    if (linkDocs.isEmpty && !transfersStarted) {
-      getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
-        Exception(),
-        message: appLocalizations?.driveAttachmentInDevelopment,
-      ));
+    if (downloadableDocs.isNotEmpty &&
+        await transferDriveDocuments(downloadableDocs)) {
+      return;
     }
+
+    getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
+      Exception(),
+      message: appLocalizations?.driveAttachmentInDevelopment,
+    ));
   }
 
   Future<void> insertDriveLinkHtml(
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -34,12 +34,7 @@ class DriveAttachmentHandler {
     AppLocalizations? appLocalizations,
   }) async {
     if (result.isEmpty) {
-      getBinding<ToastManager>()?.showMessageFailure(
-        DrivePickFailure(
-          Exception(),
-          message: appLocalizations?.driveNoValidAttachment,
-        ),
-      );
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
       return;
     }
     final linkDocs = result
@@ -64,10 +59,23 @@ class DriveAttachmentHandler {
 
     if (linkDocs.isNotEmpty || transfersStarted) return;
 
-    getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
-      Exception(),
-      message: appLocalizations?.driveAttachmentInDevelopment,
-    ));
+    _showFailureToast(appLocalizations?.driveAttachmentInDevelopment);
+  }
+
+  /// Toasts [message], or logs when there is no [ToastManager] bound: a pick
+  /// that ends with nothing attached must not also end silently.
+  void _showFailureToast(String? message) {
+    final toastManager = getBinding<ToastManager>();
+    if (toastManager == null) {
+      logWarning(
+        'DriveAttachmentHandler::_showFailureToast: no ToastManager bound, '
+        'failure not shown to the user',
+      );
+      return;
+    }
+    toastManager.showMessageFailure(
+      DrivePickFailure(Exception(), message: message),
+    );
   }
 
   Future<void> insertDriveLinkHtml(
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -1,5 +1,6 @@
 import 'package:core/core.dart';
 import 'package:core/utils/html/file_link_card_html_builder.dart';
+import 'package:tmail_ui_user/features/composer/presentation/extensions/drive_document_extension.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
@@ -41,18 +42,12 @@ class DriveAttachmentHandler {
       );
       return;
     }
-    final linkDocs = result.where((doc) {
-      final link = doc.sharingLink;
-      return link != null && (!requireHttps || link.isScheme('https'));
-    }).toList();
-    // Exclusive with linkDocs by construction: a document carrying both links
-    // is a link, because a live document is worth more than a static copy.
-    final downloadableDocs = result.where((doc) {
-      final downloadLink = doc.downloadLink;
-      return doc.sharingLink == null &&
-          downloadLink != null &&
-          (!requireHttps || downloadLink.isScheme('https'));
-    }).toList();
+    final linkDocs = result
+        .where((doc) => doc.isAttachableAsLink(requireHttps: requireHttps))
+        .toList();
+    final downloadableDocs = result
+        .where((doc) => doc.isAttachableAsDownload(requireHttps: requireHttps))
+        .toList();
 
     // Both halves are dispatched: a mixed pick inserts its links and transfers
     // its downloadable docs.
@@ -105,9 +100,8 @@ class DriveAttachmentHandler {
     DriveDocument doc, {
     AppLocalizations? appLocalizations,
   }) {
-    final link = doc.sharingLink;
-    if (link == null) return null;
-    if (requireHttps && !link.isScheme('https')) return null;
+    if (!doc.isAttachableAsLink(requireHttps: requireHttps)) return null;
+    final link = doc.sharingLink!;
 
     final openInDriveLabel =
         appLocalizations?.openInDrive ?? _fallbackOpenInDriveLabel;
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -6,6 +6,12 @@ import 'package:tmail_ui_user/main/utils/toast_manager.dart';
 import 'package:workplace/domain/entity/drive_document.dart';
 import 'package:workplace/presentation/model/drive_pick_state.dart';
 
+/// Inserts [html] into the composer's editor.
+typedef InsertHtmlCallback = Future<void> Function(String html);
+
+/// Starts drive transfers for [docs], returning whether it took them.
+typedef StartDriveTransfersCallback = Future<bool> Function(List<DriveDocument> docs);
+
 class DriveAttachmentHandler {
   DriveAttachmentHandler();
 
@@ -24,8 +30,8 @@ class DriveAttachmentHandler {
   /// transfers existed.
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
-    required Future<void> Function(String html) insertHtml,
-    required Future<bool> Function(List<DriveDocument> docs) startDriveTransfers,
+    required InsertHtmlCallback insertHtml,
+    required StartDriveTransfersCallback startDriveTransfers,
     AppLocalizations? appLocalizations,
   }) async {
     if (result.isEmpty) {
@@ -68,7 +74,7 @@ class DriveAttachmentHandler {
 
   Future<void> insertDriveLinkHtml(
     List<DriveDocument> docs, {
-    required Future<void> Function(String html) insertHtml,
+    required InsertHtmlCallback insertHtml,
     AppLocalizations? appLocalizations,
   }) async {
     await insertHtml(
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -47,9 +47,12 @@ class DriveAttachmentHandler {
     }).toList();
     // Exclusive with linkDocs by construction: a document carrying both links
     // is a link, because a live document is worth more than a static copy.
-    final downloadableDocs = result
-        .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
-        .toList();
+    final downloadableDocs = result.where((doc) {
+      final downloadLink = doc.downloadLink;
+      return doc.sharingLink == null &&
+          downloadLink != null &&
+          (!requireHttps || downloadLink.isScheme('https'));
+    }).toList();
 
     // Both halves are dispatched: a mixed pick inserts its links and transfers
     // its downloadable docs.
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -51,21 +51,20 @@ class DriveAttachmentHandler {
         .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
         .toList();
 
-    // A pick with any link doc is handled as a link pick: mixed batches are
-    // not supported.
+    // Both halves are dispatched: a mixed pick inserts its links and transfers
+    // its downloadable docs.
     if (linkDocs.isNotEmpty) {
       await insertDriveLinkHtml(
         linkDocs,
         insertHtml: insertHtml,
         appLocalizations: appLocalizations,
       );
-      return;
     }
 
-    if (downloadableDocs.isNotEmpty &&
-        await transferDriveDocuments(downloadableDocs)) {
-      return;
-    }
+    final transfersStarted = downloadableDocs.isNotEmpty &&
+        await transferDriveDocuments(downloadableDocs);
+
+    if (linkDocs.isNotEmpty || transfersStarted) return;
 
     getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
       Exception(),
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -14,9 +14,18 @@ class DriveAttachmentHandler {
   /// Whether non-https sharing/thumbnail links should be rejected.
   bool get requireHttps => BuildUtils.isReleaseMode;
 
+  /// Splits picked documents by how they can be attached and dispatches each
+  /// half.
+  ///
+  /// [startDriveTransfers] downloads and attaches the documents that carry
+  /// only a `downloadLink`, returning whether it took them. It declines on
+  /// platforms with no staging strategy, and those documents then fall back to
+  /// the "not available yet" message — the same behaviour as before drive
+  /// transfers existed.
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
     required Future<void> Function(String html) insertHtml,
+    required Future<bool> Function(List<DriveDocument> docs) startDriveTransfers,
     AppLocalizations? appLocalizations,
   }) async {
     if (result.isEmpty) {
@@ -32,19 +41,29 @@ class DriveAttachmentHandler {
       final link = doc.sharingLink;
       return link != null && (!requireHttps || link.isScheme('https'));
     }).toList();
-    // TODO: Update logic here after implement 103. Attach Drive File as Attachment
-    if (linkDocs.isEmpty) {
+    // Exclusive with linkDocs by construction: a document carrying both links
+    // is a link, because a live document is worth more than a static copy.
+    final downloadableDocs = result
+        .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
+        .toList();
+
+    if (linkDocs.isNotEmpty) {
+      await insertDriveLinkHtml(
+        linkDocs,
+        insertHtml: insertHtml,
+        appLocalizations: appLocalizations,
+      );
+    }
+
+    final transfersStarted = downloadableDocs.isNotEmpty &&
+        await startDriveTransfers(downloadableDocs);
+
+    if (linkDocs.isEmpty && !transfersStarted) {
       getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
         Exception(),
         message: appLocalizations?.driveAttachmentInDevelopment,
       ));
-      return;
     }
-    await insertDriveLinkHtml(
-      linkDocs,
-      insertHtml: insertHtml,
-      appLocalizations: appLocalizations,
-    );
   }
 
   Future<void> insertDriveLinkHtml(
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -37,12 +37,14 @@ class DriveAttachmentHandler {
       _showFailureToast(appLocalizations?.driveNoValidAttachment);
       return;
     }
-    final linkDocs = result
-        .where((doc) => doc.isAttachableAsLink(requireHttps: requireHttps))
-        .toList();
-    final downloadableDocs = result
-        .where((doc) => doc.isAttachableAsDownload(requireHttps: requireHttps))
-        .toList();
+    final linkDocs = <DriveDocument>[];
+    final downloadableDocs = <DriveDocument>[];
+    for (final doc in result) {
+      if (doc.isAttachableAsLink(requireHttps: requireHttps)) linkDocs.add(doc);
+      if (doc.isAttachableAsDownload(requireHttps: requireHttps)) {
+        downloadableDocs.add(doc);
+      }
+    }
 
     // Both halves are dispatched: a mixed pick inserts its links and transfers
     // its downloadable docs.
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -9,8 +9,8 @@ import 'package:workplace/presentation/model/drive_pick_state.dart';
 /// Inserts [html] into the composer's editor.
 typedef InsertHtmlCallback = Future<void> Function(String html);
 
-/// Starts drive transfers for [docs], returning whether it took them.
-typedef StartDriveTransfersCallback = Future<bool> Function(List<DriveDocument> docs);
+/// Transfers [docs] as real attachments, returning whether it took them.
+typedef TransferDriveDocumentsCallback = Future<bool> Function(List<DriveDocument> docs);
 
 class DriveAttachmentHandler {
   DriveAttachmentHandler();
@@ -23,15 +23,13 @@ class DriveAttachmentHandler {
   /// Splits picked documents by how they can be attached and dispatches each
   /// half.
   ///
-  /// [startDriveTransfers] downloads and attaches the documents that carry
-  /// only a `downloadLink`, returning whether it took them. It declines on
-  /// platforms with no staging strategy, and those documents then fall back to
-  /// the "not available yet" message — the same behaviour as before drive
-  /// transfers existed.
+  /// [transferDriveDocuments] downloads and attaches the documents that carry
+  /// only a `downloadLink`. It declines on platforms with no staging strategy,
+  /// and those documents then fall back to the "not available yet" message.
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
     required InsertHtmlCallback insertHtml,
-    required StartDriveTransfersCallback startDriveTransfers,
+    required TransferDriveDocumentsCallback transferDriveDocuments,
     AppLocalizations? appLocalizations,
   }) async {
     if (result.isEmpty) {
@@ -62,7 +60,7 @@ class DriveAttachmentHandler {
     }
 
     final transfersStarted = downloadableDocs.isNotEmpty &&
-        await startDriveTransfers(downloadableDocs);
+        await transferDriveDocuments(downloadableDocs);
 
     if (linkDocs.isEmpty && !transfersStarted) {
       getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -40,7 +40,7 @@ class DriveDocumentTransferRunner {
   }) async {
     final taskId = UploadTaskId(_uuid.v4());
     final cancelToken = CancelToken();
-    var exceededDeclaredSize = false;
+    bool exceededDeclaredSize = false;
 
     _uploadController.addDownloadingPlaceholder(
       taskId: taskId,
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -1,0 +1,78 @@
+import 'package:core/utils/app_logger.dart';
+import 'package:dio/dio.dart';
+import 'package:tmail_ui_user/features/upload/domain/model/upload_task_id.dart';
+import 'package:tmail_ui_user/features/upload/presentation/controller/upload_controller.dart';
+import 'package:uuid/uuid.dart';
+import 'package:workplace/data/datasource/drive_transfer/drive_transfer_strategy.dart';
+import 'package:workplace/data/datasource/drive_transfer/staged_drive_file.dart';
+import 'package:workplace/domain/entity/drive_document.dart';
+
+/// Resolves the `Authorization` header for the current session, or null when
+/// there is none to send.
+typedef ResolveAuthHeader = String? Function();
+
+/// Runs one drive document through both legs of its transfer: chip placeholder,
+/// download and upload progress, then completion or failure.
+class DriveDocumentTransferRunner {
+  DriveDocumentTransferRunner({
+    required UploadController uploadController,
+    required Uuid uuid,
+    required ResolveAuthHeader resolveAuthHeader,
+  })  : _uploadController = uploadController,
+        _uuid = uuid,
+        _resolveAuthHeader = resolveAuthHeader;
+
+  final UploadController _uploadController;
+  final Uuid _uuid;
+  final ResolveAuthHeader _resolveAuthHeader;
+
+  Future<void> run({
+    required DriveDocument doc,
+    required Uri uploadUri,
+    required DriveTransferStrategy<StagedDriveFile> strategy,
+  }) async {
+    final taskId = UploadTaskId(_uuid.v4());
+    final cancelToken = CancelToken();
+
+    _uploadController.addDownloadingPlaceholder(
+      taskId: taskId,
+      fileName: doc.name,
+      fileSize: doc.size,
+      mimeType: doc.mimeType,
+      cancelToken: cancelToken,
+    );
+
+    try {
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _resolveAuthHeader() ?? '',
+        onDownloadProgress: (received, total) => _uploadController.updateDownloadProgress(
+          taskId: taskId,
+          received: received,
+          total: total,
+        ),
+        onUploadProgress: (sent, total) => _uploadController.updateUploadProgress(
+          taskId: taskId,
+          sent: sent,
+          total: total,
+        ),
+        cancelToken: cancelToken,
+      ));
+      _uploadController.completeUploadedFile(
+        taskId: taskId,
+        attachment: attachment,
+      );
+    } catch (error) {
+      // A failing file drops its own chip and nothing else: an expired link or
+      // a cancelled transfer must not take its siblings down with it.
+      logWarning('DriveDocumentTransferRunner::run(${doc.name}): $error');
+      if (cancelToken.isCancelled) {
+        // The user asked for this one to stop, so no error toast.
+        _uploadController.deleteFileUploaded(taskId);
+      } else {
+        _uploadController.failDriveTransfer(taskId);
+      }
+    }
+  }
+}
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -177,7 +177,11 @@ class DriveDocumentTransferRunner {
 
   /// A stable, PII-free label for a transfer failure: no document name, no
   /// download URI, no error message.
-  String _errorCategory(Object error) => error is DioException
-      ? 'DioException.${error.type.name}(${error.response?.statusCode})'
-      : error.runtimeType.toString();
+  String _errorCategory(Object error) {
+    if (error is! DioException) return error.runtimeType.toString();
+
+    final category = 'DioException.${error.type.name}';
+    final statusCode = error.response?.statusCode;
+    return statusCode == null ? category : '$category($statusCode)';
+  }
 }
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -13,6 +13,9 @@ typedef ResolveAuthHeader = String? Function();
 
 /// Runs one drive document through both legs of its transfer: chip placeholder,
 /// download and upload progress, then completion or failure.
+///
+/// Every failure — staging, upload, oversized download — drops that file's chip
+/// and toasts; a user cancel drops the chip silently. Siblings are unaffected.
 class DriveDocumentTransferRunner {
   DriveDocumentTransferRunner({
     required UploadController uploadController,
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -31,7 +31,10 @@ class DriveDocumentTransferRunner {
   /// so no file is downloaded only to be rejected by the upload endpoint.
   bool get canAuthenticate => resolveAuthHeader()?.trim().isNotEmpty == true;
 
-  Future<void> run({
+  /// Returns whether the document became an attachment: `false` covers both a
+  /// failed transfer and one the user cancelled, so a batch only reports what
+  /// actually landed.
+  Future<bool> run({
     required DriveDocument doc,
     required Uri uploadUri,
     required DriveTransferStrategy<StagedDriveFile> strategy,
@@ -81,6 +84,7 @@ class DriveDocumentTransferRunner {
         taskId: taskId,
         attachment: attachment,
       );
+      return true;
     } catch (error) {
       _handleTransferFailure(
         taskId: taskId,
@@ -88,6 +92,7 @@ class DriveDocumentTransferRunner {
         exceededDeclaredSize: exceededDeclaredSize,
         error: error,
       );
+      return false;
     }
   }
 
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -51,15 +51,10 @@ class DriveDocumentTransferRunner {
     );
 
     try {
-      final authHeader = _resolveAuthHeader();
-      if (authHeader == null || authHeader.isEmpty) {
-        throw StateError('No authorization header for drive transfer');
-      }
-
       final attachment = await strategy.transfer(DriveTransferRequest(
         doc: doc,
         uploadUri: uploadUri,
-        authHeader: authHeader,
+        authHeader: _requireAuthHeader(),
         onDownloadProgress: (received, total) {
           // The link is sending more than the backend declared, so the size
           // the batch was validated against no longer holds: stop now.
@@ -86,15 +81,41 @@ class DriveDocumentTransferRunner {
         attachment: attachment,
       );
     } catch (error) {
-      // A failing file drops its own chip and nothing else: an expired link or
-      // a cancelled transfer must not take its siblings down with it.
-      logWarning('DriveDocumentTransferRunner::run(${doc.name}): $error');
-      if (cancelToken.isCancelled && !exceededDeclaredSize) {
-        // The user asked for this one to stop, so no error toast.
-        _uploadController.deleteFileUploaded(taskId);
-      } else {
-        _uploadController.failDriveTransfer(taskId);
-      }
+      _handleTransferFailure(
+        taskId: taskId,
+        doc: doc,
+        cancelToken: cancelToken,
+        exceededDeclaredSize: exceededDeclaredSize,
+        error: error,
+      );
+    }
+  }
+
+  /// The `Authorization` header for the current session, or a thrown
+  /// [StateError] when there is none to send.
+  String _requireAuthHeader() {
+    final authHeader = _resolveAuthHeader();
+    if (authHeader == null || authHeader.isEmpty) {
+      throw StateError('No authorization header for drive transfer');
+    }
+    return authHeader;
+  }
+
+  /// A failing file drops its own chip and nothing else: an expired link or
+  /// a cancelled transfer must not take its siblings down with it.
+  void _handleTransferFailure({
+    required UploadTaskId taskId,
+    required DriveDocument doc,
+    required CancelToken cancelToken,
+    required bool exceededDeclaredSize,
+    required Object error,
+  }) {
+    logWarning('DriveDocumentTransferRunner::run(${doc.name}): $error');
+    if (cancelToken.isCancelled && !exceededDeclaredSize) {
+      // The user asked for this one to stop, so no error toast.
+      _uploadController.deleteFileUploaded(taskId);
+    } else {
+      _uploadController.failDriveTransfer(taskId);
     }
   }
 }
--- a/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
@@ -53,24 +53,30 @@ class DriveDocumentTransferRunner {
   /// Puts a chip on screen for every [docs] entry at once, before any slot is
   /// free, and returns the handles to transfer them by.
   List<PendingDriveTransfer> enqueue(List<DriveDocument> docs) {
-    final pendingTransfers = docs
-        .map((doc) => PendingDriveTransfer(
-              doc: doc,
-              taskId: UploadTaskId(uuid.v4()),
-              cancelToken: CancelToken(),
-            ))
-        .toList();
+    if (docs.isEmpty) return const [];
 
-    uploadController.addDownloadingPlaceholders(pendingTransfers
-        .map((pending) => DriveTransferPlaceholder(
-              taskId: pending.taskId,
-              fileName: pending.doc.name,
-              fileSize: pending.doc.size,
-              mimeType: pending.doc.mimeType,
-              cancelToken: pending.cancelToken,
-            ))
-        .toList());
+    final pendingTransfers = <PendingDriveTransfer>[];
+    final placeholders = <DriveTransferPlaceholder>[];
 
+    for (final doc in docs) {
+      final taskId = UploadTaskId(uuid.v4());
+      final cancelToken = CancelToken();
+
+      pendingTransfers.add(PendingDriveTransfer(
+        doc: doc,
+        taskId: taskId,
+        cancelToken: cancelToken,
+      ));
+      placeholders.add(DriveTransferPlaceholder(
+        taskId: taskId,
+        fileName: doc.name,
+        fileSize: doc.size,
+        mimeType: doc.mimeType,
+        cancelToken: cancelToken,
+      ));
+    }
+
+    uploadController.addDownloadingPlaceholders(placeholders);
     return pendingTransfers;
   }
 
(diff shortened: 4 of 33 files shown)


lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _resolveAuthHeader() ?? '',
+        onDownloadProgress: (received, total) => _uploadController.updateDownloadProgress(
+          taskId: taskId,
+          received: received,
+          total: total,
+        ),
```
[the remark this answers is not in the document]
tddang-linagora: Handled in https://github.com/linagora/tmail-flutter/pull/4752

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -90,21 +90,34 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
         }
-        break;
+        return null;
       case AuthenticationType.oidc:
         if (_token != null && _token?.isTokenValid() == true) {
-          options.headers[HttpHeaders.authorizationHeader] = _getTokenAsBearerHeader(_token!.token);
+          return _getTokenAsBearerHeader(_token!.token);
         }
-        break;
+        return null;
```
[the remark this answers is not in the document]
tddang-linagora: Out of scope for this PR

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +213,96 @@ class UploadController extends BaseController {
...
+  /// A drive transfer is one operation to the user, so its chip fills a single
+  /// bar across both legs: staging maps onto 0-50, uploading onto 50-100.
+  static const int _downloadProgressCeiling = 50;
+
+  static int _mapProgress(int processed, int total, {required int from, required int to}) {
+    if (total <= 0) return from;
+    final ratio = processed / total;
+    final mapped = from + (ratio * (to - from)).round();
+    return mapped.clamp(from, to);
+  }
```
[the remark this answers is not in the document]
tddang-linagora: from, to and mapped are all int, so clamp will return int.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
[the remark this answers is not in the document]
tddang-linagora: New commit pushed to #4752

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,19 +41,29 @@ class DriveAttachmentHandler {
...
     }).toList();
-    // TODO: Update logic here after implement 103. Attach Drive File as Attachment
-    if (linkDocs.isEmpty) {
+    // Exclusive with linkDocs by construction: a document carrying both links
+    // is a link, because a live document is worth more than a static copy.
+    final downloadableDocs = result
+        .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
+        .toList();
+
+    if (linkDocs.isNotEmpty) {
```
hoangdat, line 51: why not return here when finish insert link ?

integration_test/scenarios/search/search_email_with_tag_scenario.dart
```diff
@@ -50,8 +50,6 @@ class SearchEmailWithTagScenario extends BaseTestScenario
 
       await commonRobot.selectContextMenuItemByName(labelDisplayName);
       await searchRobot.expectEmailListCountAtLeast(emailCount);
-
-      await $.pumpAndSettle(duration: const Duration(seconds: 1));
```
dab246, line 54: Should create new PR to fix it.

lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
```diff
@@ -0,0 +1,24 @@
...
+  int maxConcurrent,
+  Future<void> Function(T item) task,
+) async {
+  final iterator = items.iterator;
+
+  Future<void> worker() async {
+    while (iterator.moveNext()) {
+      await task(iterator.current);
+    }
+  }
```
dab246: we avoid use nested function

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -14,9 +14,18 @@ class DriveAttachmentHandler {
...
+  ///
+  /// [startDriveTransfers] downloads and attaches the documents that carry
+  /// only a `downloadLink`, returning whether it took them. It declines on
+  /// platforms with no staging strategy, and those documents then fall back to
+  /// the "not available yet" message — the same behaviour as before drive
+  /// transfers existed.
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
     required Future<void> Function(String html) insertHtml,
+    required Future<bool> Function(List<DriveDocument> docs) startDriveTransfers,
```
dab246: Use typedef to strong type function

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +213,96 @@ class UploadController extends BaseController {
     _refreshListUploadAttachmentState();
   }
 
+  /// A drive transfer is one operation to the user, so its chip fills a single
+  /// bar across both legs: staging maps onto 0-50, uploading onto 50-100.
+  static const int _downloadProgressCeiling = 50;
+
+  static int _mapProgress(int processed, int total, {required int from, required int to}) {
```
dab246: Why do we need static here

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -90,21 +90,34 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    }
+  }
+
+  @override
+  void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
+    final authorizationHeader = currentAuthorizationHeader;
+    if (authorizationHeader != null) {
+      options.headers[HttpHeaders.authorizationHeader] = authorizationHeader;
     }
```
dab246, line 121: Why do we need this?

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,206 @@
...
+    }
+  }
+
+  /// Uploads a staged file whose bytes live in memory or on disk through the
+  /// shared [FileUploader]. The OPFS strategy never calls this — it streams
+  /// its own raw XHR so the file never reaches the Dart heap.
+  Future<Attachment> _uploadStagedFile({
+    required StagedDriveFile staged,
+    required Uri uploadUri,
+    required void Function(int sent, int total) onUploadProgress,
```
dab246: strong type it

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,206 @@
...
+      ));
+      _uploadController.completeUploadedFile(
+        taskId: taskId,
+        attachment: attachment,
+      );
+    } catch (error) {
+      // A failing file drops its own chip and nothing else: an expired link or
+      // a cancelled transfer must not take its siblings down with it.
+      logWarning('DriveTransferPipeline::_transferOne(${doc.name}): $error');
+      _uploadController.deleteFileUploaded(taskId);
```
dab246: If an error occurs here, aren't we going to notify the user or handle it in the UI?

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4751

integration_test/scenarios/search/search_email_with_tag_scenario.dart
tddang-linagora: Done in #4754 , prefer to keep the fix here too.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,206 @@
...
+  final Uri? Function() _resolveUploadUri;
+  final String? Function() _resolveAuthHeader;
+
+  /// Held for this pipeline's lifetime: the factory caches its capability
+  /// detection and runs the stale-staging sweep once per instance.
+  ///
+  /// Not `const`: only the mobile branch of the conditional export has a const
+  /// constructor, which is the only branch the analyzer sees.
+  // ignore: prefer_const_constructors
+  final DriveTransferStrategyFactory _strategyFactory = DriveTransferStrategyFactory();
```
hoangdat: why not inject it via getx?

lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: The OPFS drive-transfer upload streams the file with a raw `XMLHttpRequest`, not Dio, so no interceptor runs on it — it sets the header itself at `workplace/lib/data/datasource/drive_transfer/opfs_xhr_upload.dart:131` and reads the value from here via `composer_controller.dart:208` → `DriveTransferPipeline._resolveAuthHeader`. Exposing the getter keeps basic/OIDC header construction in one place; `onRequest` just applies whatever it returns instead of duplicating the switch.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Updated

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -14,9 +20,18 @@ class DriveAttachmentHandler {
...
+  /// [startDriveTransfers] downloads and attaches the documents that carry
+  /// only a `downloadLink`, returning whether it took them. It declines on
+  /// platforms with no staging strategy, and those documents then fall back to
+  /// the "not available yet" message — the same behaviour as before drive
+  /// transfers existed.
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
-    required Future<void> Function(String html) insertHtml,
+    required InsertHtmlCallback insertHtml,
+    required StartDriveTransfersCallback startDriveTransfers,
```
hoangdat: as I understand, this will handle `transfer` not only start point, so how about

```suggestion
    required TransferDriveDocuments transferDriveDocuments,
```

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,24 +47,34 @@ class DriveAttachmentHandler {
...
+    final downloadableDocs = result
+        .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
+        .toList();
+
+    if (linkDocs.isNotEmpty) {
+      await insertDriveLinkHtml(
+        linkDocs,
+        insertHtml: insertHtml,
+        appLocalizations: appLocalizations,
+      );
```
hoangdat, line 56: how about explicit return here to make it easier to understand ?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
dab246: Please make sure it not side effect to old logic. Should write unit test this

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,220 @@
...
+    required CancelToken cancelToken,
+  }) async {
+    final fileInfo = switch (staged) {
+      BytesStagedFile(:final bytes) => FileInfo.fromBytes(
+          bytes: bytes,
+          name: staged.fileName,
+          size: staged.fileSize,
+          type: staged.mimeType,
+        ),
+      FileBackedStagedFile(:final filePath) => FileInfo(
```
hoangdat, line 181: I still not find the validator execute for the case of file or byte?

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Should in the future drive allow users to select multiple pick purposes in one pick session, we don't need to return here to edit the condition.

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,102 @@
...
+      await runWithConcurrency(
+        List.generate(10, (index) => index),
+        3,
+        (item) async {
+          processed.add(item);
+        },
+      );
+
+      expect(processed..sort(), List.generate(10, (index) => index));
+    });
```
hoangdat:
```suggestion
    test('Should process every item exactly once', () async {
      final processed = <int>[];
    
      await runWithConcurrency(
        List.generate(100, (index) => index),
        5,
        (item) async {
          processed.add(item);
        },
      );
    
      expect(processed, hasLength(100));
      expect(processed.toSet(), hasLength(100));
      expect(
        processed.toSet(),
        containsAll(List.generate(100, (index) => index)),
      );
    });
```

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,102 @@
...
+
+    test('Should never exceed maxConcurrent tasks in flight', () async {
```
hoangdat, line 58: change to

```dart
test('Should never exceed maxConcurrent tasks in flight', () async {
  const maxConcurrent = 3;
  const itemCount = 9;

  var inFlight = 0;
  var peakInFlight = 0;

  final completers = <Completer<void>>[];

  final run = runWithConcurrency(
    List.generate(itemCount, (index) => index),
    maxConcurrent,
    (item) async {
      inFlight++;
      peakInFlight = inFlight > peakInFlight ? inFlight : peakInFlight;

      final completer = Completer<void>();
      completers.add(completer);

      await completer.future;

      inFlight--;
    },
  );

  // Allow the first wave of workers to start.
  await Future<void>.delayed(Duration.zero);

  expect(completers, hasLength(maxConcurrent));
  expect(peakInFlight, maxConcurrent);

  // Release the current tasks so the workers can pick up
  // the next items.
  while (completers.length < itemCount) {
    final pending = completers
        .where((completer) => !completer.isCompleted)
        .toList();

    for (final completer in pending) {
      completer.complete();
    }

    await Future<void>.delayed(Duration.zero);
  }

  // Release the final wave.
  for (final completer in completers) {
    if (!completer.isCompleted) {
      completer.complete();
    }
  }

  await run;

  expect(peakInFlight, lessThanOrEqualTo(maxConcurrent));
  expect(completers, hasLength(itemCount));
});
```

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,102 @@
...
+
+      await Future<void>.delayed(Duration.zero);
+      expect(started, [0], reason: 'the single slot is taken by the first item');
+
+      gate.complete();
+      await run;
+      expect(started, [0, 1, 2]);
+    });
+
+    test('Should keep other workers running when one task throws', () async {
```
hoangdat: replace it with

```dart
test('Should keep other workers running when one task throws', () async {
  final processed = <int>[];
  final worker1Started = Completer<void>();

  await expectLater(
    runWithConcurrency(
      [0, 1, 2, 3],
      2,
      (item) async {
        if (item == 0) {
          worker1Started.complete();
          throw StateError('boom');
        }

        if (item == 1) {
          processed.add(item);
          await worker1Started.future;
        } else {
          processed.add(item);
        }
      },
    ),
    throwsA(isA<StateError>()),
  );

  expect(processed, containsAll([1, 2, 3]));
});
```

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,102 @@
...
+      expect(processed, containsAll([1, 2, 3]));
+    });
+
+    test('Should treat a concurrency below one as a single worker', () async {
+      final processed = <int>[];
+
+      await runWithConcurrency([0, 1], 0, (item) async => processed.add(item));
+
+      expect(processed, [0, 1]);
+    });
```
hoangdat, line 405: add some tests 

```dart
test('Should complete when items are empty', () async {
  var taskCalled = false;

  await runWithConcurrency(
    <int>[],
    3,
    (item) async {
      taskCalled = true;
    },
  );

  expect(taskCalled, isFalse);
});
```

```dart
test('Should treat negative concurrency as a single worker', () async {
  final processed = <int>[];

  await runWithConcurrency(
    [0, 1, 2],
    -1,
    (item) async => processed.add(item),
  );

  expect(processed, [0, 1, 2]);
});
```

```dart 
test('Should process all items when maxConcurrent exceeds item count', () async {
  final processed = <int>[];

  await runWithConcurrency(
    [0, 1, 2],
    100,
    (item) async => processed.add(item),
  );

  expect(processed, containsAllInOrder([0, 1, 2]));
  expect(processed, hasLength(3));
});
```

lib/features/upload/presentation/model/upload_file_status.dart
```diff
@@ -1,6 +1,9 @@
 
 enum UploadFileStatus {
   waiting,
+  /// Staging leg of a drive transfer, before its upload starts. Only drive
+  /// attachments enter this state.
+  downloading,
```
hoangdat:
```suggestion
  wait_to_download,
```

lib/features/composer/presentation/styles/attachment_progress_loading_composer_widget_style.dart
```diff
@@ -7,4 +7,8 @@ class AttachmentProgressLoadingComposerWidgetStyle {
 
   static const Color backgroundColor = AppColor.colorProgressLoadingBackground;
   static const Color progressColor = AppColor.primaryColor;
+
+  /// Marks the staging leg of a drive transfer, which fills the first half of
+  /// the same bar [progressColor] then finishes.
+  static const Color downloadProgressColor = AppColor.toastSuccessBackgroundColor;
```
hoangdat: keep the same value for both progress

lib/features/upload/presentation/controller/upload_controller.dart
dab246: not fixed

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
dab246: Not fixed

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
dab246: not fixed

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,220 @@
...
+  // ignore: prefer_const_constructors
+  final DriveTransferStrategyFactory _strategyFactory = DriveTransferStrategyFactory();
+
+  /// Returns whether this pipeline took responsibility for [docs]. `false`
+  /// means nothing was attempted — no staging strategy on this platform, or
+  /// no upload endpoint to send to — so the caller should fall back.
+  ///
+  /// A batch the user declines at the size gate still counts as taken: the
+  /// documents were handled and the user was told why nothing was attached.
+  Future<bool> transfer(List<DriveDocument> docs) async {
```
hoangdat, line 66: what is the error handling for some error case
- no upload uri
- stage failed
- upload failed

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,220 @@
...
+          received: received,
+          total: total,
+        ),
+        onUploadProgress: (sent, total) => _uploadController.updateUploadProgress(
+          taskId: taskId,
+          sent: sent,
+          total: total,
+        ),
+        cancelToken: cancelToken,
+      ));
```
hoangdat: I found it quite specific for Opfs, how about to keep it out the pipeline?

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,220 @@
...
+    DriveTransferStrategy<StagedDriveFile> strategy,
+  ) {
+    return runWithConcurrency(
+      docs,
+      strategy.maxConcurrentTransfers,
+      (doc) => _transferOne(doc, uploadUri, strategy),
+    );
+  }
+
+  Future<void> _transferOne(
```
hoangdat:
```suggestion
  Future<void> _transferOpfsFile(
```

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: There's no other static in this file as of the time of this comment

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: fixed in 59c6da18c0b4872d05f5becf0cf0df61a26075f2

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: fixed in 8e65e56fc5de7ba132180cd945f8cfeab3c5b630 , `failDriveTransfer` show toast.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,220 @@
...
+      fileSize: doc.size,
+      mimeType: doc.mimeType,
+      cancelToken: cancelToken,
+    );
+
+    try {
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _resolveAuthHeader() ?? '',
```
dab246: We should throw early, if authHeader is empty

hoangdat: drop the `?? ' '` to ensure we have the exactly issue from server to easier recover

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,24 +47,34 @@ class DriveAttachmentHandler {
       final link = doc.sharingLink;
       return link != null && (!requireHttps || link.isScheme('https'));
     }).toList();
-    // TODO: Update logic here after implement 103. Attach Drive File as Attachment
-    if (linkDocs.isEmpty) {
+    // Exclusive with linkDocs by construction: a document carrying both links
+    // is a link, because a live document is worth more than a static copy.
+    final downloadableDocs = result
+        .where((doc) => doc.sharingLink == null && doc.downloadLink != null)
```
hoangdat: should also check in case of `http` link

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: keep supporting mixed-batch capability, no early return here

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
hoangdat, line 66: how we will handle in case of user pick another file when the previous one in progress?

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -453,4 +539,15 @@ class UploadController extends BaseController {
...
     }
   }
+}
+
+/// A drive transfer is one operation to the user, so its chip fills a single
+/// bar across both legs: staging maps onto 0-50, uploading onto 50-100.
+const int _downloadProgressCeiling = 50;
+
+int _mapProgress(int processed, int total, {required int from, required int to}) {
+  if (total <= 0) return from;
```
hoangdat: from is the floor of the band. Download band is 0→50, so unknown length pins it at 0. Upload band is 50→100, so unknown length pins it at 50 — and if it goes non-computable partway through, the bar rewinds 80% → 50%.

Fix: 
```suggestion
  if (total <= 0) return current.clamp(from, to);
```

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Already dropped. That code moved into `DriveDocumentTransferRunner`: the batch is refused up front
when there is no session (`canAuthenticate`, `drive_document_transfer_runner.dart:34`, checked at
`drive_transfer_pipeline.dart:77`), and per file a null/empty header throws
`StateError('No authorization header for drive transfer')`
(`drive_document_transfer_runner.dart:54-57`) instead of sending a blank header. No `?? ''` left
on the auth path.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Handled. `addDownloadingPlaceholder` (`upload_controller.dart:219`) inserts the in-flight file into
`_uploadingStateFiles` carrying its declared size as `FileInfo`, so `regularAttachmentsTotalBytes`
(`upload_controller.dart:353`) counts it and the second pick's `validateBytes` is gated against the
first batch's bytes. Each batch runs its own `runWithConcurrency`, so a failing or cancelled file
never touches the other batch's files.
One accepted wrinkle, same class as the one documented at
`attachment_upload_validation_service.dart:20-25`: two overlapping batches each get
`strategy.maxConcurrentTransfers`, so peak parallelism can double. The send guard and the server
still enforce the size limit.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

lib/features/upload/presentation/model/upload_file_status.dart
tddang-linagora: Done

lib/features/composer/presentation/styles/attachment_progress_loading_composer_widget_style.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,100 @@
...
+    try {
+      final authHeader = _resolveAuthHeader();
+      if (authHeader == null || authHeader.isEmpty) {
+        throw StateError('No authorization header for drive transfer');
+      }
+
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: authHeader,
```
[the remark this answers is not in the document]
tddang-linagora: Out of scope

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,24 +45,39 @@ class DriveAttachmentHandler {
...
-      ));
-      return;
+    // Exclusive with linkDocs by construction: a document carrying both links
+    // is a link, because a live document is worth more than a static copy.
+    final downloadableDocs = result.where((doc) {
+      final downloadLink = doc.downloadLink;
+      return doc.sharingLink == null &&
+          downloadLink != null &&
+          (!requireHttps || downloadLink.isScheme('https'));
+    }).toList();
```
dab246: Should create extension for this.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,24 +45,39 @@ class DriveAttachmentHandler {
...
-    await insertDriveLinkHtml(
-      linkDocs,
-      insertHtml: insertHtml,
-      appLocalizations: appLocalizations,
-    );
+
+    final transfersStarted = downloadableDocs.isNotEmpty &&
+        await transferDriveDocuments(downloadableDocs);
+
+    if (linkDocs.isNotEmpty || transfersStarted) return;
```
dab246, line 62: `linkDocs.isNotEmpty` this condition is not necessary here.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -32,24 +45,39 @@ class DriveAttachmentHandler {
...
+
+    getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
+      Exception(),
+      message: appLocalizations?.driveAttachmentInDevelopment,
+    ));
```
dab246: To avoid silent failures, add a log here when `getBinding<ToastManager>()` is nullable.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,128 @@
...
+        _uuid = uuid,
+        _resolveAuthHeader = resolveAuthHeader;
+
+  final UploadController _uploadController;
+  final Uuid _uuid;
+  final ResolveAuthHeader _resolveAuthHeader;
+
+  /// Whether there is a session to upload with. Checked before a batch starts,
+  /// so no file is downloaded only to be rejected by the upload endpoint.
+  bool get canAuthenticate => _resolveAuthHeader()?.isNotEmpty == true;
```
dab246: should trim the result of `_resolveAuthHeader()` before checking `isNotEmpty`.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,128 @@
...
+
+  Future<void> run({
+    required DriveDocument doc,
+    required Uri uploadUri,
+    required DriveTransferStrategy<StagedDriveFile> strategy,
+  }) async {
+    final taskId = UploadTaskId(_uuid.v4());
+    final cancelToken = CancelToken();
+    var exceededDeclaredSize = false;
```
dab246: should `bool exceededDeclaredSize = false;`

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,128 @@
...
+
+    try {
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _requireAuthHeader(),
+        onDownloadProgress: (received, total) {
+          // The link is sending more than the backend declared, so the size
+          // the batch was validated against no longer holds: stop now.
+          if (doc.size > 0 && received > doc.size) {
```
dab246: what's about if `doc.size == 0`?

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,128 @@
...
+        error: error,
+      );
+    }
+  }
+
+  /// The `Authorization` header for the current session, or a thrown
+  /// [StateError] when there is none to send.
+  String _requireAuthHeader() {
+    final authHeader = _resolveAuthHeader();
+    if (authHeader == null || authHeader.isEmpty) {
```
dab246:
```suggestion
    if (authHeader == null || authHeader.trim().isEmpty) {
```

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
hoangdat: I still not found the fix for this comment

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,128 @@
...
+    required UploadController uploadController,
+    required Uuid uuid,
+    required ResolveAuthHeader resolveAuthHeader,
+  })  : _uploadController = uploadController,
+        _uuid = uuid,
+        _resolveAuthHeader = resolveAuthHeader;
+
+  final UploadController _uploadController;
+  final Uuid _uuid;
+  final ResolveAuthHeader _resolveAuthHeader;
```
dab246: Why pass parameters in and then reassign them in contructor? Why not just pass and use them directly?

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,174 @@
...
+
+  final AttachmentUploadValidationService _validationService;
+  final FileUploader _fileUploader;
+  final Uuid _uuid;
+  final DriveDocumentTransferRunner _transferRunner;
+
+  /// App-wide singleton from `CoreBindings`: it caches capability detection
+  /// and sweeps stale staging once, so it outlives any one composer.
+  final DriveTransferStrategyFactory _strategyFactory;
+  final ResolveUploadUri _resolveUploadUri;
```
dab246: Why pass parameters in and then reassign them in contructor? Why not just pass and use them directly?

```dart
DriveTransferPipeline({
    required this.validationService,
    required this.fileUploader,
    required this.uuid,
    required this.strategyFactory,
    required this.ransferRunner,
    required this.solveUploadUri,
  })
```

lib/main/bindings/core/core_bindings.dart
```diff
@@ -53,6 +55,14 @@ class CoreBindings extends Bindings {
     Get.put(DriveAttachmentHandler());
   }
 
+  /// One per app: the factory caches capability detection and sweeps stale
+  /// staging once per instance. Not `const`: only the mobile branch of the
+  /// conditional export has a const constructor.
+  void _bindingDriveTransferStrategyFactory() {
+    // ignore: prefer_const_constructors
+    Get.put(DriveTransferStrategyFactory());
+  }
```
dab246, line 64: If the app doesn't support picking files from Drive, do we still have to inject it?

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,179 @@
...
+/// so a task started by one completion cannot mutate it mid-iteration.
+void _completePending(List<Completer<void>> completers) {
+  for (final completer in List.of(completers)) {
+    if (!completer.isCompleted) completer.complete();
+  }
+}
+
+void main() {
+  group('runWithConcurrency::', () {
+    test('Should process every item exactly once', () async {
```
dab246, line 46: Try adding test cases with extremely large values ​​for `maxConcurrent` and `itemCounts`.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: In 9d430d85c4ba906491a126ddccea3ae7ad7598b3

hoangdat:
- [x] please scan memory profile before and after transfering drive to attachment

dab246:
- [x] Always display a warning when the total file size exceeds 10MB, while the maximum is 20MB.


https://github.com/user-attachments/assets/53c3b523-df56-4ce7-b7e2-11dc86a24f32


Screenshot 2026-08-14 at 12 23 56

dab246:
- [x] The app stutters, the Drive picker doesn't close immediately after selecting a file and adding it as an attachment.

dab246:
- [x] The files don't appear immediately after picking them from Drive Picker and adding them as attachments.


https://github.com/user-attachments/assets/c09ef874-fc12-4235-a5ef-7548280850b9

dab246:
- [x] There is no toast message after successfully attaching a file from Drive Picker.

tddang-linagora:
> * [x]  The app stutters, the Drive picker doesn't close immediately after selecting a file and adding it as an attachment.

The app only cares about intent message to close the dialog. It's not the app concern if drive delay the post message.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: It is load-bearing. `transfersStarted` is false whenever the pick has no download-only docs, so with `linkDocs.isNotEmpty` dropped, a links-only pick would fall through to the `driveAttachmentInDevelopment` toast right after `insertDriveLinkHtml` succeeded (`drive_attachment_handler.dart:54-65`). The test "Should not show the in-development toast when at least one doc has sharingLink" covers exactly that path.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Updated

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Updated

lib/features/composer/presentation/manager/drive_attachment_handler.dart
dab246: This is incorrect, because we checked earlier that if `linkDocs.isNotEmpty=true` it would have exited the previous logic prematurely and wouldn't have reached this logic.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Updated

lib/main/bindings/core/core_bindings.dart
tddang-linagora: The injection costs effectively nothing: the mobile branch is a `const` constructor returning `null` from `create()`, and the web branch only assigns two seam objects. OPFS capability detection and the stale-file sweep are both deferred into `create()` (`drive_transfer_strategy_factory_web.dart:36-44`), which only runs on a real drive pick. Making the binding conditional would break `Get.find<DriveTransferStrategyFactory>()` on mobile (`composer_controller.dart:208`), and making it lazy would put a lazily-resolved dependency inside the drive picker's dialog route — the route-linked auto-dispose case we already hit on this feature.

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

tddang-linagora:
>Always display a warning when the total file size exceeds 10MB, while the maximum is 20MB.

Same behavior in prod. If it's a problem, please open a github issue. Will not address in this PR.


https://github.com/user-attachments/assets/0be117c6-9a5b-4b3d-acf4-f204f56fc0ca

tddang-linagora:
>The files don't appear immediately after picking them from Drive Picker and adding them as attachments.

Current concurrency of drive file handling is 3. If you wish to increase it, please provide a number

tddang-linagora:
>There is no toast message after successfully attaching a file from Drive Picker.

Done


https://github.com/user-attachments/assets/31d11b68-2b5a-4701-b754-4790a1441c10

hoangdat:
- [x] files are still upload even composer close, we should cancel the work also, fix with the test 

https://github.com/user-attachments/assets/4f541032-cf75-4409-9e8c-b4a5c2837712

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +213,94 @@ class UploadController extends BaseController {
...
+        uploadingProgress: _mapProgress(
+          received,
+          total,
+          current: currentState.uploadingProgress,
+          from: 0,
+          to: _downloadProgressCeiling,
+        ),
+      ),
+    );
+    _refreshListUploadAttachmentState();
```
hoangdat: GIVEN a multi-MB drive file WHEN chunks stream in THEN every chunk reassigns + refreshes the whole RxList, rebuilding all chips for one progress value. Throttle, or make the chip observe its own state?

tddang-linagora:
>please scan memory profile before and after transfering drive to attachment

No problem detected

Screenshot 2026-08-14 at 2 32 26 PM

dab246:
> > The files don't appear immediately after picking them from Drive Picker and adding them as attachments.
> 
> Current concurrency of drive file handling is 3. If you wish to increase it, please provide a number

We shouldn't place a limit on this; we should display exactly as many items as the user selects, while we decide internally how many are actually being processed. However, the user needs to know immediately that the correct quantity has been selected, so this must be reflected on the UI right away.

dab246:
> > * [x]   The app stutters, the Drive picker doesn't close immediately after selecting a file and adding it as an attachment.
> 
> The app only cares about intent message to close the dialog. It's not the app concern if drive delay the post message.

Ensure that clicking the "Attach file" button does not require waiting for any operations from `tmail-flutter`. Do not execute any `await` functions within its callback.

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
hoangdat, line 1: should have test for this file in some cases: 3 decline paths

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
hoangdat, line 1: add test too
- Oversize abort, cancel-vs-fail split, 
- StateError on missing header
- ...

tddang-linagora:
> > > * [x]    The app stutters, the Drive picker doesn't close immediately after selecting a file and adding it as an attachment.
> > 
> > 
> > The app only cares about intent message to close the dialog. It's not the app concern if drive delay the post message.
> 
> Ensure that clicking the "Attach file" button does not require waiting for any operations from `tmail-flutter`. Do not execute any `await` functions within its callback.

Confirmed

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
```diff
@@ -76,6 +76,27 @@ class AttachmentUploadValidationService {
     );
   }
 
+  /// Validates a proposed upload known only by its byte totals — no
+  /// [FileInfo] and no [Attachment] yet — then runs [onAllowed] when the
+  /// upload may proceed. Used by the drive transfer, which gates a batch on
+  /// the sizes the backend declared before downloading anything.
+  Future<void> validateBytes({
```
hoangdat, line 83: write tests for this

tddang-linagora:
>files are still upload even composer close, we should cancel the work also, fix with the test

Done


https://github.com/user-attachments/assets/cfd32591-b5b8-46a0-849e-0963a200f1be

lib/features/upload/presentation/controller/upload_controller.dart

tddang-linagora:
> > > The files don't appear immediately after picking them from Drive Picker and adding them as attachments.
> > 
> > 
> > Current concurrency of drive file handling is 3. If you wish to increase it, please provide a number
> 
> We shouldn't place a limit on this; we should display exactly as many items as the user selects, while we decide internally how many are actually being processed. However, the user needs to know immediately that the correct quantity has been selected, so this must be reflected on the UI right away.

Done


https://github.com/user-attachments/assets/dd7281f1-1509-4d98-9500-42c99797009c

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4751

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: There is no early exit above. `drive_attachment_handler.dart:49-55` is an `if` with no `return`
— it inserts the link HTML and falls through to line 57, so line 60 is reached with
`linkDocs.isNotEmpty == true` on every pick that contains a link doc.

The condition is load-bearing: drop it and a link-only pick (`downloadableDocs` empty, so
`transfersStarted == false`) falls into `_showFailureToast(driveAttachmentInDevelopment)` at
line 62 after having just inserted its links. `drive_attachment_handler_test.dart:246` and `:92`
both fail without it.

Early return comment was reverted.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _requireAuthHeader(),
+        onDownloadProgress: (received, total) {
+          // The link is sending more than the backend declared, so the size
+          // the batch was validated against no longer holds: stop now.
+          // `doc.size` is 0 when the backend declared nothing, and the
+          // response's own total is then the only ceiling there is.
+          final declaredSize = doc.size > 0 ? doc.size : total;
+          if (declaredSize > 0 && received > declaredSize) {
```
dab246, line 112: When doc.size == 0, declaredSize falls back to the response's own total. Therefore, received > total cannot protect against an unexpectedly large file.

The batch-level size gate has the same problem because declaredTotalBytes sums the document sizes, so documents reporting 0 contribute nothing.

On browsers without OPFS, BufferedWebDriveTransferStrategy can therefore buffer the entire file into the Dart heap without an effective size limit, potentially causing an OOM

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,198 @@
...
+    if (strategy == null) return false;
+
+    final declaredTotalBytes =
+        docs.fold<int>(0, (total, doc) => total + doc.size);
+
+    await validationService.validateBytes(
+      // Drive documents are never inline, so both totals are the same sum.
+      proposedAllAttachmentBytes: declaredTotalBytes,
+      proposedRegularAttachmentBytes: declaredTotalBytes,
+      onAllowed: () => unawaited(_runBatch(docs, uploadUri, strategy)),
```
dab246, line 103: `runWithConcurrency()` ultimately relies on `Future.wait()`, so any uncaught worker exception propagates.

Although transferRunner.run() catches errors from its main body, addDownloadingPlaceholder() currently executes outside that try. Batch-level operations such as counting results and _showSuccessToast() are also outside an error boundary.

Any exception from those paths can therefore become an unhandled root-zone error, producing Sentry noise while giving the user no feedback.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+  }) {
+    logWarning(
+      'DriveDocumentTransferRunner::run: transfer failed | '
+      'taskId=${taskId.id} | error=${_errorCategory(error)}',
+    );
+    if (cancelToken.isCancelled && !exceededDeclaredSize) {
+      // The user asked for this one to stop, so no error toast.
+      uploadController.deleteFileUploaded(taskId);
+    } else {
+      uploadController.failDriveTransfer(taskId);
```
dab246: `failDriveTransfer()` removes the failed item and shows an error toast for each file independently.

This means five failed transfers can produce five identical error toasts, while successful transfers are summarized into a single batch toast.

We should:

Count failures in `_runBatch()` alongside `attachedCount`, then display one batch-level failure toast. `failDriveTransfer()` should only update/remove the corresponding upload state.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+    required UploadTaskId taskId,
+    required CancelToken cancelToken,
+    required bool exceededDeclaredSize,
+    required Object error,
+  }) {
+    logWarning(
+      'DriveDocumentTransferRunner::run: transfer failed | '
+      'taskId=${taskId.id} | error=${_errorCategory(error)}',
+    );
+    if (cancelToken.isCancelled && !exceededDeclaredSize) {
```
dab246: When the downloaded content exceeds the document's declared size, the transfer ends with the same generic `can_not_upload_this_file_as_attachments` message used for ordinary network/upload failures. The user therefore cannot distinguish an oversize/mismatched file from a generic transfer failure.

We should:

Introduce a dedicated error message for the received size exceeds declared size case.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -42,13 +43,15 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
         sharingLink: WorkplaceActionConfigRequest(label: l10n.addAsLink),
-        downloadLink: addAsAttachmentTitle == null
-            ? null
-            : const WorkplaceActionConfigRequest(label: addAsAttachmentTitle),
+        // Web only: it is the one platform with a staging strategy, so
+        // offering the action elsewhere would only ever reach the
+        // "not available yet" message.
+        downloadLink: kIsWeb
+            ? WorkplaceActionConfigRequest(label: l10n.addAsAttachment)
+            : null,
```
dab246: Whether the platform supports Drive file staging is currently decided independently in two places:

1. `drive_picker_state_mixin.dart` via `kIsWeb`
2. `DriveTransferStrategyFactory.create()` via whether it returns a strategy or null

When mobile staging support is added, both locations must be updated together. Otherwise, the UI can expose an unsupported action or hide an action that is already supported.

Suggestion:

Expose a single capability such as:
```dart
bool get canStage;
```
from the strategy factory/platform implementation and let the picker use that source of truth.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -14,42 +21,66 @@ class DriveAttachmentHandler {
...
+      );
+    }
+
+    final transfersStarted = downloadableDocs.isNotEmpty &&
+        await transferDriveDocuments(downloadableDocs);
+
+    if (linkDocs.isNotEmpty || transfersStarted) return;
+
+    _showFailureToast(appLocalizations?.driveAttachmentInDevelopment);
+  }
```
dab246: Should write:

```dart
  final linkDocs = <DriveDocument>[];
  final downloadableDocs = <DriveDocument>[];

  for (final doc in result) {
    if (doc.isAttachableAsLink(requireHttps: requireHttps)) {
      linkDocs.add(doc);
    }

    if (doc.isAttachableAsDownload(requireHttps: requireHttps)) {
      downloadableDocs.add(doc);
    }
  }

  final transferFuture = downloadableDocs.isNotEmpty
      ? transferDriveDocuments(downloadableDocs)
      : null;

  if (linkDocs.isNotEmpty) {
    await insertDriveLinkHtml(
      linkDocs,
      insertHtml: insertHtml,
      appLocalizations: appLocalizations,
    );
  }

  final transfersStarted = await transferFuture ?? false;

  if (linkDocs.isEmpty && !transfersStarted) {
    _showFailureToast(appLocalizations?.driveAttachmentInDevelopment);
  }
```

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+              taskId: pending.taskId,
+              fileName: pending.doc.name,
+              fileSize: pending.doc.size,
+              mimeType: pending.doc.mimeType,
+              cancelToken: pending.cancelToken,
+            ))
+        .toList());
+
+    return pendingTransfers;
+  }
```
dab246, line 81:
```suggestion
List<PendingDriveTransfer> enqueue(List<DriveDocument> docs) {
  if (docs.isEmpty) return const [];

  final pendingTransfers = <PendingDriveTransfer>[];
  final placeholders = <DriveTransferPlaceholder>[];

  for (final doc in docs) {
    final taskId = UploadTaskId(uuid.v4());
    final cancelToken = CancelToken();

    pendingTransfers.add(
      PendingDriveTransfer(
        doc: doc,
        taskId: taskId,
        cancelToken: cancelToken,
      ),
    );

    placeholders.add(
      DriveTransferPlaceholder(
        taskId: taskId,
        fileName: doc.name,
        fileSize: doc.size,
        mimeType: doc.mimeType,
        cancelToken: cancelToken,
      ),
    );
  }

  uploadController.addDownloadingPlaceholders(placeholders);

  return pendingTransfers;
}
```

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+    required this.resolveAuthHeader,
+  });
+
+  final UploadController uploadController;
+  final Uuid uuid;
+  final ResolveAuthHeader resolveAuthHeader;
+
+  /// Whether there is a session to upload with. Checked before a batch starts,
+  /// so no file is downloaded only to be rejected by the upload endpoint.
+  bool get canAuthenticate => resolveAuthHeader()?.trim().isNotEmpty == true;
```
dab246: Should improve some functions:

```dart
bool get canAuthenticate => _resolveAuthHeader() != null;

String? _resolveAuthHeader() {
  final authHeader = resolveAuthHeader()?.trim();
  return authHeader?.isNotEmpty == true ? authHeader : null;
}

String _requireAuthHeader() {
  return _resolveAuthHeader() ??
      (throw StateError('No authorization header for drive transfer'));
}
```

to ensure the logic maintains a single source of truth

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+    bool exceededDeclaredSize = false;
+
+    // Cancelled while queued, so there is nothing left to download.
+    if (cancelToken.isCancelled) return false;
+
+    try {
+      final attachment = await strategy.transfer(DriveTransferRequest(
+        doc: doc,
+        uploadUri: uploadUri,
+        authHeader: _requireAuthHeader(),
```
dab246, line 105: The auth header is being resolved too early.  If `strategy.transfer()` executes:
```text
download large file
      ↓
stage file
      ↓
upload
```
then the token is retrieved before the download. For example:
```text
17:00  resolve token
17:00  start 2 GB download
17:20  download finished
17:20  upload using token from 17:00
                      ↑
                   expired
```
This isn't purely a performance issue, it's a case of wasted I/O, the user downloads the entire file, only for the upload to fail with a 401 error.

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+    required Uri uploadUri,
+    required DriveTransferStrategy<StagedDriveFile> strategy,
+  }) async {
+    final doc = pending.doc;
+    final taskId = pending.taskId;
+    final cancelToken = pending.cancelToken;
+    bool exceededDeclaredSize = false;
+
+    // Cancelled while queued, so there is nothing left to download.
+    if (cancelToken.isCancelled) return false;
```
dab246: Should add defensive like this:

```dart
if (cancelToken.isCancelled) {
  uploadController.deleteFileUploaded(taskId);
  return false;
}
```

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+  void _handleTransferFailure({
+    required UploadTaskId taskId,
+    required CancelToken cancelToken,
+    required bool exceededDeclaredSize,
+    required Object error,
+  }) {
+    logWarning(
+      'DriveDocumentTransferRunner::run: transfer failed | '
+      'taskId=${taskId.id} | error=${_errorCategory(error)}',
+    );
```
dab246: Meaning the user voluntarily cancels:

```text
Cancel
  ↓
strategy throws cancellation
  ↓
catch
  ↓
logWarning("transfer failed")
```

Expected user actions should not trigger warnings. If many transfers are cancelled, logs or Sentry could become flooded with "failures" that aren't actually failures.

Reverse the order:

```dart
void _handleTransferFailure({
  required UploadTaskId taskId,
  required CancelToken cancelToken,
  required bool exceededDeclaredSize,
  required Object error,
}) {
  if (cancelToken.isCancelled && !exceededDeclaredSize) {
    uploadController.deleteFileUploaded(taskId);
    return;
  }

  logWarning(
    'DriveDocumentTransferRunner::run: transfer failed | '
    'taskId=${taskId.id} | error=${_errorCategory(error)}',
  );

  uploadController.failDriveTransfer(taskId);
}
```

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+    } else {
+      uploadController.failDriveTransfer(taskId);
+    }
+  }
+
+  /// A stable, PII-free label for a transfer failure: no document name, no
+  /// download URI, no error message.
+  String _errorCategory(Object error) => error is DioException
+      ? 'DioException.${error.type.name}(${error.response?.statusCode})'
+      : error.runtimeType.toString();
```
dab246:
```suggestion
String _errorCategory(Object error) {
  if (error is! DioException) {
    return error.runtimeType.toString();
  }

  final category = 'DioException.${error.type.name}';
  final statusCode = error.response?.statusCode;

  return statusCode == null
      ? category
      : '$category($statusCode)';
}
```

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
```diff
@@ -0,0 +1,170 @@
...
+            ))
+        .toList());
+
+    return pendingTransfers;
+  }
+
+  /// Returns whether the document became an attachment: `false` covers both a
+  /// failed transfer and one the user cancelled, so a batch only reports what
+  /// actually landed.
+  Future<bool> run({
```
dab246: One architectural change I particularly recommend get rid of `bool`. 
With the contract:
```text
true  = attached
false = failed OR cancelled
```
This used to make sense because the batch only needed to know: `attachedCount`. The class currently has the following requirement:
```text
failure → toast
cancel → silent
``
and if handle the previous finding: `Failure toasts are per-file, success toast is per-batch` then the pipeline needs to know:

```text
attached
failed
cancelled
```
At that point, a boolean was no longer sufficient. Suggestion use enum

```dart
enum DriveTransferResult {
  attached,
  failed,
  cancelled,
}

Future<DriveTransferResult> run(...)
```
Updated:

```dart
if (cancelToken.isCancelled) {
  return DriveTransferResult.cancelled;
}

Success => return DriveTransferResult.attached;
Catch => return _handleTransferFailure(...)
    ? DriveTransferResult.cancelled
    : DriveTransferResult.failed;
```

The pipeline can then:

```dart
var attachedCount = 0;
var failedCount = 0;

switch (result) {
  case DriveTransferResult.attached:
    attachedCount++;
  case DriveTransferResult.failed:
    failedCount++;
  case DriveTransferResult.cancelled:
    break;
}
```

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
```diff
@@ -0,0 +1,198 @@
...
+            staged: staged,
+            uploadUri: uploadUri,
+            onUploadProgress: onUploadProgress,
+            cancelToken: cancelToken,
+          ),
+    );
+    if (strategy == null) return false;
+
+    final declaredTotalBytes =
+        docs.fold<int>(0, (total, doc) => total + doc.size);
```
hoangdat, line 97: I also worry about the `doc.size = 0` here, it will bypass our validation, how about we prevent it here?

hoangdat: IMO, we can check doc.size <= 0 directly in runner to drop the place holder and show toast consistent with other fail. WDYT?

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
hoangdat, line 1: I reorganized the test and add some test reflect the error contract also in file, please add and improve the error handling for this class 

[bounded_concurrency_runner_test.dart.txt](https://github.com/user-attachments/files/31071722/bounded_concurrency_runner_test.dart.txt)

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: That's why we have #4752

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: We have #4752 for that

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: planned in #4755

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: should be in #4752

lib/features/composer/presentation/manager/drive_transfer_pipeline.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_document_transfer_runner.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

tddang-linagora:
Close by #4760

PR #4756, opened 2026-08-13
feat(labels): render JMAP standard keywords as label chips

mmaudet:
## Summary

Any [RFC 8621](https://datatracker.ietf.org/doc/html/rfc8621) custom keyword set via `Email/set { keywords/… : true }` by an external tool (mail sentinel, filter, IMAP/JMAP client) is currently silently discarded from the Twake Mail UI — only keywords with a matching server-side `Label` object (`com:linagora:params:jmap:labels`) render as chips.

This PR extends `PresentationEmail.getLabelList(labels)` so that JMAP standard keywords render as read-only label chips even without a registered `Label` object, complementing the existing proprietary Linagora label system without any change to it.

Motivation: I built a companion mail sentinel that tags incoming mail via standard JMAP (`Email/set keywords/newsletter: true`, `keywords/github-notif: true`, etc.). The keywords were correctly stored server-side and returned in `Email/get`, but the UI never surfaced them — the operator had no visibility into the sentinel's classifications.

## What the PR does

1. **Renders orphan keywords as label chips.** In `getLabelList()`, after the existing registered-label collection, synthesise a `Label` chip for every enabled keyword that (a) is not covered by a registered `Label` and (b) is not in the [system-keyword deny-list](#deny-list). The synthesised chip has `id: null`, `displayName: keyword.value`, `color: HexColor(<deterministic hash-based color>)`.

2. **Deterministic hash-based color** from a 12-color palette (subset of the tmail label picker's palette, filtered for good contrast against white chip text). Same keyword string → same color across sessions and devices. No server round-trip, no storage. FNV-1a-ish hash on the keyword's UTF-16 code units modulo the palette length.

3. **System-keyword deny-list** (`KeyWordIdentifierExtension.systemKeywords`) as a single source of truth: RFC 8621 system keywords (`$seen`, `$flagged`, `$answered`, `$forwarded`, `$draft`, `$junk`, `$notjunk`, `$phishing`, `$mdnsent`) + tmail-custom UI-mapped keywords (`\$unsubscribe`, `needs-action`, `event`). System keywords keep their dedicated UI treatment (star icon, read badge, spam-folder move, etc.) and are never rendered as chips.

4. **Interaction with the Linagora label extension** is fully preserved. Registered `Label` objects keep their user-chosen `displayName` and `color`; they take precedence over orphan-chip synthesis for the same keyword. The Linagora capability gate (`isLabelAvailable`) continues to gate the *management* UI unchanged — the orphan-chip *display* path is independent and works on any JMAP server.

5. **All four `getLabelList` callsites benefit automatically** — thread view, single email view, thread detail, search. The synthesis lives in the shared extension in `lib/features/email/presentation/extensions/presentation_email_extension.dart`; no per-callsite change was needed.

## Backwards compatibility

- **Server side:** unchanged. `Email.keywords` is already fetched in every `Email/get` call (`ThreadConstants`).
- **Existing `Label` objects:** untouched — their color and display name still win over any orphan synthesis for the same keyword.
- **System keywords:** always suppressed from chip rendering — dedicated UI icons remain the sole visual representation.
- **Older client behavior:** clients that predate this change continue to ignore orphan keywords. Rolling back the change is a UI-only revert.

## Files changed

| Path | Change |
| :--- | :--- |
| `model/lib/extensions/keyword_identifier_extension.dart` | + `systemKeywords` deny-list, + `isSystemKeyword` getter, + `deterministicHexColor` getter |
| `lib/features/email/presentation/extensions/presentation_email_extension.dart` | `getLabelList()` extended to synthesise orphan chips |
| `test/model/lib/extensions/get_label_list_in_keyword_email_test.dart` | +5 new tests for orphan synthesis, +2 for color determinism |
| `docs/dev/jmap-standard-keywords.md` | New — feature + API + color docs for integrators |

**519 insertions, 4 deletions** across 4 files.

## Tests

Unit tests in `test/model/lib/extensions/get_label_list_in_keyword_email_test.dart` cover:

- orphan keyword surfaces as a chip when no matching `Label` exists,
- the full RFC 8621 system keyword set is filtered out,
- registered `Label` wins over orphan synthesis for the same keyword,
- mixed registered + orphan + system-filtered input produces the right chip set,
- disabled keywords (`keywords/foo: false`) are excluded,
- deterministic color — same keyword string always resolves to the same hex value,
- basic well-formedness (`#RRGGBB`) of the assigned color.

Run:
```bash
flutter test test/model/lib/extensions/get_label_list_in_keyword_email_test.dart
```

## Manual verification

Built with `Dockerfile` unchanged (`FLUTTER_VERSION=3.38.9`) and deployed to a dev Twake Mail instance. Confirmed with the browser:

1. Set several test keywords via JMAP `Email/set` (`jmap-standard-test`, `newsletter`, `action-required`, `urgent`, `project-alpha`).
2. Reloaded the Twake Mail web UI.
3. Verified each keyword rendered as a distinct chip with a deterministic color, on both the thread view and the email detail view.
4. Verified `$seen`, `$flagged`, `$answered` on the same emails did NOT render as chips.
5. Verified an existing user-created `Label` with matching keyword still won (kept its custom color / display name).

## Design doc

See `docs/dev/jmap-standard-keywords.md` for the full feature + API documentation (audience: integrators building tools that read from or write to a Twake Mail account via JMAP).

## Notes

- Marked as **draft** — happy to iterate on any of the design choices before flipping to ready-for-review.
- The palette is a subset of the existing tmail color picker for visual consistency with user-created labels.
- Read-only chips only — removing an orphan keyword still requires an `Email/set { keywords/foo: null }` call or registering a matching `Label` and using the standard remove UI. Adding an in-chip remove action for orphans is a follow-up if desired.

Related: this unblocks external JMAP-compliant mail tooling (mail sentinels, filters, standards-based clients) from surfacing their tags in Twake Mail.

## Summary by CodeRabbit

* **New Features**
  * Custom email keywords now appear as visual label chips, even when no registered label exists.
  * Unregistered keyword labels receive consistent colors across the app.
  * Registered labels take precedence, while system keywords remain excluded.
  * Disabled keywords are omitted from displayed labels.

* **Documentation**
  * Added guidance covering keyword labels, compatibility, rollout behavior, limitations, and testing.

### DIFF
--- a/lib/features/email/presentation/extensions/presentation_email_extension.dart
+++ b/lib/features/email/presentation/extensions/presentation_email_extension.dart
@@ -3,6 +3,7 @@ import 'package:jmap_dart_client/jmap/mail/email/keyword_identifier.dart';
 import 'package:labels/model/label.dart';
 import 'package:model/email/email_action_type.dart';
 import 'package:model/email/presentation_email.dart';
+import 'package:model/extensions/keyword_identifier_extension.dart';
 import 'package:model/extensions/list_email_address_extension.dart';
 import 'package:tmail_ui_user/features/email/presentation/utils/email_utils.dart';
 import 'package:tmail_ui_user/features/thread/data/extensions/map_keywords_extension.dart';
@@ -158,16 +159,37 @@ extension PresentationEmailExtension on PresentationEmail {
   }
 
   List<Label> getLabelList(List<Label> labels) {
-    if (keywords?.isNotEmpty != true || labels.isEmpty) return const [];
+    if (keywords?.isNotEmpty != true) return const [];
 
     final enabledKeywords = keywords!.enabledKeywords;
-
     if (enabledKeywords.isEmpty) return const [];
 
-    return labels
+    // 1) Registered labels: keywords that have a server-side `Label` object
+    //    (Linagora `com:linagora:params:jmap:labels` extension) — carry
+    //    displayName + color set by the user.
+    final registered = labels
         .where((label) =>
             label.keyword != null && enabledKeywords.contains(label.keyword))
         .toList();
+
+    // 2) Orphan / standard JMAP keywords: any keyword set via the RFC 8621
+    //    `Email/set keywords` API by an external tool (mail sentinel,
+    //    filter, IMAP client) that is NOT already a registered Label AND is
+    //    NOT a UI-mapped system keyword (star, seen, draft, junk, etc.).
+    //    Rendered as read-only chips with the default primary color so
+    //    keywords from any standards-compliant JMAP producer become visible.
+    final registeredKeywords = <String>{
+      for (final label in registered)
+        if (label.keyword != null) label.keyword!.value,
+    };
+    final orphanLabels = <Label>[];
+    for (final keyword in enabledKeywords) {
+      if (keyword.isSystemKeyword) continue;
+      if (registeredKeywords.contains(keyword.value)) continue;
+      orphanLabels.add(Label(keyword: keyword, displayName: keyword.value));
+    }
+
+    return [...registered, ...orphanLabels];
   }
 
   PresentationEmail toggleKeyword(KeyWordIdentifier keyword, bool remove) {
--- a/lib/features/email/presentation/extensions/presentation_email_extension.dart
+++ b/lib/features/email/presentation/extensions/presentation_email_extension.dart
@@ -1,5 +1,6 @@
 import 'package:jmap_dart_client/jmap/mail/email/email_address.dart';
 import 'package:jmap_dart_client/jmap/mail/email/keyword_identifier.dart';
+import 'package:labels/model/hex_color.dart';
 import 'package:labels/model/label.dart';
 import 'package:model/email/email_action_type.dart';
 import 'package:model/email/presentation_email.dart';
@@ -186,7 +187,14 @@ extension PresentationEmailExtension on PresentationEmail {
     for (final keyword in enabledKeywords) {
       if (keyword.isSystemKeyword) continue;
       if (registeredKeywords.contains(keyword.value)) continue;
-      orphanLabels.add(Label(keyword: keyword, displayName: keyword.value));
+      orphanLabels.add(Label(
+        keyword: keyword,
+        displayName: keyword.value,
+        // Deterministic hash-based color from the tmail palette so each
+        // orphan keyword renders in a distinct, stable color across
+        // sessions and devices without any server-side storage.
+        color: HexColor(keyword.deterministicHexColor),
+      ));
     }
 
     return [...registered, ...orphanLabels];
--- a/docs/dev/jmap-standard-keywords.md
+++ b/docs/dev/jmap-standard-keywords.md
@@ -1,0 +1,324 @@
+# JMAP Standard Keywords as Label Chips
+
+> **Audience:** integrators and developers building tools that read from or
+> write to a Twake Mail account via JMAP (mail sentinels, filters, custom
+> IMAP/JMAP clients, backfill scripts).
+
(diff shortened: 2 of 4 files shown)


chibenwa:
As discussed on signal, raw flags are technical and not meant to be exposed directly.

As such I oppose firmly this work.

Let me propose two alternatives for your issue @mmaudet :

 - ease label création either with IMAP metadata extension or even automatically for your `keywords/...` keywords 

That's a backend listener and by far what I prefer: can change color, no double-concept leaking through all the stack, and we can "list" those labels.

- or restrict raw hemp keyword representation to a prefix eg it shall start with labels/... 

Functional downside here is: no color change, no listing of those labels.

chibenwa:
TLDR

@mmaudet if Linagora but also the thunderbird team came along with this "label" concept instead of raw IMAP keyword uses it is for very good reason.

Our model is very versatile and expressive. I have no doubt we succeed to extend it in a way that allows you to fully leverage it.

I thus remain fully at your disposal to discuss this topic

hoangdat:
hi, very nice to see your contribution

it quite work on my side

image

But I am not really understand the case where we have the function. I tried to hack the function to add the keyword when sending an email but this keyword always be droped.

lib/features/email/presentation/extensions/presentation_email_extension.dart
```diff
@@ -158,16 +160,44 @@ extension PresentationEmailExtension on PresentationEmail {
   }
 
   List<Label> getLabelList(List<Label> labels) {
-    if (keywords?.isNotEmpty != true || labels.isEmpty) return const [];
+    if (keywords?.isNotEmpty != true) return const [];
 
     final enabledKeywords = keywords!.enabledKeywords;
-
     if (enabledKeywords.isEmpty) return const [];
```
dab246, line 166:
```suggestion
final enabledKeywords = keywords?.enabledKeywords;
if (enabledKeywords == null || enabledKeywords.isEmpty) {
  return const [];
}
```

lib/features/email/presentation/extensions/presentation_email_extension.dart
```diff
@@ -158,16 +160,44 @@ extension PresentationEmailExtension on PresentationEmail {
...
+      if (registeredKeywords.contains(keyword.value)) continue;
+      orphanLabels.add(Label(
+        keyword: keyword,
+        displayName: keyword.value,
+        // Deterministic hash-based color from the tmail palette so each
+        // orphan keyword renders in a distinct, stable color across
+        // sessions and devices without any server-side storage.
+        color: HexColor(keyword.deterministicHexColor),
+      ));
+    }
```
dab246, line 198:
```suggestion
List<Label> getLabelList(List<Label> labels) {
  final enabledKeywords = keywords.enabledKeywordSet;
  if (enabledKeywords.isEmpty) {
    return const [];
  }

  final result = <Label>[];
  final registeredKeywords = <KeyWordIdentifier>{};

  // Keep server-side labels that are currently enabled.
  // These preserve user-defined metadata such as display name and color.
  for (final label in labels) {
    final keyword = label.keyword;
    if (keyword == null || !enabledKeywords.contains(keyword)) {
      continue;
    }

    result.add(label);
    registeredKeywords.add(keyword);
  }

  // Add enabled custom keywords that are not backed by a server-side Label.
  // System keywords are handled separately by the UI.
  for (final keyword in enabledKeywords) {
    if (keyword.isSystemKeyword || registeredKeywords.contains(keyword)) {
      continue;
    }

    result.add(
      Label(
        keyword: keyword,
        displayName: keyword.value,
        // Generate a stable color so orphan keywords remain visually
        // consistent across sessions without server-side metadata.
        color: HexColor(keyword.deterministicHexColor),
      ),
    );
  }

  return result;
}
```

PR #4766, opened 2026-08-19
Fix mobile Search action state synchronization

dab246:
## Issue

On mobile Search in v0.35.0, actions on selected emails did not update the visible result list immediately. The UI only refreshed after selecting the email again. The Archive action was also not handled from Search.

## Root cause

Search uses a dedicated Riverpod presentation state, but action-success reconciliation did not update that state correctly. On native mobile, it mutated the existing `PresentationEmail.keywords` map and re-published the same email instances, so no observable immutable state transition was emitted to rebuild the UI. On web mobile/tablet layouts, browser runtime detection routed updates to the mailbox GetX list instead of the Search state. Move, delete and clear reconciliation had the same state-owner gap. Archive was also missing from the single- and multi-selection Search action handlers.

## Solution

- Resolve the active presentation-state owner: Search state on native mobile and web mobile/tablet layouts; mailbox list on web desktop.
- Replace changed Search emails immutably and publish a new result list, rather than mutating existing email instances.
- Route move, delete and clear reconciliation to the active Search presentation state.
- Handle Archive for both single and multi-selected Search emails.
- Add regression tests for mobile, web-mobile, web-desktop, partial move success, delete/clear and Archive flows.

## Demo

https://github.com/user-attachments/assets/a11ad456-47f8-49ff-a640-a5cbf88a9ab5

https://github.com/user-attachments/assets/e97a8f6f-3d07-4413-a05c-a32d0db62d03

## New Features
- Added archive actions for individual emails and multiple selected emails.
- Search results now update immediately after read-status, flag, and archive actions.

## Bug Fixes
- Improved synchronization when emails are moved, deleted, cleared, or updated.
- Empty deletion requests no longer trigger unnecessary updates.
- Email changes now work consistently across mobile, tablet, desktop, and responsive layouts.
- Fixed selection mode behavior after archiving multiple emails.

### DIFF
--- a/integration_test/robots/abstract/abstract_common_robot.dart
+++ b/integration_test/robots/abstract/abstract_common_robot.dart
@@ -84,7 +84,9 @@ abstract class AbstractCommonRobot extends CoreRobot {
           .last;
     }));
 
-    // Refresh view after provisioning emails
+    // Web e2e browser storage can retain mailbox data from a preceding test.
+    // A normal refresh may return that cache while the server-side provisioning
+    // change is still being reconciled, so force a fresh query on web only.
     if (refreshEmailView) {
       await threadController.refreshAllEmail(
         shouldClearCache: PlatformInfo.isWeb,
--- a/integration_test/robots/abstract/abstract_common_robot.dart
+++ b/integration_test/robots/abstract/abstract_common_robot.dart
@@ -1,4 +1,5 @@
 import 'package:get/get.dart';
+import 'package:core/utils/platform_info.dart';
 import 'package:jmap_dart_client/jmap/identities/identity.dart';
 import 'package:jmap_dart_client/jmap/mail/email/email_address.dart';
 import 'package:jmap_dart_client/jmap/mail/mailbox/mailbox.dart';
@@ -85,7 +86,9 @@ abstract class AbstractCommonRobot extends CoreRobot {
 
     // Refresh view after provisioning emails
     if (refreshEmailView) {
-      await threadController.refreshAllEmail();
+      await threadController.refreshAllEmail(
+        shouldClearCache: PlatformInfo.isWeb,
+      );
     }
 
     ComposerBindings().dispose();
--- a/integration_test/robots/web/web_search_robot.dart
+++ b/integration_test/robots/web/web_search_robot.dart
@@ -7,6 +7,8 @@ import 'package:tmail_ui_user/features/base/model/ui_keys.dart';
 import 'package:tmail_ui_user/features/base/widget/popup_menu/popup_menu_item_action_widget.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/widgets/quick_search/email_quick_search_item_tile_widget.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/widgets/search_input_form_widget.dart';
+import 'package:tmail_ui_user/features/search/email/presentation/search_email_view.dart';
+import 'package:core/presentation/views/search/search_bar_view.dart';
 import 'package:tmail_ui_user/features/thread/presentation/widgets/email_tile_web_builder.dart';
 
 import '../../extensions/patrol_finder_extension.dart';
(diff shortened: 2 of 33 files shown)


tddang-linagora (dismissed):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4766

lib/features/mailbox_dashboard/presentation/extensions/search_email_presentation_owner_extension.dart
```diff
@@ -0,0 +1,22 @@
+import 'package:core/utils/platform_info.dart';
+import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/features/search/email/presentation/notifier/search_email_presentation_notifier.dart';
+import 'package:tmail_ui_user/main/providers/app_provider_container.dart';
+import 'package:tmail_ui_user/main/routes/route_navigation.dart';
+
+extension SearchEmailPresentationOwnerExtension on MailboxDashBoardController {
+  bool get isSearchEmailPresentationOwner {
```
hoangdat, line 8: how about to reuse `SearchEmailController._ownsSearchResults` or `ThreadSearchExecutionObserver._ownsSearchResults` instead of write new flag?

hoangdat, line 8: Idea: one function to update whatever email list is on screen.

  We have two lists:

  • Inbox / web desktop search → emailsInCurrentMailbox
  • Phone / tablet search → listResultSearch

  Today each action (delete, mark-read, move) repeats: “if search, write list A, else write list B.” Easy to miss on the next action.

  Instead:

  applyToVisibleEmailList((emails) => /* change emails */);

  That function picks the list. Callers only describe the change:

```
  // delete
  applyToVisibleEmailList(
    (emails) => emails.where((e) => !deletedIds.contains(e.id)).toList(),
  );
  // mark read — same function, different transform
  applyToVisibleEmailList(
    (emails) => emails.map((e) => e.id == id ? e.markRead() : e).toList(),
  );
```

  Please: add applyToVisibleEmailList, use it in delete / flags / move, drop the repeated if (isSearchEmailPresentationOwner) blocks.

integration_test/robots/abstract/abstract_common_robot.dart
```diff
@@ -85,7 +86,9 @@ abstract class AbstractCommonRobot extends CoreRobot {
 
     // Refresh view after provisioning emails
     if (refreshEmailView) {
-      await threadController.refreshAllEmail();
+      await threadController.refreshAllEmail(
+        shouldClearCache: PlatformInfo.isWeb,
```
hoangdat, line 92: why need this?

integration_test/robots/web/web_search_robot.dart
```diff
@@ -7,22 +7,40 @@ import 'package:tmail_ui_user/features/base/model/ui_keys.dart';
...
   Future<void> tapOnSearchField() async {
+    if ($(SearchBarView).evaluate().isNotEmpty) {
+      // Responsive web uses the mobile search entry point below 1200px.
+      await super.tapOnSearchField();
+      await waitForCondition(
+        () async => $(SearchEmailView).evaluate().isNotEmpty,
+        timeout: TestTimeouts.short,
+      );
+      return;
+    }
```
hoangdat: how about to have `WebResponsiveSearchRobot` instead of using if/else?

lib/features/mailbox_dashboard/presentation/extensions/search_email_presentation_owner_extension.dart
dab246: Direct reuse was not possible because both existing ownership getters are private and represent complementary owners. I extracted their shared layout predicate into isSearchEmailPresentationLayoutOwner and now reuse it from SearchEmailController, ThreadSearchExecutionObserver, and the dashboard state-owner extension.

dab246: Added applyToVisibleEmailList to centralize selecting and publishing the visible email list. Flag, move, delete, and clear actions now only provide immutable list transforms. The helper preserves the existing mailbox guard for delete/clear through shouldApplyToMailboxList. Added coverage for Search ownership, mailbox ownership, and rejected mailbox guards.

integration_test/robots/abstract/abstract_common_robot.dart
dab246: This is e2e-only. On web, provisioning updates the server while the browser can still hold the previous email/state cache. refreshAllEmail() starts an asynchronous refresh and may initially read that cache, so the test can proceed without the newly provisioned email. Clearing the email/state cache forces the next refresh to query the server; this stabilizes the web attachment and search scenarios. Native does not reproduce this cache timing issue.

integration_test/robots/web/web_search_robot.dart
dab246: The current condition is repeated across the web search input flow and the active layout can change during the same test after a viewport resize. I’ll extract a WebResponsiveSearchRobot and centralize the runtime layout selection in one facade/selector, so responsive-specific interactions no longer leak into each desktop operation.

hoangdat (approved):

tddang-linagora (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4766

PR #4760, opened 2026-08-17
ADR Drive attachment by jmap

tddang-linagora:
## Summary by CodeRabbit

* **Documentation**
  * Rejected the previous client-side approach for attaching Drive files.
  * Added a proposal for attaching Drive files through a server-mediated upload flow.
  * Documented capability checks, upload limits, error handling, timeouts, logging, and bounded processing.
  * Clarified that links and files are handled separately, with files processed through authenticated uploads.
  * Documented a single cancellable indeterminate attachment state while processing.
  * Recorded the relationship between the rejected proposal and the new approach.

PR #4767, opened 2026-08-19
TF-4728 Drive picker jmap implementation

tddang-linagora:
## Issue
- #4728 

## ADR
- #4760

## Dependency
- https://github.com/linagora/tmail-backend/issues/2556

## Demo

https://github.com/user-attachments/assets/1a8a4e6f-1a9c-4588-96a0-8f8e9ab6ef54

## Summary by CodeRabbit

- **New Features**
  - Attach Drive documents directly in the composer.
  - Pending attachments appear immediately while transfers complete.
  - Uploads support cancellation and clearly show success or failure states.
  - Transfers are limited for improved reliability and performance.
  - Picker availability reflects attachment limits and remaining capacity.
  - Mixed selections support both link attachments and file transfers.
  - Successful Drive attachments display a confirmation message.

- **Bug Fixes**
  - Improved validation blocks unsupported or unsafe links.
  - Transfer and insertion failures show localized feedback.
  - Upload cancellation and cleanup are handled more reliably.

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -104,6 +104,7 @@ import 'package:tmail_ui_user/features/email/domain/usecases/print_email_interac
 import 'package:tmail_ui_user/features/email/domain/usecases/save_template_email_interactor.dart';
 import 'package:tmail_ui_user/features/email/domain/usecases/transform_html_email_content_interactor.dart';
 import 'package:tmail_ui_user/features/email/presentation/extensions/presentation_email_extension.dart';
+import 'package:tmail_ui_user/features/home/domain/extensions/session_extensions.dart';
 import 'package:tmail_ui_user/features/email/presentation/model/composer_arguments.dart';
 import 'package:tmail_ui_user/features/mailbox/domain/model/create_new_mailbox_request.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/domain/usecases/remove_composer_cache_by_id_interactor.dart';
@@ -128,14 +129,20 @@ import 'package:tmail_ui_user/features/upload/domain/state/local_image_picker_st
 import 'package:tmail_ui_user/features/upload/domain/usecases/local_file_picker_interactor.dart';
 import 'package:tmail_ui_user/features/upload/domain/usecases/local_image_picker_interactor.dart';
 import 'package:tmail_ui_user/features/upload/presentation/controller/upload_controller.dart';
+import 'package:tmail_ui_user/features/upload/presentation/providers/upload_from_url_providers.dart';
 import 'package:tmail_ui_user/features/upload/presentation/validator/attachment_upload_validation_service.dart';
 import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
+import 'package:tmail_ui_user/main/providers/app_provider_container.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/universal_import/html_stub.dart' as html;
 import 'package:workplace/domain/entity/drive_document.dart';
 import 'package:tmail_ui_user/features/composer/presentation/manager/drive_attachment_handler.dart';
+import 'package:tmail_ui_user/features/composer/presentation/manager/concurrency_gate.dart';
+import 'package:tmail_ui_user/features/composer/presentation/manager/drive_attachment_transfer_runner.dart';
 import 'package:tmail_ui_user/main/utils/app_config.dart';
+import 'package:tmail_ui_user/main/utils/toast_manager.dart';
+import 'package:workplace/presentation/model/drive_pick_state.dart';
 
 class ComposerController extends BaseController
     with
@@ -1018,18 +1025,72 @@ class ComposerController extends BaseController
         result,
         insertHtml: (html) async {
           if (PlatformInfo.isWeb) {
-            richTextWebController?.editorController.insertHtml(html);
-          } else {
-            await richTextMobileTabletController?.restoreMobileEditorFocus();
-            await htmlEditorApi?.insertHtml(html);
-            await SchedulerBinding.instance.endOfFrame;
+            final editorController = richTextWebController?.editorController;
+            if (editorController == null) return false;
+            editorController.insertHtml(html);
+            return true;
           }
+          final editorApi = htmlEditorApi;
+          if (editorApi == null) return false;
+          await richTextMobileTabletController?.restoreMobileEditorFocus();
+          await editorApi.insertHtml(html);
+          await SchedulerBinding.instance.endOfFrame;
+          return true;
         },
+        transferDriveDocuments: (docs) => _transferDriveDocuments(docs),
         appLocalizations: currentContext != null ? AppLocalizations.of(currentContext!) : null,
       );
     } catch (e) {
       logWarning('ComposerController::handleDrivePickResult:Exception = $e');
+      // A throw here can strand waiting chips, so the failure must be visible.
+      getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
+        e,
+        message: currentContext != null
+            ? AppLocalizations.of(currentContext!).driveAttachmentTransferFailed
+            : null,
+      ));
     }
+  }
+
+  static const _maxConcurrentDriveTransfers = 3;
+
+  /// One gate per composer: reopening the picker mid-transfer queues the new
+  /// batch behind the running one instead of multiplying in-flight requests.
+  late final ConcurrencyGate _driveTransferGate =
+      ConcurrencyGate(_maxConcurrentDriveTransfers);
+
+  Future<DriveTransferOutcome> _transferDriveDocuments(List<DriveDocument> docs) async {
+    final jmapUrl = dynamicUrlInterceptors.jmapUrl;
+    if (jmapUrl == null || jmapUrl.isEmpty) {
+      logWarning('ComposerController::_transferDriveDocuments: jmapUrl is unavailable');
+      return DriveAttachmentTransferRunner.notStartedOutcome;
+    }
+    final session = mailboxDashBoardController.sessionCurrent;
+    final accountId = mailboxDashBoardController.accountId.value;
+    final uploadUri = session?.getUploadFromUrlUri(
+      accountId,
+      jmapUrl: jmapUrl,
+    );
+    if (uploadUri == null || accountId == null) {
+      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      return DriveAttachmentTransferRunner.notStartedOutcome;
+    }
+
+    // Resolved once per batch: each read rebuilds interactor -> repo -> datasource -> dio.
+    final interactor =
+        appProviderContainer.read(uploadDriveDocumentFromUrlInteractorProvider);
+    final runner = DriveAttachmentTransferRunner(
+      uploadFromUrl: (request) => interactor.execute(request),
+      gate: _driveTransferGate,
+    );
+    return runner.transfer((
+      docs: docs,
+      accountId: accountId,
+      uploadUri: uploadUri,
+      onPlaceholdersReady: uploadController.addDownloadingPlaceholders,
+      onSuccess: uploadController.resolveDriveTransferSuccess,
+      onFailure: uploadController.resolveDriveTransferFailure,
+    ));
   }
 
   Future<dynamic> _showSendingMessageDialog({
--- a/lib/features/composer/presentation/extensions/drive_document_extension.dart
+++ b/lib/features/composer/presentation/extensions/drive_document_extension.dart
@@ -1,0 +1,20 @@
+import 'package:workplace/domain/entity/drive_document.dart';
+
+extension DriveDocumentExtension on DriveDocument {
+  /// HTML-embedded link: scheme still validated since the browser follows it directly.
+  bool isAttachableAsLink({required bool requireHttps}) =>
+      sharingLink?.isSafeLinkScheme(requireHttps: requireHttps) ?? false;
+
+  /// Backend fetches this via JMAP, so no scheme gate is needed.
+  /// An empty link is rejected here so it never reaches the backend.
+  /// Link precedence lives at the call site, not here.
+  bool isAttachableAsDownload() =>
+      downloadLink?.toString().trim().isNotEmpty ?? false;
+}
+
+extension _SafeLinkSchemeExtension on Uri {
+  /// Allowlist, not a denylist: anything else (`javascript:`, `data:`, ...) is
+  /// executable once embedded in composer HTML.
+  bool isSafeLinkScheme({required bool requireHttps}) =>
+      isScheme('https') || (!requireHttps && isScheme('http'));
+}
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -1,10 +1,18 @@
 import 'package:core/core.dart';
 import 'package:core/utils/html/file_link_card_html_builder.dart';
+import 'package:tmail_ui_user/features/composer/presentation/extensions/drive_document_extension.dart';
+import 'package:tmail_ui_user/features/composer/presentation/manager/drive_attachment_transfer_runner.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
 import 'package:workplace/domain/entity/drive_document.dart';
 import 'package:workplace/presentation/model/drive_pick_state.dart';
+
+/// Transfers [docs] as real attachments, reporting the batch outcome.
+typedef TransferDriveDocumentsCallback = Future<DriveTransferOutcome> Function(List<DriveDocument> docs);
+
+/// Inserts [html] at the caret; false when no editor is bound to insert into.
+typedef InsertHtmlCallback = Future<bool> Function(String html);
 
 class DriveAttachmentHandler {
   DriveAttachmentHandler();
@@ -16,43 +24,120 @@ class DriveAttachmentHandler {
 
   Future<void> handleDrivePickResult(
     List<DriveDocument> result, {
-    required Future<void> Function(String html) insertHtml,
+    required InsertHtmlCallback insertHtml,
+    required TransferDriveDocumentsCallback transferDriveDocuments,
     AppLocalizations? appLocalizations,
   }) async {
     if (result.isEmpty) {
-      getBinding<ToastManager>()?.showMessageFailure(
-        DrivePickFailure(
-          Exception(),
-          message: appLocalizations?.driveNoValidAttachment,
-        ),
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+      return;
+    }
+    final (linkDocs, downloadableDocs, droppedCount) =
+        _splitByAttachability(result);
+
+    // A mixed pick inserts its links and transfers its downloadable docs.
+    var linkInserted = false;
+    if (linkDocs.isNotEmpty) {
+      linkInserted = await insertDriveLinkHtml(
+        linkDocs,
+        insertHtml: insertHtml,
+        appLocalizations: appLocalizations,
+      );
+    }
+
+    if (downloadableDocs.isNotEmpty) {
+      await _transferDownloadableDocs(
+        downloadableDocs,
+        linkInserted: linkInserted,
+        transferDriveDocuments: transferDriveDocuments,
+        appLocalizations: appLocalizations,
+      );
+    } else if (linkInserted) {
+      _showSuccessToast(appLocalizations?.driveAttachmentAddedSuccessfully);
+    } else if (linkDocs.isEmpty) {
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+    } else {
+      // Docs were linkable but the editor was unbound — nothing reached the body.
+      _showFailureToast(appLocalizations?.driveAttachmentTransferFailed);
+    }
+
+    // Dropped docs would otherwise vanish silently from a pick whose siblings worked.
+    if (droppedCount > 0 && (linkDocs.isNotEmpty || downloadableDocs.isNotEmpty)) {
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+    }
+  }
+
+  /// Splits [result] into (linkable docs, downloadable docs, dropped count).
+  (List<DriveDocument>, List<DriveDocument>, int) _splitByAttachability(
+    List<DriveDocument> result,
+  ) {
+    final linkDocs = <DriveDocument>[];
+    final downloadableDocs = <DriveDocument>[];
+    var droppedCount = 0;
+    for (final doc in result) {
+      if (doc.isAttachableAsLink(requireHttps: requireHttps)) {
+        linkDocs.add(doc);
+      } else if (doc.isAttachableAsDownload()) {
+        downloadableDocs.add(doc);
+      } else {
+        droppedCount++;
+      }
+    }
+    return (linkDocs, downloadableDocs, droppedCount);
+  }
+
+  /// Transfers [downloadableDocs], toasting the batch outcome. A partial
+  /// failure stays silent here: each failed chip already toasts its own error.
+  Future<void> _transferDownloadableDocs(
+    List<DriveDocument> downloadableDocs, {
+    required bool linkInserted,
+    required TransferDriveDocumentsCallback transferDriveDocuments,
+    AppLocalizations? appLocalizations,
+  }) async {
+    final outcome = await transferDriveDocuments(downloadableDocs);
+    if (!outcome.started) {
+      _showFailureToast(appLocalizations?.driveAttachmentTransferFailed);
+      return;
+    }
+    if (outcome.failed == 0 && (linkInserted || outcome.succeeded > 0)) {
+      _showSuccessToast(appLocalizations?.driveAttachmentAddedSuccessfully);
+    }
+  }
+
+  /// Toasts [message], or logs if no [ToastManager] is bound.
+  void _showFailureToast(String? message) {
+    final toastManager = getBinding<ToastManager>();
+    if (toastManager == null) {
+      logWarning(
+        'DriveAttachmentHandler::_showFailureToast: no ToastManager bound, '
+        'failure not shown to the user',
       );
       return;
     }
-    final linkDocs = result.where((doc) {
-      final link = doc.sharingLink;
-      return link != null && (!requireHttps || link.isScheme('https'));
-    }).toList();
-    // TODO: Update logic here after implement 103. Attach Drive File as Attachment
-    if (linkDocs.isEmpty) {
-      getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
-        Exception(),
-        message: appLocalizations?.driveAttachmentInDevelopment,
-      ));
-      return;
-    }
-    await insertDriveLinkHtml(
-      linkDocs,
-      insertHtml: insertHtml,
-      appLocalizations: appLocalizations,
+    toastManager.showMessageFailure(
+      DrivePickFailure(Exception(), message: message),
     );
   }
 
-  Future<void> insertDriveLinkHtml(
+  /// Toasts [message], or logs if no [ToastManager] is bound.
+  void _showSuccessToast(String? message) {
+    final toastManager = getBinding<ToastManager>();
+    if (toastManager == null) {
+      logWarning(
+        'DriveAttachmentHandler::_showSuccessToast: no ToastManager bound, '
+        'success not shown to the user',
+      );
+      return;
+    }
+    toastManager.showMessageSuccess(DrivePickSuccess(message: message));
+  }
+
+  Future<bool> insertDriveLinkHtml(
     List<DriveDocument> docs, {
-    required Future<void> Function(String html) insertHtml,
+    required InsertHtmlCallback insertHtml,
     AppLocalizations? appLocalizations,
-  }) async {
-    await insertHtml(
+  }) {
+    return insertHtml(
       buildDriveLinksHtml(docs, appLocalizations: appLocalizations),
     );
   }
@@ -77,9 +162,8 @@ class DriveAttachmentHandler {
     DriveDocument doc, {
     AppLocalizations? appLocalizations,
   }) {
-    final link = doc.sharingLink;
-    if (link == null) return null;
-    if (requireHttps && !link.isScheme('https')) return null;
+    if (!doc.isAttachableAsLink(requireHttps: requireHttps)) return null;
+    final link = doc.sharingLink!;
 
     final openInDriveLabel =
         appLocalizations?.openInDrive ?? _fallbackOpenInDriveLabel;
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -36,14 +36,13 @@ class DriveAttachmentHandler {
         _splitByAttachability(result);
 
     // A mixed pick inserts its links and transfers its downloadable docs.
-    var linkInserted = false;
-    if (linkDocs.isNotEmpty) {
-      linkInserted = await insertDriveLinkHtml(
-        linkDocs,
-        insertHtml: insertHtml,
-        appLocalizations: appLocalizations,
-      );
-    }
+    final linkInserted = linkDocs.isEmpty
+        ? false
+        : await insertDriveLinkHtml(
+            linkDocs,
+            insertHtml: insertHtml,
+            appLocalizations: appLocalizations,
+          );
 
     if (downloadableDocs.isNotEmpty) {
       await _transferDownloadableDocs(
@@ -52,19 +51,45 @@ class DriveAttachmentHandler {
         transferDriveDocuments: transferDriveDocuments,
         appLocalizations: appLocalizations,
       );
-    } else if (linkInserted) {
+    } else {
+      _showLinkOnlyOutcome(
+        linkInserted: linkInserted,
+        hasLinkDocs: linkDocs.isNotEmpty,
+        appLocalizations: appLocalizations,
+      );
+    }
+
+    _showDroppedDocsFailure(
+      droppedCount,
+      attachedAny: linkDocs.isNotEmpty || downloadableDocs.isNotEmpty,
+      appLocalizations: appLocalizations,
+    );
+  }
+
+  /// Toasts the outcome of a pick that carried no downloadable docs.
+  void _showLinkOnlyOutcome({
+    required bool linkInserted,
+    required bool hasLinkDocs,
+    AppLocalizations? appLocalizations,
+  }) {
+    if (linkInserted) {
       _showSuccessToast(appLocalizations?.driveAttachmentAddedSuccessfully);
-    } else if (linkDocs.isEmpty) {
+    } else if (!hasLinkDocs) {
       _showFailureToast(appLocalizations?.driveNoValidAttachment);
     } else {
       // Docs were linkable but the editor was unbound — nothing reached the body.
       _showFailureToast(appLocalizations?.driveAttachmentTransferFailed);
     }
+  }
 
-    // Dropped docs would otherwise vanish silently from a pick whose siblings worked.
-    if (droppedCount > 0 && (linkDocs.isNotEmpty || downloadableDocs.isNotEmpty)) {
-      _showFailureToast(appLocalizations?.driveNoValidAttachment);
-    }
+  /// Dropped docs would otherwise vanish silently from a pick whose siblings worked.
+  void _showDroppedDocsFailure(
+    int droppedCount, {
+    required bool attachedAny,
+    AppLocalizations? appLocalizations,
+  }) {
+    if (droppedCount == 0 || !attachedAny) return;
+    _showFailureToast(appLocalizations?.driveNoValidAttachment);
   }
 
   /// Splits [result] into (linkable docs, downloadable docs, dropped count).
--- a/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
@@ -132,6 +132,9 @@ class DriveAttachmentTransferRunner {
       request.onFailure(taskId);
       return;
     }
+    // Chip deleted while this upload was in flight: a late response must not
+    // resolve a task the user already dropped.
+    if (task.placeholder.cancelToken?.isCancelled == true) return;
     // fold runs outside the try so a throw from onSuccess/onFailure isn't
     // mistaken for an upload failure and double-reported; the callbacks are
     // guarded by _notify instead, so one bad chip can't abort its siblings.
--- a/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
@@ -69,6 +69,9 @@ class DriveAttachmentTransferRunner {
               ),
             ))
         .toList();
+    final placeholdersById = {
+      for (final task in tasks) task.placeholder.taskId: task.placeholder,
+    };
 
     try {
       request.onPlaceholdersReady(tasks.map((task) => task.placeholder).toList());
@@ -92,11 +95,19 @@ class DriveAttachmentTransferRunner {
       onPlaceholdersReady: request.onPlaceholdersReady,
       onSuccess: (UploadTaskId taskId, Attachment attachment) {
         succeeded++;
-        _notify('onSuccess', () => request.onSuccess(taskId, attachment));
+        _notify(
+          'onSuccess',
+          placeholdersById[taskId],
+          () => request.onSuccess(taskId, attachment),
+        );
       },
       onFailure: (UploadTaskId taskId) {
         failed++;
-        _notify('onFailure', () => request.onFailure(taskId));
+        _notify(
+          'onFailure',
+          placeholdersById[taskId],
+          () => request.onFailure(taskId),
+        );
       },
     );
 
@@ -151,7 +162,11 @@ class DriveAttachmentTransferRunner {
   }
 
   /// Runs a caller callback; a throw is logged and contained to this task.
-  void _notify(String label, void Function() callback) {
+  void _notify(
+    String label,
+    DriveTransferPlaceholder? placeholder,
+    void Function() callback,
+  ) {
     try {
       callback();
     } catch (e, s) {
@@ -159,6 +174,12 @@ class DriveAttachmentTransferRunner {
         'DriveAttachmentTransferRunner::_notify: $label failed',
         exception: e,
         stackTrace: s,
+        // File name stays out: extras reach Sentry.
+        extras: {
+          'task_id': placeholder?.taskId.id,
+          'file_size': placeholder?.fileSize,
+          'mime_type': placeholder?.mimeType,
+        },
       );
     }
   }
--- a/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
@@ -1,5 +1,6 @@
 import 'package:core/presentation/state/failure.dart';
 import 'package:core/presentation/state/success.dart';
+import 'package:core/utils/app_logger.dart';
 import 'package:dartz/dartz.dart';
 import 'package:dio/dio.dart';
 import 'package:jmap_dart_client/jmap/account_id.dart';
@@ -69,7 +70,17 @@ class DriveAttachmentTransferRunner {
             ))
         .toList();
 
-    request.onPlaceholdersReady(tasks.map((task) => task.placeholder).toList());
+    try {
+      request.onPlaceholdersReady(tasks.map((task) => task.placeholder).toList());
+    } catch (e, s) {
+      // Without chips nothing can resolve, so the batch never starts.
+      logError(
+        'DriveAttachmentTransferRunner::transfer: onPlaceholdersReady failed',
+        exception: e,
+        stackTrace: s,
+      );
+      return notStartedOutcome;
+    }
 
     // Counted here so the caller can tell a fully clean batch from a partial one.
     var succeeded = 0;
@@ -81,11 +92,11 @@ class DriveAttachmentTransferRunner {
       onPlaceholdersReady: request.onPlaceholdersReady,
       onSuccess: (UploadTaskId taskId, Attachment attachment) {
         succeeded++;
-        request.onSuccess(taskId, attachment);
+        _notify('onSuccess', () => request.onSuccess(taskId, attachment));
       },
       onFailure: (UploadTaskId taskId) {
         failed++;
-        request.onFailure(taskId);
+        _notify('onFailure', () => request.onFailure(taskId));
       },
     );
 
@@ -122,7 +133,8 @@ class DriveAttachmentTransferRunner {
       return;
     }
     // fold runs outside the try so a throw from onSuccess/onFailure isn't
-    // mistaken for an upload failure and double-reported.
+    // mistaken for an upload failure and double-reported; the callbacks are
+    // guarded by _notify instead, so one bad chip can't abort its siblings.
     either.fold(
       (failure) {
         // a user-cancelled transfer is not a failure; the chip is already gone.
@@ -133,5 +145,18 @@ class DriveAttachmentTransferRunner {
           ? request.onSuccess(taskId, success.attachment)
           : request.onFailure(taskId),
     );
+  }
+
+  /// Runs a caller callback; a throw is logged and contained to this task.
+  void _notify(String label, void Function() callback) {
+    try {
+      callback();
+    } catch (e, s) {
+      logError(
+        'DriveAttachmentTransferRunner::_notify: $label failed',
+        exception: e,
+        stackTrace: s,
+      );
+    }
   }
 }
--- a/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
@@ -1,0 +1,137 @@
+import 'package:core/presentation/state/failure.dart';
+import 'package:core/presentation/state/success.dart';
+import 'package:dartz/dartz.dart';
+import 'package:dio/dio.dart';
+import 'package:jmap_dart_client/jmap/account_id.dart';
+import 'package:model/email/attachment.dart';
+import 'package:tmail_ui_user/features/composer/presentation/manager/concurrency_gate.dart';
+import 'package:tmail_ui_user/features/upload/domain/model/upload_task_id.dart';
+import 'package:tmail_ui_user/features/upload/domain/repository/upload_from_url_request.dart';
+import 'package:tmail_ui_user/features/upload/domain/state/upload_drive_document_from_url_state.dart';
+import 'package:tmail_ui_user/features/upload/presentation/model/drive_transfer_placeholder.dart';
+import 'package:uuid/uuid.dart';
+import 'package:workplace/domain/entity/drive_document.dart';
+
+/// Runs one gateway call; injected so this class never touches GetX.
+typedef UploadFromUrlCall = Future<Either<Failure, Success>> Function(
+  UploadFromUrlRequest request,
+);
+
+typedef OnPlaceholdersReady = void Function(List<DriveTransferPlaceholder> placeholders);
+typedef OnDriveTransferSuccess = void Function(UploadTaskId taskId, Attachment attachment);
+typedef OnDriveTransferFailure = void Function(UploadTaskId taskId);
+
+/// One batch's input, bundled into a single param for [DriveAttachmentTransferRunner.transfer].
+typedef DriveTransferRequest = ({
+  List<DriveDocument> docs,
+  AccountId accountId,
+  Uri uploadUri,
+  OnPlaceholdersReady onPlaceholdersReady,
+  OnDriveTransferSuccess onSuccess,
+  OnDriveTransferFailure onFailure,
+});
+
+/// One batch's result. [started] is false when the batch never ran; cancelled
+/// tasks resolve neither counter.
+typedef DriveTransferOutcome = ({bool started, int succeeded, int failed});
+
+/// One in-flight unit of work: the doc plus its pre-minted placeholder.
+typedef _DriveTransferTask = ({DriveDocument doc, DriveTransferPlaceholder placeholder});
+
+/// Fans a batch of downloadable drive documents out to [uploadFromUrl] through
+/// [gate]. The gate is owned by the caller, so concurrency stays bounded across
+/// every batch rather than per runner. Pure logic, no GetX — fully unit-testable.
+class DriveAttachmentTransferRunner {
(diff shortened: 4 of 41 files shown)


lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +23,56 @@ class DriveAttachmentHandler {
...
-    if (linkDocs.isEmpty) {
-      getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
-        Exception(),
-        message: appLocalizations?.driveAttachmentInDevelopment,
-      ));
+    final linkDocs = <DriveDocument>[];
+    final downloadableDocs = <DriveDocument>[];
+    for (final doc in result) {
+      if (doc.isAttachableAsLink(requireHttps: requireHttps)) linkDocs.add(doc);
+      if (doc.isAttachableAsDownload) downloadableDocs.add(doc);
```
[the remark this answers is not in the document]
tddang-linagora: read the definition of isAttachableAsDownload

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
```diff
@@ -0,0 +1,30 @@
+import 'dart:async';
+
+/// Runs [task] over [items] with at most [maxConcurrent] in flight, workers
+/// pulling from one shared iterator so a slot frees the instant it's done.
+/// [task] must handle its own failures; [items] must not throw while spawning.
+Future<void> runWithConcurrency<T>(
```
hoangdat: need tests for this method

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,102 @@
...
+      ));
+    } catch (_) {
+      // uploadFromUrl must resolve every task; a thrown error still counts as failure.
+      request.onFailure(taskId);
+      return;
+    }
+    // fold runs outside the try so a throw from onSuccess/onFailure isn't
+    // mistaken for an upload failure and double-reported.
+    either.fold(
+      (_) => request.onFailure(taskId),
```
hoangdat: `cancel` will be have the same message with failure?

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +214,47 @@ class UploadController extends BaseController {
...
+        uploadingProgress: 100,
+        uploadStatus: UploadFileStatus.succeed,
+        attachment: attachment,
+      ),
+    );
+    _refreshListUploadAttachmentState();
+  }
+
+  /// Resolves a drive-transfer chip to failed; kept, not deleted.
+  void resolveDriveTransferFailure(UploadTaskId taskId) {
```
hoangdat, line 248: this function still keep failed chips?

test/features/composer/presentation/manager/drive_attachment_transfer_runner_test.dart
hoangdat, line 1: please add more test to cover better

```
  1. uploadFromUrl throws — the catch at lines 88–91 is production code with no test.
  2. Right that is not UploadDriveDocumentFromUrlSuccess — the ternary at 97–99 is untested.
  3. UploadDriveDocumentFromUrlCancelled is Left — runner does not distinguish cancel from fail. Interactor tests Cancelled; this file never does.
     That is why chip-delete toasts an error.
  4. Request payload — never asserts attachmentUrl / name / mimeType / cancelToken / uploadUri were forwarded.
  5. Placeholder contents — only placeholders.length. No unique taskId, no fileName/fileSize/cancelToken non-null.
  6. Mixed batch — one success + one fail + one null link in the same transfer. No test that each taskId maps to the right callback.
  7. Return value — non-empty all-fail still returns true. Untested; handler depends on it.
  8. Concurrency test — asserts maxInFlight <= 2 only. Never uploadCalls == 5 or 5 onSuccess. Could pass if only 2 of 5 started.
```

workplace/lib/presentation/widget/drive_attachment_context_menu_tile.dart
```diff
@@ -10,6 +10,7 @@ import 'package:workplace/presentation/model/drive_intent_image_assets.dart';
 class DriveAttachmentContextMenuTile extends StatefulWidget {
   final ImagePaths imagePaths;
   final Uri workplaceUri;
+  final ValueGetter<bool> uploadFromUrlSupported;
```
hoangdat: how about to create a  `DrivePickerSession` to keep all the thing inside, and maybe a `DrivePickerSessionResolver` with `resolve()` method to get the things on-demand?

lib/features/composer/presentation/manager/bounded_concurrency_runner.dart
tddang-linagora: Added

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Updated

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: Fixed

test/features/composer/presentation/manager/drive_attachment_transfer_runner_test.dart
tddang-linagora: Added

workplace/lib/presentation/widget/drive_attachment_context_menu_tile.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1028,42 @@ class ComposerController extends BaseController
...
+      jmapUrl: dynamicUrlInterceptors.jmapUrl,
+    );
+    if (uploadUri == null || accountId == null) {
+      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      return false;
+    }
+
+    final runner = DriveAttachmentTransferRunner(
+      uploadFromUrl: (request) =>
+          Get.find<UploadDriveDocumentFromUrlInteractor>(tag: composerId)
```
dab246: The new flow retrieves `UploadDriveDocumentFromUrlInteractor` via `Get.find`, but the interactor exists only as a Riverpod provider and has no GetX registration.

tddang-linagora: Done

hoangdat:
- [ ] demo on cancel progress transfer in case of close composer

lib/features/composer/presentation/extensions/drive_document_extension.dart
```diff
@@ -0,0 +1,15 @@
...
+  /// HTML-embedded link: scheme still validated since the browser follows it directly.
+  bool isAttachableAsLink({required bool requireHttps}) {
+    final link = sharingLink;
+    return link != null && (!requireHttps || link.isScheme('https'));
+  }
+
+  /// Backend fetches this via JMAP, so no scheme gate is needed.
+  bool get isAttachableAsDownload {
+    final link = downloadLink;
+    return sharingLink == null && link != null;
```
hoangdat: how about the case of sharing not https and not null?

lib/features/composer/presentation/manager/drive_attachment_handler.dart
hoangdat: should not support both action if doc have 2 links

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +23,56 @@ class DriveAttachmentHandler {
...
+        linkDocs,
+        insertHtml: insertHtml,
+        appLocalizations: appLocalizations,
+      );
+    }
+
+    final transfersStarted = downloadableDocs.isNotEmpty &&
+        await transferDriveDocuments(downloadableDocs);
+
+    if (linkDocs.isNotEmpty || transfersStarted) return;
```
hoangdat: we should have dedicated toast for the case of error in transfer drive document?

lib/features/composer/presentation/extensions/drive_document_extension.dart
tddang-linagora: Updated

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Updated

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -18,6 +19,17 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+      // Read at picker-open time, so it's always the current session's answer.
+      uploadFromUrlSupported: () {
+        final dashboard = getBinding<MailboxDashBoardController>();
+        final jmapUrl = dashboard?.dynamicUrlInterceptors.jmapUrl;
+        if (jmapUrl == null || jmapUrl.isEmpty) return false;
+        return dashboard?.sessionCurrent?.isUploadFromUrlSupported(
+              dashboard.accountId.value,
+              jmapUrl: jmapUrl,
+            ) ??
+            false;
```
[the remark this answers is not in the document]
hoangdat: please address it too @tddang-linagora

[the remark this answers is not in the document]
tddang-linagora: `isUploadFromUrlSupported` already returns `false` when `accountId` is null. It delegates to `getUploadFromUrlUri` (session_extensions.dart:161), which early-returns `null` at line 162 (`if (accountId == null) return null;`) before ever consulting the session capability. `isUploadFromUrlSupported` (session_extensions.dart:136) is just `getUploadFromUrlUri(...) != null`, so a null `accountId` can never make it return `true`. The described failure path — picker exposing an upload action that fails downstream — cannot occur through this callback; `_transferDriveDocuments`'s own `accountId == null` guard (composer_controller.dart:1052) is redundant defense-in-depth, not compensation for a bug here. No change made.

@coderabbitai

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1029,48 @@ class ComposerController extends BaseController
...
+    );
+    return runner.transfer((
+      docs: docs,
+      accountId: accountId,
+      uploadUri: uploadUri,
+      onPlaceholdersReady: uploadController.addDownloadingPlaceholders,
+      onSuccess: uploadController.resolveDriveTransferSuccess,
+      onFailure: uploadController.resolveDriveTransferFailure,
+    ));
+  }
```
hoangdat, line 1094: How about

```suggestion
  final interactor = appProviderContainer
      .read(uploadDriveDocumentFromUrlInteractorProvider);
  return DriveAttachmentTransferRunner(
    uploadFromUrl: (request) => interactor.execute(request).first,
  ).transfer((
    docs: docs,
    accountId: accountId,
    uploadUri: uploadUri,
    onPlaceholdersReady: uploadController.addDownloadingPlaceholders,
    onSuccess: uploadController.resolveDriveTransferSuccess,
    onFailure: uploadController.resolveDriveTransferFailure,
   ));
```

less lookup, less instance create for interactor -> repo -> datasource -> dio. WDYT?

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1029,48 @@ class ComposerController extends BaseController
...
+      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      return false;
+    }
+
+    final runner = DriveAttachmentTransferRunner(
+      uploadFromUrl: (request) =>
+          appProviderContainer
+              .read(uploadDriveDocumentFromUrlInteractorProvider)
+              .execute(request)
+              .first,
```
hoangdat: why not use Future for interactor?

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +23,82 @@ class DriveAttachmentHandler {
...
-      getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
-        Exception(),
-        message: appLocalizations?.driveAttachmentInDevelopment,
-      ));
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+    }
+  }
+
+  /// Splits [result] into (linkable docs, downloadable docs).
+  (List<DriveDocument>, List<DriveDocument>) _splitByAttachability(
```
hoangdat: good idea to handling the error of link in transfer logic if it is not good sharing link, but should more detail at gate before push it to BE

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,109 @@
...
+      maxConcurrent,
+      (task) => _runOne(task, request),
+    );
+    return true;
+  }
+
+  Future<void> _runOne(_DriveTransferTask task, DriveTransferRequest request) async {
+    final taskId = task.placeholder.taskId;
+    final downloadLink = task.doc.downloadLink;
+    if (downloadLink == null) {
```
hoangdat: prevent the empty link here also

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

lib/features/composer/presentation/extensions/drive_document_extension.dart
```diff
@@ -0,0 +1,20 @@
+import 'package:workplace/domain/entity/drive_document.dart';
+
+extension DriveDocumentExtension on DriveDocument {
+  /// HTML-embedded link: scheme still validated since the browser follows it directly.
+  bool isAttachableAsLink({required bool requireHttps}) {
+    final link = sharingLink;
```
dab246: Why do we need to assign this to a new variable instead of using `sharingLink` directly?

lib/features/composer/presentation/extensions/drive_document_extension.dart
```diff
@@ -0,0 +1,20 @@
...
+    return link != null && (!requireHttps || link.isScheme('https'));
+  }
+
+  /// Backend fetches this via JMAP, so no scheme gate is needed; only
+  /// applies when the doc isn't already handled as a link, keeping the two
+  /// buckets mutually exclusive.
+  /// An empty link is rejected here so it never reaches the backend.
+  bool isAttachableAsDownload({required bool requireHttps}) {
```
dab246: Reject insecure download URLs in release.  `isAttachableAsDownload` only checks that `downloadLink` is non-null and non-empty. In release, an `http://` link is therefore accepted, then `DriveAttachmentTransferRunner` forwards it as `Content-Location` for the JMAP backend to download. This is inconsistent with the existing data-layer invariant in `workplace/lib/domain/entity/drive_document_extension.dart`, where a staging download rejects non-HTTPS links in release. A signed Drive URL fetched over HTTP can expose the token and the file contents.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1029,47 @@ class ComposerController extends BaseController
...
+    final uploadUri = session?.getUploadFromUrlUri(
+      accountId,
+      jmapUrl: jmapUrl,
+    );
+    if (uploadUri == null || accountId == null) {
+      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      return false;
+    }
+
+    // Resolved once per batch: each read rebuilds interactor -> repo -> datasource -> dio.
```
dab246, line 1079: Bound concurrency globally, not per batch. `_transferDriveDocuments` constructs a new `DriveAttachmentTransferRunner` for every picker result. Each runner has three workers, but picker results are delivered through a `void` callback, so a user can reopen the picker while a previous batch is still transferring. The actual in-flight request count is therefore `3 × active batches`, not three. This can overload the browser, backend and Drive service, while causing many simultaneous UI updates.

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +214,44 @@ class UploadController extends BaseController {
...
+    _uploadingStateFiles.updateElementByUploadTaskId(
+      taskId,
+      (state) => state?.copyWith(
+        uploadingProgress: 100,
+        uploadStatus: UploadFileStatus.succeed,
+        attachment: attachment,
+      ),
+    );
+    _refreshListUploadAttachmentState();
+  }
```
dab246, line 244: Avoid O(N²) state and UI work for Drive batches.


For each completed file, `updateElementByUploadTaskId` performs a linear `indexWhere`. Then `_refreshListUploadAttachmentState` copies the full list twice and refreshes the GetX observable. The web attachment widget responds by recreating every attachment item in a `Wrap`.

For a batch of N files, N completions result in approximately O(N²) lookup, list-copy and widget-construction work. The picker has no count limit in this flow; a byte quota does not prevent many small files from being selected. The runner also eagerly creates task and placeholder lists, and the UI eagerly renders all waiting chips, so memory is O(N) for the lifetime of the batch.

**Recommended fix:** store attachment state in an ordered map/index keyed by `UploadTaskId`; retain a separate order list only if necessary. Coalesce observable refreshes to one per frame or a short batch window. For the web UI, use a lazy/virtualized list or impose a file-count limit if all chips must remain visible.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -43,17 +42,19 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
       final filePickerConfig = WorkplaceFilePickerConfigRequest(
         sharingLink: WorkplaceActionConfigRequest(label: l10n.addAsLink),
         downloadLink: addAsAttachmentTitle == null
             ? null
             : WorkplaceActionConfigRequest(
                 label: addAsAttachmentTitle,
-                maxFileSize: maxAttachmentSizeBytes,
-                availableSize: maxAttachmentSizeBytes,
+                maxFileSize: snapshot.maxAttachmentSizeBytes,
+                availableSize: snapshot.maxAttachmentSizeBytes,
```
dab246: `maxFileSize` and `availableSize` are both set to the full server attachment limit. The composer validator already knows the bytes used by existing regular and inline attachments, but that remaining capacity is not passed to the Drive picker.

For example, with a 100 MB limit and 80 MB already attached, the picker advertises 100 MB available. It can upload another 50 MB before the composer refuses to send the email. This wastes upload bandwidth and leaves unnecessary blobs on the account.

**Recommended fix:** inject a composer-specific remaining-capacity getter: `max(0, hardLimit - currentAllAttachmentBytes)`. Use this value for `availableSize` while retaining the per-file maximum for `maxFileSize`.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,110 @@
...
+
+    await runWithConcurrency(
+      tasks,
+      maxConcurrent,
+      (task) => _runOne(task, request),
+    );
+    return true;
+  }
+
+  Future<void> _runOne(_DriveTransferTask task, DriveTransferRequest request) async {
```
dab246, line 120: Skip cancelled queued tasks before invoking upload

Deleting a waiting chip or closing the composer cancels its token, but queued workers still call `uploadFromUrl` when they reach that task. Dio will normally resolve the request as cancelled, but every task still traverses the interactor/repository/datasource chain.

**Recommended fix:** return immediately when `task.placeholder.cancelToken?.isCancelled == true`, while retaining the token in the request to handle a cancellation race after this check.

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
```diff
@@ -0,0 +1,197 @@
...
+    final items = [0, 1];
+
+    await runWithConcurrency<int>(items, 5, (item) async {
+      await Future<void>.delayed(Duration.zero);
+      processed.add(item);
+    });
+
+    expect(processed.toSet(), items.toSet());
+    expect(processed, hasLength(items.length));
+  });
```
dab246: The “items fit in one wave” scenario appears three times (lines 52–80). The “other workers drain after one throws” and “single worker stops after throw” scenarios are each repeated at lines 148–196.

**Recommended fix:** retain one test per behaviour and keep the helper-based versions. This reduces maintenance noise without reducing coverage.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +23,82 @@ class DriveAttachmentHandler {
...
+    List<DriveDocument> result,
+  ) {
+    final linkDocs = <DriveDocument>[];
+    final downloadableDocs = <DriveDocument>[];
+    for (final doc in result) {
+      if (doc.isAttachableAsLink(requireHttps: requireHttps)) linkDocs.add(doc);
+      if (doc.isAttachableAsDownload(requireHttps: requireHttps)) downloadableDocs.add(doc);
```
dab246: Should be:
```dart
if (doc.isAttachableAsLink(requireHttps: requireHttps)) {
  linkDocs.add(doc);
} else if (doc.isAttachableAsDownload(requireHttps: requireHttps)) {
  downloadableDocs.add(doc);
}
```

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,110 @@
...
+
+/// One in-flight unit of work: the doc plus its pre-minted placeholder.
+typedef _DriveTransferTask = ({DriveDocument doc, DriveTransferPlaceholder placeholder});
+
+/// Fans a batch of downloadable drive documents out to [uploadFromUrl] with
+/// bounded concurrency. Pure logic, no GetX — fully unit-testable.
+class DriveAttachmentTransferRunner {
+  DriveAttachmentTransferRunner({
+    required this.uploadFromUrl,
+    this.maxConcurrent = 3,
```
dab246: Bound Drive transfers across picker batches

`maxConcurrent` is currently enforced per `DriveAttachmentTransferRunner.transfer()` call only. `ComposerController` creates a new runner for every picker result, while `onPickState` is a `void` callback, so a user can reopen the picker before a previous batch finishes.

For example, two consecutive selections can run 3 + 3 = 6 upload-from-URL requests concurrently; more selections increase that number further.

Could we move the queue/semaphore to composer or `UploadController` scope so all Drive batches share one concurrency limit? Please also add a test that starts two batches and verifies the aggregate in-flight request count never exceeds the configured limit.

dab246:
Add regression tests:
- [x] reject an HTTP `downloadLink` when HTTPS is required;
- [x] two concurrent picker batches never exceed the global concurrency limit;
- [x] `availableSize` equals remaining composer capacity;
- [x] a task cancelled while waiting in the queue never calls `uploadFromUrl`;
- [x] a late success after chip deletion/controller close does not restore the attachment;
- [x] remove duplicated cases in `bounded_concurrency_runner_test.dart`.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

lib/features/composer/presentation/extensions/drive_document_extension.dart
tddang-linagora: Removed

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: Refresh coalescing and a virtualized/count-limited web attachment list are deferred. Both touch upload paths shared with regular and inline uploads, well outside this PR, and N is already bounded by the server attachment byte quota. The index map removes the quadratic term; if per-frame refresh coalescing is still wanted it is better done as its own change against `UploadController._refreshListUploadAttachmentState` with the non-Drive callers in scope.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

test/features/composer/presentation/manager/bounded_concurrency_runner_test.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

tddang-linagora:
> - [ ]  reject an HTTP `downloadLink` when HTTPS is required;

`downloadLink` is never followed by the browser — the backend fetches it server-side via JMAP upload-from-URL — so the scheme allowlist that protects the HTML-embedded `sharingLink` does not apply to it. Gating it on `https` would reject working links on http-only Drive deployments (the same deployments where `requireHttps` is false for sharing links). The null/empty guard in `drive_attachment_transfer_runner.dart:79-83` still keeps malformed links off the backend. Happy to add the gate if the product contract requires https-only Drive endpoints.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

dab246:
- [ ] No toast message when pick file from Drive picker as attachment Success and Cancel

dab246:
- [ ] No dialog warning message when pick file from Drive as attachment 

https://github.com/user-attachments/assets/e3ef4429-6c4c-44b1-9bc8-3d9631f6147c

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1030,55 @@ class ComposerController extends BaseController
             await SchedulerBinding.instance.endOfFrame;
           }
         },
+        transferDriveDocuments: (docs) => _transferDriveDocuments(docs),
         appLocalizations: currentContext != null ? AppLocalizations.of(currentContext!) : null,
       );
     } catch (e) {
       logWarning('ComposerController::handleDrivePickResult:Exception = $e');
```
hoangdat, line 1044: Silent drop: catch only logs. insertHtml/Get.find/onSuccess throws leave waiting chips and no user toast.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +24,107 @@ class DriveAttachmentHandler {
...
+    List<DriveDocument> result,
+  ) {
+    final linkDocs = <DriveDocument>[];
+    final downloadableDocs = <DriveDocument>[];
+    for (final doc in result) {
+      if (doc.isAttachableAsLink(requireHttps: requireHttps)) {
+        linkDocs.add(doc);
+      } else if (doc.isAttachableAsDownload()) {
+        downloadableDocs.add(doc);
+      }
```
hoangdat, line 109: Silent drop: docs that are neither a safe link nor a download are skipped with no toast when siblings are valid.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +24,107 @@ class DriveAttachmentHandler {
...
+    }
+    if (outcome.failed == 0 && (linkInserted || outcome.succeeded > 0)) {
+      _showSuccessToast(appLocalizations?.driveAttachmentAddedSuccessfully);
+    }
+  }
+
+  /// Toasts [message], or logs if no [ToastManager] is bound.
+  void _showFailureToast(String? message) {
+    final toastManager = getBinding<ToastManager>();
+    if (toastManager == null) {
```
hoangdat, line 135: Silent drop: no ToastManager means the user never sees this failure (log only).

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +214,44 @@ class UploadController extends BaseController {
...
+
+  /// Resolves a drive-transfer chip to failed by removing it, matching the
+  /// plain-upload failure path.
+  void resolveDriveTransferFailure(UploadTaskId taskId) {
+    deleteFileUploaded(taskId);
+    _showToastMessageWhenUploadAttachmentsFailure(
```
hoangdat, line 250: Silent drop: chip is already gone; if overlay is null the user gets no reason the Drive transfer failed.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -18,9 +19,25 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+      },
       oidcTokenGetter: () => getBinding<AuthorizationInterceptors>()?.currentOidcIdToken,
       maxAttachmentSizeBytesGetter: () =>
           getBinding<MailboxDashBoardController>()?.maxSizeAttachmentsPerEmail?.value,
+      // Read at picker-open time from the composer that owns the picker.
+      remainingAttachmentCapacityBytesGetter: (composerId) =>
+          getBinding<ComposerController>(tag: composerId)
+              ?.attachmentUploadValidationService
+              .remainingCapacityBytes,
       onPickState: (composerId, state) async {
```
hoangdat, line 41: Silent drop: composer missing (closed/wrong tag) discards DrivePickResult with no log or toast.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -18,9 +19,25 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
         if (state is DrivePickResult) {
           try {
```
hoangdat: Silent drop: this catch never runs — handleDrivePickResult already swallows. Toast in the composer catch instead.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -43,17 +42,22 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
       final l10n = AppLocalizations.of(context)!;
```
hoangdat, line 64: Silent drop: disposed-before-modal (early return just above) is log-only — no DrivePickFailure toast. User tapped, nothing happens.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,40 +24,107 @@ class DriveAttachmentHandler {
...
-          message: appLocalizations?.driveNoValidAttachment,
-        ),
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+      return;
+    }
+    final (linkDocs, downloadableDocs) = _splitByAttachability(result);
+
+    // A mixed pick inserts its links and transfers its downloadable docs.
+    if (linkDocs.isNotEmpty) {
+      await insertDriveLinkHtml(
```
hoangdat: Silent drop: insertHtml no-ops if the editor is unbound (`?.` in ComposerController), then success still toasts.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -161,17 +165,23 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
         if (uri == null) return const SizedBox.shrink();
         return DriveAttachmentContextMenuTile(
           imagePaths: imagePaths,
-          workplaceUri: uri,
+          session: _sessionFor(uri, null),
           onPickCallback: onPickState == null
               ? null
               : (state) => onPickState!(null, state),
```
hoangdat, line 171: Silent drop: context menu passes composerId null. Tagged web composers never get the pick; result is discarded.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,137 @@
...
+                taskId: UploadTaskId(const Uuid().v4()),
+                fileName: doc.name,
+                fileSize: doc.size,
+                mimeType: doc.mimeType,
+                cancelToken: CancelToken(),
+              ),
+            ))
+        .toList();
+
+    request.onPlaceholdersReady(tasks.map((task) => task.placeholder).toList());
```
hoangdat: Out of control: if onPlaceholdersReady throws, uploads never start, no chips, and the error escapes to the composer catch (log only).

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,137 @@
...
+        succeeded++;
+        request.onSuccess(taskId, attachment);
+      },
+      onFailure: (UploadTaskId taskId) {
+        failed++;
+        request.onFailure(taskId);
+      },
+    );
+
+    await Future.wait(
```
hoangdat, line 114: Out of control: onSuccess/onFailure throwing aborts Future.wait. That chip can stay waiting; the batch outcome is never returned.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,137 @@
...
+        cancelToken: task.placeholder.cancelToken,
+      ));
+    } catch (_) {
+      // uploadFromUrl must resolve every task; a thrown error still counts as failure.
+      request.onFailure(taskId);
+      return;
+    }
+    // fold runs outside the try so a throw from onSuccess/onFailure isn't
+    // mistaken for an upload failure and double-reported.
+    either.fold(
```
hoangdat, line 152: Out of control: fold is outside try, so a callback throw is not a per-task failure — it bubbles as an uncaught batch error.

lib/features/upload/presentation/validator/attachment_upload_validation_service.dart
```diff
@@ -39,6 +39,14 @@ class AttachmentUploadValidationService {
             validator ?? AttachmentUploadValidator([const AttachmentSizeLimitRule()])),
         _feedbackBuilder = feedbackBuilder ?? AttachmentValidationFeedbackImpl.new;
 
+  /// Bytes still attachable under the server cap, `null` when it advertises none.
+  int? get remainingCapacityBytes {
```
hoangdat, line 43: Out of control: transfer never re-checks this. Picker availableSize is a hint; oversized Drive files fail later at send.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -43,17 +42,22 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
             ? null
             : WorkplaceActionConfigRequest(
                 label: addAsAttachmentTitle,
-                maxFileSize: maxAttachmentSizeBytes,
-                availableSize: maxAttachmentSizeBytes,
+                maxFileSize: snapshot.maxAttachmentSizeBytes,
+                // What is left after the composer's existing attachments, so
+                // the picker can't offer bytes the send guard would reject.
+                availableSize: snapshot.remainingAttachmentCapacityBytes ??
+                    snapshot.maxAttachmentSizeBytes,
```
hoangdat, line 82: Out of control: availableSize is picker UX only. Upload-from-url does not enforce remainingCapacityBytes.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -135,16 +144,11 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
         return DriveAttachmentPickerButton(
           composerId: composerId,
           imagePaths: imagePaths,
-          workplaceUri: uri,
           style: style,
+          session: _sessionFor(uri, composerId),
           onPickCallback: onPickState == null
               ? null
               : (state) => onPickState!(composerId, state),
```
hoangdat, line 151: Out of control: void callback discards the async onPickState Future. A throw on DrivePickFailure is an unhandled zone error.

hoangdat (commented):
Out-of-control anchors from the Drive pick error-handling review (uncaught / stuck / deferred).

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1030,55 @@ class ComposerController extends BaseController
             await SchedulerBinding.instance.endOfFrame;
```
hoangdat: Out of control: insertHtml (web path not awaited) — iframe/JS failures never surface as Dart errors. A throw here is log-only.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1025,13 +1030,55 @@ class ComposerController extends BaseController
...
+        appProviderContainer.read(uploadDriveDocumentFromUrlInteractorProvider);
+    final runner = DriveAttachmentTransferRunner(
+      uploadFromUrl: (request) => interactor.execute(request),
+      gate: _driveTransferGate,
+    );
+    return runner.transfer((
+      docs: docs,
+      accountId: accountId,
+      uploadUri: uploadUri,
+      onPlaceholdersReady: uploadController.addDownloadingPlaceholders,
```
hoangdat, line 1090: Out of control: addDownloadingPlaceholders throwing here aborts transfer before any upload; composer catch only logs.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -7,6 +7,7 @@ import 'package:workplace/l10n/workplace_localizations.dart';
 import 'package:workplace/presentation/model/drive_intent_image_assets.dart';
 import 'package:workplace/presentation/model/drive_pick_outcome.dart';
 import 'package:workplace/presentation/model/drive_pick_state.dart';
+import 'package:workplace/presentation/model/drive_picker_session.dart';
 import 'package:workplace/presentation/view/drive_intent_web_view_modal.dart';
 
 typedef OnPickDriveCallback = void Function(DrivePickState state);
```
hoangdat: Out of control: void callback cannot await onPickState. A throw on DrivePickFailure becomes an unhandled zone error.

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -328,6 +328,9 @@ class UploadController extends BaseController {
         AppLocalizations.of(currentContext!).can_not_upload_this_file_as_attachments,
         leadingSVGIconColor: Colors.white,
         leadingSVGIcon: imagePaths.icAttachment);
+    } else {
+      // The chip is already gone, so an unshowable toast loses the reason entirely.
+      logError('UploadController::_showToastMessageWhenUploadAttachmentsFailure: no context to show failure for ${failure.uploadId}');
```
hoangdat, line 334: why not keep the failure in the log also

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -161,17 +165,23 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
         if (uri == null) return const SizedBox.shrink();
         return DriveAttachmentContextMenuTile(
           imagePaths: imagePaths,
-          workplaceUri: uri,
+          session: _sessionFor(uri, null),
```
dab246, line 168: The Drive context menu does not know the target composer. The user can open the “Attach from Drive” picker from the context menu, but the selected result has nowhere to be applied. The toolbar path is not affected because it receives a `composerId`.

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -213,6 +214,44 @@ class UploadController extends BaseController {
...
+      file: FileInfo(
+        fileName: placeholder.fileName,
+        fileSize: placeholder.fileSize,
+        type: placeholder.mimeType,
+      ),
+      uploadStatus: UploadFileStatus.waiting,
+      cancelToken: placeholder.cancelToken,
+    )));
+    _refreshListUploadAttachmentState();
+  }
```
dab246, line 231: An in-progress transfer is lost when attachments are refreshed after saving a draf. 
A realistic sequence is: select a Drive file and immediately save the draft. The draft is saved without the waiting attachment; refresh removes the chip and drops its cancel token from state while the upload-from-URL request keeps running. If the backend later succeeds, no chip remains to resolve. The attachment disappears for the user, and the JMAP blob may be orphaned while the unnecessary request consumes resources.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -36,24 +35,33 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
             ? null
             : WorkplaceActionConfigRequest(
                 label: addAsAttachmentTitle,
-                maxFileSize: maxAttachmentSizeBytes,
-                availableSize: maxAttachmentSizeBytes,
+                maxFileSize: snapshot.maxAttachmentSizeBytes,
+                // What is left after the composer's existing attachments, so
+                // the picker can't offer bytes the send guard would reject.
+                availableSize: snapshot.remainingAttachmentCapacityBytes ??
+                    snapshot.maxAttachmentSizeBytes,
```
dab246, line 82: `availableSize` is a UX hint, not a trust boundary. A stale result, picker bug, or payload that violates the contract can upload files over the quota. The send guard/server will reject the message only after network and storage have already been used, with a possible orphaned blob.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -0,0 +1,137 @@
...
+      (failure) {
+        // a user-cancelled transfer is not a failure; the chip is already gone.
+        if (failure is UploadDriveDocumentFromUrlCancelled) return;
+        request.onFailure(taskId);
+      },
+      (success) => success is UploadDriveDocumentFromUrlSuccess
+          ? request.onSuccess(taskId, success.attachment)
+          : request.onFailure(taskId),
+    );
+  }
```
dab246, line 162: A cancelled upload can still be counted as successful. If the user removes the chip or closes the composer while `uploadFromUrl()` is pending, a late `Right` response can still trigger the success callback and show an incorrect success toast.

**Fix:** check `cancelToken.isCancelled` immediately after `await uploadFromUrl(...)` and add a regression test for this race condition.

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: the validator is already removed from the drive attachment flow, which meant we already accept working with what we have and let server validate the rest. If any information is reported not as it should be, we should let the server be our safety net instead of readdressing validation problem.

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: `RichTextWebController.insertHtml` is a synchronous fire-and-forget bridge call; there is no Future to await, so awaiting would not surface iframe/JS failures either. A synchronous throw is already caught by the `handleDrivePickResult` try/catch at `composer_controller.dart:1043`, which toasts `driveAttachmentTransferFailed` rather than only logging. Making the JS side report insertion failures is a separate editor-bridge change, not something this call site can observe.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
hoangdat: please ensure `onSuccess/onFailure` throw were catched

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: `buildContextMenuTile` is only rendered from `composer_view.dart:498`, the mobile/tablet view, which is a `GetWidget<ComposerController>` with no tag — that composer is registered untagged (`open_and_close_composer_extension.dart:72`, `AppRoutes.composer`). So `getBinding<ComposerController>(tag: null)` in `composer_attachment_extension_registry_provider.dart:43` resolves exactly the composer that opened the menu. Web composers are tagged and use `buildToolbarButton`, which passes its `composerId`; they never build this tile. Passing `null` here is the correct tag, not a missing one.

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: Done

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: Done

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -134,4 +146,17 @@ class DriveAttachmentTransferRunner {
           : request.onFailure(taskId),
     );
   }
+
+  /// Runs a caller callback; a throw is logged and contained to this task.
+  void _notify(String label, void Function() callback) {
+    try {
+      callback();
+    } catch (e, s) {
+      logError(
```
hoangdat, line 173: IMO, not all the failure need to push to Sentry

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,43 +24,120 @@ class DriveAttachmentHandler {
 
   Future<void> handleDrivePickResult(
```
dab246, line 25: Should update this function because CodeScene is flagging it: `Complex Method: DriveAttachmentHandler.handleDrivePickResult has a cyclomatic complexity of 9, threshold = 9`.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -16,43 +24,120 @@ class DriveAttachmentHandler {
...
+      _showSuccessToast(appLocalizations?.driveAttachmentAddedSuccessfully);
+    } else if (linkDocs.isEmpty) {
+      _showFailureToast(appLocalizations?.driveNoValidAttachment);
+    } else {
+      // Docs were linkable but the editor was unbound — nothing reached the body.
+      _showFailureToast(appLocalizations?.driveAttachmentTransferFailed);
+    }
+
+    // Dropped docs would otherwise vanish silently from a pick whose siblings worked.
+    if (droppedCount > 0 && (linkDocs.isNotEmpty || downloadableDocs.isNotEmpty)) {
```
dab246: CodeScene is flagging it: `Complex Conditional: DriveAttachmentHandler.handleDrivePickResult has 1 complex conditionals with 2 branches, threshold = 2`

dab246:
- [x] Add border radius 6px for Drive Picker dialog

https://github.com/user-attachments/assets/ef3adce7-8da5-499f-b630-f2031c75b903

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
```diff
@@ -36,24 +35,33 @@ mixin DrivePickerStateMixin<T extends StatefulWidget> on State<T> {
...
         logError('DrivePickerStateMixin::onPickerTap: state disposed before opening modal');
+        await _dispatch(
+          DrivePickFailure(StateError('drive picker state disposed before modal')),
+        );
         return;
       }
       final l10n = AppLocalizations.of(context)!;
       // Captured up front: the caller may pop this context (e.g. a context
       // menu tile) before the intent future settles, disposing this state.
+      final snapshot = DrivePickerSessionResolver(session).resolve();
```
dab246, line 68: **Handle and cover failures while preparing the picker configuration**
`resolve()` runs before the modal's error boundary. If a session capability or the remaining-size getter throws here, `onPickerTap()` completes with an unhandled asynchronous error instead of dispatching `DrivePickFailure`; this also makes the retry behaviour unverified. Could we convert failures from the pre-modal configuration step into the same visible failure outcome, and add a regression test that verifies no unhandled error is emitted and a subsequent tap can retry successfully?

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1018,18 +1025,72 @@ class ComposerController extends BaseController
...
     }
+    final session = mailboxDashBoardController.sessionCurrent;
+    final accountId = mailboxDashBoardController.accountId.value;
+    final uploadUri = session?.getUploadFromUrlUri(
+      accountId,
+      jmapUrl: jmapUrl,
+    );
+    if (uploadUri == null || accountId == null) {
+      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      return DriveAttachmentTransferRunner.notStartedOutcome;
```
dab246, line 1076: **Add controller-level coverage for unavailable upload endpoints**

The existing `ComposerController` Drive tests cover link insertion, but not the attachment-only path when `jmapUrl` is empty or `Session.getUploadFromUrlUri(...)` returns `null`. Please add a regression test asserting that this path shows the transfer-failure toast, does not add download placeholders, and never starts an upload. This protects the production wiring between `Session`, `AccountId`, the upload-from-URL capability, and the UI layer.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: _notify only fires its logError when a caller's onSuccess/onFailure callback throws (drive_attachment_transfer_runner.dart:146-158) — not on upload failures, which resolve through onFailure and never log. A throwing callback is a programming defect that is otherwise fully invisible: it's swallowed to protect sibling tasks, shows no toast, leaves no chip state. Sentry is the only channel that surfaces it, and volume is bounded by "should never happen", so it isn't quota noise.

hoangdat: please put more information to the log, attachment metadata WDYT?

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

workplace/lib/presentation/mixin/drive_picker_state_mixin.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

dab246 (approved):

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4767

PR #4776, opened 2026-08-24
TF-4728 Sentry logging drive attachment failure

tddang-linagora:
## Issue
- #4728

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when adding drive attachments from links.
  * Upload cancellations now stop cleanly when leaving the upload screen.
  * Pending attachments are preserved during attachment-state refreshes.
  * Drive attachment transfers now update correctly on success or failure.
  * Missing or unavailable transfer services are handled more consistently.

* **Improvements**
  * Upload failures now provide more consistent diagnostic information for support and troubleshooting.
  * Attachment transfer state updates are more responsive and reliable.
  * Upload progress is cleaned up correctly when uploads finish or are cancelled.

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -120,6 +120,7 @@ import 'package:tmail_ui_user/features/network_connection/presentation/network_c
   if (dart.library.html) 'package:tmail_ui_user/features/network_connection/presentation/web_network_connection_controller.dart';
 import 'package:tmail_ui_user/features/server_settings/domain/usecases/get_server_setting_interactor.dart';
 import 'package:tmail_ui_user/features/upload/domain/exceptions/pick_file_exception.dart';
+import 'package:tmail_ui_user/features/upload/domain/exceptions/upload_exception.dart';
 import 'package:tmail_ui_user/features/upload/domain/extensions/file_info_extension.dart';
 import 'package:tmail_ui_user/features/upload/domain/extensions/list_file_upload_extension.dart';
 import 'package:tmail_ui_user/features/upload/domain/model/upload_task_id.dart';
@@ -1069,6 +1070,8 @@ class ComposerController extends BaseController
     if (jmapUrl == null || jmapUrl.isEmpty) {
       logError(
         'ComposerController::_transferDriveDocuments: jmapUrl is unavailable',
+        exception: const UploadFromUrlEndpointUnavailableException('jmapUrl is unavailable'),
+        stackTrace: StackTrace.current,
         extras: {'docCount': docs.length},
       );
       return DriveAttachmentTransferRunner.notStartedOutcome;
@@ -1082,6 +1085,8 @@ class ComposerController extends BaseController
     if (uploadUri == null || accountId == null) {
       logError(
         'ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable',
+        exception: const UploadFromUrlEndpointUnavailableException('upload-from-url uri could not be resolved'),
+        stackTrace: StackTrace.current,
         extras: {
           'docCount': docs.length,
           'accountId': accountId?.id.value,
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1040,8 +1040,13 @@ class ComposerController extends BaseController
         transferDriveDocuments: (docs) => _transferDriveDocuments(docs),
         appLocalizations: currentContext != null ? AppLocalizations.of(currentContext!) : null,
       );
-    } catch (e) {
-      logWarning('ComposerController::handleDrivePickResult:Exception = $e');
+    } catch (e, s) {
+      logError(
+        'ComposerController::handleDrivePickResult failed',
+        exception: e,
+        stackTrace: s,
+        extras: {'docCount': result.length},
+      );
       // A throw here can strand waiting chips, so the failure must be visible.
       getBinding<ToastManager>()?.showMessageFailure(DrivePickFailure(
         e,
@@ -1062,7 +1067,10 @@ class ComposerController extends BaseController
   Future<DriveTransferOutcome> _transferDriveDocuments(List<DriveDocument> docs) async {
     final jmapUrl = dynamicUrlInterceptors.jmapUrl;
     if (jmapUrl == null || jmapUrl.isEmpty) {
-      logWarning('ComposerController::_transferDriveDocuments: jmapUrl is unavailable');
+      logError(
+        'ComposerController::_transferDriveDocuments: jmapUrl is unavailable',
+        extras: {'docCount': docs.length},
+      );
       return DriveAttachmentTransferRunner.notStartedOutcome;
     }
     final session = mailboxDashBoardController.sessionCurrent;
@@ -1072,7 +1080,14 @@ class ComposerController extends BaseController
       jmapUrl: jmapUrl,
     );
     if (uploadUri == null || accountId == null) {
-      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      logError(
+        'ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable',
+        extras: {
+          'docCount': docs.length,
+          'accountId': accountId?.id.value,
+          'hasSession': session != null,
+        },
+      );
       return DriveAttachmentTransferRunner.notStartedOutcome;
     }
 
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1087,9 +1087,9 @@ class ComposerController extends BaseController
         'ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable',
         exception: const UploadFromUrlEndpointUnavailableException('upload-from-url uri could not be resolved'),
         stackTrace: StackTrace.current,
+        // Account id stays out: extras reach Sentry.
         extras: {
           'docCount': docs.length,
-          'accountId': accountId?.id.value,
           'hasSession': session != null,
         },
       );
--- a/lib/features/composer/presentation/extensions/drive_document_extension.dart
+++ b/lib/features/composer/presentation/extensions/drive_document_extension.dart
@@ -10,6 +10,23 @@ extension DriveDocumentExtension on DriveDocument {
   /// Link precedence lives at the call site, not here.
   bool isAttachableAsDownload() =>
       downloadLink?.toString().trim().isNotEmpty ?? false;
+
+  /// Root cause for why a doc failed both attachability checks; used for Sentry triage.
+  String dropReason({required bool requireHttps}) {
+    final hasSharingLink = sharingLink != null;
+    final downloadLinkText = downloadLink?.toString().trim();
+    final hasDownloadLink = downloadLinkText?.isNotEmpty ?? false;
+
+    if (!hasSharingLink && !hasDownloadLink) return 'missing_links';
+    if (hasSharingLink &&
+        !sharingLink!.isSafeLinkScheme(requireHttps: requireHttps)) {
+      return 'unsafe_sharing_scheme';
+    }
+    if (downloadLinkText != null && downloadLinkText.isEmpty) {
+      return 'blank_download_link';
+    }
+    return 'unsupported_link_scheme';
+  }
 }
 
 extension _SafeLinkSchemeExtension on Uri {
--- a/lib/features/composer/presentation/extensions/drive_document_extension.dart
+++ b/lib/features/composer/presentation/extensions/drive_document_extension.dart
@@ -17,13 +17,13 @@ extension DriveDocumentExtension on DriveDocument {
     final downloadLinkText = downloadLink?.toString().trim();
     final hasDownloadLink = downloadLinkText?.isNotEmpty ?? false;
 
+    if (downloadLinkText != null && downloadLinkText.isEmpty) {
+      return 'blank_download_link';
+    }
     if (!hasSharingLink && !hasDownloadLink) return 'missing_links';
     if (hasSharingLink &&
         !sharingLink!.isSafeLinkScheme(requireHttps: requireHttps)) {
       return 'unsafe_sharing_scheme';
-    }
-    if (downloadLinkText != null && downloadLinkText.isEmpty) {
-      return 'blank_download_link';
     }
     return 'unsupported_link_scheme';
   }
--- a/lib/features/composer/presentation/manager/drive_attachment_handler.dart
+++ b/lib/features/composer/presentation/manager/drive_attachment_handler.dart
@@ -2,6 +2,7 @@ import 'package:core/core.dart';
 import 'package:core/utils/html/file_link_card_html_builder.dart';
 import 'package:tmail_ui_user/features/composer/presentation/extensions/drive_document_extension.dart';
 import 'package:tmail_ui_user/features/composer/presentation/manager/drive_attachment_transfer_runner.dart';
+import 'package:tmail_ui_user/features/upload/domain/exceptions/upload_exception.dart';
 import 'package:tmail_ui_user/main/localizations/app_localizations.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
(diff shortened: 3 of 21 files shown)


lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
```diff
@@ -102,6 +113,15 @@ class DriveAttachmentTransferRunner {
...
         if (failure is UploadDriveDocumentFromUrlCancelled) return;
+        logError(
+          'DriveAttachmentTransferRunner::_runOne: upload failed',
+          exception: failure is UploadDriveDocumentFromUrlFailure ? failure.exception : failure,
+          extras: {
+            'taskId': taskId.id,
+            'fileName': task.doc.name,
+            'mimeType': task.doc.mimeType,
+          },
+        );
```
dab246: Avoid reporting upload errors twice. The interactor already reports the exception and returns `UploadDriveDocumentFromUrlFailure`; this branch then reports the same exception again. As a result, a single failed upload generates two Sentry events, including for recoverable network errors. Use a single logging funnel and report only unexpected errors.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1058,7 +1066,14 @@ class ComposerController extends BaseController
...
     if (uploadUri == null || accountId == null) {
-      logWarning('ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable');
+      logError(
+        'ComposerController::_transferDriveDocuments: upload-from-url endpoint is unavailable',
+        extras: {
+          'docCount': docs.length,
+          'accountId': accountId?.id.value,
+          'hasSession': session != null,
+        },
+      );
```
dab246, line 1095: Missing exception & stacktrace

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -232,21 +232,36 @@ class UploadController extends BaseController {
...
         attachment: attachment,
       ),
     );
+    if (!found) {
+      // The chip is gone (composer disposed, list cleared) so the
+      // successful upload never reaches the screen - surface it.
+      logError(
+        'UploadController::resolveDriveTransferSuccess: taskId not found in state list',
+        extras: {'taskId': taskId.id, 'fileName': attachment.name},
+      );
```
dab246: idem

lib/features/upload/presentation/controller/upload_controller.dart
```diff
@@ -232,21 +232,36 @@ class UploadController extends BaseController {
...
   /// Resolves a drive-transfer chip to failed by removing it, matching the
   /// plain-upload failure path.
   void resolveDriveTransferFailure(UploadTaskId taskId) {
-    deleteFileUploaded(taskId);
+    final found = _uploadingStateFiles.deleteElementByUploadTaskId(taskId);
+    if (!found) {
+      logError(
+        'UploadController::resolveDriveTransferFailure: taskId not found in state list',
+        extras: {'taskId': taskId.id},
+      );
```
dab246: idem

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -78,8 +78,10 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
     )) {
       either.fold(
         (failure) {
-          logWarning(
-            'WorkplaceComposerAttachmentExtension::_exchangeAccessToken failed: $failure',
+          logError(
+            'WorkplaceComposerAttachmentExtension::_exchangeAccessToken failed',
+            exception: failure is FeatureFailure ? failure.exception : failure,
+            extras: {'platformUrl': platformUrl.toString()},
           );
```
dab246: Intent failures already have a logging funnel. This exception is rethrown to `DriveIntentMessageHandlerMixin._failWith`, which already reports it to Sentry with the original stack trace. Logging it here creates a duplicate event, so reporting should be handled by a single funnel.

lib/features/composer/presentation/manager/drive_attachment_transfer_runner.dart
tddang-linagora: Done

lib/features/composer/presentation/composer_controller.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart

lib/features/upload/presentation/controller/upload_controller.dart

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

lib/features/upload/domain/usecases/upload_drive_document_from_url_interactor.dart
```diff
@@ -14,9 +15,25 @@ class UploadDriveDocumentFromUrlInteractor {
...
+      // Single logging point for everything the repository/datasource/API
+      // layers throw (account mismatch, Dio errors, JSON parse errors).
+      if (!isCancelled) {
+        logError(
+          'UploadDriveDocumentFromUrlInteractor::execute failed',
+          exception: e,
+          stackTrace: s,
+          extras: {
+            'accountId': request.accountId.id.value,
+            'uploadUri': request.uploadUri.toString(),
```
dab246: IMO, we don't send Upload Uri to Sentry.

lib/features/upload/presentation/controller/upload_controller.dart
dab246: Missing Exception, Stachtrace

dab246: idem

dab246, line 335: A single upload failure can generate multiple Sentry events. The backend error is first logged in the interactor, then `onFailure` propagates to `UploadController._showToastMessageWhenUploadAttachmentsFailure`, which logs the same failure again at here. The `uploadFromUrl` exception path in the runner has the same issue. There should be a single owner for durable error logging—preferably the interactor, where sanitized metadata is available, while the controller should only handle displaying the failure toast.

dab246 (commented):
Shoukl add more regression tests:

- [x] `UploadDriveDocumentFromUrlInteractor` — when the repository throws for a request containing a sensitive upload URL, account identifier, and filename, the emitted error event must not contain any of those values in either `extras` or the log message. The event should retain only allow-listed technical metadata.

- [x] `UploadDriveDocumentFromUrlInteractor` — when the request is cancelled and the repository throws, it must return `UploadDriveDocumentFromUrlCancelled` and emit no error event.

- [x] Drive upload failure flow — when `interactor -> runner -> UploadController.resolveDriveTransferFailure` handles one backend failure, exactly one durable error event must be emitted. This prevents duplicate Sentry issues for the same failed upload.

- [x] `DriveAttachmentTransferRunner` — when the injected `uploadFromUrl` callback throws unexpectedly, it must resolve the matching chip as failed and emit one sanitized error event without the attachment filename or URL.

- [x] `ComposerController` — when `jmapUrl` or the upload-from-URL endpoint is unavailable, the user must see the existing transfer-failure toast and one sanitized diagnostic event must be emitted without the account identifier.

- [x] `UploadController.resolveDriveTransferSuccess` — when a late success arrives after its placeholder has been removed, the chip must not be restored and any diagnostic event must not include the attachment filename.

- [x] `UploadController.resolveDriveTransferFailure` — when the task ID is no longer present, the method must not throw or create duplicate error events; it should produce at most one sanitized diagnostic.

- [x] `UploadFileStateList` — verify the new boolean contract explicitly: `updateElementBy`, `updateElementByUploadTaskId`, and `deleteElementByUploadTaskId` return `true` for an existing item and `false` for an unknown item.

lib/features/upload/domain/usecases/upload_drive_document_from_url_interactor.dart
tddang-linagora: Done

lib/features/upload/presentation/controller/upload_controller.dart

tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4776

lib/features/upload/domain/model/upload_attachment.dart
```diff
@@ -61,10 +61,21 @@ class UploadAttachment with EquatableMixin {
 
       _updateEvent(Right(SuccessAttachmentUploadState(uploadTaskId, attachment, fileInfo)));
     } catch (error, stackTrace) {
-      logWarning('UploadAttachment::upload():ERROR: $error');
       if (error is DioException && error.type == DioExceptionType.cancel) {
         _updateEvent(Left(CancelAttachmentUploadState(uploadTaskId)));
       } else {
+        // Single logging point for attachment upload failures; the controller
+        // only shows the toast.
+        logError(
```
dab246, line 69: Raw `DioException` can still expose upload request URLs in Sentry

This logs the raw `DioException` before it is mapped by `exceptionThrower`. Because `dio.addSentry()` registers `DioEventProcessor`, Sentry can extract `requestOptions.uri` from the exception and attach it as `SentryRequest.url`. Our sanitizer currently preserves `req.url`, so upload endpoints containing account ids can still reach Sentry even though `extras` and `rawMessage` avoid filename/account/url. Please sanitize or map the exception before `logError`, or redact sensitive path/query values in `SentryInitializer._sanitizeRequest`. Add a regression test with a real `DioException` carrying a sentinel upload URL/account id.

Why this matters:

The PR intentionally avoids putting upload URI, download link, account id, and filename into `extras`. That is good, but for `DioException` the sensitive data can travel through the exception object itself and be reconstructed by Sentry integrations.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -106,6 +107,19 @@ class DriveAttachmentHandler {
         downloadableDocs.add(doc);
       } else {
         droppedCount++;
+        final downloadLink = doc.downloadLink;
+        logError(
```
dab246, line 111: Dropped Drive documents are logged one event per document.

This emits one Sentry error event for every dropped Drive document. A picker result containing many invalid documents can create many almost-identical events, while the event intentionally excludes identifying values such as filename/link/doc id. Consider aggregating this per pick result instead: one event with `droppedCount`, reason counts, `requireHttps`, and safe technical metadata. That would reduce Sentry noise and make the signal easier to triage.

Why this matters:

The event is useful because dropped docs would otherwise disappear silently, especially when valid siblings are attached successfully. However, reporting per document can inflate Sentry volume without adding much actionable detail, since privacy rules correctly keep identifying fields out.

lib/features/composer/presentation/manager/drive_attachment_handler.dart
```diff
@@ -106,6 +107,19 @@ class DriveAttachmentHandler {
...
+          'DriveAttachmentHandler::_splitByAttachability: document is not attachable',
+          exception: const DriveDocumentNotAttachableException(),
+          stackTrace: StackTrace.current,
+          extras: {
```
dab246, line 115: Dropped-document events need a safe root-cause reason.

The current metadata says whether sharing/download links exist, but it does not tell why the document was dropped. For Sentry triage, please add a safe `dropReason` or reason counters, for example `missing_links`, `blank_download_link`, `unsafe_sharing_scheme`, or `unsupported_link_scheme`. That keeps URLs out of Sentry while making the event actionable.

Why this matters:

When a Drive attachment is dropped, the likely fixes differ depending on whether the picker returned no links, an empty download link, or a link rejected by release-mode HTTPS requirements. The current boolean metadata is a good start, but it requires more inference than necessary during incident triage.

dab246 (commented):
Add focused tests for the Sentry privacy edge that is not covered yet:

- [x] `UploadAttachment.upload` logs a real `DioException` whose `RequestOptions` contains a sentinel upload URL/account id, and the Sentry-bound request data is redacted.
- [x] `DriveAttachmentTransferRunner._runOne` handles an injected callback that throws a real `DioException` with sentinel request metadata.
- [x]  `SentryInitializer._sanitizeRequest` redacts sensitive URL path/query/header values, including `content-location`, account ids, tokens, and session-like parameters.

lib/features/upload/domain/model/upload_attachment.dart
hoangdat: please igonore it in the scope of upload from drive

lib/features/composer/presentation/manager/drive_attachment_handler.dart
hoangdat: not a big deal, we accept these messy case

lib/features/composer/presentation/manager/drive_attachment_handler.dart
hoangdat: okie, please fix it

tddang-linagora:
>UploadAttachment.upload logs a real DioException whose RequestOptions contains a sentinel upload URL/account id, and the Sentry-bound request data is redacted.

Out of scope per https://github.com/linagora/tmail-flutter/pull/4776#discussion_r3877729124

tddang-linagora:
>DriveAttachmentTransferRunner._runOne handles an injected callback that throws a real DioException with sentinel request metadata.

Done

lib/features/composer/presentation/manager/drive_attachment_handler.dart
tddang-linagora: Done

hoangdat (dismissed):

lib/features/composer/presentation/extensions/drive_document_extension.dart
```diff
@@ -10,6 +10,23 @@ extension DriveDocumentExtension on DriveDocument {
...
+    if (!hasSharingLink && !hasDownloadLink) return 'missing_links';
+    if (hasSharingLink &&
+        !sharingLink!.isSafeLinkScheme(requireHttps: requireHttps)) {
+      return 'unsafe_sharing_scheme';
+    }
+    if (downloadLinkText != null && downloadLinkText.isEmpty) {
+      return 'blank_download_link';
+    }
+    return 'unsupported_link_scheme';
+  }
```
dab246, line 29: The blank download-link case is swallowed by `missing_links`. For a doc with `sharingLink == null` and `downloadLink == Uri.parse('')`, `hasDownloadLink` is false, so line 20 returns `missing_links` before `blank_download_link` can run. Please check `downloadLinkText != null && downloadLinkText.isEmpty` before the missing-links branch, and add a regression assertion for `emptyDownloadLinkDoc`.

test/features/composer/presentation/manager/drive_attachment_transfer_runner_test.dart
```diff
@@ -625,4 +629,167 @@ void main() {
...
+        expect(failedTaskIds, hasLength(1));
+        expect(result, (started: true, succeeded: 0, failed: 1));
+
+        expect(logHandler.errorRecords, hasLength(1));
+        final record = logHandler.errorRecords.single;
+        // The raw DioException is intentionally forwarded as `exception:` so Sentry
+        // can attach it; extras/rawMessage are the only fields this call controls
+        // directly, and neither may carry the sentinel URL. The exception's own
+        // request URL is redacted downstream by
+        // SentryInitializer.sanitizeRequestForTesting, not here.
```
dab246: Please remove or reword this comment. `SentryInitializer.sanitizeRequestForTesting` does not exist in the current tree, so this documents a safety net that the code does not actually provide. Since the team agreed raw `DioException` redaction is out of scope here, the test comment should only say that this unit test verifies the runner-controlled fields (`extras` and `rawMessage`) do not contain the sentinel URL.

lib/features/composer/presentation/extensions/drive_document_extension.dart
tddang-linagora: Done

test/features/composer/presentation/manager/drive_attachment_transfer_runner_test.dart
tddang-linagora: Done

dab246 (approved):
LGTM

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4776

PR #4783, opened 2026-08-27
Sync `flutter_localization` in `Podfile.lock` for iOS

dab246:
## Summary

Synchronize `ios/Podfile.lock` with the Flutter plugin graph by adding the native iOS pod entry for `flutter_localization`.

## Why this is needed

The internal `workplace` package depends on `flutter_localization`, so the root Flutter dependency graph resolves version `0.3.2` and the generated iOS plugin configuration includes it.

Before this change, `ios/Podfile.lock` did not record the corresponding CocoaPods dependency even though Flutter exposes `flutter_localization` as an iOS plugin.

As a result, running the iOS dependency synchronization locally can modify the tracked lockfile unexpectedly. Keeping the lockfile aligned makes the CocoaPods resolution reproducible across developer machines and CI, and prevents unrelated iOS dependency diffs from being mixed into feature or bug-fix pull requests.

### DIFF
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -115,6 +115,8 @@ PODS:
     - Flutter
   - flutter_local_notifications (0.0.1):
     - Flutter
+  - flutter_localization (0.0.1):
+    - Flutter
   - flutter_native_splash (0.0.1):
     - Flutter
   - flutter_secure_storage (6.0.0):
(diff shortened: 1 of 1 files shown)


tddang-linagora (approved):

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4783

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4783

PR #4787, opened 2026-08-28
TF-4790 Upload failed in sequence after a upload failed 

dab246:
## Issue

#4790

After an attachment upload fails, no further upload succeeds. Reopening the app logs the user out.

```
PlatformException(token_failed, Concurrent operations detected: token, token)
DioException [connection error]: null  uri: https://.../upload/<blobId>
HiveError: Cannot read, unknown typeId: 74. Did you forget to register an adapter?
```

## Root cause

`FileUploader` ran each upload in a `worker_manager` isolate. Inside that isolate it called `HiveCacheConfig.instance.setUp()` and used the isolate-tagged `DioClient`, which carries its own `AuthorizationInterceptors`.

So an expired token during upload meant several worker isolates each ran their own token refresh:

- Native AppAuth was invoked concurrently from background isolates → `Concurrent operations detected: token, token`. The refresh never completed, so every later upload kept failing on the stale token.
- Each isolate opened its own Hive owner to persist the refreshed token → box corruption → the next launch could not read the session and fell back to logout.

Two smaller defects came from the same design. The 401 retry rebuilt the request by hand and dropped `onSendProgress` and `CancelToken`, so progress froze after a refresh. And the worker path never passed `CancelToken` to Dio at all, so cancelling an upload only changed the UI while the file kept uploading.

## Solution

- **Uploads run on the root isolate.** `FileUploader` uses the root `DioClient`, so all upload 401s share one authorization owner and one AppAuth caller. No Hive or native auth is initialised from a worker isolate. Uploads stay concurrent — the work is I/O-bound, so the isolate bought nothing.

- **The 401 replay keeps the original request.** On mobile the retry replays `RequestOptions` through `Dio.fetch`, preserving `onSendProgress`, `CancelToken` and timeouts. The replay body is assigned explicitly, because `copyWith` silently falls back to the already-consumed original stream when the rebuilt body is null. Web keeps its existing `Dio.request` path unchanged.

- **Guards restored after merging the two code paths.** Web and mobile previously had separate implementations; merging them lost web's structural guarantee of never touching `dart:io`, so an attachment is treated as byte-backed whenever it runs on web. An attachment with neither a path nor bytes now raises `MissingAttachmentSourceException` instead of uploading an empty body. Charset detection reads at most 256 KiB rather than materialising a whole text attachment on the root isolate, and a probe failure no longer discards an upload the server already accepted.

- **Dead code removed.** `UploadFileArguments`, the isolate closures, and the `static` entry points are gone. `HtmlAnalyzer` moved to `NetworkBindings` since it only lived in `NetworkIsolateBindings` to pick up the isolate `FileUploader`.

See `docs/adr/0106-root-isolate-attachment-upload-authentication.md`.

## New Features
* Uploads can automatically retry after authentication refresh on mobile and web.
* Upload progress and cancellation are preserved during mobile retries.
* Web uploads consistently use byte-based attachment data.
* Clear errors are reported when an attachment source is unavailable.

## Bug Fixes
* Improved handling of large text attachments and unavailable source files.
* Upload failures now provide safer error details.

## Documentation
* Added architectural documentation for the upload authentication flow.


test/features/upload/data/network/file_uploader_test.dart
```diff
@@ -0,0 +1,385 @@
...
+  }) async {
+    final directory = await Directory.systemTemp.createTemp('charset-upload-');
+    addTearDown(() async {
+      await directory.delete(recursive: true);
+    });
+    final file = File('${directory.path}/$fileName');
+    await file.writeAsString(content);
+    final server = await startRecordingUploadServer(receivedBodies);
+
+    return FileUploader(DioClient(Dio()), FileUtils()).uploadAttachment(
```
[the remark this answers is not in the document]
dab246: Is this comment done?

lib/features/upload/data/network/file_uploader.dart
```diff
@@ -46,149 +46,21 @@ class FileUploader {
...
-      } else {
-        return _parsingResponse(
-          resultJson: resultJson,
-          fileName: argsUpload.fileInfo.fileName);
-      }
-    } on DioException catch (exception) {
-      logWarning('FileUploader::_handleUploadAttachmentAction():DioException: $exception');
-
-      throw exception.copyWith(
-        requestOptions: exception.requestOptions.copyWith(data: ''));
```
tddang-linagora, line 154: Ensure to carry the data scrub over so exception doesn't carry the files' bytes

dab246: I kept the scrub at the FileUploader boundary so upload failures no longer propagate a DioException carrying the attachment body. The sanitized exception now clears requestOptions.data, removes the upload extra that can retain the byte-backed stream, and also updates response.requestOptions to avoid keeping a second reference. I added a regression test covering a byte-backed upload failure.

tddang-linagora (dismissed):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4787

test/features/interceptor/authorization_interceptor_upload_retry_test.dart
```diff
@@ -0,0 +1,476 @@
...
+      'http://${server.address.address}:${server.port}/upload/account-id',
+      data: file.openRead(),
+      options: filePathUploadOptions(file.path, sourceBytes.length),
+    ).timeout(const Duration(seconds: 30));
+
+    expect(response.statusCode, HttpStatus.ok);
+    expectReplayedWithRefreshedToken(
+      receivedBodies: receivedBodies,
+      authorizationHeaders: authorizationHeaders,
+      expectedBodyCounts: {base64Encode(sourceBytes): 2},
```
hoangdat, line 326: It make the reader confuse with the product behavior. IMO we should separate the real expectation of product with the technical expectation.

test/features/interceptor/authorization_interceptor_upload_retry_test.dart
```diff
@@ -0,0 +1,476 @@
...
+      dio.post(
+        'http://${server.address.address}:${server.port}/upload/account-id',
+        data: file.openRead(),
+        options: filePathUploadOptions(file.path, sourceBytes.length),
+      ).timeout(const Duration(seconds: 30)),
+      throwsA(isA<DioException>()),
+    );
+
+    // The replay carries _refreshAttemptedKey through copyWith, so a second 401
+    // must give up rather than refresh and retry forever.
```
hoangdat: IMO, it is not true, because right now we retry in a different Dio

test/features/upload/domain/model/upload_attachment_test.dart
hoangdat, line 1: No test for the success path (SuccessAttachmentUploadState), no test for progress events reaching progressState, and no test for a plain non-cancel DioException (a 500) landing as an error state. Those are the common cases; only the two edge cases are covered.

test/features/upload/data/network/file_uploader_test.dart
```diff
@@ -0,0 +1,462 @@
...
+        ),
+        Uri.parse('http://${server.address.address}:${server.port}/upload/account-id'),
+      ).timeout(const Duration(seconds: 30));
+
+      expect(capturedUploadExtra.containsKey(FileUploader.filePathExtraKey), isFalse);
+      expect(capturedUploadExtra[FileUploader.streamDataExtraKey], isA<Stream<List<int>>>());
+    });
+
+    test('resolves the charset of a text/plain attachment from bytes', () async {
+      final sourceBytes = Uint8List.fromList(utf8.encode('hello charset'));
```
hoangdat: 13-byte payload never reaches the sublistView cap — bump past 256 KiB so truncation is covered.

test/features/upload/data/network/file_uploader_test.dart
```diff
@@ -0,0 +1,462 @@
...
+        bytes: Uint8List.fromList(<int>[1, 2, 3]),
+        type: 'application/pdf',
+      ),
+      Uri.parse('http://${server.address.address}:${server.port}/upload/account-id'),
+    ).timeout(const Duration(seconds: 30));
+
+    expect(attachment.charset, isNull);
+  });
+
+  test('keeps a completed upload when the charset probe fails', () async {
```
hoangdat, line 407: Covers the file-read failure only. Add a case where getCharsetFromBytes throws to cover the rest.

test/features/interceptor/authorization_interceptor_upload_retry_test.dart
dab246: The product expectation is already asserted separately by checking that both uploads complete with 200 OK. expectReplayedWithRefreshedToken intentionally verifies the transport-level regression contract: each original payload is delivered again, unchanged, with the refreshed token. The body-count map also avoids making the concurrent-upload test order-dependent. I kept this structure because splitting it would duplicate the same technical assertions without improving the observable behavior under test.

test/features/interceptor/authorization_interceptor_upload_retry_test.dart
dab246: _refreshAttemptedKey is propagated, but it is not what prevents another retry in this path: the replay uses a separate Dio instance without interceptors. I updated the comment to state that the second 401 is propagated after the single replay rather than starting another refresh cycle. The test still verifies the intended one-refresh/one-replay behavior.

test/features/upload/domain/model/upload_attachment_test.dart
dab246: I added production-path tests using a real FileUploader and local HTTP server to verify: successful upload emits SuccessAttachmentUploadState, send progress reaches progressState, and a non-cancel HTTP 500 emits ErrorAttachmentUploadState.

test/features/upload/data/network/file_uploader_test.dart
dab246: The byte-backed web charset test now uses a payload larger than 256 KiB and asserts that the detector receives exactly the first 256 KiB, while the server still receives the complete payload.

test/features/upload/data/network/file_uploader_test.dart
dab246: I added a FileUtils test double that throws from getCharsetFromBytes. The test verifies that the completed upload is preserved and the resulting attachment has an unknown (null) charset.

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4787

PR #4792, opened 2026-08-28
[WIP] [ADR] Save to Drive

hoangdat:
## Summary by CodeRabbit

* **Documentation**
  * Added an architectural proposal for saving email attachments directly to Drive.
  * Describes a streamlined flow that transfers attachments without routing file data through the client.
  * Documents the “Save to Drive” and “See in Drive” experience, along with related design considerations and open questions.

### DIFF
--- a/docs/adr/0106-save-email-attachment-to-drive.md
+++ b/docs/adr/0106-save-email-attachment-to-drive.md
@@ -1,0 +1,57 @@
+# 106. Save Email Attachment to Drive
+
+Date: 2026-08-28
+
+## Status
+
+Proposed
+
+## Reference
+
+- Builds on [ADR-0095](0095-external-drive-file-picker-integration.md)
+  (Drive token, workplace FQDN). Inverse of
+  [ADR-0105](0105-attach-drive-file-via-jmap-mediated-upload.md)
+  (Drive→Mail fetch). Mail BE shipped
+  `unauthenticatedBlobAccess.adoc` (tmail-backend jmap-extensions).
+  Drive import-from-URL is not documented yet.
+
+## Context
+
+Users save email attachments to Drive. Bytes must not transit the
+Mail client (memory, double hop).
+
+## Decision
+
+- **BE-to-BE transfer.** Client mints a short-lived blob URL, then
+  asks Drive to fetch and store it. Client never downloads or
+  re-uploads bytes.
+- **Mint via Mail BE.** Gate on session capability
+  `com:linagora:params:jmap:unauthenticated:blob:access`. Call
+  `UnauthenticatedBlobAccess/set` create keyed by `blobId` (empty
+  object). Build GET URL from capability `endpoint` + returned
+  `token`. TTL is server-set (`tokenTtlInSeconds`).
+- **Hand off to Drive.** Authenticated Drive call with blob URL. 
+  Drive GETs Mail within TTL and writes the file. 
+- **No folder picker.** Drive writes to a well-known auto folder
+  it owns. User organizes later in Drive.
+- **Attachment UX.** Each attachment shows `Save to Drive` until
+  lookup/save says it exists, then `See in Drive` (opens
+  `openUrl`). Mail should store flags `save_to_drive_1`, 
+  `save_to_drive_2`, `save_to_drive_*` to indicate attachments were save to Drive
+
+## Consequences
+
+- Client stays memory-flat; large files are a Drive/Mail TTL
+  concern, not a client buffer.
+- Blocked on Drive shipping import-from-URL, source lookup, and
+  auto-folder (request/response schema TBD).
+
+## Open questions
+
+- Exact Drive endpoint, auto-folder name, and `openUrl` shape.
+- Whether a later re-save should copy instead of being
+  idempotent; default is idempotent.
+
+## Sources
+
+- tmail-backend `unauthenticatedBlobAccess.adoc`: https://github.com/linagora/tmail-backend/blob/master/docs/modules/ROOT/pages/tmail-backend/jmap-extensions/unauthenticatedBlobAccess.adoc


docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+- **BE-to-BE transfer.** Client mints a short-lived blob URL, then
+  asks Drive to fetch and store it. Client never downloads or
+  re-uploads bytes.
+- **Mint via Mail BE.** Gate on session capability
+  `com:linagora:params:jmap:unauthenticated:blob:access`. Call
+  `UnauthenticatedBlobAccess/set` create keyed by `blobId` (empty
+  object). Build GET URL from capability `endpoint` + returned
+  `token`. TTL is server-set (`tokenTtlInSeconds`).
+- **Hand off to Drive.** Authenticated Drive call with blob URL. 
+  Drive GETs Mail within TTL and writes the file.
```
Crash--, line 34: do you want a callback / status of this upload ? or is it a fire and forget from your side?

docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+## Decision
+
+- **BE-to-BE transfer.** Client mints a short-lived blob URL, then
+  asks Drive to fetch and store it. Client never downloads or
+  re-uploads bytes.
+- **Mint via Mail BE.** Gate on session capability
+  `com:linagora:params:jmap:unauthenticated:blob:access`. Call
+  `UnauthenticatedBlobAccess/set` create keyed by `blobId` (empty
+  object). Build GET URL from capability `endpoint` + returned
+  `token`. TTL is server-set (`tokenTtlInSeconds`).
```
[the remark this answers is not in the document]
chibenwa: UnauthenticatedBlobAccess/set was added a validUntil property

https://github.com/linagora/tmail-backend/pull/2594

docs/adr/0106-save-email-attachment-to-drive.md
chibenwa: IMO drive modal is responsible of the transfer (as mail gives it the download link) so it's a drive modal concern.

BTW backend to backend is fast and reliable so fire - success is likely OK

docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+- **Hand off to Drive.** Authenticated Drive call with blob URL. 
+  Drive GETs Mail within TTL and writes the file. 
+- **No folder picker.** Drive writes to a well-known auto folder
```
Crash--, line 35: @poupotte I just checked the figma, but we only have the scenario where the user choose the folder destination. 

can you've something for the use case without any folder selection?

poupotte: Yes it's in progress on design team. I'll ping you when it will be added on the figma

docs/adr/0106-save-email-attachment-to-drive.md
hoangdat: IMO, to make the best UX for user, 
- we should also `retry` when `save to drive` failed
- we dont want to block user with the `transfer dialog`, user can easily to do other actions in mail during the transfering

So maybe we should depend on API call than the iframe for this case?

docs/adr/0106-save-email-attachment-to-drive.md
chibenwa: Gmail proposes a concevnience "Mail" folder to save the user 1 click...

docs/adr/0106-save-email-attachment-to-drive.md
dab246, line 34: Please define a stable idempotency key for each attachment. A retry generates a new blob URL/token, so a lost response after a successful import can create duplicate files.

docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+  it owns. User organizes later in Drive.
+- **Attachment UX.** Each attachment shows `Save to Drive` until
+  lookup/save says it exists, then `See in Drive` (opens
+  `openUrl`). Mail should store flags `save_to_drive_1`, 
+  `save_to_drive_2`, `save_to_drive_*` to indicate attachments were save to Drive
```
dab246, line 40: save_to_drive_1 / save_to_drive_2 are not stable attachment identifiers. Use a stable sourceRef (scoped by account, email, and blob) and let Drive return the document ID and openUrl

docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+
+Users save email attachments to Drive. Bytes must not transit the
+Mail client (memory, double hop).
+
+## Decision
+
+- **BE-to-BE transfer.** Client mints a short-lived blob URL, then
+  asks Drive to fetch and store it. Client never downloads or
+  re-uploads bytes.
+- **Mint via Mail BE.** Gate on session capability
```
dab246, line 28: The blob URL is a bearer secret. Explicitly forbid logging, telemetry, caching, or persisting this URL/token, and require HTTPS with a strict redirect policy.

docs/adr/0106-save-email-attachment-to-drive.md
dab246, line 34: The Drive import contract is still TBD. Define the request/response schema, ownership checks, timeout/retry behaviour, and failures before accepting this ADR.

docs/adr/0106-save-email-attachment-to-drive.md
```diff
@@ -0,0 +1,57 @@
...
+  `com:linagora:params:jmap:unauthenticated:blob:access`. Call
+  `UnauthenticatedBlobAccess/set` create keyed by `blobId` (empty
+  object). Build GET URL from capability `endpoint` + returned
+  `token`. TTL is server-set (`tokenTtlInSeconds`).
+- **Hand off to Drive.** Authenticated Drive call with blob URL. 
+  Drive GETs Mail within TTL and writes the file. 
+- **No folder picker.** Drive writes to a well-known auto folder
+  it owns. User organizes later in Drive.
+- **Attachment UX.** Each attachment shows `Save to Drive` until
```
dab246, line 37: Define idle → saving → saved | failed state per attachment and coalesce repeated taps while saving.

PR #4798, opened 2026-08-29
fix: highlight Bcc recipient field during drag-and-drop

### LINKED ISSUE #3074: [Bug] When drag and dropping contact to Bcc, the field isn't highlighted
0.11.4, tmail.linagora.com, tested on Safari, Opera, Firefox

1. Open composer 
2. Enter some recipient address 
4. Drag and drop the address to the Bcc field 
5. The field isn't highlighted the same way is is for drag-and-drop to "To" and "Cc" fields

Expectation: the field is highlighted when I drag the address over this field 

https://github.com/user-attachments/assets/5fb40091-a46e-46da-b4b5-1577f74c61df

usmamalik345:
## Summary
- Fix [#3074](https://github.com/linagora/tmail-flutter/issues/3074): Bcc recipient field did not show the drop highlight while To/Cc did.
- Drive the highlight from `DragTarget.candidateData` instead of manual `_isDragging` updates in `onMove`/`onLeave`, which could miss rebuilds (especially on a focused empty Bcc field).
- Keep `focusedBorderColor` aligned with the highlight so TagEditor does not clear the border while focused.

## Test plan
- [x] `flutter test test/features/composer/recipient_composer_widget_test.dart` (17 tests, including regression for #3074)
- [x] `flutter analyze` on the changed widget file
- [x] Manual: open composer on web → add a To recipient → enable Cc and Bcc → drag the To chip over Bcc and confirm the Bcc field border highlights the same way as To/Cc

Fixes #3074

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recipient drag-and-drop behavior so the destination field visibly highlights while an address is dragged over it.
  * Fixed highlighting for Bcc recipient fields during address movement.

* **Tests**
  * Added coverage to verify drag-and-drop highlighting appears correctly across recipient fields.

### DIFF
--- a/lib/features/composer/presentation/widgets/recipient_composer_widget.dart
+++ b/lib/features/composer/presentation/widgets/recipient_composer_widget.dart
@@ -127,7 +127,6 @@ class _RecipientComposerWidgetState extends State<RecipientComposerWidget> {
 
   Timer? _gapBetweenTagChangedAndFindSuggestion;
   int _tagIndexFocused = -1;
-  bool _isDragging = false;
   late List<EmailAddress> _currentListEmailAddress;
 
   @override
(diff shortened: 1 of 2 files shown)


hoangdat:
Hi @usmamalik345 , thank for your contrinution.

We did not reproduce the issue in master branch, and the issue was fixed in the commit: https://github.com/linagora/tmail-flutter/commit/69d2e59f3

So I propose to close the PR.

Thanks and BRs

tddang-linagora (commented):
Thanks for the PR. The targeting gh issue is quite old and hasn't been able to be reproduced for a while. Can the bug be reproduced and confirm fixed from your side?

usmamalik345:
Thanks for the review @tddang-linagora.

I agree #3074 is quite old. I also could not reliably reproduce the missing Bcc highlight on the current app UI in a full manual session from this environment.

What I did verify:

1. **Code-level cause**  
   The highlight previously depended on a manual `_isDragging` flag updated in `onMove` / `onLeave` + `setState`. That is fragile compared with Flutter’s intended `DragTarget.candidateData` path, and it can miss rebuilds (especially with a focused empty Bcc/`TagEditor` field). To/Cc could still look fine while Bcc did not.

2. **Automated regression**  
   Added a widget test that drags a To recipient over the Bcc `DragTarget` and asserts `TagEditor.enableBorder` becomes `true`. It passes with this change:
   `flutter test test/features/composer/recipient_composer_widget_test.dart` (includes the #3074 case).

3. **CI**  
   Prebuild / Patrol / Jenkins checks on this PR are green.

So this PR is mainly a correctness fix for the drop-highlight mechanism (aligned with how `DragTarget` is meant to be used), plus a regression guard for Bcc.

Happy to:
- close this if you prefer not to land it without a live UI repro, or
- keep it as a hardening fix if that is acceptable.

Please let me know which you prefer.

test/features/composer/recipient_composer_widget_test.dart
```diff
@@ -634,5 +636,95 @@ void main() {
...
+                  prefix: PrefixEmailAddress.cc,
+                  listEmailAddress: <EmailAddress>[],
+                  imagePaths: imagePaths,
+                  maxWidth: 600,
+                  isTestingForWeb: true,
+                  composerId: composerId,
+                ),
+                RecipientComposerWidget(
+                  prefix: PrefixEmailAddress.bcc,
+                  listEmailAddress: <EmailAddress>[],
```
tddang-linagora, line 676:
```suggestion
                  listEmailAddress: const <EmailAddress>[],
                  imagePaths: imagePaths,
                  maxWidth: 600,
                  isTestingForWeb: true,
                  composerId: composerId,
                ),
                RecipientComposerWidget(
                  prefix: PrefixEmailAddress.bcc,
                  listEmailAddress: const <EmailAddress>[],
```

tddang-linagora (approved):
Just to reduce Dart analyzer's noise. Logic and functionality are tested and worked fine.

dab246:
Hi @usmamalik345, thank you for contributing this fix to the project. Moving the drag-highlight state to `DragTarget.candidateData` is a sensible direction: Flutter owns the candidate lifecycle, the duplicated `_isDragging` state is removed, and I do not see a production memory leak or a recipient-transfer regression introduced by this refactor.

Before merging, could we please strengthen and correct the regression coverage?

The current test does not yet prove the reported issue is fixed:

1. It would also pass with the previous implementation. Flutter invokes `onMove` after a target is entered, and the old code already set `_isDragging = true` in that callback. Please add a reproducible case that fails before this change and passes after it. If that cannot be reproduced, the change should be described as a state-management refactor rather than a verified bug fix.

2. `isTestingForWeb: true` on `RecipientComposerWidget` does not enable `PlatformInfo.isTestingForWeb`. The Bcc autofocus path reads `PlatformInfo.isWeb` directly, so the existing VM test does not reproduce an empty, focused Bcc field on web.

3. The added test currently fails on Chrome: after `gesture.cancel()`, a 40 ms gesture timer remains pending at teardown. Please advance the clock (for example, `await tester.pump(const Duration(milliseconds: 50));`) or settle the tree, then assert that the highlight is cleared.

Please cover these cases:

- Web focused-empty-Bcc: enable and reset `PlatformInfo.isTestingForWeb`, give Bcc an explicit `FocusNode`, ensure it has focus, drag a To recipient over Bcc, and verify the enabled border / highlight is shown. Avoid another empty auto-focused field such as Cc in this fixture, or explicitly request Bcc focus.
- Leave and cancel: after leaving Bcc or cancelling the gesture, verify its highlight is disabled. This prevents a stale highlighted field.
- Target transition: drag To → Cc → Bcc and verify Cc is unhighlighted as Bcc becomes highlighted.
- Rebuild while hovering: trigger a parent rebuild or a focus transition while Bcc remains a drag candidate; its highlight must remain visible.
- Same-composer transfer: dropping To → Bcc updates the Bcc list and removes from To exactly once.
- Same-field drop: dropping To → To must not update lists or invoke source removal, while the highlight still clears.
- Duplicate target recipient: preserve and explicitly assert the current product contract, including whether the source recipient is removed when the destination already contains the same address.
- Cross-composer transfer: verify destination add and originating-composer removal each occur exactly once.
- Disposal during drag: remove the highlighted Bcc widget, then cancel the gesture; assert no exception, no `setState() after dispose`, and no pending timer.
- Native platform coverage: run the hover/cancel scenario with both `TargetPlatform.android` and `TargetPlatform.iOS`, covering the mobile `keyboardType` and `childWhenDragging` branches.

The `candidateData` logic itself is shared by all recipient fields, so a complete Bcc scenario covers the new highlight branch; there is no need to duplicate every case for To/Cc/Bcc/Reply-To unless their behavior differs.

lib/features/composer/presentation/widgets/recipient_composer_widget.dart
```diff
@@ -194,132 +193,128 @@ class _RecipientComposerWidgetState extends State<RecipientComposerWidget> {
...
+                              }
+                              _updateListEmailAddressAction();
+                              tagEditorState.resetTextField();
+                              tagEditorState.closeSuggestionBox();
+                            },
+                          );
+                        },
+                      );
                     },
+                    onWillAcceptWithDetails: (_) => true,
```
dab246, line 312: `onWillAcceptWithDetails: (_) => true` intentionally retained to reject a nullable `Draggable.data` value from candidateData? All current `DraggableEmailAddress` sources provide non-null data, so omitting this callback preserves accepted-drag behavior for supported payloads. If the null-payload guard is intentional, please add a short comment explaining that contract, otherwise, it can be removed as a small simplification.

hoangdat:
hi @usmamalik345 , please continue on this too

hoangdat:
Hi @usmamalik345 

Thanks again for your contribution! Since we haven’t had any updates for a while, we’ll close this PR for now to keep things tidy.

You’re always welcome to come back to it or open a new PR whenever you have time. We’d be very happy to have your contribution!

Issue #4713, opened 2026-07-16
Reduce loading time in file picker


poupotte:
# Summary

* [Related EPIC](#related-epic)
* [Definition](#definition)

## Related EPIC

* https://github.com/linagora/tmail-flutter/issues/4376

## Definition

Reduce time between moment that user clic on button and moent where user can select his file

[Back to Summary](#summary)

chibenwa:
I triggered an ongoing discussion by mail on that very topic yesterday.

To quote @Crash-- ( cc @zatteo )

```
We knew we would eventually run into performance issues. We've been talking about it for a few weeks already, and we only started collecting meaningful metrics last week now that we have something functional.

We've also dedicated someone to work on this full time since this morning (Théo).

First, have a look at this benchmark for file pickers: https://github.com/Crash--/webmail-cloud-picker-benchmark#results-so-far

Google is far from achieving a 2-second loading time, so we're definitely not the only ones facing this challenge. That said, I think we should be able to get close to that target (2s), at least for warm loads.

Then, I looked at your screenshot and was quite surprised (!!!) to see all those 304 responses... The former Cozy team recently migrated the build tooling from Webpack to Rsbuild, and it looks like this changed the length of the generated asset hashes. As a result, Cozy Stack stopped applying the proper cache headers because it expected hashes longer than 10 characters: https://github.com/linagora/cozy-libs/pull/3079 I'm also pretty sure we have an issue with font.css: we shouldn't be making two requests for it.

Beyond that, there are still plenty of optimizations we can make:

We should reduce the number of HTTP requests made by TMail. Ideally, we'd only make the request needed to load the iframe.
We should investigate whether we can reduce the duration of the token_exchange request. Even if we get down to a single HTTP request, we'll still need to perform that exchange on the backend.
We'll reduce the download size of the File Picker to improve cold-start performance.
We'll reduce the number of requests made by Drive when running in File Picker mode.
Regarding the _find request taking around 500 ms, let's wait for the real implementation of the File Picker before drawing conclusions. The current PoC may simply be using a suboptimal index.
Only after we've done all of that should we consider adding caching on the TMail side. If we can't get down to a single HTTP request, then it might make sense for Mail to cache the exchanged token.

So today I don't think we should organize a meeting on the subject yet. Let's wait few days. 
```

TLDR: that is not a tmail issue, more a platform / drive issue @poupotte

zatteo:
[...]

PR #4800, opened 2026-08-31
TF-4713 ADR File picker speed up

tddang-linagora:
## Issue
- #4713 

## Summary by CodeRabbit

* **Documentation**
  * Added ADR-0106 documenting Drive token prefetching and caching.
  * Describes session-load prefetch, shared token reuse, request coordination, and recovery from authorization failures.
  * Documents cache invalidation behavior, error handling, scope exclusions, and related architectural decisions.

### DIFF
--- a/docs/adr/0106-drive-picker-token-prefetch-and-cache.md
+++ b/docs/adr/0106-drive-picker-token-prefetch-and-cache.md
@@ -1,0 +1,42 @@
+# 106. Drive Picker Token Prefetch and Cache
+
+Date: 2026-08-31
+
+## Status
+
+Proposed
+
+## Reference
+
+- Builds on [ADR-0095](0095-external-drive-file-picker-integration.md) and [ADR-0105](0105-attach-drive-file-via-jmap-mediated-upload.md), both unchanged.
+
+## Context
+
+- Every Drive tap pays two sequential round-trips before the picker renders: `POST /auth/token_exchange` (OIDC `id_token` → Drive access token), then `POST /intents`.
+- The exchanged token is discarded after each tap; the skeleton stays on screen for both calls.
+- TF-4713.
+
+## Decision
+
+- **Prefetch trigger is session load.** Fire as soon as the Drive platform URI is available, not at composer open, not at login as a separate hook.
+- **One in-memory store.** `WorkplaceTokenStore` (`obtain` / `recoverAfterUnauthorized` / `prime` / `clear`) lives on the keepAlive `WorkplaceComposerAttachmentExtension`. Composers never own tokens.
+- **No TTL.** cozy-stack does not return `expires_in`. Cache the session until Drive rejects it.
+- **Refresh is reactive only.** It runs after a failed `POST /intents`, never on a timer, never during prefetch.
+- **Coordination is cache-or-join.** Return the cached session, or await the single in-flight `Future`. No generation counters, no `force` flags, no `_isRefreshing` booleans.
+- **401 handling (`recoverAfterUnauthorized`):** refresh via `POST /auth/access_token` (`grant_type=refresh_token`) when refresh token + client id/secret are all present; otherwise fall back to a full re-exchange. Either path retries that composer's `/intents` exactly once. A second 401 is not recovered — it surfaces as the existing modal failure.
+- **Token sharing, not intent sharing.** The store shares one access token across composers; each composer still gets its own `/intents` call and its own modal.
+- **Invalidation is URI-null, not composer-close.** `clear()` fires on logout, account switch, Drive preference off, or capability off. Composer close, all-composers-close, and OIDC `id_token` refresh are no-ops for the store.
+- **Double-tap stays a UI concern.** The existing `DrivePickerStateMixin._modalOpen` guard owns same-button double-click; the store is not involved.
+- **Out of scope:** Dio interceptor, disk persistence, intent prefetch, TTL/`expiresAt`/JWT `exp`, priming from `ComposerController.onInit`, a dedicated refresh interactor/state (the store calls the repository directly).
+
+## Consequences
+
+- First Drive tap after session load is faster: it only pays for `POST /intents` when the store is warm.
+- New failure mode: a cached token can go stale between prefetch and tap; mitigated by the 401 recovery path, with no third attempt.
+- `workplace` gains one new store abstraction (`WorkplaceTokenStore` / `InMemoryWorkplaceTokenStore`); `lib/` stays token-agnostic and only reads the registry.
+- Worst case matches today's behavior: if prefetch failed or never ran, the tap pays for the exchange itself.
+
+## Sources
+
+- [ADR-0095: External drive file picker integration](0095-external-drive-file-picker-integration.md)
+- [ADR-0105: Attach Drive File via JMAP-mediated Upload](0105-attach-drive-file-via-jmap-mediated-upload.md)


docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+- **Token sharing, not intent sharing.** The store shares one access token across composers; each composer still gets its own `/intents` call and its own modal.
+- **Invalidation is URI-null, not composer-close.** `clear()` fires on logout, account switch, Drive preference off, or capability off. Composer close, all-composers-close, and OIDC `id_token` refresh are no-ops for the store.
+- **Double-tap stays a UI concern.** The existing `DrivePickerStateMixin._modalOpen` guard owns same-button double-click; the store is not involved.
+- **Out of scope:** Dio interceptor, disk persistence, intent prefetch, TTL/`expiresAt`/JWT `exp`, priming from `ComposerController.onInit`, a dedicated refresh interactor/state (the store calls the repository directly).
+
+## Consequences
+
+- First Drive tap after session load is faster: it only pays for `POST /intents` when the store is warm.
```
chibenwa, line 34: I guess that in practice the drive token has a TTL of its own and that more often by not in long twake mail usage session it would be invalidated.

And instead of 1 token_exchange + 1 intent call I'd get 1 failed intent call 1 token_exchange + 1 intent call making the problem actually worse.

So I am sceptical about true benefits of this approach appart during demo day.

Cc @Crash--

chibenwa: To make it viable a background renewal of the drive token would be needed but it sucks.

BTW that means that every twake mail page load triggers a token exchange even if 1% of users use file picker. This dis-symetry needs to be accounted for in the performance profile.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Context
+
+- Every Drive tap pays two sequential round-trips before the picker renders: `POST /auth/token_exchange` (OIDC `id_token` → Drive access token), then `POST /intents`.
+- The exchanged token is discarded after each tap; the skeleton stays on screen for both calls.
+- TF-4713.
+
+## Decision
+
+- **Prefetch trigger is session load.** Fire as soon as the Drive platform URI is available, not at composer open, not at login as a separate hook.
```
dab246, line 21: How is this started at session load? Today `WorkplaceComposerAttachmentExtension` is created only when composer UI watches the registry. Until a composer opens, no store exists and `prime()` cannot run. Please define a session-scoped bootstrap owner; otherwise the first picker tap is still cold.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+- **Refresh is reactive only.** It runs after a failed `POST /intents`, never on a timer, never during prefetch.
+- **Coordination is cache-or-join.** Return the cached session, or await the single in-flight `Future`. No generation counters, no `force` flags, no `_isRefreshing` booleans.
+- **401 handling (`recoverAfterUnauthorized`):** refresh via `POST /auth/access_token` (`grant_type=refresh_token`) when refresh token + client id/secret are all present; otherwise fall back to a full re-exchange. Either path retries that composer's `/intents` exactly once. A second 401 is not recovered — it surfaces as the existing modal failure.
+- **Token sharing, not intent sharing.** The store shares one access token across composers; each composer still gets its own `/intents` call and its own modal.
+- **Invalidation is URI-null, not composer-close.** `clear()` fires on logout, account switch, Drive preference off, or capability off. Composer close, all-composers-close, and OIDC `id_token` refresh are no-ops for the store.
+- **Double-tap stays a UI concern.** The existing `DrivePickerStateMixin._modalOpen` guard owns same-button double-click; the store is not involved.
```
dab246, line 29: URI-null is not an account boundary. Two users on the same Workplace tenant can have the same URI, while this keepAlive store survives account changes. Scope the cache by authenticated session/account identity and clear it before auth teardown. Please add an `A → logout/switch → B`, same-FQDN regression test.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+- The exchanged token is discarded after each tap; the skeleton stays on screen for both calls.
+- TF-4713.
+
+## Decision
+
+- **Prefetch trigger is session load.** Fire as soon as the Drive platform URI is available, not at composer open, not at login as a separate hook.
+- **One in-memory store.** `WorkplaceTokenStore` (`obtain` / `recoverAfterUnauthorized` / `prime` / `clear`) lives on the keepAlive `WorkplaceComposerAttachmentExtension`. Composers never own tokens.
+- **No TTL.** cozy-stack does not return `expires_in`. Cache the session until Drive rejects it.
+- **Refresh is reactive only.** It runs after a failed `POST /intents`, never on a timer, never during prefetch.
+- **Coordination is cache-or-join.** Return the cached session, or await the single in-flight `Future`. No generation counters, no `force` flags, no `_isRefreshing` booleans.
```
dab246, line 25: `clear()` is unsafe while prefetch or refresh is in flight: a request started by account A can complete after clear and repopulate the cache for account B. Cache-or-join does not cover late completions. Add a session generation/epoch or cancellation, and commit results only when the scope is still current.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+- **401 handling (`recoverAfterUnauthorized`):** refresh via `POST /auth/access_token` (`grant_type=refresh_token`) when refresh token + client id/secret are all present; otherwise fall back to a full re-exchange. Either path retries that composer's `/intents` exactly once. A second 401 is not recovered — it surfaces as the existing modal failure.
```
dab246, line 26: The current exchange contract returns only String `accessToken`, `refreshToken`, `clientId`, and `clientSecret` are discarded. Define a typed token-grant contract for exchange and refresh, including refresh-token rotation and error handling. Also confirm that retaining `client_secret` is safe on Flutter Web, confidential secrets cannot be protected in a browser.

dab246, line 26: A stale cached token makes the first tap perform POST `/intents → 401 → refresh/exchange → retry /intents`: three sequential requests, compared with two today. This contradicts the “worst case matches today” claim. Request expires_in/expires_at from Platform, or explicitly measure and accept this idle-session regression.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+- **Token sharing, not intent sharing.** The store shares one access token across composers; each composer still gets its own `/intents` call and its own modal.
+- **Invalidation is URI-null, not composer-close.** `clear()` fires on logout, account switch, Drive preference off, or capability off. Composer close, all-composers-close, and OIDC `id_token` refresh are no-ops for the store.
+- **Double-tap stays a UI concern.** The existing `DrivePickerStateMixin._modalOpen` guard owns same-button double-click; the store is not involved.
+- **Out of scope:** Dio interceptor, disk persistence, intent prefetch, TTL/`expiresAt`/JWT `exp`, priming from `ComposerController.onInit`, a dedicated refresh interactor/state (the store calls the repository directly).
```
dab246, line 30: A presentation-owned store calling the repository directly bypasses the project dependency direction. Please inject a token interactor/gateway through bindings instead of letting the extension own repository access.

docs/adr/0106-drive-picker-token-prefetch-and-cache.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Consequences
+
+- First Drive tap after session load is faster: it only pays for `POST /intents` when the store is warm.
+- New failure mode: a cached token can go stale between prefetch and tap; mitigated by the 401 recovery path, with no third attempt.
+- `workplace` gains one new store abstraction (`WorkplaceTokenStore` / `InMemoryWorkplaceTokenStore`); `lib/` stays token-agnostic and only reads the registry.
+- Worst case matches today's behavior: if prefetch failed or never ran, the tap pays for the exchange itself.
```
dab246, line 37: This improves only warm token-exchange latency. The issue discussion also identifies Platform/Drive bottlenecks: asset cache headers, duplicate font requests, picker bundle size, and Drive requests. Add measurable acceptance criteria for tap → picker interactive, request count/bytes, cache-hit rate, and session-start regression before accepting this ADR.

dab246, line 37: Please define prime() failure semantics. It must not create an unhandled async error, must release the in-flight state after failure/cancellation, and a later tap must retry exchange normally. Add coverage for this path.

tddang-linagora:
Close for #4804

PR #4801, opened 2026-09-01
Fix reply-to defaulting to identity email when unset

### LINKED ISSUE #2651: Reply-to in Identities should match the identity email if empty
### Description

When creating an other identity with reply-to field to null, if I use that identity to send a mail, the reply-to field seems to be set to the default identity one. 

### Expected result

I think it should default to the mail of the identity instead.

### Current behavior

reply-to-tm-issue

#### Reproduction Steps

- let's say my default identity is bob
- create a new identity (not default) -> alice
- leave the reply-to field to none
- send a mail selecting that new identity
- the reply-to field is set to bob and not alice

usmamalik345:
## Summary
- Sync reply-to recipients when switching identities in the composer (mirrors existing BCC behavior)
- Default Reply-To to the selected identity email when a non-default identity inherits the account primary reply-to address
- Add regression tests for identity switching and reply-to header generation

Fixes #2651

## Test plan
- [x] \lutter test test/features/composer/presentation/extensions/create_email_request_extension_test.dart\
- [x] \lutter test test/features/composer/presentation/composer_controller_test.dart\
- [x] \lutter analyze\ on changed files
- [ ] Manual: create non-default identity with reply-to = none, send mail, verify Reply-To header matches identity email

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

- **Bug Fixes**
  - Switching identities now removes Reply-To addresses associated with the former identity.
  - Selecting a new identity no longer automatically copies its configured Reply-To addresses into the composer.
  - Explicitly configured Reply-To addresses matching the account email are preserved.
  - The Reply-To field is disabled when no addresses remain after switching identities.
  - Identity selection is protected against slower, outdated selections overwriting newer choices.

### DIFF
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1818,8 +1818,19 @@ class ComposerController extends BaseController
     if (replyToRecipientState.value == PrefixRecipientState.disabled) {
       replyToRecipientState.value = PrefixRecipientState.enabled;
     }
-    listReplyToEmailAddress = listEmailAddress.toList();
+    listReplyToEmailAddress = _normalizeReplyToFromIdentity(listEmailAddress);
     updateStatusEmailSendButton();
+  }
+
+  List<EmailAddress> _normalizeReplyToFromIdentity(Set<EmailAddress> listEmailAddress) {
+    final identity = identitySelected.value;
+    final identityEmail = identity?.email;
+    if (identityEmail?.isNotEmpty == true &&
+        identityEmail != ownEmailAddress &&
+        listEmailAddress.every((address) => address.email == ownEmailAddress)) {
+      return [EmailAddress(identity!.name, identityEmail)];
+    }
+    return listEmailAddress.toList();
   }
 
   void _removeReplyToEmailAddressFromFormerIdentity(Set<EmailAddress> listEmailAddress) {
--- a/lib/features/composer/presentation/composer_controller.dart
+++ b/lib/features/composer/presentation/composer_controller.dart
@@ -1769,27 +1769,40 @@ class ComposerController extends BaseController
       if (formerIdentity.bcc?.isNotEmpty == true) {
         _removeBccEmailAddressFromFormerIdentity(formerIdentity.bcc!);
       }
-      if (formerIdentity.replyTo?.isNotEmpty == true) {
-        _removeReplyToEmailAddressFromFormerIdentity(formerIdentity.replyTo!);
-      }
+      _removeReplyToEmailAddressFromFormerIdentity(formerIdentity);
       await _removeSignature();
     }
+
+    // Identity-change callbacks are not awaited by the UI; a slower earlier
+    // selection must not resume and overwrite a newer selection.
+    if (!_isCurrentlySelectedIdentity(newIdentity)) return;
 
     if (newIdentity.bcc?.isNotEmpty == true) {
       _applyBccEmailAddressFromIdentity(newIdentity.bcc!);
     }
 
-    if (newIdentity.replyTo?.isNotEmpty == true) {
-      _applyReplyToEmailAddressFromIdentity(newIdentity.replyTo!);
-    }
+    // Do not copy identity reply-to into listReplyToEmailAddress. That list is
+    // for user-entered recipients; createReplyToRecipients already defaults from
+    // identity.replyTo / identity.email when the list is empty.
 
     if (newIdentity.signatureAsString.isNotEmpty == true) {
       await applySignature(newIdentity.signatureAsString.asSignatureHtml());
     }
 
+    if (!_isCurrentlySelectedIdentity(newIdentity)) return;
+
     if (PlatformInfo.isMobile) {
       await htmlEditorApi?.onDocumentChanged();
     }
+  }
+
+  bool _isCurrentlySelectedIdentity(Identity identity) {
+    final selected = identitySelected.value;
+    if (selected == null) return false;
+    if (identity.id != null && selected.id != null) {
+      return selected.id == identity.id;
+    }
+    return identical(selected, identity);
   }
 
   void _applyBccEmailAddressFromIdentity(Set<EmailAddress> listEmailAddress) {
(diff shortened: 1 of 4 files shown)


lib/features/composer/presentation/extensions/create_email_request_extension.dart
```diff
@@ -53,6 +53,10 @@ extension CreateEmailRequestExtension on CreateEmailRequest {
     if (isNotReplyTo) return null;
 
     if (identity?.replyTo?.isNotEmpty == true) {
+      if (_shouldDefaultReplyToIdentityEmail()) {
+        return {EmailAddress(identity!.name, identity!.email)};
+      }
```
dab246: Explicit Reply-To is overwritten. This treats an explicitly configured Reply-To pointing to the primary address as an inherited value. The identity editor allows that configuration, so replies for an identity intentionally configured to use `ownEmailAddress` will be redirected to the identity email. We need an explicit inherited/unset signal; this predicate cannot infer user intent.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1769,13 +1769,20 @@ class ComposerController extends BaseController
       if (formerIdentity.bcc?.isNotEmpty == true) {
         _removeBccEmailAddressFromFormerIdentity(formerIdentity.bcc!);
       }
+      if (formerIdentity.replyTo?.isNotEmpty == true) {
+        _removeReplyToEmailAddressFromFormerIdentity(formerIdentity.replyTo!);
+      }
```
dab246: Remove the effective value, not the raw value. The apply path can normalize `{ownEmailAddress}` into `{formerIdentity.email}`, but this cleanup removes only the raw identity value. Switching from that identity to one with an empty Reply-To leaves the former identity address in the composer, and request generation sends it as an explicit Reply-To. Track and remove the effective value that was applied.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1769,13 +1769,20 @@ class ComposerController extends BaseController
...
       await _removeSignature();
     }
 
     if (newIdentity.bcc?.isNotEmpty == true) {
       _applyBccEmailAddressFromIdentity(newIdentity.bcc!);
     }
 
+    if (newIdentity.replyTo?.isNotEmpty == true) {
+      _applyReplyToEmailAddressFromIdentity(newIdentity.replyTo!);
+    }
```
dab246: Protect against stale identity selection. This runs after `await _removeSignature()`. Identity-change callbacks are not awaited by the UI, so a slower earlier selection can resume after a later selection and overwrite its Reply-To. Serialize selections or use a generation/id check before applying side effects.

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1807,6 +1814,35 @@ class ComposerController extends BaseController
     updateStatusEmailSendButton();
   }
 
+  void _applyReplyToEmailAddressFromIdentity(Set<EmailAddress> listEmailAddress) {
+    if (replyToRecipientState.value == PrefixRecipientState.disabled) {
+      replyToRecipientState.value = PrefixRecipientState.enabled;
+    }
+    listReplyToEmailAddress = _normalizeReplyToFromIdentity(listEmailAddress);
+    updateStatusEmailSendButton();
+  }
```
dab246: Identity defaults become manual overrides. Adding an identity-derived value to `listReplyToEmailAddress` makes request generation treat it as a manual override. This bypasses `isNotReplyTo` for drafts/templates and skips the existing display-name normalization for nameless Reply-To addresses. Keep automatic identity defaults separate from user-entered recipients.

test/features/composer/presentation/composer_controller_test.dart
```diff
@@ -1781,6 +1782,77 @@ void main() {
...
+          );
+
+          await composerController?.selectIdentity(workIdentity);
+
+          expect(
+            composerController?.listReplyToEmailAddress,
+            [EmailAddress('Work', 'work@domain.tld')],
+          );
+        },
+      );
```
dab246, line 1861: Missing transition regression coverage. This verifies normalization only immediately after selecting the identity. Please add the failing transition: select an identity whose Reply-To is `{accountPrimary}`, then select an identity with an empty Reply-To; the list must be empty and the generated JMAP payload must not retain the former identity address.

dab246 (commented):
Hi @usmamalik345, thank you for your contribution to the project. I’ve done a quick review and have a few comments; please take a look and fix them.

usmamalik345:
Thanks @dab246 for the review — addressed in 6303ff1:

1. **Explicit Reply-To preserved** — Removed `_shouldDefaultReplyToIdentityEmail`. Configured reply-to (including one pointing at the account primary) is no longer rewritten.
2. **No identity defaults in composer recipients** — Identity reply-to is no longer copied into `listReplyToEmailAddress`. Send-time defaults stay in `createReplyToRecipients` (`identity.replyTo` / `identity.email`).
3. **Cleanup on switch** — Former identity configured reply-to addresses are removed from the composer list when switching identities.
4. **Stale selection guard** — After `await _removeSignature()` / `applySignature()`, side effects are skipped if a newer identity is already selected.
5. **Tests** — Updated coverage for (1)–(3), including the `{accountPrimary}` → empty reply-to transition.

Happy to adjust further if needed.

dab246:
> Thanks @dab246 for the review — addressed in [6303ff1](https://github.com/linagora/tmail-flutter/commit/6303ff1faf6e47bc4d3817da8d56b6c10cef8193):
> 
> 1. **Explicit Reply-To preserved** — Removed `_shouldDefaultReplyToIdentityEmail`. Configured reply-to (including one pointing at the account primary) is no longer rewritten.
> 2. **No identity defaults in composer recipients** — Identity reply-to is no longer copied into `listReplyToEmailAddress`. Send-time defaults stay in `createReplyToRecipients` (`identity.replyTo` / `identity.email`).
> 3. **Cleanup on switch** — Former identity configured reply-to addresses are removed from the composer list when switching identities.
> 4. **Stale selection guard** — After `await _removeSignature()` / `applySignature()`, side effects are skipped if a newer identity is already selected.
> 5. **Tests** — Updated coverage for (1)–(3), including the `{accountPrimary}` → empty reply-to transition.
> 
> Happy to adjust further if needed.

Thanks @usmamalik345 for the quick turnaround — 6303ff1 is a real improvement. Removing `_shouldDefaultReplyToIdentityEmail` and keeping identity defaults out of `listReplyToEmailAddress` were both the right calls.

Before going further on details, though, I re-validated the premise of the PR, and I think we should stop here. Two findings.

## 1. #2651 is already fixed on master

`9b30b4268` (an ancestor of master) added an identity-email fallback to `createReplyToRecipients`, *before* the `ownEmailAddress` fallback. The ticket's exact scenario, run against master:

```dart
identity = Identity(name: 'Alice', email: 'alice@domain.tld', replyTo: {});
ownEmailAddress = 'bob@domain.tld';        // default identity
// createReplyToRecipients() -> {Alice <alice@domain.tld>}   PASSES on master
```

The precondition holds too: `identity_creator_controller.dart:532` writes `replyTo = {}` when the user leaves the field at **None**. So the reproduction steps in #2651 cannot fail today.

Consistent with that, this PR no longer changes that path at all — after 6303ff1 the diff to `create_email_request_extension.dart` is a single deleted blank line, and the composer changes are off the path too: with nothing typed, `listReplyToEmailAddress` is empty and `createReplyToRecipients` falls straight through to the identity default.

## 2. The remaining cleanup-on-switch is a regression

With the apply side removed, the controller only ever *removes* from `listReplyToEmailAddress`, never adds. But that list is fed solely by draft/reply restore (`composer_controller.dart:765`) and by the user typing (`composer_controller.dart:791`) — so the removal can only delete user data:

- Type `bob@domain.tld` into Reply-To under identity Bob (`replyTo={bob}`), switch to Alice -> the typed address is silently dropped.
- Open a draft with `Reply-To: bob@domain.tld`, switch Bob -> Alice -> the header is dropped and the field collapses.

I wrote both as tests: they fail on this branch, pass on master. `editAsNewEmail` is a third case — `identitySelected.value` is set *before* `selectIdentity()` (`setup_selected_identity_extension.dart:36-37`), so `formerIdentity == newIdentity` and the removal runs against the identity being selected.

Separately, `_isCurrentlySelectedIdentity` can just be `identitySelected.value == identity`, since `Identity` uses `EquatableMixin`. The `identical()` fallback returns `false` for equal-but-distinct instances, silently skipping BCC and signature application.

## Proposal

Close this PR, and close #2651 as already fixed by `9b30b4268`. None of the three production changes should land:

- `_removeReplyToEmailAddressFromFormerIdentity` — off the #2651 path, and it can only delete user-entered or draft-restored addresses.
- `_isCurrentlySelectedIdentity` — unrelated to #2651 and untested. If the race is real, it deserves its own ticket, with a plain `identitySelected.value == identity`.
- Deleted blank line in `create_email_request_extension.dart` — no behaviour change.

The tests are worth salvaging. A small test-only PR against master would be very welcome:

- **Keep** your guard test `should preserve explicitly configured replyTo pointing at account primary...` — it passes on master and pins what your first commit would have broken.
- **Add** the real #2651 regression: identity with empty `replyTo` and `email != ownEmailAddress`, asserting the identity address wins. The existing `...when identity has no replyTo configured` uses `identity.email == ownEmail`, so it would still pass against the pre-`9b30b4268` code — that's the gap that let this look unfixed.
- **Drop** the three `selectIdentity reply-to sync` tests — they seed `listReplyToEmailAddress` with an identity's reply-to, which production never does.

Sorry for the wasted cycles on the composer side — that's on the missing regression test, not on you.

usmamalik345:
Thanks @dab246 for the careful re-check — that makes sense.

I'll close this PR. I'll also leave a note on #2651 about it already being fixed by `9b30b4268`, and open a small test-only PR with the regression coverage you suggested.

usmamalik345:
Closing per maintainer feedback: #2651 is already fixed on master. Follow-up will be a test-only PR.

usmamalik345:
Test-only follow-up: https://github.com/linagora/tmail-flutter/pull/4808

Issue #4799, opened 2026-08-31
HTML Mail be truncated in width


hoangdat:
### Desc 
- HTML mail be truncated 
- cannot read the whole information
- mail be hard to scroll in horizontal 

### DoD
- [x] well display like gmail 

(mail be sent to team mailbox)

Issue #3863, opened 2025-07-08
Text mail or fixed size parts of HTML mails partially unreadable on app


guimard:
### Description

When a mail contains a `<pre>` part, we can not scroll to the right, then it's unreadable. Same with some `<table>`

### Current situation

![Image](https://github.com/user-attachments/assets/eef49a0f-cb3d-4631-9e95-32735efe07f0)

### Acceptance criteria

Maybe include such part into a `<div>` with automatic left/right scroll bar ?

### Context

Android 15

chibenwa:
Please provide the EML for the team @guimard

guimard:
> Please provide the EML for the team [@guimard](https://github.com/guimard)

Done via front channel

PR #4806, opened 2026-09-04
TF-4799 Fix HTML Mail be truncated in width

dab246:
## Issue

- [#4799](https://github.com/linagora/tmail-flutter/issues/4799)
- Related: [#3863](https://github.com/linagora/tmail-flutter/issues/3863)

## Root cause

Some emails use fixed-width/nested tables, cells, padding, and images designed for desktop. On native mobile, these constraints can exceed the WebView width. The existing logic normalized images but could not reflow the whole overflowing hierarchy or remeasure correctly after images and quoted content changed size.

## Solution

- Add a responsive script enabled only for native mobile `EmailView`; web and other HTML viewer screens are unchanged.
- Detect only overflowing blocks, reflow wrappers/tables/cells to available width, and use `zoom` only when reflow cannot fit the table.
- Preserve content that already fits. On narrow phones, adjust padding and typography only in reflowed content; text is never reduced below 12px.
- Remeasure after load, resize, image load, and quote toggle. Layout work is batched per animation frame, and temporary styles are restored before every new pass.

## Summary by CodeRabbit

- **New Features**
  - Improved mobile email viewing with responsive reflow for tables, images, fixed-width content, and preformatted text.
  - Enhanced HTML content display with configurable sizing, typography, scrolling, and layout behavior.
  - Improved support for email links, attachments, previews, and horizontal content navigation.
  - Added overflow detection and layout adjustments to preserve readability on narrow screens.

- **Bug Fixes**
  - Improved content sizing and loading stability across supported platforms.
  - Added safeguards for asynchronous HTML viewer updates.

- **Tests**
  - Added comprehensive coverage for responsive layouts, sizing, overflow, typography, and browser rendering.


lib/features/email/presentation/widgets/html_attachment_previewer.dart
```diff
@@ -142,13 +142,25 @@ class _HtmlAttachmentPreviewerState extends State<HtmlAttachmentPreviewer> {
...
+                constraints: BoxConstraints.tightFor(width: width),
+              ),
+            ),
+            behavior: HtmlContentViewerBehavior(features: {
+              HtmlContentViewerFeature.keepWidthWhileLoading,
+            }),
+            callbacks: HtmlContentViewerCallbacks(
+              onMailto: (uri) async {
+                widget.mailToClicked(uri);
+              },
```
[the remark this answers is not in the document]
dab246: Thanks for flagging this. We verified the native implementation before this migration: its `mailto` callback did not call `popBack()`, so this PR preserves the existing native behavior and does not introduce a regression.

The web previewer already closes before opening the composer, while native has a different pre-existing flow. Aligning both behaviors is a separate UX change that needs mobile validation, so it should be handled in a dedicated ticket rather than in TF-4799, which is scoped to responsive email rendering.

core/lib/utils/html/mobile_email_responsive_layout_script.dart
hoangdat, line 1: we should carefully on the scan of element to avoid bad performance in large email 

  1. Quick exit when content.scrollWidth <= content.clientWidth.
  2. Detect boundaries directly: rect.left < content.left or rect.right > content.right; this handles RTL better.
  3. Check element.scrollWidth > element.clientWidth for long text.
  4. Scan likely candidates first: tables, images, pre, fixed/min-width elements.
  5. Cache viewport/content dimensions and skip unchanged reruns.
  6. Add real WebView tests for large HTML, RTL, nested tables, and long text.

core/lib/utils/html/mobile_email_responsive_layout_script.dart
```diff
@@ -0,0 +1,434 @@
...
+          for (var i = 0; i < roots.length; i++) {
+            var root = roots[i];
+            var tables = getTopLevelTables(root);
+            if (tables.length === 0) {
+              markResponsiveQuote(root);
+              markResponsiveLayout(root);
+              if (root.tagName === 'TABLE') {
+                reflowTableToAvailableWidth(content, root, root.getBoundingClientRect().width);
+              } else {
+                scaleElementToAvailableWidth(content, root, root.getBoundingClientRect().width);
```
hoangdat: Could we attempt wrapper reflow before scaling here? A fixed-width non-table root (for example, `<div style="width:790px">`) reaches this branch and goes directly to `zoom`, so text can become very small and the advertised “reflow first, zoom as fallback” behavior is skipped. Suggested flow: call `makeWrapperResponsive(root)`, remeasure, and scale only if it still overflows, with a minimum effective text-size/scale floor. Please add a real-layout regression test for this case.

core/lib/utils/html/mobile_email_responsive_layout_script.dart
```diff
@@ -0,0 +1,434 @@
...
+            setResponsiveStyle(
+              fixedWidthCell.element,
+              'width',
+              fixedWidthCell.width.toFixed(3) + '%',
+            );
+          }
+
+          var availableWidth = getAvailableWidth(content, table);
+          if (table.getBoundingClientRect().width > availableWidth + 1) {
+            scaleElementToAvailableWidth(content, table, table.getBoundingClientRect().width);
```
hoangdat: Could we use local horizontal scrolling when scaling makes text smaller than 12px? It preserves table readability. Please prevent table swipes from triggering thread navigation and test on iOS/Android.

core/lib/utils/html/mobile_email_responsive_layout_script.dart
dab246: Benchmarked first: an 8000-element email at 360px takes ~6ms per pass, so the scan isn't the problem — the number of passes is.

1. **Quick exit** — tried it, had to drop it. It skips content clipped by an `overflow:hidden` ancestor: a 900px wrapper stayed 900px and got cut off, which is this ticket's bug. Guarded the `resize` listener instead (skip when `clientWidth` is unchanged). That kills the WebView height feedback loop, where the repeated passes actually come from, with no false negatives.
2. **Boundary detection / RTL** — RTL already works: a 790px wrapper sits at `left: -290` and gets reflowed today. The real gap is an element narrower than the content but pushed outside it, and it fails the same way in LTR. Couldn't build a realistic email that hits it, and changing root detection affects every email, so I left it. Send me a sample if you have one.
3. **`element.scrollWidth`** — real bug, fixed. `white-space: nowrap` was never detected (`scrollWidth 4421` vs `clientWidth 360`). Added `relaxNoWrapContent()`: `nowrap` → `normal`, `pre` → `pre-wrap` so spacing survives. Long words and `<pre>` were already covered by existing CSS.
4. **Candidate-first scan** — skipped. Worst case is unchanged and a tag allowlist would miss a plain `<div>`, which is the case in your next comment. The scan is already read-only with writes batched, so it's one layout flush total.
5. **Caching** — already there via `createLayoutMetrics()`. A blanket rerun-skip would be wrong since images finish loading and change layout without changing width; the resize guard covers the real case.
6. **Tests** — added `mobile_email_responsive_layout_browser_test.dart`, which renders the real document in an iframe and asserts layout instead of script source. 12 cases: large HTML, RTL, nested tables, nowrap, unbreakable cell text, stylesheet widths, clipped ancestors, images. Wired into `CHROME_TEST_PATHS`.

core/lib/utils/html/mobile_email_responsive_layout_script.dart
dab246: Good catch, biggest one of the three. On `<div style="width:790px">` with 16px text at 360px:

| | applied | effective text |
|---|---|---|
| before | `zoom: 0.456` | 7.3px |
| after | reflow only | 16px |

Reflow clears the overflow completely, so the scale fallback never runs. Now calls `makeWrapperResponsive(root)`, remeasures, scales only if it still overflows.

One addition: replaced elements (`IMG/IFRAME/VIDEO/…`) still go straight to scaling — forcing `width: 100%` on them breaks the aspect ratio (an 800×400 image overflowed by 120px while I was testing).

core/lib/utils/html/mobile_email_responsive_layout_script.dart
dab246: Right symptom, but the cause was elsewhere and the fix is simpler than local scrolling.

Our own stylesheet has `table, td, th { word-break: normal !important }`, so a long token in a `<td>` can never wrap and scaling was the only way to fit it — that's how 16px text ended up at 4.1px.

Fix: relax `word-break: break-word` on the cells first, remeasure, scale only if it still overflows. Same fixture now renders at 16px with no overflow, and tables that already fit are untouched. Used `word-break: break-word` over `overflow-wrap: anywhere` since the latter needs Safari 15.4+.

This avoids the nested scroller, so no gesture conflict. Worth flagging though: `runScriptsHandleScrollEvent` listens on `.tmail-content`, and now that horizontal overflow is gone `maxOffset` is always 0 — every horizontal swipe already navigates between emails. A scroll container would have needed a propagation guard.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4806

tddang-linagora (dismissed):

hoangdat (approved):

tddang-linagora (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4806

PR #4810, opened 2026-09-07
Allow workplace to refresh jmap token

tddang-linagora:
## Issue
- Due to workplace (where file picker and other workplace requests are made) does not use the app's common Dio, it does not has access to the refresh token mechanism, hence the exchange token process failing with 401 will forever stuck users in an unauthenticated state, until users trigger other process that require network to jmap.

## Solution

- Expose the refresh token mechanism publicly so other services can trigger it by their own.
- Dedup the refresh token mechanism so concurrent unauthenticated failure does not refresh token multiple times.

## Summary by CodeRabbit

* **New Features**
  * Workplace attachment requests can refresh authentication and retry once after unauthorized responses.
  * Existing refreshed credentials are reused when available.
  * Token refresh outcomes distinguish duplicate, stale-session, and unavailable-token scenarios.

* **Bug Fixes**
  * Prevented duplicate refreshes during concurrent requests.
  * Improved reliability when attachment and app requests refresh credentials simultaneously.
  * Preserved valid credentials when refresh responses omit token values.
  * Prevented stale refresh results from restoring cleared sessions.
  * Authentication failures now follow consistent recovery handling, including urgent error flows.

### DIFF
--- a/docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
+++ b/docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
@@ -1,0 +1,25 @@
+# 0107. Route Workplace's OIDC refresh through the main interceptor
+
+Date: 2026-09-07
+
+## Status
+
+Accepted
+
+## Context
+
+Workplace's Drive `token_exchange` request runs on `WorkplaceDio`, a bare `Dio()` instance with no interceptors. A 401 on that request (a stale OIDC id token) has no refresh-and-retry path, unlike the main app's Dio, which refreshes via `AuthorizationInterceptors`. `AuthorizationInterceptors` lives in the main app package; `workplace` is a separate package that cannot depend on it.
+
+## Decision
+
+- `AuthorizationInterceptors` exposes `requestTokenRefresh()`, a dedup-guarded public entry point backed by an in-flight `Future<TokenOIDC>`. Any caller — the interceptor's own reactive `onError` path or an external caller — joins the same in-flight refresh instead of starting a second one.
+- `AuthorizationInterceptors` exposes `isRefreshFailureFatal()`, reusing the existing web/mobile rejection classifiers, so an external caller can tell a dead session (RFC 6749 400/401-equivalent) from a transient failure and decide whether to call the existing public `clear()`.
+- `ComposerAttachmentExtensionRegistry`'s Workplace wiring passes an `oidcRefreshTrigger` callback into `WorkplaceComposerAttachmentExtension`, alongside the existing `oidcTokenGetter`, mirroring the same callback-injection pattern across the package boundary.
+- `WorkplaceComposerAttachmentExtension._exchangeAccessToken` catches a 401 from the token-exchange call and retries exactly once with a token refreshed via `oidcRefreshTrigger`, guarded by a `refreshAttempted` flag to prevent looping.
+
+## Consequences
+
+- A stale OIDC id token on the Workplace exchange-token request self-heals via one retry instead of surfacing as a generic toast.
+- A 401 on the main Dio and a 401 on Workplace's Dio racing at the same time still trigger exactly one refresh call against the server; both callers resolve to the same new token.
+- `WorkplaceDio` remains a plain, interceptor-free `Dio()`; only the exchange-token call site gains retry logic, not every Workplace request.
+- Workplace's own refresh failures follow the same fatal-vs-transient classification as the main app's reactive path, so a dead refresh token forces logout consistently regardless of which request discovered it.
--- a/docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
+++ b/docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
@@ -13,7 +13,8 @@ Workplace's Drive `token_exchange` request runs on `WorkplaceDio`, a bare `Dio()
 ## Decision
 
 - `AuthorizationInterceptors` exposes `requestTokenRefresh()`, a dedup-guarded public entry point backed by an in-flight `Future<TokenOIDC>`. Any caller — the interceptor's own reactive `onError` path or an external caller — joins the same in-flight refresh instead of starting a second one.
-- `AuthorizationInterceptors` exposes `isRefreshFailureFatal()`, reusing the existing web/mobile rejection classifiers, so an external caller can tell a dead session (RFC 6749 400/401-equivalent) from a transient failure and decide whether to call the existing public `clear()`.
+- `requestTokenRefresh()` owns the outcome of a refresh for every caller: a server rejection (RFC 6749 400/401-equivalent, classified by the existing web/mobile classifiers) clears the session and throws `RefreshTokenFailedException`; a same-token response throws `RefreshTokenDuplicatedException` before anything is persisted; transient failures are rethrown untouched.
+- Callers never classify refresh failures themselves. The Workplace wiring forwards a `DrivePickFailure` carrying an urgent exception into `MailboxDashBoardController`'s `BaseController` handling (`validateUrgentException` / `handleUrgentException`), the same path the interceptor's own rejected requests take.
 - `ComposerAttachmentExtensionRegistry`'s Workplace wiring passes an `oidcRefreshTrigger` callback into `WorkplaceComposerAttachmentExtension`, alongside the existing `oidcTokenGetter`, mirroring the same callback-injection pattern across the package boundary.
 - `WorkplaceComposerAttachmentExtension._exchangeAccessToken` catches a 401 from the token-exchange call and retries exactly once with a token refreshed via `oidcRefreshTrigger`, guarded by a `refreshAttempted` flag to prevent looping.
 
@@ -22,4 +23,4 @@ Workplace's Drive `token_exchange` request runs on `WorkplaceDio`, a bare `Dio()
 - A stale OIDC id token on the Workplace exchange-token request self-heals via one retry instead of surfacing as a generic toast.
 - A 401 on the main Dio and a 401 on Workplace's Dio racing at the same time still trigger exactly one refresh call against the server; both callers resolve to the same new token.
 - `WorkplaceDio` remains a plain, interceptor-free `Dio()`; only the exchange-token call site gains retry logic, not every Workplace request.
-- Workplace's own refresh failures follow the same fatal-vs-transient classification as the main app's reactive path, so a dead refresh token forces logout consistently regardless of which request discovered it.
+- Fatal-vs-transient classification lives in one place, so a dead refresh token forces logout consistently regardless of which request discovered it.
--- a/lib/features/base/base_controller.dart
+++ b/lib/features/base/base_controller.dart
@@ -163,7 +163,8 @@ abstract class BaseController extends GetxController
     return exception is NoNetworkError
       || exception is BadCredentialsException
       || exception is ConnectionError
-      || exception is RefreshTokenFailedException;
+      || exception is RefreshTokenFailedException
+      || exception is RefreshTokenDuplicatedException;
   }
 
   void handleErrorViewState(Object error, StackTrace stackTrace) {}
@@ -185,7 +186,9 @@ abstract class BaseController extends GetxController
       _handleConnectionErrorException();
     } else if (exception is BadCredentialsException) {
       handleBadCredentialsException();
-    } else if (exception is RefreshTokenFailedException) {
+    } else if (exception is RefreshTokenFailedException
+        || exception is RefreshTokenDuplicatedException) {
+      // A refresh that returns the same token cannot clear the 401 either.
       handleRefreshTokenFailedException();
     }
   }
@@ -202,7 +205,9 @@ abstract class BaseController extends GetxController
       _handleConnectionErrorException();
     } else if (exception is BadCredentialsException) {
       handleBadCredentialsException();
-    } else if (exception is RefreshTokenFailedException) {
+    } else if (exception is RefreshTokenFailedException
+        || exception is RefreshTokenDuplicatedException) {
+      // A refresh that returns the same token cannot clear the 401 either.
       handleRefreshTokenFailedException();
     } else {
       logWarning(
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -19,18 +19,9 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
   return ComposerAttachmentExtensionRegistry([
     WorkplaceComposerAttachmentExtension(
       workplaceUri: uriNotifier,
-      // Read at picker-open time, so it's always the current session's answer.
-      uploadFromUrlSupported: () {
-        final dashboard = getBinding<MailboxDashBoardController>();
-        final jmapUrl = dashboard?.dynamicUrlInterceptors.jmapUrl;
-        if (jmapUrl == null || jmapUrl.isEmpty) return false;
-        return dashboard?.sessionCurrent?.isUploadFromUrlSupported(
-              dashboard.accountId.value,
-              jmapUrl: jmapUrl,
-            ) ??
-            false;
-      },
+      uploadFromUrlSupported: _isUploadFromUrlSupported,
       oidcTokenGetter: () => getBinding<AuthorizationInterceptors>()?.currentOidcIdToken,
+      oidcRefreshTrigger: _refreshWorkplaceOidcToken,
       maxAttachmentSizeBytesGetter: () =>
           getBinding<MailboxDashBoardController>()?.maxSizeAttachmentsPerEmail?.value,
       // Read at picker-open time from the composer that owns the picker.
@@ -38,23 +29,50 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
           getBinding<ComposerController>(tag: composerId)
               ?.attachmentUploadValidationService
               .remainingCapacityBytes,
-      onPickState: (composerId, state) async {
-        if (state is DrivePickResult) {
-          final composer = getBinding<ComposerController>(tag: composerId);
-          if (composer == null) {
-            // Composer closed or wrong tag — the pick has nowhere to land.
-            logError('ComposerAttachmentExtensionRegistry::onPickState: no ComposerController for tag=$composerId, drive pick discarded');
-            getBinding<ToastManager>()?.showMessageFailure(
-              DrivePickFailure(Exception('ComposerController unavailable')),
-            );
-            return;
-          }
-          // No catch here: handleDrivePickResult swallows and toasts its own errors.
-          await composer.handleDrivePickResult(state.documents);
-        } else if (state is DrivePickFailure) {
-          getBinding<ToastManager>()?.showMessageFailure(state);
-        }
-      },
+      onPickState: _onDrivePickState,
     ),
   ]);
+}
+
+/// Read at picker-open time, so it's always the current session's answer.
+bool _isUploadFromUrlSupported() {
+  final dashboard = getBinding<MailboxDashBoardController>();
+  final jmapUrl = dashboard?.dynamicUrlInterceptors.jmapUrl;
+  if (jmapUrl == null || jmapUrl.isEmpty) return false;
+  return dashboard?.sessionCurrent?.isUploadFromUrlSupported(
+        dashboard.accountId.value,
+        jmapUrl: jmapUrl,
+      ) ??
+      false;
+}
+
+/// Triggers the main app's OIDC refresh for Workplace's own (unwired) Dio.
+Future<String?> _refreshWorkplaceOidcToken() async {
+  final interceptor = getBinding<AuthorizationInterceptors>();
+  if (interceptor == null) return null;
+  try {
+    final newToken = await interceptor.requestTokenRefresh();
+    return newToken.tokenId.uuid;
+  } catch (e) {
+    if (interceptor.isRefreshFailureFatal(e)) interceptor.clear();
+    rethrow;
+  }
+}
+
+Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
+  if (state is DrivePickResult) {
+    final composer = getBinding<ComposerController>(tag: composerId);
+    if (composer == null) {
+      // Composer closed or wrong tag — the pick has nowhere to land.
+      logError('ComposerAttachmentExtensionRegistry::onPickState: no ComposerController for tag=$composerId, drive pick discarded');
+      getBinding<ToastManager>()?.showMessageFailure(
+        DrivePickFailure(Exception('ComposerController unavailable')),
+      );
+      return;
+    }
+    // No catch here: handleDrivePickResult swallows and toasts its own errors.
+    await composer.handleDrivePickResult(state.documents);
+  } else if (state is DrivePickFailure) {
+    getBinding<ToastManager>()?.showMessageFailure(state);
+  }
 }
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -5,6 +5,7 @@ import 'package:tmail_ui_user/features/composer/presentation/composer_controller
 import 'package:tmail_ui_user/features/home/domain/extensions/session_extensions.dart';
 import 'package:tmail_ui_user/features/login/data/network/interceptors/authorization_interceptors.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/providers/workplace/drive_attachment_uri_value_notifier_provider.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
@@ -52,8 +53,14 @@ bool _isUploadFromUrlSupported() {
 Future<String?> _refreshWorkplaceOidcToken() async {
   final interceptor = getBinding<AuthorizationInterceptors>();
   if (interceptor == null) return null;
-  final newToken = await interceptor.requestTokenRefresh();
-  return newToken.tokenId.uuid;
+  try {
+    final newToken = await interceptor.requestTokenRefresh();
+    return newToken.tokenId.uuid;
+  } on RefreshTokenFailedException catch (e, st) {
+    // The JMAP path logs its own; without this the Drive journey is silent.
+    interceptor.logFatalRefreshRejection(e, st);
+    rethrow;
+  }
 }
 
 Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -5,6 +5,7 @@ import 'package:tmail_ui_user/features/composer/presentation/composer_controller
 import 'package:tmail_ui_user/features/home/domain/extensions/session_extensions.dart';
 import 'package:tmail_ui_user/features/login/data/network/interceptors/authorization_interceptors.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
+import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/providers/workplace/drive_attachment_uri_value_notifier_provider.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
@@ -54,8 +55,11 @@ Future<String?> _refreshWorkplaceOidcToken() async {
     final newToken = await interceptor.requestTokenRefresh();
     return newToken.tokenId.uuid;
   } catch (e) {
-    if (interceptor.isRefreshFailureFatal(e)) interceptor.clear();
-    rethrow;
+    if (!interceptor.isRefreshFailureFatal(e)) rethrow;
+    interceptor.clear();
+    // Same funnel as the interceptor's own reject(RefreshTokenFailedException) path.
+    getBinding<MailboxDashBoardController>()?.handleRefreshTokenFailedException();
+    throw RefreshTokenFailedException();
   }
 }
 
@@ -73,6 +77,8 @@ Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
     // No catch here: handleDrivePickResult swallows and toasts its own errors.
     await composer.handleDrivePickResult(state.documents);
   } else if (state is DrivePickFailure) {
+    // Logout already triggered by _refreshWorkplaceOidcToken; no second report.
+    if (state.error is RefreshTokenFailedException) return;
     getBinding<ToastManager>()?.showMessageFailure(state);
   }
 }
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -5,7 +5,6 @@ import 'package:tmail_ui_user/features/composer/presentation/composer_controller
 import 'package:tmail_ui_user/features/home/domain/extensions/session_extensions.dart';
 import 'package:tmail_ui_user/features/login/data/network/interceptors/authorization_interceptors.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
-import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/providers/workplace/drive_attachment_uri_value_notifier_provider.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
@@ -48,19 +47,13 @@ bool _isUploadFromUrlSupported() {
 }
 
 /// Triggers the main app's OIDC refresh for Workplace's own (unwired) Dio.
-/// The interceptor owns the outcome; a dead session surfaces as
-/// RefreshTokenFailedException and is routed by [_onDrivePickState].
+/// The interceptor owns the outcome, logging included; a dead session surfaces
+/// as RefreshTokenFailedException and is routed by [_onDrivePickState].
 Future<String?> _refreshWorkplaceOidcToken() async {
   final interceptor = getBinding<AuthorizationInterceptors>();
   if (interceptor == null) return null;
-  try {
-    final newToken = await interceptor.requestTokenRefresh();
-    return newToken.tokenId.uuid;
-  } on RefreshTokenFailedException catch (e, st) {
-    // The JMAP path logs its own; without this the Drive journey is silent.
-    interceptor.logFatalRefreshRejection(e, st);
-    rethrow;
-  }
+  final newToken = await interceptor.requestTokenRefresh();
+  return newToken.tokenId.uuid;
 }
 
 Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
--- a/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
+++ b/lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
@@ -5,7 +5,6 @@ import 'package:tmail_ui_user/features/composer/presentation/composer_controller
 import 'package:tmail_ui_user/features/home/domain/extensions/session_extensions.dart';
 import 'package:tmail_ui_user/features/login/data/network/interceptors/authorization_interceptors.dart';
 import 'package:tmail_ui_user/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller.dart';
-import 'package:tmail_ui_user/main/exceptions/remote/authentication_exception.dart';
 import 'package:tmail_ui_user/main/providers/workplace/drive_attachment_uri_value_notifier_provider.dart';
 import 'package:tmail_ui_user/main/routes/route_navigation.dart';
 import 'package:tmail_ui_user/main/utils/toast_manager.dart';
@@ -48,19 +47,13 @@ bool _isUploadFromUrlSupported() {
 }
 
 /// Triggers the main app's OIDC refresh for Workplace's own (unwired) Dio.
+/// The interceptor owns the outcome; a dead session surfaces as
+/// RefreshTokenFailedException and is routed by [_onDrivePickState].
 Future<String?> _refreshWorkplaceOidcToken() async {
   final interceptor = getBinding<AuthorizationInterceptors>();
   if (interceptor == null) return null;
-  try {
-    final newToken = await interceptor.requestTokenRefresh();
-    return newToken.tokenId.uuid;
-  } catch (e) {
-    if (!interceptor.isRefreshFailureFatal(e)) rethrow;
-    interceptor.clear();
-    // Same funnel as the interceptor's own reject(RefreshTokenFailedException) path.
-    getBinding<MailboxDashBoardController>()?.handleRefreshTokenFailedException();
-    throw RefreshTokenFailedException();
-  }
+  final newToken = await interceptor.requestTokenRefresh();
+  return newToken.tokenId.uuid;
 }
 
 Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
@@ -77,8 +70,13 @@ Future<void> _onDrivePickState(String? composerId, DrivePickState state) async {
     // No catch here: handleDrivePickResult swallows and toasts its own errors.
     await composer.handleDrivePickResult(state.documents);
   } else if (state is DrivePickFailure) {
-    // Logout already triggered by _refreshWorkplaceOidcToken; no second report.
-    if (state.error is RefreshTokenFailedException) return;
+    final dashboard = getBinding<MailboxDashBoardController>();
+    final error = state.error;
+    if (dashboard != null && error is Exception && dashboard.validateUrgentException(error)) {
+      // Session-level failures take BaseController's standard path (logout, reconnect).
+      dashboard.handleUrgentException(failure: state, exception: error);
+      return;
+    }
     getBinding<ToastManager>()?.showMessageFailure(state);
   }
 }
--- a/lib/features/login/data/extensions/token_response_extension.dart
+++ b/lib/features/login/data/extensions/token_response_extension.dart
@@ -4,11 +4,17 @@ import 'package:model/model.dart';
 
 extension TokenResponseExtension on TokenResponse {
 
-  TokenOIDC toTokenOIDC({String? maybeAvailableRefreshToken}) {
+  // A refresh response may omit refresh_token, and OIDC Core 12.2 lets it omit
+  // id_token too; keep whichever the current token already holds.
+  TokenOIDC toTokenOIDC({required TokenOIDC currentToken}) {
     return TokenOIDC(
       accessToken ?? '',
-      TokenId(idToken ?? ''),
-      refreshToken ?? maybeAvailableRefreshToken ?? '',
+      TokenId(_orCurrent(idToken, currentToken.tokenId.uuid)),
+      _orCurrent(refreshToken, currentToken.refreshToken),
       expiredTime: accessTokenExpirationDateTime ?? DateTime.now());
   }
+
+  // The plugin returns '' rather than null when a field is absent.
+  String _orCurrent(String? value, String current) =>
+      value?.isNotEmpty == true ? value! : current;
 }
--- a/lib/features/login/data/extensions/token_response_extension.dart
+++ b/lib/features/login/data/extensions/token_response_extension.dart
@@ -9,12 +9,13 @@ extension TokenResponseExtension on TokenResponse {
   TokenOIDC toTokenOIDC({required TokenOIDC currentToken}) {
     return TokenOIDC(
       accessToken ?? '',
-      TokenId(_orCurrent(idToken, currentToken.tokenId.uuid)),
-      _orCurrent(refreshToken, currentToken.refreshToken),
+      TokenId(_idTokenOrCurrent(currentToken.tokenId.uuid)),
+      refreshToken ?? currentToken.refreshToken,
       expiredTime: accessTokenExpirationDateTime ?? DateTime.now());
   }
 
-  // The plugin returns '' rather than null when a field is absent.
-  String _orCurrent(String? value, String current) =>
-      value?.isNotEmpty == true ? value! : current;
+  // flutter_appauth_web stringifies the absent field, so it arrives as "null";
+  // mobile sends '' or null for the same thing.
+  String _idTokenOrCurrent(String current) =>
+      (idToken?.isNotEmpty == true && idToken != 'null') ? idToken! : current;
 }
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -576,8 +576,15 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
   String _getTokenAsBearerHeader(String token) => 'Bearer $token';
 
-  Future<PersonalAccount> _updateCurrentAccount({required TokenOIDC tokenOIDC}) async {
+  Future<PersonalAccount> _updateCurrentAccount({
+    required TokenOIDC tokenOIDC,
+    required int generation,
+  }) async {
     final currentAccount = await _accountCacheManager.getCurrentAccount();
+
+    // The session died while we read the account; these writes land between
+    // clearAll() and closeHive(), resurrecting it on the next launch.
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
 
     // Persist the new token BEFORE mutating the account cache. persistOneTokenOidc
     // is crash-safe (write-before-prune), so the token box always holds a usable
@@ -668,7 +675,10 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
     }
     _updateNewToken(acquired);
 
-    final personalAccount = await _updateCurrentAccount(tokenOIDC: acquired);
+    final personalAccount = await _updateCurrentAccount(
+      tokenOIDC: acquired,
+      generation: generation,
+    );
     if (PlatformInfo.isIOS) {
       await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
     }
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -365,22 +365,18 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         'AuthorizationInterceptors::onError: Perform get New Token',
         webConsoleEnabled: true,
       );
-      final previousToken = _token?.token;
-      final newTokenOidc = await requestTokenRefresh();
-
-      if (newTokenOidc.token == previousToken) {
-        // Refresh returned the SAME token — retrying cannot clear the 401, so it
-        // propagates to logout. Real auth death, but tag it for forensics.
-        _logForcedLogoutFor401(
-          authErrorType: 'forced_logout_401_refreshed_token_duplicated',
-          err: err,
-          hasAttemptedRefresh: true,
-        );
-        return super.onError(err, handler);
-      }
+      await requestTokenRefresh();
 
       requestOptions.extra[_refreshAttemptedKey] = true;
       return await _performRetry(requestOptions, err, handler);
+    } on RefreshTokenDuplicatedException {
+      // Retrying cannot clear the 401, so it propagates to logout; tag it for forensics.
+      _logForcedLogoutFor401(
+        authErrorType: 'forced_logout_401_refreshed_token_duplicated',
+        err: err,
+        hasAttemptedRefresh: true,
+      );
+      return super.onError(err, handler);
     } on DioException catch (refreshError, st) {
       // Web routes ALL refresh failures (Dio or non-Dio) through the single
       // web handler, so the session decision is uniform regardless of how the
@@ -643,6 +639,9 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
     final newTokenOidc = _withCurrentIdTokenIfMissing(PlatformInfo.isIOS
         ? await _getNewTokenForIOSPlatform()
         : await _getNewTokenForOtherPlatform());
+    if (newTokenOidc.token == _token?.token) {
+      throw const RefreshTokenDuplicatedException();
+    }
     _updateNewToken(newTokenOidc);
 
     final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -39,6 +39,7 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
   OIDCConfiguration? _configOIDC;
   TokenOIDC? _token;
   String? _authorization;
+  Future<TokenOIDC>? _refreshInFlight;
 
   final RefreshTokenErrorClassifier? _injectedErrorClassifier;
 
@@ -90,6 +91,19 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
   TokenOIDC? get currentToken =>
       _authenticationType == AuthenticationType.oidc ? _token : null;
+
+  /// Triggers a refresh, or joins one already running — dedupes an external
+  /// caller (e.g. Workplace's own Dio) against this interceptor's [onError].
+  Future<TokenOIDC> requestTokenRefresh() {
+    return _refreshInFlight ??= _acquireAndPersistNewToken()
+        .whenComplete(() => _refreshInFlight = null);
+  }
+
+  /// True when [error] is a confirmed server-side refresh rejection (session
+  /// dead). Mirrors [_handleRefreshErrorOnWeb]/[_handleRefreshErrorOnMobile].
+  bool isRefreshFailureFatal(Object error) => PlatformInfo.isWeb
+      ? _errorClassifier.isServerRejection(error)
+      : _isRefreshRejectedByTokenEndpoint(error);
 
   @override
   void onRequest(RequestOptions options, RequestInterceptorHandler handler) {
@@ -351,11 +365,10 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         'AuthorizationInterceptors::onError: Perform get New Token',
         webConsoleEnabled: true,
       );
-      final newTokenOidc = PlatformInfo.isIOS
-        ? await _getNewTokenForIOSPlatform()
-        : await _getNewTokenForOtherPlatform();
+      final previousToken = _token?.token;
+      final newTokenOidc = await requestTokenRefresh();
 
-      if (newTokenOidc.token == _token?.token) {
+      if (newTokenOidc.token == previousToken) {
         // Refresh returned the SAME token — retrying cannot clear the 401, so it
         // propagates to logout. Real auth death, but tag it for forensics.
         _logForcedLogoutFor401(
@@ -364,13 +377,6 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
           hasAttemptedRefresh: true,
         );
         return super.onError(err, handler);
-      }
-      _updateNewToken(newTokenOidc);
-
-      final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
-
-      if (PlatformInfo.isIOS) {
-        await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
       }
 
       requestOptions.extra[_refreshAttemptedKey] = true;
@@ -631,6 +637,20 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
   Future<TokenOIDC> _getNewTokenForOtherPlatform() {
     return _invokeRefreshTokenFromServer();
+  }
+
+  Future<TokenOIDC> _acquireAndPersistNewToken() async {
+    final newTokenOidc = PlatformInfo.isIOS
+        ? await _getNewTokenForIOSPlatform()
+        : await _getNewTokenForOtherPlatform();
+    _updateNewToken(newTokenOidc);
+
+    final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
+    if (PlatformInfo.isIOS) {
+      await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
+    }
+
+    return newTokenOidc;
   }
 
   Future<Response> _retryRequest(
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -13,7 +13,6 @@ import 'package:model/account/personal_account.dart';
 import 'package:model/oidc/oidc_configuration.dart';
 import 'package:model/oidc/token_oidc.dart';
 import 'package:tmail_ui_user/features/base/extensions/object_extensions.dart';
-import 'package:tmail_ui_user/features/login/data/extensions/token_oidc_extension.dart';
 import 'package:tmail_ui_user/features/login/data/local/account_cache_manager.dart';
 import 'package:tmail_ui_user/features/login/data/local/token_oidc_cache_manager.dart';
 import 'package:tmail_ui_user/features/login/data/network/authentication_client/authentication_client_base.dart';
@@ -609,7 +608,7 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
       _configOIDC!.redirectUrl,
       _configOIDC!.discoveryUrl,
       _configOIDC!.scopes,
-      _token!.refreshToken
+      _token!
     );
   }
 
@@ -644,18 +643,17 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
     // logged-out account and re-persist it over wiped caches.
     if (generation != _sessionGeneration) throw RefreshTokenFailedException();
 
-    final newTokenOidc = acquired.withIdTokenFallback(_token?.tokenId);
-    if (newTokenOidc.token == _token?.token) {
+    if (acquired.token == _token?.token) {
       throw const RefreshTokenDuplicatedException();
     }
-    _updateNewToken(newTokenOidc);
+    _updateNewToken(acquired);
 
-    final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
+    final personalAccount = await _updateCurrentAccount(tokenOIDC: acquired);
     if (PlatformInfo.isIOS) {
       await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
     }
 
-    return newTokenOidc;
+    return acquired;
   }
 
   Future<Response> _retryRequest(
--- a/lib/features/login/data/network/interceptors/authorization_interceptors.dart
+++ b/lib/features/login/data/network/interceptors/authorization_interceptors.dart
@@ -365,6 +365,14 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         hasAttemptedRefresh: true,
       );
       return super.onError(err, handler);
+    } on StaleSessionRefreshException catch (staleError) {
+      // Another session owns the interceptor now; this answer is not its verdict.
+      logWarning(
+        'AuthorizationInterceptors::onError: '
+        'refresh answered for a replaced session, keeping the current one',
+        webConsoleEnabled: true,
+      );
+      return _propagateKeepingSession(staleError, err, handler);
     } on RefreshTokenFailedException catch (refreshError, st) {
       // Session already cleared by requestTokenRefresh; surface the dead session.
       logFatalRefreshRejection(refreshError, st);
@@ -641,7 +649,7 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
 
     // The session died while we were away; re-arming it would resurrect a
     // logged-out account and re-persist it over wiped caches.
-    if (generation != _sessionGeneration) throw RefreshTokenFailedException();
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
 
     if (acquired.token == _token?.token) {
       throw const RefreshTokenDuplicatedException();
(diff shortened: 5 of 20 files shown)


tddang-linagora:
- [x] Add ADR

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +19,60 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+  final interceptor = getBinding<AuthorizationInterceptors>();
+  if (interceptor == null) return null;
+  try {
+    final newToken = await interceptor.requestTokenRefresh();
+    return newToken.tokenId.uuid;
+  } catch (e) {
+    if (interceptor.isRefreshFailureFatal(e)) interceptor.clear();
+    rethrow;
+  }
+}
```
dab246, line 57: Refresh token rejection does not trigger the logout flow.  Failure scenario:

1. The Workplace token exchange returns 401.
2. The OIDC refresh token has also expired or been revoked.
3. The callback identifies the error as fatal and calls AuthorizationInterceptors.clear().
4. The user remains on the mailbox screen and only sees an "attach from Drive failed" toast.
5. The interceptor has already lost its credentials; logout can only happen indirectly when another request subsequently fails.

dab246: Regression tests should cover:

- A refresh rejection from the Workplace server triggers the logout/reconnect handler exactly once.
- Transient or network refresh failures preserve the current session and only report the Drive operation failure.
- The same exception is not reported by both the refresh layer and the picker layer.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -91,6 +92,19 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
   TokenOIDC? get currentToken =>
       _authenticationType == AuthenticationType.oidc ? _token : null;
 
+  /// Triggers a refresh, or joins one already running — dedupes an external
+  /// caller (e.g. Workplace's own Dio) against this interceptor's [onError].
+  Future<TokenOIDC> requestTokenRefresh() {
+    return _refreshInFlight ??= _acquireAndPersistNewToken()
+        .whenComplete(() => _refreshInFlight = null);
+  }
```
dab246, line 140: Two staggered Workplace 401 responses can trigger two separate token refreshes. Concurrency scenario:

1. Two pickers/requests send the old ID token concurrently.
2. Request A receives `401`, refreshes successfully, and `_refreshInFlight` is cleared.
3. Request B's `401` arrives after step 2.
4. B starts a new refresh even though a valid new token is already available.

dab246: Regression test should cover: two token exchanges using the same old token, with the second `401` arriving after the first refresh completes. Both requests should succeed with exactly **one** refresh call.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +20,65 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+  if (jmapUrl == null || jmapUrl.isEmpty) return false;
+  return dashboard?.sessionCurrent?.isUploadFromUrlSupported(
+        dashboard.accountId.value,
+        jmapUrl: jmapUrl,
+      ) ??
+      false;
+}
+
+/// Triggers the main app's OIDC refresh for Workplace's own (unwired) Dio.
+Future<String?> _refreshWorkplaceOidcToken() async {
```
dab246, line 52: When the refresh token is rejected by the server:

1. `_refreshWorkplaceOidcToken()` only calls `interceptor.clear()`.
2. The exception is converted to `DrivePickFailure`.
3. `_onDrivePickState()` only shows a toast.

As a result, `handleRefreshTokenFailedException()` is never called. The user may remain on the dashboard with cleared auth state and only be logged out after another request fails, which violates the ADR.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -67,29 +73,74 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
...
+      throw failure;
+    }
+
+    // Mirrors AuthorizationInterceptors.validateToRetryTheRequestWithNewToken.
+    final currentToken = oidcTokenGetter();
+    if (currentToken != null && currentToken != failedToken) {
+      return _exchangeAccessToken(platformUrl, currentToken, refreshAttempted: true);
+    }
+
+    final refreshedToken = await oidcRefreshTrigger!();
```
dab246: Scenario:

1. Workplace sends token `T0`.
2. A JMAP request refreshes `T0 → T1`.
3. Workplace receives a delayed `401`.
4. The callback calls `requestTokenRefresh()` and generates `T2`, even though `T1` is still valid.

This causes unnecessary refresh/rotation. If the second refresh fails, a healthy session could be incorrectly cleared or logged out.

dab246 (commented):
The current test only verifies that the callback is called once, but does not confirm that the retried request actually uses the refreshed ID token. Please add coverage for:

* A refresh completes before the Workplace `401` arrives: the token endpoint should not be called again, and the retry should use the current token.
* A fatal Workplace refresh triggers logout exactly once, even when a JMAP `401` occurs concurrently.
* Assert that the body of the second `token_exchange` request contains the refreshed token.

tddang-linagora:
> The current test only verifies that the callback is called once, but does not confirm that the retried request actually uses the refreshed ID token. Please add coverage for:
> 
>     * A refresh completes before the Workplace `401` arrives: the token endpoint should not be called again, and the retry should use the current token.
> 
>     * A fatal Workplace refresh triggers logout exactly once, even when a JMAP `401` occurs concurrently.
> 
>     * Assert that the body of the second `token_exchange` request contains the refreshed token.

Done

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: Addressed in fd7958d52. `_refreshWorkplaceOidcToken` (lines 58-62) calls `interceptor.clear()`, then `MailboxDashBoardController.handleRefreshTokenFailedException()`, and throws `RefreshTokenFailedException`; `_onDrivePickState` (line 81) swallows that exception so the picker layer does not toast a second time. Covered by `fatal rejection triggers the logout handler exactly once…` and now by the real-interceptor race test.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Addressed in bb5c4fa98. `_retryAfterRefreshOrThrow` (lines 130-133) re-reads `oidcTokenGetter()` first; when the current id token already differs from the one that got the 401 it retries with T1 and never calls `oidcRefreshTrigger`, mirroring `AuthorizationInterceptors.validateToRetryTheRequestWithNewToken`. Test `reuses the already-refreshed token when a second 401 arrives late, with one refresh` asserts one refresh and, since 0ca877b8c, that both retries send the refreshed token.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +20,65 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+  final interceptor = getBinding<AuthorizationInterceptors>();
+  if (interceptor == null) return null;
+  try {
+    final newToken = await interceptor.requestTokenRefresh();
+    return newToken.tokenId.uuid;
+  } catch (e) {
+    if (!interceptor.isRefreshFailureFatal(e)) rethrow;
+    interceptor.clear();
+    // Same funnel as the interceptor's own reject(RefreshTokenFailedException) path.
+    getBinding<MailboxDashBoardController>()?.handleRefreshTokenFailedException();
```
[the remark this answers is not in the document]
tddang-linagora: Duplicate calls are already absorbed downstream: `clearDataAndGoToLoginPage` funnels into `TwakeAppManager.runClearDataOnce` (`lib/main/utils/twake_app_manager.dart:11`), so a second caller joins the in-flight teardown instead of starting another. A guard here could not dedupe the JMAP path either, which reaches the handler via `BaseController.handleUrgentException*`, and two concurrent JMAP 401s already yield two handler calls today. Not adding a second dedup layer.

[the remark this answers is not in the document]
tddang-linagora: The second navigation is already idempotent. On web `navigateToLoginPage` (base_controller.dart:487) returns early when `Get.currentRoute == AppRoutes.login`, and `pushAndPopAll` is `Get.offAllNamed`, which pushes synchronously, so the second resumer of the shared `clearAllData` future hits that guard. On mobile it re-runs the login-info lookup and `offAllNamed(login)` again with the same result and no data touched. Two Workplace callbacks on one rejected refresh need two Drive pickers open at once, and the picker is a non-dismissible modal (drive_picker_state_mixin.dart:123), so that pair cannot occur; the JMAP+Workplace pair is the same shape the app already has when a queued second JMAP 401 surfaces as `BadCredentialsException` after `clear()`, which a guard in this provider cannot reach. Adding a one-shot flag to a `keepAlive` provider also needs a reset on re-login or the next session never force-logs-out; that is more risk than the duplicate it removes.

hoangdat:
- [x] should take care the case of `refresh Token` not response new `id_token`, should reuse the old one

Ref: https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -365,13 +378,6 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
         );
         return super.onError(err, handler);
       }
```
hoangdat: why not keep the code inside `_acquireAndPersistNewToken` (intead of return we throw in case of same-token)?

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +20,65 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+    throw RefreshTokenFailedException();
```
hoangdat: why workplace need to handle this instead of interceptor?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -351,30 +324,25 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    } on RefreshTokenDuplicatedException {
+      // Retrying cannot clear the 401, so it propagates to logout; tag it for forensics.
+      _logForcedLogoutFor401(
+        authErrorType: 'forced_logout_401_refreshed_token_duplicated',
+        err: err,
+        hasAttemptedRefresh: true,
+      );
+      return super.onError(err, handler);
+    } on RefreshTokenFailedException catch (refreshError) {
+      // Session already cleared by requestTokenRefresh; surface the dead session.
```
hoangdat: why not `logError` here instead of inside _acquireAndPersistNewToken?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +596,54 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
+    if (PlatformInfo.isIOS) {
+      await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
+    }
+
+    return newTokenOidc;
+  }
+
+  // OIDC Core 12.2: a refresh response may omit id_token; keep the one we have.
+  TokenOIDC _withCurrentIdTokenIfMissing(TokenOIDC fresh) {
```
hoangdat: why not keep it in `lib/features/login/data/extensions/token_response_extension.dart`, same as `maybeAvailableRefreshToken`?

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -91,6 +92,20 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
       _authenticationType == AuthenticationType.oidc ? _token : null;
 
+  /// Triggers a refresh, or joins one already running — dedupes an external
+  /// caller (e.g. Workplace's own Dio) against this interceptor's [onError].
+  /// Owns the outcome: throws [RefreshTokenDuplicatedException] on a same-token
+  /// response, clears the session and throws [RefreshTokenFailedException] on a
+  /// server rejection, rethrows transient failures untouched.
+  Future<TokenOIDC> requestTokenRefresh() {
+    return _refreshInFlight ??= _acquireAndPersistNewToken()
+        .whenComplete(() => _refreshInFlight = null);
```
hoangdat: `clear()` leaves `_refreshInFlight` set. If we log out while a refresh awaits AppAuth, the next login can join that old Future → wrong tokens or RefreshTokenFailedException. JMAP+Drive 401 sharing is OK. Fix: `_refreshInFlight=null` in `clear()` + identity-guard `whenComplete`.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -67,29 +73,74 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
...
+    if (refreshedToken == null) throw failure;
+
+    return _exchangeAccessToken(platformUrl, refreshedToken, refreshAttempted: true);
+  }
+
+  bool _isUnauthorized(Object failure) =>
```
hoangdat: I worry about this condition, please ensure with the cozy and write the test to docs it.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -67,29 +73,74 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
...
       );
     }
-    return accessToken;
+    return caughtFailure == null ? Right(accessToken) : Left(caughtFailure!);
   }
 
+  /// Retries once on a 401 with the current token if another request already
+  /// refreshed it, else with a freshly refreshed one (Workplace's Dio has no
+  /// refresh interceptor of its own).
+  Future<String?> _retryAfterRefreshOrThrow({
```
hoangdat: it quite a confused name


```suggestion
  Future<String?> _triggerRefreshOIDCToken({
```

hoangdat: should add a test to ensure no forever loop on it

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -67,29 +73,74 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
...
+    required Uri platformUrl,
+    required String failedToken,
+    required Object failure,
+    required bool refreshAttempted,
+  }) async {
+    if (refreshAttempted || oidcRefreshTrigger == null || !_isUnauthorized(failure)) {
```
hoangdat: please ensure we have the dio exception in this `failure`?

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
hoangdat: If the IdP omits id_token on refresh, this returns the same token that just 401'd, so the retry is guaranteed to fail. Please log the id token before/after here so prod tells us which case happened.

hoangdat (commented):
Three anchor notes below. Separately, two red tests were added on the branch for the cleared-session refresh crash and the duplicate-token classification gap.

workplace/test/presentation/extension/workplace_composer_attachment_extension_test.dart
```diff
@@ -569,5 +560,148 @@ void main() {
...
+      WorkplaceDio.setInstance(Dio()..httpClientAdapter = adapter);
+
+      var refreshCallCount = 0;
+      var currentToken = 'old-oidc-token';
+      var tokenReads = 0;
+      final notifier = ValueNotifier<Uri?>(_platformUri);
+      final ext = _makeExtension(
+        notifier,
+        // Reads 1-2: A start + A retry check; read 3: B start (old); read 4: B retry check.
+        oidcTokenGetter: () => ++tokenReads <= 3 ? 'old-oidc-token' : currentToken,
```
hoangdat: This counter hard-codes how many times _fetchIntent reads oidcTokenGetter. A refactor that changes the read count silently flips which branch this exercises. Prefer a fake with an explicit stale flag.

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -1640,6 +1641,231 @@ void main() {
...
+        expect(response.statusCode, equals(HttpStatus.ok));
+        expect(workplaceToken.token, equals(OIDCFixtures.newTokenOidc.token));
+      },
+    );
+  });
+
+  // ============================================================
+  // requestTokenRefresh: owns the fatal-vs-transient decision
+  // ============================================================
+  group('requestTokenRefresh: owns the fatal-vs-transient decision', () {
```
hoangdat, line 1755: All new requestTokenRefresh groups run non-web only, so the web branch of _isRefreshRejectedByServer (ArgumentError from flutter_appauth_web) is untested for direct callers. Drive attach ships on web.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -91,6 +92,20 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
   TokenOIDC? get currentToken =>
       _authenticationType == AuthenticationType.oidc ? _token : null;
 
+  /// Triggers a refresh, or joins one already running — dedupes an external
+  /// caller (e.g. Workplace's own Dio) against this interceptor's [onError].
+  /// Owns the outcome: throws [RefreshTokenDuplicatedException] on a same-token
+  /// response, clears the session and throws [RefreshTokenFailedException] on a
+  /// server rejection, rethrows transient failures untouched.
+  Future<TokenOIDC> requestTokenRefresh() {
```
dab246, line 106: **Blocking — red test.** `clear()` nulls `_configOIDC`/`_token` and `_invokeRefreshTokenFromServer` dereferences both, so a Workplace 401 arriving after the JMAP path already killed the session throws `_TypeError: Null check operator used on a null value`. That's an `Error`, so it can't be classified urgent and ends in a generic toast.

Guard the entry point on the session preconditions and throw `RefreshTokenFailedException` when there is nothing to refresh with.

test/features/base/base_controller_test.dart
```diff
@@ -146,6 +161,22 @@ void main() {
...
+    test('should return true when exception is RefreshTokenFailedException', () {
+      expect(mockBaseController.validateUrgentException(RefreshTokenFailedException()), isTrue);
+    });
+
+    // Journey A (JMAP 401): a same-token refresh reaches super.onError, becomes a
+    // BadCredentialsException and forces logout. Journey B (Drive token_exchange
+    // 401) surfaces the RefreshTokenDuplicatedException itself, so it must be
+    // classified urgent here or the dead session only produces a toast.
+    // KNOWN FAILING: RefreshTokenDuplicatedException is not in validateUrgentException.
+    test('should return true when exception is RefreshTokenDuplicatedException', () {
```
dab246, line 177: **Blocking — red test.** Add `RefreshTokenDuplicatedException` to `validateUrgentException`, otherwise a same-token refresh discovered by Drive only toasts, while the same condition on the JMAP path forces logout — the asymmetry the ADR says this PR removes.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +596,54 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+        exception: e,
+        stackTrace: st,
+        extras: _errorClassifier.buildSentryExtras(e),
+        webConsoleEnabled: true,
+      );
+      clear();
+      throw RefreshTokenFailedException();
+    }
+
+    final newTokenOidc = _withCurrentIdTokenIfMissing(acquired);
```
dab246: Nothing ties the in-flight refresh to the session that started it. `_clearAllData()` runs `cachingManager.clearAll()` → `clear()` → `closeHive()`; a Workplace-initiated refresh landing in that window re-arms `_token` on a cleared interceptor and re-persists it through `_updateCurrentAccount` after the caches were wiped.

Capture a session generation before the awaits and drop the result if it changed; reset `_refreshInFlight` in `clear()`.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
dab246: The branch above skips a pointless retry when the token already changed, but the post-refresh path doesn't repeat that check. With `_withCurrentIdTokenIfMissing` now deliberately keeping the current id token when the refresh response omits `id_token`, `refreshedToken == failedToken` is the expected case — and it re-sends the exact token that just 401'd.

Compare against `failedToken` and throw the original failure instead.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
dab246: Does `token_exchange` really answer 401 for an expired `id_token`? RFC 8693 specifies 400 `invalid_grant` for a bad `subject_token`. If it's 400, the retry never fires. Worth confirming against a captured response.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -29,6 +31,9 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
   /// Read when the picker opens, so the JMAP capability is always current.
   final ValueGetter<bool> uploadFromUrlSupported;
   final String? Function() oidcTokenGetter;
+
+  /// Triggers the main app's OIDC refresh; returns the refreshed id token.
+  final Future<String?> Function()? oidcRefreshTrigger;
```
dab246: One production call site, always passing the trigger — `required` removes the degraded mode and the `== null` branch only tests reach. Also inline `Future<String?> Function()?` breaks this file's typedef convention (`OnDrivePickStateChanged`, `ValueGetter`).

workplace/test/presentation/extension/workplace_composer_attachment_extension_test.dart
```diff
@@ -569,5 +560,148 @@ void main() {
       expect(downloadLink!.maxFileSize, equals(5000));
       expect(downloadLink.availableSize, equals(5000));
     });
+
+    testWidgets('retries once via oidcRefreshTrigger after a 401, then succeeds', (tester) async {
```
dab246, line 621: Two cheap edges still uncovered: non-401 failure → trigger never called; trigger returns null → the original 401 is thrown, not a `StateError` from the null access token.

dab246 (changes requested):
Two committed tests are red: `requestTokenRefresh` on a cleared session, and `RefreshTokenDuplicatedException` not classified urgent. Confirmed locally — 2 failures in the main package, workplace suite green.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +19,64 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+    // No catch here: handleDrivePickResult swallows and toasts its own errors.
+    await composer.handleDrivePickResult(state.documents);
+  } else if (state is DrivePickFailure) {
+    final dashboard = getBinding<MailboxDashBoardController>();
+    final error = state.error;
+    if (dashboard != null && error is Exception && dashboard.validateUrgentException(error)) {
+      // Session-level failures take BaseController's standard path (logout, reconnect).
+      dashboard.handleUrgentException(failure: state, exception: error);
+      return;
+    }
```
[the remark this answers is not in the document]
tddang-linagora: First finding: done. `RefreshTokenDuplicatedException` now in `validateUrgentException` + both `handleUrgentException` branches.

Second finding (coalesce recovery): no. Same thing you opened as #3966038622 and withdrew — nothing changed since. `runClearDataOnce` shares teardown, web `navigateToLoginPage` returns early on the login route, and two Workplace callbacks need two pickers open at once while the picker is a non-dismissible modal (`drive_picker_state_mixin.dart:125`). One-shot flag on a `keepAlive` provider also needs reset on re-login or next session never force-logs-out. More risk than the duplicate it removes.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Already covered. `workplace_composer_attachment_extension_test.dart` "does not retry twice when the refreshed token also gets a 401" feeds two 401s and asserts `refreshCallCount == 1`.

Bound is structural, not luck: `_exchangeAccessToken` recurses with `refreshAttempted: true`, and the guard throws immediately on that flag. One retry per exchange whatever the server returns.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Yes, it is a `DioException`. `WorkplaceDataSourceImpl.exchangeToken` does not catch, so Dio throws straight into `ExchangeDriveTokenInteractor.execute`, which wraps it verbatim as `ExchangeWorkplaceTokenFailure(exception: e)`; `_requestAccessToken` then unwraps `failure.exception`.

The `_fail401` / `_fail400` adapters in the retry tests exercise that chain end to end.

test/features/base/base_controller_test.dart
```diff
@@ -219,6 +252,59 @@ void main() {
...
+        when(mockTwakeAppManager.hasComposer).thenReturn(true);
+
+        mockBaseController.handleUrgentException(exception: RefreshTokenFailedException());
+        await pumpEventQueue();
+
+        verifyInOrder([
+          mockTwakeAppManager.setExecutingBeforeReconnect(true),
+          mockBeforeReconnectManager.executeBeforeReconnectListeners(),
+        ]);
+        expect(mockBaseController.logoutCalls, 1);
```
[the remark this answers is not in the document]
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/test/presentation/extension/workplace_composer_attachment_extension_test.dart
tddang-linagora: Done

test/features/interceptor/authorization_interceptor_test.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

test/features/base/base_controller_test.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

workplace/test/presentation/extension/workplace_composer_attachment_extension_test.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +627,37 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+      clear();
+      // Each caller logs it — this one method serves both JMAP and Workplace.
+      throw RefreshTokenFailedException(cause: e);
+    }
+
+    // The session died while we were away; re-arming it would resurrect a
+    // logged-out account and re-persist it over wiped caches.
+    if (generation != _sessionGeneration) throw RefreshTokenFailedException();
+
+    final newTokenOidc = acquired.withIdTokenFallback(_token?.tokenId);
```
hoangdat: why not handle it in the same place with `maybeAvailableRefreshToken`? in side `AuthenticationClientBase::refreshingTokensOIDC`?

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -132,14 +135,26 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
       return _exchangeAccessToken(platformUrl, currentToken, refreshAttempted: true);
     }
 
-    final refreshedToken = await oidcRefreshTrigger!();
-    if (refreshedToken == null) throw failure;
+    final refreshedToken = await oidcRefreshTrigger();
+    logWarning(
```
hoangdat, line 159: should enable console web = true

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -1640,6 +1641,417 @@ void main() {
...
+        expect(authorizationInterceptors.authenticationType, AuthenticationType.oidc);
+      },
+    );
+
+    test(
+      'GIVEN a request that 401ed and is awaiting a refresh from its own session\n'
+      'WHEN the user logs out and a new session signs in before that refresh lands\n'
+      'THEN the leftover request fails without a session verdict\n'
+      'SO it cannot force the freshly signed-in session to log out.\n'
+      'KNOWN FAILING: the generation guard throws RefreshTokenFailedException,\n'
```
dab246: This expectation fails today: the refresh resolves to `RefreshTokenFailedException`, which is exactly the session verdict a leftover request must not carry.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +627,37 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    } catch (e) {
+      if (!_isRefreshRejectedByServer(e)) rethrow;
+      clear();
+      // Each caller logs it — this one method serves both JMAP and Workplace.
+      throw RefreshTokenFailedException(cause: e);
+    }
+
+    // The session died while we were away; re-arming it would resurrect a
+    // logged-out account and re-persist it over wiped caches.
+    if (generation != _sessionGeneration) throw RefreshTokenFailedException();
```
dab246: A stale generation means the answer is irrelevant, not that the session is dead — `RefreshTokenFailedException` here reaches `validateUrgentException` and logs out whichever session is signed in by then.

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
```diff
@@ -19,42 +20,70 @@ ComposerAttachmentExtensionRegistry composerAttachmentExtensionRegistry(Ref ref)
...
+/// RefreshTokenFailedException and is routed by [_onDrivePickState].
+Future<String?> _refreshWorkplaceOidcToken() async {
+  final interceptor = getBinding<AuthorizationInterceptors>();
+  if (interceptor == null) return null;
+  try {
+    final newToken = await interceptor.requestTokenRefresh();
+    return newToken.tokenId.uuid;
+  } on RefreshTokenFailedException catch (e, st) {
+    // The JMAP path logs its own; without this the Drive journey is silent.
+    interceptor.logFatalRefreshRejection(e, st);
```
dab246: `DriveIntentMessageHandlerMixin._failWith` already reports this exception to Sentry once it reaches the intent loader, so every fatal Drive refresh emits two events for the same error.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -91,6 +96,46 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+  /// caller (e.g. Workplace's own Dio) against this interceptor's [onError].
+  /// Owns the outcome: throws [RefreshTokenDuplicatedException] on a same-token
+  /// response, clears the session and throws [RefreshTokenFailedException] on a
+  /// server rejection, rethrows transient failures untouched.
+  Future<TokenOIDC> requestTokenRefresh() {
+    final inFlight = _refreshInFlight;
+    if (inFlight != null) return inFlight;
+
+    // Nothing left to refresh with — the session already died elsewhere.
+    if (_configOIDC == null || _token == null) {
```
dab246, line 111: This guard is weaker than `validateToRefreshToken`: an OIDC session with an empty refresh token still reaches `_invokeRefreshTokenFromServer`, and the resulting rejection clears the session, so a Drive-only 401 ends in a full logout.

dab246 (commented):
Four issues on the refresh path, details inline.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

test/features/interceptor/authorization_interceptor_test.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/composer/presentation/providers/composer_attachment_extension_registry_provider.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/extensions/token_response_extension.dart
```diff
@@ -4,11 +4,17 @@ import 'package:model/model.dart';
...
+      TokenId(_orCurrent(idToken, currentToken.tokenId.uuid)),
+      _orCurrent(refreshToken, currentToken.refreshToken),
       expiredTime: accessTokenExpirationDateTime ?? DateTime.now());
   }
-}
\ No newline at end of file
+
+  // The plugin returns '' rather than null when a field is absent.
+  String _orCurrent(String? value, String current) =>
+      value?.isNotEmpty == true ? value! : current;
```
hoangdat: we miss the case empty for `refreshToken`?

`refreshToken ?? maybeAvailableRefreshToken ?? ''`

tddang-linagora: `_orCurrent` is the empty case — it tests `value?.isNotEmpty == true`, which `??` alone misses.
The trailing `?? ''` is unreachable: `TokenOIDC.refreshToken` is non-nullable
(`model/lib/oidc/token_oidc.dart:17`).

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +640,38 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
 
+  Future<TokenOIDC> _acquireAndPersistNewToken(int generation) async {
+    final TokenOIDC acquired;
+    try {
+      acquired = PlatformInfo.isIOS
+          ? await _getNewTokenForIOSPlatform()
+          : await _getNewTokenForOtherPlatform();
+    } catch (e, st) {
+      if (!_isRefreshRejectedByServer(e)) rethrow;
+      clear();
```
dab246, line 662: The `generation != _sessionGeneration` guard below does not cover this branch, so a rejection for an already-replaced session still clears the session that signed in after it.

dab246 (commented):
One issue on the shared refresh path, details inline.

- [x] `authorization_interceptors.dart:407` has the same missing generation check on its `clear()`.

test/features/interceptor/authorization_interceptor_test.dart
```diff
@@ -1640,6 +1641,375 @@ void main() {
...
+          newToken: OIDCFixtures.tokenOidcExpiredTime,
+          newConfig: OIDCFixtures.oidcConfiguration,
+        );
+        when(authenticationClient.refreshingTokensOIDC(any, any, any, any, any))
+            .thenAnswer((_) async => OIDCFixtures.tokenOidcNotExpiredYet);
+        when(accountCacheManager.getCurrentAccount())
+            .thenAnswer((_) async => AccountFixtures.aliceAccount);
+
+        final fresh = authorizationInterceptors.requestTokenRefresh();
+        gate.complete(OIDCFixtures.tokenOidcNotExpiredYet);
```
dab246, line 1981: All three stale-session tests complete this gate successfully, so a refresh rejected after the session was replaced is untested.

dab246 (commented):
One note on the stale-session coverage, inline.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

test/features/interceptor/authorization_interceptor_test.dart
tddang-linagora: Done

tddang-linagora:
> One issue on the shared refresh path, details inline.
> 
>     * [x]  `authorization_interceptors.dart:407` has the same missing generation check on its `clear()`.

`:408` can no longer see a stale-generation error. Every refresh failure goes through
`_acquireAndPersistNewToken`'s catch (`authorization_interceptors.dart:650`), which now throws
`StaleSessionRefreshException` before both the `clear()` at :655 and the rethrow that feeds the
DioException branch; `_refreshTokenThenRetry` catches that at :375 and keeps the session, above the
`on DioException` clause at :389. `clear()` at :408 runs only for a 400 on a refresh whose
generation is still current — the session's own verdict. A second guard there would be unreachable,
and any test for it would have to fake the invariant it asserts.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +641,41 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    // The session died while we were away; re-arming it would resurrect a
+    // logged-out account and re-persist it over wiped caches.
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
+
+    if (acquired.token == _token?.token) {
+      throw const RefreshTokenDuplicatedException();
+    }
+    _updateNewToken(acquired);
+
+    final personalAccount = await _updateCurrentAccount(tokenOIDC: acquired);
```
dab246: `_updateCurrentAccount` can re-persist the refreshed token over caches `_clearAllData` already wiped: the guard above runs before these awaits, and `clear()` lands only after `cachingManager.clearAll()`. Master runs this same sequence with no guard at all, so the race predates this PR and the new guard narrows it — re-checking the generation immediately before the writes would close it.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +641,41 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+          : await _getNewTokenForOtherPlatform();
+    } catch (e, st) {
+      // The session that started this refresh is gone; its answer is not a
+      // verdict on the session that replaced it.
+      if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
+      if (!_isRefreshRejectedByServer(e)) rethrow;
+      clear();
+      // Logged here, not per caller: this one method serves JMAP and Workplace.
+      final fatal = RefreshTokenFailedException(cause: e);
+      _logFatalRefreshRejection(fatal, st);
```
dab246, line 665: Centralising the fatal log here did remove the caller-side duplicate, but `DriveIntentMessageHandlerMixin._failWith` still reports the `RefreshTokenFailedException` that `requestTokenRefresh` throws once it reaches the intent loader, so a fatal Drive refresh is two Sentry events for one incident. The direct-caller test asserting exactly one error event only observes the interceptor, not the Workplace funnel above it.

test/features/composer/presentation/composer_controller_test.dart
```diff
@@ -139,10 +140,14 @@ class MockMailboxDashBoardController extends Mock implements MailboxDashBoardCon
   @override
   InternalFinalCallback<void> get onDelete => mockControllerCallback();
 
+  // Overridable so a test can reproduce a session torn down mid-logout.
+  AccountId? currentAccountId = AccountFixtures.aliceAccountId;
```
dab246, line 145: `currentAccountId` and `currentSession` shadow members `MailboxDashBoardController` already declares, so these two lines add the only new analyzer output this branch produces: `annotate_overrides` on both. Add `@override`, or rename the fields so they are not overrides.

dab246 (commented):
Three notes on the refresh path, details inline.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Two events, two stages: interceptor logs session cleared, `_failWith` logs Drive stage died.
To dedupe, `_failWith` must know `RefreshTokenFailedException` — it lives in app `lib/`, and
`workplace` depends only on `core` + `model`. Needs a new "already reported" marker in `core`
across both packages, for duplicated telemetry with no correctness/UX impact. Not worth it.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

test/features/composer/presentation/composer_controller_test.dart
tddang-linagora: Done

dab246 (dismissed):
lgtm

lib/features/login/data/network/interceptors/authorization_interceptors.dart
dab246: These are two propagation stages of the same incident, not two independent failures. The interceptor reports the token-endpoint rejection, then _failWith reports the resulting RefreshTokenFailedException again; both become Sentry events.  I agree workplace must not import the app exception. It only needs a package-neutral, per-instance “already reported” contract in core, or an injected reporting policy. Please add an interceptor → Workplace loader → _failWith regression test asserting one error event.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -633,6 +648,44 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+      throw fatal;
+    }
+
+    // The session died while we were away; re-arming it would resurrect a
+    // logged-out account and re-persist it over wiped caches.
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
+
+    if (acquired.token == _token?.token) {
+      throw const RefreshTokenDuplicatedException();
+    }
```
dab246, line 675: Drive uses `tokenId.uuid`, but this check compares the access token. A refresh returning the same access token with a new ID token is incorrectly rejected and can log the user out. Please make duplicate detection caller-specific.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -568,9 +576,16 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
-  Future<PersonalAccount> _updateCurrentAccount({required TokenOIDC tokenOIDC}) async {
+  Future<PersonalAccount> _updateCurrentAccount({
+    required TokenOIDC tokenOIDC,
+    required int generation,
+  }) async {
     final currentAccount = await _accountCacheManager.getCurrentAccount();
 
+    // The session died while we read the account; these writes land between
+    // clearAll() and closeHive(), resurrecting it on the next launch.
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
```
dab246, line 587: This guard only runs before persistence. `clear()` can still run during the following awaits, allowing the refresh to write session data after logout and return success. Please guard the complete persistence sequence and cover this race with a test.

Crash--:
From my discussions with @hoangdat  a while ago, I understood we had dropped this implementation because it was too costly.

I also raised a point then: if we still had to do it, it should be an external SDK rather than something specific to TMail. That way we could reuse it in Chat too, and the responsibilities would be clearly split.

I'd much rather see a `twake_workplace` package that handles the Workplace side (token exchange, detecting a stale token, retrying once, Cozy bridge, the intent creation). TMail would just consume it and plug in its own auth through a small interface. And of course, we can also add in the design system a new widget to handle the UI of the intents, but here this is bit out of scope. 

TMail would keep only what is really its own: refreshing the session without duplicate calls (requestTokenRefresh()), plus the composer integration and error handling.

tddang-linagora:
> From my discussions with @hoangdat a while ago, I understood we had dropped this implementation because it was too costly.
> 
> I also raised a point then: if we still had to do it, it should be an external SDK rather than something specific to TMail. That way we could reuse it in Chat too, and the responsibilities would be clearly split.
> 
> I'd much rather see a `twake_workplace` package that handles the Workplace side (token exchange, detecting a stale token, retrying once, Cozy bridge, the intent creation). TMail would just consume it and plug in its own auth through a small interface. And of course, we can also add in the design system a new widget to handle the UI of the intents, but here this is bit out of scope.
> 
> TMail would keep only what is really its own: refreshing the session without duplicate calls (requestTokenRefresh()), plus the composer integration and error handling.

That's already agreed. This PR does not add the refresh capability for workplace token. It enables the external refresh capability of the jmap session, should any third party module needs while using the app's jmap token (in this case, workplace does, as token_exchange consume the jmap token).

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/extensions/token_response_extension.dart
```diff
@@ -4,11 +4,18 @@ import 'package:model/model.dart';
 
 extension TokenResponseExtension on TokenResponse {
 
-  TokenOIDC toTokenOIDC({String? maybeAvailableRefreshToken}) {
+  // A refresh response may omit refresh_token, and OIDC Core 12.2 lets it omit
+  // id_token too; keep whichever the current token already holds.
+  TokenOIDC toTokenOIDC({required TokenOIDC currentToken}) {
```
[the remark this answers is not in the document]
tddang-linagora: if there is such compilation error, the CI analysis would be red.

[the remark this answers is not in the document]
tddang-linagora: read the definition in `lib/features/login/data/extensions/authentication_token_extension.dart`

hoangdat:
hi @Crash-- , 

We applied the way to communicate with `Drive picker` via `Cozy bridge`.

But this PR addresses different problem:  tmail oidc token expired before communicate with external services. Fx: `token_exchange` (for mobile drive picker) or future functions. Before this PR, the refresh OIDC mechanism only work with JMAP requests. 

So we refactoring the logic to reuse the refresh handling for not only JMAP requests also other requests which use tmail OIDC token.

lib/features/base/base_controller.dart
```diff
@@ -163,7 +163,8 @@ abstract class BaseController extends GetxController
     return exception is NoNetworkError
       || exception is BadCredentialsException
       || exception is ConnectionError
-      || exception is RefreshTokenFailedException;
+      || exception is RefreshTokenFailedException
+      || exception is RefreshTokenDuplicatedException;
```
dab246: The `RefreshTokenDuplicatedException` is not thrown outside the interceptor. So the code handling this exception has become dead code.

lib/features/base/base_controller.dart
```diff
@@ -185,7 +186,9 @@ abstract class BaseController extends GetxController
       _handleConnectionErrorException();
     } else if (exception is BadCredentialsException) {
       handleBadCredentialsException();
-    } else if (exception is RefreshTokenFailedException) {
+    } else if (exception is RefreshTokenFailedException
+        || exception is RefreshTokenDuplicatedException) {
```
dab246: idem

lib/features/base/base_controller.dart
```diff
@@ -202,7 +205,9 @@ abstract class BaseController extends GetxController
...
```
dab246: idem

docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
```diff
@@ -0,0 +1,26 @@
...
+Accepted
+
+## Context
+
+Workplace's Drive `token_exchange` request runs on `WorkplaceDio`, a bare `Dio()` instance with no interceptors. A 401 on that request (a stale OIDC id token) has no refresh-and-retry path, unlike the main app's Dio, which refreshes via `AuthorizationInterceptors`. `AuthorizationInterceptors` lives in the main app package; `workplace` is a separate package that cannot depend on it.
+
+## Decision
+
+- `AuthorizationInterceptors` exposes `requestTokenRefresh()`, a dedup-guarded public entry point backed by an in-flight `Future<TokenOIDC>`. Any caller — the interceptor's own reactive `onError` path or an external caller — joins the same in-flight refresh instead of starting a second one.
+- `requestTokenRefresh()` owns the outcome of a refresh for every caller: a server rejection (RFC 6749 400/401-equivalent, classified by the existing web/mobile classifiers) clears the session and throws `RefreshTokenFailedException`; a same-token response throws `RefreshTokenDuplicatedException` before anything is persisted; transient failures are rethrown untouched.
```
dab246: Update ADR to reflect the new implementation.

test/features/composer/presentation/providers/composer_attachment_extension_registry_provider_test.dart
```diff
@@ -0,0 +1,288 @@
...
+      const duplicated = RefreshTokenDuplicatedException();
+      when(interceptor.requestTokenRefresh()).thenThrow(duplicated);
+
+      await expectLater(
+        readExtension().oidcRefreshTrigger(),
+        throwsA(same(duplicated)),
+      );
+
+      verifyNever(interceptor.clear());
+    });
```
dab246: This test does not yet align with the production code logic.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -585,6 +607,7 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
       userName: currentAccount.userName
     );
     await _accountCacheManager.setCurrentAccount(personalAccount);
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
```
dab246: When `clear()` happens while `setCurrentAccount()` is still in flight:

```dart
await setCurrentAccount(personalAccount);
```

the sequence can be:

1. The previous generation guard has already passed.
2. `setCurrentAccount()` starts writing the account.
3. Logout runs `clearAll()` and wipes both the account and token.
4. `setCurrentAccount()` completes and persists the account again.
5. Only then does the following guard detect the stale generation and throw.

The exception is thrown, but the account has already been written back.

Therefore, the check after `setCurrentAccount()` is only a **detector**, not a **guard** — it can detect that the session became stale, but it cannot prevent the write that has already completed.

### Impact

After logout, the cache can end up in an inconsistent state:

* The account still exists with `isSelected: true`.
* The OIDC token has already been removed.

On the next launch:

1. The app reads the persisted account and considers the user authenticated.
2. It then fails to find the corresponding token.
3. The app enters a **half-authenticated state** instead of cleanly returning to the login screen.

tddang-linagora: Good catch, the post-write check is a detector not a guard. Three ways to close it, ranked by how much they cost:

1. **Delete-on-detect** — when the post-`setCurrentAccount` check finds a stale generation, call `_accountCacheManager.deleteCurrentAccount(personalAccount.hashId)` before throwing. `deleteCurrentAccount` already exists and is keyed by hash id, so it can't clobber a different account a subsequent login wrote in the meantime. Smallest diff, no new primitive; there's still a brief window where the stale account is visible before the undo runs.
   ```dart
   await _accountCacheManager.setCurrentAccount(personalAccount);
   if (generation != _sessionGeneration) {
     // The write raced a clear(); undo it so the cache doesn't say "authenticated" with no token.
     await _accountCacheManager.deleteCurrentAccount(personalAccount.hashId);
     throw const StaleSessionRefreshException();
   }
   ```
2. **`clear()` waits then re-wipes** — `clear()` awaits `_refreshInFlight` (if any) to settle, then unconditionally re-wipes the account+token cache. Guarantees the final state is clean regardless of interleaving; delays logout completion by however long the in-flight refresh takes; `clear()` becomes async, so all call sites need checking.
   ```dart
   void clear() async {
     final inFlight = _refreshInFlight;
     _authorization = null;
     _token = null;
     _configOIDC = null;
     _authenticationType = AuthenticationType.none;
     _sessionGeneration++;
     _refreshInFlight = null;
     if (inFlight != null) {
       await inFlight.catchError((_) {});
       await _accountCacheManager.clear();
       await _tokenOidcCacheManager.clear();
     }
   }
   ```
3. **Mutual-exclusion lock** — wrap the persist-token+set-account critical section in `_updateCurrentAccount` and `clear()`'s cache wipe in the same `package:synchronized` `Lock` (already a transitive dep). Makes the interleave impossible rather than repairing it after the fact; strongest guarantee, but changes `clear()`'s signature to async and adds a new sync primitive to the interceptor — more surface area for a bug fix.
   ```dart
   final _cacheLock = Lock();

   Future<PersonalAccount> _updateCurrentAccount(...) =>
     _cacheLock.synchronized(() async {
       if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
       await _tokenOidcCacheManager.persistOneTokenOidc(tokenOIDC);
       ...
     });

   Future<void> clear() => _cacheLock.synchronized(() async {
     ...
     await _accountCacheManager.clear();
     await _tokenOidcCacheManager.clear();
   });
   ```

Leaning toward (1) for this PR — smallest, keyed precisely, no signature changes — with (3) as a follow-up if this class of race shows up elsewhere. Open to your take.

**File that would be touched (whichever direction is picked):** `lib/features/login/data/network/interceptors/authorization_interceptors.dart` (`_updateCurrentAccount` / `clear()`)

lib/features/base/base_controller.dart
tddang-linagora: Done

lib/features/base/base_controller.dart
tddang-linagora: Done

lib/features/base/base_controller.dart
tddang-linagora: Done

docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
tddang-linagora: Done

test/features/composer/presentation/providers/composer_attachment_extension_registry_provider_test.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
dab246: **Option 1:** The right direction, but not sufficient as-is. `personalAccount.hashId` should be `personalAccount.id`, and the same delete-on-detect logic is also needed for `persistOneTokenOidc`. More importantly, the generation bump must happen **before** the wipe to close the main race window.

**Option 2:** Should be dropped. It can deadlock because `clear()` is called from inside `_refreshInFlight` and then waits for the same in-flight future. It is also fire-and-forget, so `closeHive()` can race with the cleanup.

**Option 3:** Technically sound and provides the strongest guarantee, but it is more complex than necessary. It requires adding `package:synchronized` as an explicit dependency, and `clear()` is still not awaited by its production caller.

**Recommendation:** Move the generation bump before `Future.wait`, then apply Option 1's delete-on-detect to both account and token writes. This is the simplest correct fix without introducing a new locking primitive.

tddang-linagora: Done

Crash--:
Thanks @hoangdat, that's clear now. So the refresh stays purely reactive: a 401 on JMAP, or a 400/401 on `token_exchange` for the Workplace path, and `requestTokenRefresh()` only makes the existing JMAP refresh reachable from outside the interceptor. No objection to that goal.

What still bothers me is that the policy around it now exists twice: detect the stale token, check whether another request already refreshed it, refresh once, retry once. It lives both in `AuthorizationInterceptors` and in `WorkplaceComposerAttachmentExtension`. That is the same thing you ask inline: _"why workplace need to handle this instead of interceptor?"_

I understand a plain interceptor on `WorkplaceDio` would not be enough, since `token_exchange` carries the id token in the body rather than in the `Authorization` header, so the retry has to rebuild the request. But that is an argument for extracting that policy as one small reusable piece, not for duplicating it. Something like "given a request, a staleness test, and a way to get a fresh token, run it and retry once" — used by both the interceptor and Workplace.

That piece is also exactly what would move into `twake_workplace` later, with TMail providing only `refreshIdToken()`. Happy to go with the current shape for this PR if you prefer to unblock the bug, as long as we agree the dedup of that logic is the follow-up.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

lib/features/base/base_controller.dart
```diff
@@ -649,6 +649,10 @@ abstract class BaseController extends GetxController
 
   Future<void> _clearAllData() async {
     try {
+      // Bump the session generation BEFORE wiping caches, not after: an
+      // in-flight refresh must see itself as stale before its writes race the wipe.
+      authorizationInterceptors.clear();
+      authorizationIsolateInterceptors.clear();
       await Future.wait([
         if (isAuthenticatedWithOidc)
```
dab246: `isAuthenticatedWithOidc` is always false.

lib/features/base/base_controller.dart
```diff
@@ -649,6 +649,10 @@ abstract class BaseController extends GetxController
 
   Future<void> _clearAllData() async {
```
dab246, line 650: Should add unit test for this. add a test to assert the call order.

docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
```diff
@@ -0,0 +1,26 @@
...
+- Same-token detection is not centralized: JMAP's own `onError` path and the Workplace wiring each compare the token they had before the refresh against the one `requestTokenRefresh()` returns, since they check different fields. A `RefreshTokenFailedException` (the only fatal type that crosses the caller boundary) reaches `MailboxDashBoardController`'s `BaseController` handling (`validateUrgentException` / `handleUrgentException`), the same path the interceptor's own rejected requests take.
+- `ComposerAttachmentExtensionRegistry`'s Workplace wiring passes an `oidcRefreshTrigger` callback into `WorkplaceComposerAttachmentExtension`, alongside the existing `oidcTokenGetter`, mirroring the same callback-injection pattern across the package boundary.
+- `WorkplaceComposerAttachmentExtension._exchangeAccessToken` catches a 401 from the token-exchange call and retries exactly once with a token refreshed via `oidcRefreshTrigger`, guarded by a `refreshAttempted` flag to prevent looping.
+
+## Consequences
+
+- A stale OIDC id token on the Workplace exchange-token request self-heals via one retry instead of surfacing as a generic toast.
```
dab246, line 25: Needs to be updated accurately according to the new logic.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -346,35 +356,44 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
-        return super.onError(err, handler);
-      }
-      _updateNewToken(newTokenOidc);
-
-      final personalAccount = await _updateCurrentAccount(tokenOIDC: newTokenOidc);
-
-      if (PlatformInfo.isIOS) {
-        await _iosSharingManager.saveKeyChainSharingSession(personalAccount);
+      await requestTokenRefresh();
+      if (_token?.token == tokenBeforeRefresh) {
```
dab246: `requestTokenRefresh()` returns a `TokenOIDC`, but the code ignores the return value and reads the mutable `_token` instead.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -568,13 +582,26 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
+
     // Persist the new token BEFORE mutating the account cache. persistOneTokenOidc
     // is crash-safe (write-before-prune), so the token box always holds a usable
     // token even if the process is killed mid-update.
     await _tokenOidcCacheManager.persistOneTokenOidc(tokenOIDC);
+    if (generation != _sessionGeneration) {
+      // A clear() landed mid-write; undo it so the token box doesn't say
+      // "authenticated" for a session that was just wiped.
+      await _tokenOidcCacheManager.deleteTokenOidc(tokenOIDC.tokenIdHash);
```
dab246: This can throw an exception, so should wrap it in a try-catch block.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -585,6 +612,11 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
       userName: currentAccount.userName
     );
     await _accountCacheManager.setCurrentAccount(personalAccount);
+    if (generation != _sessionGeneration) {
+      // Same as above, but for the account write.
+      await _accountCacheManager.deleteCurrentAccount(personalAccount.id);
```
dab246: idem

lib/features/base/base_controller.dart
tddang-linagora: Done

lib/features/base/base_controller.dart
tddang-linagora: Done

docs/adr/0107-workplace-oidc-refresh-via-shared-interceptor.md
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
```diff
@@ -88,27 +96,85 @@ class WorkplaceComposerAttachmentExtension implements ComposerAttachmentPlugin {
...
+      'WorkplaceComposerAttachmentExtension::_triggerRefreshOIDCToken: '
+      'failedIdTokenHash=${failedToken.hashCode} | '
+      'refreshedIdTokenHash=${refreshedToken?.hashCode}',
+      webConsoleEnabled: true,
+    );
+    // The IdP may omit id_token on refresh, in which case the current one is
+    // kept — retrying would re-send the token that just 401'd.
+    if (refreshedToken == null || refreshedToken == failedToken) throw failure;
```
dab246, line 171:
```suggestion
if (refreshedToken == null || refreshedToken == failedToken) {
  logError(
    'WorkplaceComposerAttachmentExtension::_triggerRefreshOIDCToken: '
    'auth_error_type=drive_exchange_id_token_not_refreshed | will_logout=false | '
    'refreshReturnedToken=${refreshedToken != null}',
    exception: failure,
    webConsoleEnabled: true,
  );
  throw failure;
}
```

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -91,6 +95,54 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
+  /// [StaleSessionRefreshException] when the session that started it was
+  /// replaced meanwhile, rethrows transient failures untouched.
+  Future<TokenOIDC> requestTokenRefresh() {
+    final inFlight = _refreshInFlight;
+    if (inFlight != null) return inFlight;
+
+    // Nothing left to refresh with — the session already died elsewhere.
+    if (_configOIDC == null || _token == null) {
+      return Future.error(RefreshTokenFailedException());
+    }
```
dab246, line 121:
```suggestion
if (_configOIDC == null || _token == null) {
  logError(
    'AuthorizationInterceptors::requestTokenRefresh: '
    'auth_error_type=refresh_on_cleared_session | will_logout=true | '
    'hasConfig=${_configOIDC != null} | hasToken=${_token != null}',
    webConsoleEnabled: true,
  );
  return Future.error(RefreshTokenFailedException());
}
```

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: This path is already logged twice — `logWarning` directly above (line ~158) carries both token hashes, and the rethrown `failure` lands in `DriveIntentMessageHandlerMixin._failWith` (drive_intent_message_handler_mixin.dart:65), the single Sentry funnel this PR set up. A third log here duplicates that funnel rather than adding signal.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
tddang-linagora: Done

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
dab246: IMO, `logWarning` is unnecessary and adds no value here; we should drop it and use `logError` instead, as that is more accurate.

lib/features/login/data/network/interceptors/authorization_interceptors.dart
```diff
@@ -568,13 +590,30 @@ class AuthorizationInterceptors extends QueuedInterceptorsWrapper {
...
 
+    // Re-checked before every write below — a clear() landing during any
+    // of these awaits must not resurrect a session it just wiped.
+    if (generation != _sessionGeneration) throw const StaleSessionRefreshException();
+
     // Persist the new token BEFORE mutating the account cache. persistOneTokenOidc
     // is crash-safe (write-before-prune), so the token box always holds a usable
     // token even if the process is killed mid-update.
     await _tokenOidcCacheManager.persistOneTokenOidc(tokenOIDC);
+    if (generation != _sessionGeneration) {
```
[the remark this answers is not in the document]
tddang-linagora: Race is real but needs old refresh in flight + full logout + re-login + persist inside one network RTT; worst case is a forced re-login, no data loss. Fix as described spans the interceptor, `TokenOidcCacheManager`, `AuthenticationOIDCDataSourceImpl:88/95` and `IOSSharingManager` — a lock inside the interceptor alone would not close it. Cost outweighs benefit. On the keychain: no rollback after `:716` is deliberate — that entry is keyed by `accountId`, not `tokenIdHash`, so rolling back would delete the *new* session's entry.

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
dab246: Oke, so update logWarning again:

```dart
// Trace level so this rides into Sentry as a breadcrumb on the _failWith
  // event; warning reaches the console only, and nothing in mobile release.
logTrace(
  'WorkplaceComposerAttachmentExtension::_triggerRefreshOIDCToken',
  extras: {
    'refreshReturnedToken': refreshedToken != null,
    'refreshReturnedSameToken': refreshedToken == failedToken,
  },
  webConsoleEnabled: true,
);
```

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

workplace/lib/presentation/extension/workplace_composer_attachment_extension.dart
tddang-linagora: Done

dab246 (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4810

hoangdat (approved):

PR #4818, opened 2026-09-08
Hide the premium CTA when no paywall url is configured

### LINKED ISSUE #4817: Premium CTA is shown when no paywall url is configured
### Description

The "Increase your space" button is displayed as soon as the account has the SaaS premium capability, without checking that the deployment actually has a paywall url. On a deployment where none is configured, clicking it only shows the error toast `Paywall url not available`.

### Expected result

* No premium CTA when the deployment has no paywall url, meaning no workplace FQDN and no `paywallUrlTemplate` in the Linagora ecosystem well-known.

### Current behavior

* The CTA is displayed in the sidebar, and clicking it ends on the `Paywall url not available` toast with a Retry action that fails again.
* The quota banner and the over quota dialogs in the composer offer the same CTA and behave the same way, they share the premium check.

#### Reproduction Steps

1. Log in with an account that has the SaaS premium capability, on a deployment with no workplace FQDN and no `paywallUrlTemplate`.
2. Look at the bottom of the left sidebar.
3. Click "Increase your space".

### Acceptance criteria

* The premium CTA is hidden when no paywall url can be built for the current deployment.
* The CTA still shows and still opens the paywall when a workplace FQDN or a `paywallUrlTemplate` is available.

### Context

Web, version 0.35.1

### Additional information

Paywall url not available

MontaGhanmy:
Closes #4817

## Description

The premium CTA was gated on the SaaS premium capability only. On a deployment with no workplace FQDN and no `paywallUrlTemplate` in the Linagora ecosystem well-known, the button was still displayed and clicking it could only end on the `Paywall url not available` toast.

## Changes

- `PaywallController` now tells whether a paywall url can be built, from the same two sources it already uses to navigate.
- The premium check behind the sidebar button, the quota banner and the composer over quota dialogs relies on it, so all of them hide together when the deployment has no paywall.
- The cached ecosystem is now observable, so the sidebar reacts when the well-known answer arrives after the first build.
- Unit tests on the reachability rule.

The CTA stays hidden until a paywall url is known, so it also stays hidden when the well-known call fails, since there is nothing to open in that case.

The upgrade button of the settings storage page is not covered here, its controller has no ecosystem to check.

Note: no Flutter toolchain on the machine where this was written, so the tests were not run locally, CI will.

## Summary by CodeRabbit

* **New Features**
  * Added provider-based premium upgrade CTAs across storage management, quota banners, mailbox sidebar, and composer flows.
  * Added ecosystem-aware premium availability and paywall destination resolution.
  * Added secure paywall URL validation and launch handling.
  * Added ecosystem state synchronization for dashboard integrations.

* **Bug Fixes**
  * Prevented invalid or unsafe paywall links from opening.
  * Improved handling of blank or malformed ecosystem paywall templates.
  * Prevented scrolling actions after related widgets are unmounted.

* **Tests**
  * Expanded coverage for premium CTA visibility, navigation, ecosystem states, URL validation, and lifecycle behavior.


Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4818

chibenwa:
Seen with @hoangdat @poupotte @Crash-- 

Exactly the kind of good common sense that is needed. Thanks!

hoangdat:
As I understand, we added the `premium` capability in linagora with the `canUpgrade = true`, then the CTA button displayed? 
So now we should rely on the paywallURL to display it or not?

Crash--:
> As I understand, we added the premium capability in linagora with the canUpgrade = true, then the CTA button displayed?

Kind of, yes...

In fact, we enabled `twakeIntegration` on the backend to get access to the Common Settings.

By default, this integration enables Common Settings, SaaS capabilities, and PGP.

So we ended up with `isPaying: false` and `canUpgrade: true` for LNG users, because that's the default behavior on the backend when you have SaaS capabilities enabled without any specific configuration.

So I see two things here:

- Relying on a well-known property to enable or disable this feature on the frontend could prevent this kind of side effect. The deployment would be a little more complex, but at least we'd avoid this kind of issue.
- We should disable the SaaS capability on the backend for LNG. And maybe we should rework `twakeIntegration` to avoid similar issues in the future.

MontaGhanmy:
I guess this PR raised enough attention at this point. Please feel free to review it/add on top of it and/or merge it guys. Thank you 🙏

chibenwa:
IMO it don't hurt: it only hides brken CTA buttons...

I :+1: massively

tddang-linagora (dismissed):
Confirmed running locally with saas config on, userinfo called and no CTA button

Image

lib/features/mailbox_dashboard/domain/linagora_ecosystem/linagora_ecosystem.dart
```diff
@@ -48,6 +48,9 @@ extension LinagoraEcosystemExtension on LinagoraEcosystem {
     return listAppLinagoraEcosystem.where((app) => app.isAppAndroidEnabled).toList();
   }
 
+  String? get paywallUrlTemplate =>
+      (properties?[LinagoraEcosystemIdentifier.paywallURL] as ApiUrlLinagoraEcosystem?)?.value;
```
dab246: Suggested change:

```dart
String? get paywallUrlTemplate {
  final property =
      properties?[LinagoraEcosystemIdentifier.paywallURL];

  return property is ApiUrlLinagoraEcosystem
      ? property.value
      : null;
}
```

dab246: Add defesnsive by unit test:

```dart
group('LinagoraEcosystem.paywallUrlTemplate', () {
    test('returns configured template', () {
      final ecosystem = LinagoraEcosystem.deserialize({
        'paywallUrlTemplate':
            'https://domain.tld/paywall?email={localPart}',
      });

      expect(
        ecosystem.paywallUrlTemplate,
        'https://domain.tld/paywall?email={localPart}',
      );
    });

    test('returns null when property is missing', () {
      final ecosystem = LinagoraEcosystem.deserialize({
        'scribePromptUrl': 'https://domain.tld/scribe',
      });

      expect(ecosystem.paywallUrlTemplate, isNull);
    });

    test('returns null instead of throwing when property has wrong type', () {
      final ecosystem = LinagoraEcosystem.deserialize({
        'paywallUrlTemplate': {'url': 'invalid'},
      });

      expect(() => ecosystem.paywallUrlTemplate, returnsNormally);
      expect(ecosystem.paywallUrlTemplate, isNull);
    });
  });
```

hoangdat:
thank @MontaGhanmy for your contribution, team will take over this PR to gate the CTA in some place in storage in settings page and some further minor fix.

lib/features/paywall/presentation/paywall_controller.dart
```diff
@@ -16,8 +18,24 @@ import 'package:tmail_ui_user/main/utils/app_utils.dart';
...
   final String ownEmailAddress;
+  final Rxn<LinagoraEcosystem>? linagoraEcosystem;
 
-  PaywallController({required this.ownEmailAddress});
+  PaywallController({required this.ownEmailAddress, this.linagoraEcosystem});
+
+  static bool isReachable({String? workplaceFqdn, String? paywallUrlTemplate}) {
+    return workplaceFqdn?.trim().isNotEmpty == true ||
+        paywallUrlTemplate?.trim().isNotEmpty == true;
+  }
```
dab246: The current logic considers any non-empty FQDN as reachable:

```dart
workplaceFqdn?.trim().isNotEmpty == true
```

For example, `localhost` returns `true`, but `WebLinkGenerator.safeGenerateWebLink()` cannot generate a valid URL because the FQDN is invalid. As a result, the CTA is displayed, but clicking it falls back to the ecosystem URL or results in an error.

Consider using a shared URL builder for both the reachability predicate and navigation to avoid having two different definitions of "reachable":

```dart
import 'package:core/utils/web_link_generator.dart';

class PaywallUtils {
  const PaywallUtils._();

  static String buildWorkplacePaywallUrl(String? workplaceFqdn) {
    final normalizedFqdn = workplaceFqdn?.trim();
    if (normalizedFqdn == null || normalizedFqdn.isEmpty) return '';

    return WebLinkGenerator.safeGenerateWebLink(
      workplaceFqdn: normalizedFqdn,
      pathname: '/settings/premium',
    );
  }

  static bool isReachable({
    String? workplaceFqdn,
    String? paywallUrlTemplate,
  }) {
    return buildWorkplacePaywallUrl(workplaceFqdn).isNotEmpty ||
        paywallUrlTemplate?.trim().isNotEmpty == true;
  }
}
```

dab246: And add unit test to defensive:

```dart
group('PaywallUtils.isReachable', () {
    test('returns false when both sources are absent', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: null,
          paywallUrlTemplate: null,
        ),
        isFalse,
      );
    });

    test('returns false when both sources are blank', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: '   ',
          paywallUrlTemplate: '',
        ),
        isFalse,
      );
    });

    test('returns true for a valid Workplace FQDN', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: 'workplace.domain.tld',
          paywallUrlTemplate: null,
        ),
        isTrue,
      );
    });

    test('returns false for non-empty but invalid Workplace FQDN', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: 'localhost',
          paywallUrlTemplate: null,
        ),
        isFalse,
      );
    });

    test('returns true when ecosystem contains a paywall template', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: null,
          paywallUrlTemplate:
              'https://domain.tld/paywall?email={localPart}',
        ),
        isTrue,
      );
    });

    test('falls back to ecosystem when Workplace FQDN is invalid', () {
      expect(
        PaywallUtils.isReachable(
          workplaceFqdn: 'localhost',
          paywallUrlTemplate:
              'https://domain.tld/paywall?email={localPart}',
        ),
        isTrue,
      );
    });
  });
```

lib/features/paywall/presentation/paywall_controller.dart
```diff
@@ -16,8 +18,24 @@ import 'package:tmail_ui_user/main/utils/app_utils.dart';
 
 class PaywallController extends BaseController {
   final String ownEmailAddress;
+  final Rxn<LinagoraEcosystem>? linagoraEcosystem;
 
-  PaywallController({required this.ownEmailAddress});
+  PaywallController({required this.ownEmailAddress, this.linagoraEcosystem});
```
dab246: There are several design concerns with the current implementation:

* `PaywallController` depends on GetX state management solely to determine the CTA.
* The paywall feature has a reverse dependency on the mailbox dashboard domain model.
* The controller receives the entire ecosystem object while it only needs to know whether a paywall URL is available.
* The double nullable state (`linagoraEcosystem == null` and `linagoraEcosystem.value == null`) introduces two different semantics that are difficult to maintain.
* `ManageAccountDashBoardController` initializes the controller without an ecosystem, causing `isAvailable` to default to `true`, while the mailbox dashboard has different behavior.

Consider simplifying the dependency and passing only the information required to determine the CTA availability, rather than coupling `PaywallController` to GetX and the mailbox dashboard domain model.

lib/features/paywall/presentation/paywall_controller.dart
```diff
@@ -16,8 +18,24 @@ import 'package:tmail_ui_user/main/utils/app_utils.dart';
...
+
+  /// `true` when no ecosystem is available to tell, so the caller keeps its CTA.
+  bool get isAvailable {
+    if (linagoraEcosystem == null) return true;
+
+    return isReachable(
+      workplaceFqdn: twakeAppManager.oidcUserInfo?.workplaceFqdn,
+      paywallUrlTemplate: linagoraEcosystem!.value?.paywallUrlTemplate,
+    );
+  }
```
dab246: Currently, `Obx` watches:

```dart
linagoraEcosystem.value?.paywallUrlTemplate
```

but reads the FQDN from a value that is not reactive to `Obx`:

```dart
twakeAppManager.oidcUserInfo?.workplaceFqdn
```

This can lead to the following failure scenario:

1. The ecosystem and quota finish loading first.
2. The sidebar rebuilds while the Workplace FQDN is still `null`, so the CTA is hidden.
3. The OIDC user info finishes loading later and updates `TwakeAppManager`.
4. `Obx` does not receive any notification from `TwakeAppManager`.
5. The CTA remains hidden even though a valid Workplace paywall URL is now available.

Consider using the existing source of truth and only watching it on web:

```dart
Consumer(
  builder: (context, ref, child) {
    return MailboxSidebarFooter(
      isDesktop: isDesktop,
      showIncreaseSpaceButton: true,
      workplaceFqdn: ref.watch(workplaceFqdnProvider),
    );
  },
)
```

Then add a plain input to the footer:

```dart
final String? workplaceFqdn;

const MailboxSidebarFooter({
  super.key,
  this.isDesktop = false,
  this.showIncreaseSpaceButton = false,
  this.workplaceFqdn,
});
```

Mobile can continue using the default `showIncreaseSpaceButton: false` without watching `workplaceFqdnProvider`.

dab246 (commented):
Thank you for contributing such a great PR to the project. I’ve reviewed the branch and noticed a few issues regarding architecture and safeguards against regressions. Please take a look. Thanks.

lib/features/manage_account/presentation/storage/storage_controller.dart
```diff
@@ -1,10 +1,70 @@
...
+    if (!PlatformInfo.isWeb) return;
+
+    _loadEcosystemPaywallUrl();
+  }
+
+  void _loadEcosystemPaywallUrl() {
+    if (!validatePremiumIsAvailable()) return;
+
+    final jmapUrl = dynamicUrlInterceptors.jmapUrl;
+    if (jmapUrl == null) return;
```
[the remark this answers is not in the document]
dab246: Is this comment done?

lib/features/quotas/presentation/widget/quotas_banner_widget.dart
```diff
@@ -2,105 +2,146 @@ import 'package:core/presentation/views/button/default_close_button_widget.dart'
...
+      ),
+    );
+  }
+
+  Widget _buildSubtitle(BuildContext context, String? workplaceFqdn) {
+    final appLocalizations = AppLocalizations.of(context);
+
+    if (!PlatformInfo.isWeb) {
+      return _buildSubtitleWithoutPremium(appLocalizations);
+    }
```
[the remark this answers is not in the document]
dab246: Is this comment done?

test/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller_test.dart
```diff
@@ -430,6 +478,260 @@ void main() {
...
+      );
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(urlLauncherChannel, (call) async {
+        if (call.method == 'launch') {
+          final arguments = call.arguments as Map<dynamic, dynamic>;
+          launchedUrls.add(arguments['url'] as String);
+          return true;
+        }
+        return null;
+      });
```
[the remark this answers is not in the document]
dab246: Is this comment done?

lib/features/manage_account/presentation/storage/storage_controller.dart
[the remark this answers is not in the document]
dab246: Retry does not apply here; `jmapUrl` is initialized before `StorageController` is created in both the reload flow and the normal navigation flow.

lib/features/quotas/presentation/widget/quotas_banner_widget.dart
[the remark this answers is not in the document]
dab246: Keep the quota CTA web-only to preserve the current behavior. A regression test has been added for non-web platforms.

test/features/mailbox_dashboard/presentation/controller/mailbox_dashboard_controller_test.dart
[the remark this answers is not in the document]
dab246: Keep the `launch` method; dependency `2.3.2` still uses the MethodChannel `launch` internally. A positive control has also been added to ensure the unsafe URL test is no longer vacuous.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4818

tddang-linagora (dismissed):

lib/features/paywall/presentation/providers/premium_cta_provider.dart
hoangdat, line 1: this file is too big, we should split it. Separate the ecosystem and the payment?

lib/features/paywall/presentation/extensions/premium_cta_ref_extension.dart
```diff
@@ -0,0 +1,42 @@
...
+      read(premiumCtaProvider(context)) is PremiumCtaAvailable;
+
+  void openPremiumCta(PremiumCtaContext? context) => _openPremiumCta(
+        read(premiumCtaProvider(context)),
+        read(paywallLauncherProvider),
+      );
+}
+
+void _openPremiumCta(PremiumCtaState state, PaywallLauncher launcher) {
+  if (state is! PremiumCtaAvailable) return;
```
hoangdat, line 40: should show toast to user?

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -1172,11 +1174,10 @@ class ComposerController extends BaseController
       exception: failure.exception,
     );
 
-    final isIncreaseMySpaceIsDisabled =
-        !mailboxDashBoardController.validatePremiumIsAvailable() ||
-            mailboxDashBoardController.validateUserHasIsAlreadyHighestSubscription();
-
-    final needIncreaseMySpace = !isIncreaseMySpaceIsDisabled &&
+    final providerContainer = ProviderScope.containerOf(context, listen: false);
```
hoangdat, line 1177: I found we use `containerOf` instead of `appProviderContainer` directly, should we document it in ADR?

lib/features/paywall/presentation/paywall_launcher.dart
```diff
@@ -0,0 +1,29 @@
...
+/// resolves the same instance from [paywallLauncherProvider], and tests
+/// substitute it by overriding that provider.
+class PaywallLauncher {
+  const PaywallLauncher();
+
+  void launch(Uri destination) {
+    final paywallUrl = destination.toString();
+    // The CTA state only exposes validated destinations; re-check here because
+    // this is the boundary that hands a URL to the platform launcher.
+    if (!PaywallUtils.isValidPaywallUrl(paywallUrl)) {
```
hoangdat, line 20: toast this case also?

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4818

lib/features/composer/presentation/composer_controller.dart
```diff
@@ -2510,17 +2505,19 @@ class ComposerController extends BaseController
...
+          onConfirmAction();
+        }
+        popBack();
 
-          if (needIncreaseMySpace) {
-            mailboxDashBoardController.paywallController?.navigateToPaywall();
-          } else {
-            _autoFocusFieldWhenLauncher();
-          }
+        if (needIncreaseMySpace) {
```
tddang-linagora, line 2515: `needIncreaseMySpace` is created from `providerContainer`, which looks at the `context` parameter of this method. However, this `if` check is after a `popBack` call. Should we add `mounted` check here?

lib/features/paywall/presentation/providers/premium_cta_provider.dart
dab246: Updated. Ecosystem state/loading was extracted into `active_ecosystem_provider.dart` `premium_cta_provider.dart` now focuses on CTA resolution.

lib/features/paywall/presentation/extensions/premium_cta_ref_extension.dart
dab246: I don’t think a toast is needed here. This is a defensive guard for a stale/unavailable CTA state; normally the CTA is already hidden or disabled.

lib/features/composer/presentation/composer_controller.dart
dab246: Updated. The usage rule is now documented in ADR-0092: use the nearest ProviderScope when BuildContext is available and avoid new appProviderContainer dependencies.

lib/features/paywall/presentation/paywall_launcher.dart
dab246: I don’t think a toast is appropriate here. This is a defense-in-depth check for an internal/configuration issue; invalid destinations are already rejected earlier, and this fallback logs a warning for diagnostics.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4818

lib/features/composer/presentation/composer_controller.dart
dab246: A mounted check is not needed here. The container is resolved before opening the dialog, and the callback does not access BuildContext. Also, popBack() only closes the confirmation dialog, not the composer route.

hoangdat (approved):

tddang-linagora (approved):

PR #4841, opened 2026-09-17
TF-4838 Resolve Workplace FQDN from the Linagora ecosystem fallback

### LINKED ISSUE #4838: Enable Drive integration for Linagora
### Desc

Right now, `/userInfo` only called via `saas` flag.
We should gate `/userInfo` via other flag

### DoD
- [x] fetch `userInfo` in when it need

tddang-linagora:
Adds a fallback source for the Workplace FQDN so Drive attachment and the paywall CTA still resolve on deployments where OIDC `/userInfo` has no `workplaceFqdn` claim (or is never called, e.g. non-SaaS).

- `.well-known/linagora-ecosystem` gains an optional `workplaceFqdnFallback` template key, e.g. `{localPart}.twake.linagora.com`.
- `workplaceFqdnProvider` now tracks two sources (OIDC `/userInfo` and the ecosystem fallback) and emits `userInfo ?? ecosystem` — userInfo always wins when present, load order doesn't matter.
- New `WorkplaceFqdnEcosystemHandler` resolves the template against the JMAP session username and feeds it into the provider via the existing ecosystem handler registry. It takes `LinagoraEcosystemHandlerRegistry`'s own provider-scoped `Ref` (never `appProviderContainer`, which is frozen per ADR-0092) — not a `WidgetRef`, since the registry dispatches `onEcosystemCleared` from another widget's `dispose()`, where a captured `WidgetRef` is already unsafe to use.
- Only camelCase placeholders (`{localPart}`, `{domainName}`) are supported, matching every other ecosystem/paywall template in the codebase.

No change to `/userInfo` gating (`_shouldCallUserInfo` still SaaS-only) or `DioClient`.

Closes #4838

## Summary by CodeRabbit

- **New Features**
  - Added ecosystem-based Workplace FQDN fallback support using the account email.
  - Workplace FQDN values now resolve by priority: user information first, ecosystem fallback second.
  - Added support for `{localPart}` and `{domainName}` placeholders, including URL-encoded forms.
  - Invalid, blank, or unresolved values are ignored; valid HTTPS addresses are used for paywall and Drive features.

- **Documentation**
  - Added configuration guidance and technical documentation for Workplace FQDN source priority.


core/lib/data/network/dio_client.dart
```diff
@@ -54,6 +55,31 @@ class DioClient {
         .catchError((error) => throw error);
   }
 
+  /// Authenticated GET that never carries the JMAP accept header.
+  ///
+  /// Goes through the shared [_dio], so [AuthorizationInterceptors] still attaches the bearer token
+  /// and refreshes it on 401. Used for non-JMAP endpoints such as the OIDC `/userInfo` endpoint.
+  Future<dynamic> getWithoutJmapHeader(
```
hoangdat: how about the same with `post`, use the flag `bool useJMAPHeader = true,` in signature ?

test/features/login/data/network/oidc_http_client_test.dart
```diff
@@ -64,5 +64,21 @@ void main() {
...
+      // arrange
+      when(dioClient.getWithoutJmapHeader(any))
+          .thenAnswer((_) async => {'sub': 'user-sub', 'email': 'user@domain.tld'});
+
+      // act
+      final userInfo = await oidcHttpClient.fetchUserInfo('https://sso.domain.tld/userinfo');
+
+      // assert
+      expect(userInfo.email, 'user@domain.tld');
+      verify(dioClient.getWithoutJmapHeader('https://sso.domain.tld/userinfo')).called(1);
```
hoangdat: can we verify the header also?

chibenwa:
Why is this needed cf @Crash--

Crash--:
I don't know why we changed from saas platform to user_info. And I don't understand the changes because for me we already have this check to make the call to userinfo (for exemple for common settings). 

Have you done this change because for LNG we don't have the `workplaceFQDN` attribute? If yes, then why we don't fallback on the string literal as we do for other deployment?

test/features/login/data/network/oidc_http_client_test.dart
dab246: The `Accept` header is the behaviour this PR changes, and nothing covers it. Please add a unit test for `DioClient.jmapHeader`.

hoangdat:
hi @Crash-- ,

1. Linagora is `saas` platform or not?
2. Gate the `user-info` under `saas` is not a good idea -> so we change it
3. right now`workplaceFQDN` in userinfo is our single source of truth, so we will have the fallback value for this in linagora-ecosystem.

core/lib/data/network/dio_client.dart
tddang-linagora: done

test/features/login/data/network/oidc_http_client_test.dart
tddang-linagora: done

chibenwa:
Yes but I do not see why we need an extra config option that would make it harder to activate the feature.

hoangdat:
ok, so we should remove the change for new config, but add new fallback configuration for workplace FQDN

lib/main/providers/workplace/workplace_fqdn_notifier.dart
```diff
@@ -5,16 +5,30 @@ part 'workplace_fqdn_notifier.g.dart';
 
 @Riverpod(keepAlive: true)
 class WorkplaceFqdnNotifier extends _$WorkplaceFqdnNotifier {
+  String? _fromUserInfo;
+  String? _fromEcosystem;
```
Crash--: as said this morning to @hoangdat this idea if not really does it came frame userinfo or from ecosystem. 

It's more: 

can tmail calculate itself by on the username ? If yes, then well-know & template literal 
if not, then need to perform an API call. Today the api is the sso/well-known, but tomorrow it can be something else (we're thinking about it - it's connected to how twake will integrate / consume external SSO / LDAP).

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4841

lib/features/mailbox_dashboard/presentation/linagora_ecosystem/workplace_fqdn_ecosystem_handler.dart
```diff
@@ -0,0 +1,50 @@
...
+  String? _resolveFallbackFqdn(LinagoraEcosystem ecosystem) {
+    final template = ecosystem.workplaceFqdnFallbackTemplate;
+    if (template == null) return null;
+
+    final ownerEmail = _resolveOwnerEmail()?.trim();
+    if (ownerEmail == null || ownerEmail.isEmpty) return null;
+
+    final resolved =
+        PaywallUrlPattern(template).getQualifiedUrl(ownerEmail: ownerEmail).trim();
+    return resolved.isEmpty ? null : resolved;
```
dab246: A username with no `@` leaves `{localPart}` unresolved, so the template becomes `.twake.linagora.com`, which both `normalizeWorkplaceFqdn` and `isValidFqdn` accept. The configuration doc says such a template is ignored. Please reject a resolution that left a placeholder unfilled.

test/features/mailbox_dashboard/presentation/linagora_ecosystem/workplace_fqdn_ecosystem_handler_test.dart
```diff
@@ -0,0 +1,119 @@
...
+    description: 'leaves the provider null when the owner email is null',
+    ownerEmail: null,
+    expected: null,
+  ),
+  (
+    description: 'leaves the provider null when the owner email is blank',
+    ownerEmail: '   ',
+    expected: null,
+  ),
+];
```
dab246, line 58: These cases cover a parseable address, a null one and a blank one, but never a username that is not an address. Please add a case for a username with no `@`.

lib/features/mailbox_dashboard/presentation/delegates/ecosystem_provider_listener_delegate.dart
```diff
@@ -42,6 +43,11 @@ class EcosystemProviderListenerDelegate
     if (registry.hasHandlers) return;
     registry
       ..register(DriveAttachmentEcosystemHandler())
+      ..register(WorkplaceFqdnEcosystemHandler(
+        ref: registry.ref,
+        resolveOwnerEmail: () =>
+            getBinding<MailboxDashBoardController>()?.sessionCurrent?.username.value,
+      ))
```
dab246, line 50: Nothing asserts this handler is registered: its own test builds it directly, so deleting this line keeps every test green. Please add a delegate test that loads an ecosystem with `workplaceFqdnFallback` and checks `workplaceFqdnProvider`.

lib/features/mailbox_dashboard/presentation/linagora_ecosystem/workplace_fqdn_ecosystem_handler.dart
tddang-linagora: Done

test/features/mailbox_dashboard/presentation/linagora_ecosystem/workplace_fqdn_ecosystem_handler_test.dart
tddang-linagora: Done

lib/features/mailbox_dashboard/presentation/delegates/ecosystem_provider_listener_delegate.dart
tddang-linagora: Done

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
```diff
@@ -0,0 +1,31 @@
...
+
+/// A bare host, no path/query/fragment, https outside debug builds.
+bool _isUsableWorkplaceUri(Uri uri) {
+  if (uri.host.isEmpty) return false;
+  if (!_isBareUri(uri)) return false;
+  return uri.scheme == 'https' || kDebugMode;
+}
+
+bool _isBareUri(Uri uri) =>
+    uri.path.isEmpty && !uri.hasQuery && !uri.hasFragment;
```
dab246, line 31: `https://workplace.example.com/` is now rejected because its path is `/`, so a trailing slash silently hides the Drive button and the paywall CTA. The `startsWith('http')` test above also misreads a bare host like `httpadmin.twake.linagora.com` as already-schemed. Please normalize the value before validating it.

lib/features/paywall/domain/model/paywall_url_pattern.dart
```diff
@@ -18,6 +18,16 @@ class PaywallUrlPattern with EquatableMixin {
     );
   }
 
+  /// Resolves [pattern], or null when a placeholder cannot be filled from
+  /// [ownerEmail] — a half-filled URL points at the wrong host.
+  String? resolveQualifiedUrl({required String ownerEmail}) {
```
dab246, line 23: `premium_cta_provider.dart:188` still calls `getQualifiedUrl`, so the paywall keeps the half-filled template this was added to prevent. `isValidPaywallUrl` does not catch it: `isValidFqdn` drops empty labels before counting, so `.twake.linagora.com` passes. Please use `resolveQualifiedUrl` there too.

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4841

lib/features/mailbox_dashboard/presentation/linagora_ecosystem/linagora_ecosystem_handler_registry.dart
```diff
@@ -6,9 +6,23 @@ part 'linagora_ecosystem_handler_registry.g.dart';
...
+  /// [WidgetRef] captured from a widget that may later unmount. Set once
+  /// when the registry provider builds.
+  Ref get ref {
+    final attachedRef = _ref;
+    assert(attachedRef != null,
+        'LinagoraEcosystemHandlerRegistry.ref read before attachRef()');
+    return attachedRef!;
+  }
+
+  void attachRef(Ref ref) => _ref ??= ref;
```
hoangdat: should remove the it

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
tddang-linagora: Done

lib/features/paywall/domain/model/paywall_url_pattern.dart
tddang-linagora: Done

lib/features/mailbox_dashboard/presentation/linagora_ecosystem/linagora_ecosystem_handler_registry.dart
tddang-linagora: Done

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
```diff
@@ -0,0 +1,36 @@
...
+  if (uri == null || !_isUsableWorkplaceUri(uri)) return null;
+  // A trailing slash is the same host, so keep the value canonical.
+  return fqdn.endsWith('/') ? fqdn.substring(0, fqdn.length - 1) : fqdn;
+}
+
+/// A bare host, no path/query/fragment, https outside debug builds.
+bool _isUsableWorkplaceUri(Uri uri) {
+  if (uri.host.isEmpty) return false;
+  if (!_isBareUri(uri)) return false;
+  return uri.scheme == 'https' || kDebugMode;
```
hoangdat: `kDebugMode` is true under `flutter test`, so this rule is never exercised: replacing it with `return true;` keeps all 24 tests green. Needs a testable flag.

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
```diff
@@ -0,0 +1,36 @@
...
+/// the resolver keeps tracking that source as a dependency.
+typedef WorkplaceFqdnSourceReader = String? Function(Ref ref);
+
+/// One source of the Workplace FQDN. Every source exposes the same API so
+/// [workplaceFqdnProvider] can rank them purely by list position.
+abstract interface class WorkplaceFqdnSource {
+  void setFqdn(String? rawFqdn);
+}
+
+final _schemeRegExp = RegExp(r'^https?://', caseSensitive: false);
```
hoangdat, line 15: No test covers `caseSensitive: false` — flipping it back to case-sensitive keeps all 24 green. Add one: `HTTPS://host.com` must not be null.

hoangdat (commented):
Two test-coverage gaps on the FQDN scheme rules, both found by mutation testing. Non-blocking: the rules behave correctly, they just aren't guarded. Details inline.

lib/features/mailbox_dashboard/domain/linagora_ecosystem/linagora_ecosystem.dart
```diff
@@ -67,4 +67,12 @@ extension LinagoraEcosystemExtension on LinagoraEcosystem {
...
     return value is DriveAttachmentLinagoraEcosystem ? value : null;
   }
+
+  String? get workplaceFqdnFallbackTemplate {
+    final property = properties?[LinagoraEcosystemIdentifier.workplaceFqdnFallback];
+    if (property is! ApiUrlLinagoraEcosystem) return null;
+
+    final template = property.value.trim();
+    return template.isEmpty ? null : template;
+  }
```
hoangdat: it quite duplicated with `paywallUrlTemplate`? how about to reduce it?

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
dab246, line 15: `caseSensitive: false` lets `HTTPS://workplace.example.com` through, and `_computeUri` in `drive_attachment_uri_value_notifier_provider.dart` tests the scheme case-sensitively, so it builds `https://https//workplace.example.com` and Drive points at host `https`. Please share one scheme test with `_computeUri`.

lib/features/paywall/domain/model/paywall_url_pattern.dart
```diff
@@ -10,14 +10,40 @@ class PaywallUrlPattern with EquatableMixin {
   PaywallUrlPattern(this.pattern);
 
   String getQualifiedUrl({required String ownerEmail, String? domainName}) {
```
dab246: `getQualifiedUrl` is public and still returns the half-filled URL, and `resolveQualifiedUrl` is now its only caller. Please make it private.

lib/features/paywall/domain/model/paywall_url_pattern.dart
```diff
@@ -10,14 +10,40 @@ class PaywallUrlPattern with EquatableMixin {
...
-    final mailAddress = _getMailAddress(ownerEmail: ownerEmail);
+    final values = _resolveValues(ownerEmail: ownerEmail, domainName: domainName);
```
hoangdat, line 13: I concern the paywall URL sometime is the static URL (no local part). For example:
-  https://host.com/path
- host.com?q=1
- https://host.com#frag

If it, we will disable the CTA button, it quite dangerous

tddang-linagora: `resolveQualifiedUrl` returns null only when a placeholder is *present* and unfilled
(`paywall_url_pattern.dart:43`), so a static template passes through untouched.
`https://host.com/path` and `https://host.com#frag` both pass `isValidPaywallUrl` — it checks
scheme + host only, not path (`paywall_utils.dart:37`). `host.com?q=1` is rejected for having
no scheme, which is TF-4817 behaviour on master, not this PR.

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
tddang-linagora: Done

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
tddang-linagora: Done

lib/features/mailbox_dashboard/domain/linagora_ecosystem/linagora_ecosystem.dart
tddang-linagora: Done

lib/main/providers/workplace/fqdn/workplace_fqdn_source.dart
tddang-linagora: Done

lib/features/paywall/domain/model/paywall_url_pattern.dart
tddang-linagora: Done

lib/main/providers/workplace/fqdn/workplace_fqdn_user_info_notifier.dart
```diff
@@ -0,0 +1,15 @@
...
+/// Workplace FQDN from the OIDC `/userInfo` claim.
+@Riverpod(keepAlive: true)
+class WorkplaceFqdnUserInfoNotifier extends _$WorkplaceFqdnUserInfoNotifier
+    implements WorkplaceFqdnSource {
+  @override
+  String? build() => null;
+
+  @override
+  void setFqdn(String? rawFqdn) => state = normalizeWorkplaceFqdn(rawFqdn);
+}
```
dab246, line 15: This `keepAlive` provider persists across logouts. If the new account does not call `/userInfo`, the old account's FQDN takes precedence over the ecosystem fallback, potentially causing Drive to send the current OIDC token to the old host. Reset or invalidate this source upon logout or when initiating a new authentication/session, and add a regression test for account switching.

docs/adr/0110-workplace-fqdn-source-priority.md
```diff
@@ -0,0 +1,43 @@
...
+
+One derived provider observes every source and exposes the first non-null in list order; it is the only value consumers read.
+
+Priority is list position, not code branching — a new source is a new provider plus a list entry.
+
+Because every source is observed, a higher-priority source that resolves late takes over automatically, and clearing it falls back down the list.
+
+A source that resolves to a blank or unparseable value contributes nothing, so the next source is used.
+
+## Consequences
```
hoangdat, line 35: because Riverpod for workplaceFQDNSource is keepAlive, we should carefully check when we work with multiple account or delegation

lib/main/providers/workplace/fqdn/workplace_fqdn_user_info_notifier.dart
tddang-linagora: doc updated

docs/adr/0110-workplace-fqdn-source-priority.md
tddang-linagora: doc updated

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4841

dab246 (approved):
LGTM

PR #4847, opened 2026-09-21
Fix iOS app build failure on Xcode `27.0` and macOS `Golden Gate 27.0`

dab246:
## Problem

Xcode 27 no longer accepts iOS Simulator deployment targets below 15.0.
The app targets and `platform :ios` are already set to 15.0, but generated pod targets still use 13.0 and fail before the app code is compiled:

```text
The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 13.0,
but the range of supported deployment target versions is 15.0 to 27.0.x.
```

## Root cause

`flutter_additional_ios_build_settings` removes explicit deployment targets below iOS 13 so those pods inherit the project platform.
The custom `post_install` hook runs afterward, a missing value becomes `0.0` through `to_f`, so the hook writes 13.0 back:

```ruby
if config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'].to_f < 13.0
  config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '13.0'
end
```

## Solution

Raise the floor in the hook to 15.0 so it matches `platform :ios, '15.0'` and the app targets:

```ruby
# Xcode 27 only supports deployment targets from 15.0 upwards, which also
# matches the app targets. Override any pod with a lower deployment target.
if config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'].to_f < 15.0
  config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '15.0'
end
```

Regenerate the pod project with:

```bash
cd ios
pod install
```

This only updates the `Podfile` checksum in `Podfile.lock`, no pod versions or app-target deployment settings change.

## Summary by CodeRabbit

* **Chores**
  * Updated the minimum supported iOS version for included components to iOS 15.


tddang-linagora (approved):

hoangdat (approved):

Arsnael:
Docker image published for this PR: linagora/tmail-web-pr:4847

